PR #494, opened 2018-01-02
CFY-7620 New stage rpm

tehasdf:


.circleci/config.yml
```diff
@@ -0,0 +1,99 @@
...
+          at: *workdir
+      - run: npm test
+
+  build:
+    <<: *docker_node
+
+    steps:
+      - attach_workspace:
+          at: *workdir
+      - run: sudo npm install -g webpack
```
qooban: So even if you attach to workspace created during `beforebuild` step, you still need to install webpack?

packaging/cloudify-stage.spec
```diff
@@ -0,0 +1,67 @@
...
+
+cd ${RPM_SOURCE_DIR}
+sh scripts/preinstall.sh
+webpack --config webpack.config-prod.js --bail
+
+%install
+install -m 755 -d %{buildroot}/opt/cloudify-stage/dist
+install -m 755 -d %{buildroot}/opt/cloudify-stage/backend
+install -m 755 -d %{buildroot}/opt/cloudify-stage/conf
+install -m 755 -d %{buildroot}/opt/cloudify-stage/resources
```
qooban: I'm wondering what will happen during uninstall with directories and files created by cloudify-stage?
We create `userData` directory within cloudify-stage root folder containing custom widgets/templates/pages.

packaging/files/opt/cloudify-stage/NOTICE.txt
```diff
@@ -0,0 +1,59 @@
+---AngularJS---
```
qooban: Looks like this file is outdated. We are not using Grafana at all and Stage is based on React instead of Angular.
I'm not sure if one of the widget's (Topology widget) is not using Angular, still.
And there are a lot of libraries and tools in package.json dependencies sections, so probably we need to go through them and add here licences.

I suppose you have copied this file from somewhere, right? Do you know who is maintaining it?

packaging/files/opt/manager/scripts/restore-snapshot.py
```diff
@@ -0,0 +1,49 @@
...
+#  * See the License for the specific language governing permissions and
+#  * limitations under the License.
+
+import os
+import shutil
+import argparse
+
+
+def _restore(snapshot_root, homedir, override=False):
+    for folder in ['conf', 'dist/userData', 'dist/widgets', 'dist/templates']:
```
qooban: Is that file copied from other place?
@idob-cloudify Do we need here `dist/widgets` and `dist/templates` when we have `dist/userData`?

scripts/preinstall.sh
```diff
@@ -1,13 +1,24 @@
 #!/usr/bin/env bash
 set -x
-npm install webpack -g
-npm install bower -g
-npm install gulp -g
-npm install grunt-cli -g
+
+if [ ${CIRCLECI:false} == true ]; then
+    npm_install_g="sudo npm install -g"
```
qooban: @limor-gs After we merge those changes done by @funkyHat and @tehasdf, we could probably updated Stage-Multibranch job to use `scripts/preinstall.sh` instead of duplicating all the commands in job's configuration.

.circleci/config.yml
funkyHat: Yeah, this is because the "workspace" is the repo only, and `npm install -g` puts `webpack` outside that so we have to install it again in the new container

packaging/files/opt/cloudify-stage/NOTICE.txt
funkyHat: Yeah it's copied from https://github.com/cloudify-cosmo/cloudify-manager-install/blob/master/cfy_manager/components/stage/NOTICE.txt where it was probably copied from the old blueprint based installer... I don't know who maintains these NOTICEs.

packaging/files/opt/manager/scripts/restore-snapshot.py
funkyHat: This is a slightly manual merge of the changes in 2 different repos, keeping them all seemed sensible to me.
Further down there's a comment about ignoring failure to restore `dist/userData`. If we remove `dist/widgets` and `dist/template` here what happens when those exist in a snapshot restored from an old version (maybe it's fine?)

packaging/cloudify-stage.spec
funkyHat: I had similar questions about data left lying around in `/opt/manager` and `/opt/mgmtworker`.
It sounds like we will remove all of our folders in `/opt` during teardown, but this should happen in `cfy_manager`, not the RPM, so I'll update that PR

packaging/files/opt/manager/scripts/restore-snapshot.py
idob-cloudify: 'dist/widgets', 'dist/templates' should be removed from here, they contain only system widgets and templates that come with the release and should not be ovewritten from snapshot

packaging/files/opt/cloudify-stage/NOTICE.txt
qooban: @idob-cloudify Could you check that?

idob-cloudify: yes, it was copied from manager-install which was copied from manager-blueprints.
i also don't know who is maintaining this but it's not related to the rpm build, so let's continue with what we have a revisit to update the file content afterwards.

funkyHat:
Thanks @qooban 
@limor-gs has asked us to add the build to the Jenkinsfile so I still need to do that.
Also the system tests are unstable partly due to all the other new RPM merges, so we'll hold off on merging this (& -composer) for a few days until we've got the tests stable again

geokala:
Closed: As this PR is more than six months old without having been merged. Please reopen (and rebase if necessary) if it is still relevant.

PR #518, opened 2018-02-19
STAGE-684 / CY-121 - Separate user configuration from internal configuration

qooban:
### Configuration changes
- removed `conf/customer.json`
- removed `conf/widgets.json`
- moved user configuration parts from `conf/app.json` to `conf/userConfig.json`
- `conf/userConfig.json` is treated as default values storage for `userData/userConfig.json`
- user is able to modify configuration by updating `userData/userConfig.json`
- only properties present in `conf/userConfig.json` can be modified using `userData/userConfig.json`


qooban:
@edenpessach I have updated them according to changes I've done in package.json (added `flat` lib) and according to changes on the master (version).

PR #525, opened 2018-02-28
CY-123 - Stage widget help link & CY-138 - Allow adding ReadMe files to widgets that will be presented as widget's helpStage readme help

qooban:
# New configuration
- `hasReadme` - if widget has this property set to true (default: false), then `readme.md` file is expected to be in widget directory, this file is shown in modal after click on help icon
- `helpUrl` - if widget has this property set, then value of this field is treated as URL to which user will be redirected on help icon click
- order of setting help icon link
  - if `helpUrl` is set then it is taken no matter what is the value of `hasReadme` property, 
  - if `helpUrl` is not set, then:  
    - if `hasReadme` is set to true, then `readme.md` file from widget's folder will be shown after user click on help icon, 
    - if `hasReadme` not present or set to false, then no help icon will be shown

# Screenshots 

## Widget with readme file (without header)
image

## Widget with readme file in edit mode (without header)
image

## Widget with help link (with header)
image

## Help link in edit mode (with header)
image

## Readme modal
See [ReadmeModal_0.png](https://github.com/cloudify-cosmo/cloudify-stage/compare/CY-124?expand=1#diff-8ed56e951382b7d537358a2f37910d0a)


app/components/Widget.js
```diff
@@ -116,20 +150,34 @@ export default class Widget extends Component {
...
-                                <i className="expand link icon small"
-                                   onClick={() => this.props.onWidgetMaximize(this.props.pageId, this.props.widget.id, true)}/>
-                            }
-                        </div>
+                        this.props.widget.definition.showHeader ?
+                            <div className={`widgetViewButtons ${this.props.widget.maximized?'alwaysOnTop':''}`}>
+                                {
+                                    this.props.widget.maximized ?
+                                        [
+                                            helpIcon(null)
```
edenpessach: what does size null means?

qooban: I wanted to set normal size of the icon, bu there is no such possibility AFAIK.

There is size 'medium' in Semantic UI React:
https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/lib/SUI.js#L20
but 'medium' is not allowed for Icon component, see:
https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/elements/Icon/Icon.js#L68

What do you suggest? We can set undefined also or empty string here.

edenpessach: I see, I agree with what you did. Just add a comment 'Not setting size is not overriding icon normal size" or something like this and I'll approve and merge

PR #526, opened 2018-03-01
CY-119 plugin catalog - show matching plugins of manager distribution

edenpessach:


widgets/pluginsCatalog/src/PluginsCatalogModal.js
```diff
@@ -117,7 +117,7 @@ export default class PluginsCatalogModal extends React.Component {
             />
             Are you sure you want to upload the plugin
             {' '}
-            <b>({this.props.plugin && this.props.plugin.name})</b>
+            <b>({this.props.plugin && this.props.plugin.title})</b>
```
qooban: Why the name of the plugin is in the brackets? And what for we add `..` after that?
image
I've registered some time ago a bug - https://cloudifysource.atlassian.net/browse/STAGE-613 - maybe we can solve it, here?

edenpessach: That just how it was, what do you suggest we do instead?

qooban: I think the original intention here was to have sth like:
*NetConf (CentOS Core)*
Now - as I don't see OS name - I think it can be removed and be just:
*NetConf*.

PR #534, opened 2018-03-08
CY-172 - Hide widgets that users aren't permitted... - Filter out widgets from view solution

qooban:


app/components/Widget.js
```diff
@@ -95,26 +95,6 @@ export default class Widget extends Component {
             );
         }
 
-        if (this.props.widget.definition && !this._isUserAuthorized()) {
```
edenpessach: What is the behaviour if a user was authorised for a widget, and it puts it in a page.
then it looses the permission for the widget? does it just disappear?

qooban: Do we expect such situation? It depends on what we received from manager. That is fetched when stage's backend is started. So I suspect that it won't be handled properly until we reload 'Authorization config' from manager.

qooban: And if you are afraid of removal of this line, then check `WidgetsList` component. Similar check is just moved one level up.

edenpessach: It's not because of the authorization config, its when a user have a user apps with widgets he was authorized for and no longer is.

qooban: Closing this PR and creating new ona as new solution is completely different. See https://github.com/cloudify-cosmo/cloudify-stage/pull/542

qooban:
Continued in https://github.com/cloudify-cosmo/cloudify-stage/pull/542

qooban:
Let's discuss this solution once again as we will face another problems if go with #542.

PR #535, opened 2018-03-09
CY-179 - Minor changes to Stage Help

qooban:
* Updated ExternalLink component to have also Icon anchored instead of only text
* Updated WidgetDefinition to set default values for all optional widget's settings

### DIFF
--- a/app/utils/WidgetDefinition.js
+++ b/app/utils/WidgetDefinition.js
@@ -5,14 +5,20 @@
 export default class WidgetDefinition {
 
     constructor(data) {
-        // Set default
-        this.showHeader = true;
-        this.showBorder = true;
-        this.initialWidth = 3;
-        this.initialHeight = 12;
+        // Set default values for optional properties
         this.color = 'blue';
-        this.initialConfiguration = [];
+        this.categories = [Stage.GenericConfig.CATEGORY.OTHERS];
+        //this.description = undefined;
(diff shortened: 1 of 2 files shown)


app/utils/WidgetDefinition.js
```diff
@@ -5,14 +5,20 @@
...
-        this.showHeader = true;
-        this.showBorder = true;
-        this.initialWidth = 3;
-        this.initialHeight = 12;
+        // Set default values for optional properties
         this.color = 'blue';
-        this.initialConfiguration = [];
+        this.categories = [Stage.GenericConfig.CATEGORY.OTHERS];
+        //this.description = undefined;
+        //this.fetchUrl = undefined;
```
edenpessach: did you forget to delete the comments?

qooban: Actually, it is intentional, I wanted to explicitly list here the default values of all optional widgets settings. As assigning `undefined` makes no effect I have commented it out. What do you think?

BTW I have also updated documentation about settings on Friday: http://docs.getcloudify.org/4.3.0/manager_webui/custom-widgets/#widget-settings
Sorted all settings alphabetically and created `Required` column.

New changes related to Stage Help (`hasReadme` and `helpUrl`) are here: https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/690

PR #542, opened 2018-03-14
CY-172 - Hide widgets that users aren't permitted... - Templates update solution

qooban:
### ChangeLog
- Created initial template for viewer (no access to deployment action buttons, blueprint action buttons and composer link)
- Created initial template for operator (no access to blueprint action buttons and composer link)
- Implemented configurable drill down page
- Removed old readme file for widgets development
- Fixed some widget naming in initial templates

### Known issues
1. **Drill down page** - admin can create template with deployment/blueprint widget for user which don't have access to specific widgets in default drilldown page - eg. deployment/blueprint action buttons (`drillDownPage` configuration parameter is hidden). We can think of adding possibility to show hidden widget configuration parameters for admins. 
2. **Blueprint/Deployment sources** - viewer has access to Blueprint/Deployment sources and operator does not have. I have assumed that it is a bug on manager side. If I'm wrong, let me know, I'll update template for operator.


widgets/README-vanilajs.md
```diff
@@ -1,326 +0,0 @@
-#How to write a plugin
```
edenpessach: Why delete this file?

qooban: Why not? Do we need this?

It is not up-to-date and all the widget creation documentation was already moved to: http://docs.getcloudify.org/4.3.0/manager_webui/custom-widgets/ & http://docs.getcloudify.org/4.3.0/apis/widgets-components/. 

These 2 shall be the only place we keep for documentation about widget creation IMHO.

PR #546, opened 2018-03-19
CY-212 only allow sys_admin to install widgets

edenpessach:


backend/routes/Widgets.js
```diff
@@ -19,13 +20,22 @@ router.get('/list', function (req, res, next) {
         .catch(next);
 });
 
-router.put('/install', function (req, res, next) {
+function validateInstallWidgetsPermission(req, res, next) {
+    const permissionId = 'stage_install_widgets';
```
qooban: I'm wondering how we could extract common part of `app` and `backend` for such cases, not to duplicate, eg. consts (like this permission name) or functions (like `isAuthorized` in backend and `isUserAuthorized` in app). 

Do you have any idea? New directory in repo root dir with common files?

edenpessach: I feel that the added value of combing the code to another common dir is very low and I rather just duplicate it. We make the same logic with almost every component in the ui, we duplicate and it gives us flexibility.

`IsAuthorized` isn't doing exactly the same as `isUserAuthorized` since it's only checking for system roles - stage backend doesn't care about tenants.

qooban: We had such situations before, but I agree that added value is still low and we can live without it.

PR #554, opened 2018-04-18
CY-288 cloudify product tours

edenpessach:


app/index.tmpl.html
```diff
@@ -5,6 +5,7 @@
     <title>Cloudify UI</title>
     <link rel="stylesheet" type="text/css" href="/stage/style">
     <link rel="icon" type="image/png" href="/stage/app/images/favicon.png">
+    <link rel="stylesheet" type="text/css" href="/node_modules/hopscotch/dist/css/hopscotch.min.css">
```
edenpessach: @qooban whats the best way to load the css?

qooban: @edenpessach Maybe just `import 'hopscotch/dist/css/hopscotch.min.css';` in the place where you exactly need it (in `ToursModal`?). 

Another option which I learned recently is to go even further - to dynamically load that on demand, not add to bundle - see https://webpack.js.org/guides/code-splitting/#dynamic-imports.

I would avoid adding references to node_modules in index.html as we don't have node_modules in app package, only in backend package.

PR #555, opened 2018-04-18
CY-234 - Optimize Stage dependencies

qooban:
PR is huge, I know :-) A lot of files were modified because in React v16 there was major change in terms of prop types. There is no React.PropTypes, but PropTypes are imported from 'prop-types' library. Over 100 files had to be changed because of that.

Main 3 goals achieved:
1. Update Webpack from v1 to v4.6.0
2. Update ReactJS from v15 to v16.3.1
3. Clear warnings in `npm install`

Related changes:
1. package.json refactoring
- added valid license tag
- added node's compatible version
2. Organized depenedencies
- upgraded all necessary dependencies to be compatible with React v16
- migrated from react-addons-css-transition-group to react-transition-group
- moved gulp dependencies to development dependencies
- removed unused bower dependency
3. React Router migration
- migrated to React Router v4 as previous version was not compatible with React v16
4. Semantic UI 
- updated semantic-ui from v2.2.9 to v2.2.14
- updated semantic.json to avoid interactive installer.
- updated Dropdown to be closed on blur
5. EditableLabel component improvement
- created EditableLabel component without external dependencies as old component was not compatible with React v16
- updated unit tests
6. Small fixes
- fixed invalid URL path for drilldown pages
- React components comparison are now done using areComponentsEqual function from react-hot-loader


app/routes.js
```diff
@@ -1,128 +0,0 @@
...
-        window.location = store.getState().config.app.saml.ssoUrl;
-    };
-
-    return (
-        <Route path='/'>
-            <Route component={LogoPage}>
-                {store.getState().config.app.saml.enabled
-                    ?
-                    [
-                        <Route key='loginRoute' path='login' onEnter={redirectToSSO}/>,
```
qooban: @edenpessach I figured out that I have forgetten about this routes in new implementation. Will fix on Monday.

qooban: Fixed it https://github.com/cloudify-cosmo/cloudify-stage/pull/556/commits/110bacb44e7c8de777febef8f25106caea36b94b .

PR #564, opened 2018-05-01
CY-300 - Support hidden-value secrets in UI

qooban:
* Added new column in Secrets Table - Hidden Value:
image

* Updated secret creation modal with new checkbox - Hidden Value:
image

* Error message in popup upon clicking on eye icon to show secrets value by user without proper permissions:
image

* Error message in modal upon clicking on edit icon to update secrets value by user without proper permissions:
image

* Error message upon clicking on checkbox in table to change Hidden Value parameter value by user without proper permissions:
image
Message is directly taken from Cloudify Manager backend.


qooban:
If you want to stick to the **CY-300** description I can change error message on 'eye' icon click to 'dont' icon and popup on hover:
image

@talisela, @idob-cloudify : What do you think?

PR #567, opened 2018-05-09
CY-315 - Update deployment update modal

qooban:
image

# Features
- current (set in the context) blueprint is set at start
- deployment inputs are filled with current values if selected blueprint is the same as the old used for deployment
- you can change input values or upload YAML file to fill deployment inputs automatically
- blueprints can be chosen from installed blueprints
- when you hover on input you get description (if available in blueprint.plan.inputs dict)
- removed custom workflow input
- added Force update switch
- divided modal into 3 sections

# Minors
- unified actions for blueprints
- fixed minor styling issues


talisela:
Is the behavior when clicking "Run Custom workflow" the same? can we improve that as well, and show a list of workflows instead of making the user specify the workflow ID?

qooban:
@talisela @idob-cloudify PR updated according to last mail. Custom workflow removed and divided modal into three sections.

PR #578, opened 2018-05-23
CY-880 - Change "Number of nodes" widget to "Number of Compute Nodes"

qooban:
image


qooban:
Waiting for REST API change to get number of compute nodes with single call.

PR #579, opened 2018-05-24
CY-365 - Add a message to the list widgets, saying they are empty

qooban:
image
image


bmansheim:
Not sure if we need "on the manager" and "in the catalog" but you probably know better.

qooban:
@bmansheim I removed "on the manager" - as this we can take as default. I left "in the catalog" as this is only for two specific widgets - Blueprint Catalog and Plugins Catalog.

PR #583, opened 2018-05-29
(work-in-progress) CY-373 - Drilldown pages don't work after page name change or parent page name change or page removal

qooban:
I've spotted today another bug related to drill down pages.
I suspect that this one can help us with *CY-343 Duplicated widgets in deployment drill-down page*.

---
**UPDATE:** 
Bug is not as simple to solve as I thought it to be at the beginning. I have extended the scope of it a bit as drill down page feature is quite messy. I need to clean it

### DIFF
--- a/app/actions/widgets.js
+++ b/app/actions/widgets.js
@@ -101,7 +101,7 @@ export function addWidgetDrilldownPage(widgetId,drillDownName,drillDownPageId) {
 
 }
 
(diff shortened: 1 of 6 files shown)


geokala:
Closed: As this PR is more than six months old without having been merged. Please reopen (and rebase if necessary) if it is still relevant.

PR #606, opened 2018-06-13
Improve context of instructions

bmansheim:


README.md
```diff
@@ -11,24 +11,29 @@
...
 Open browser to see if application is running. It runs by default on `http://localhost:4000`. 
 
 Changes in the source code shall be hot loaded to the development development version of the application. For changes `app` directory you don't need to reload page, for changes in `widgets` directory you need to reload page to see your updates.
 
-## Build
-The following steps shall be done to create application package:
+### Package stage for remote deployment
+Create application package:
 - to create production build run: `npm run build`
 - to pack all necessary files into archive run: `npm run zip`
```
qooban: I would add few more lines here to present what shall be done next in order to fully update UI on specified Cloudify Manager.

Upload application package:
- define path to private SSH key to access Cloudify Manager - `export SSH_KEY_PATH=<PATH>`
- define Cloudify Manager IP adress - `export MANAGER_IP=<MANAGER_IP>`
- upload package to the Cloudify Manager and restart UI services - `npm run upload`

README.md
```diff
@@ -11,24 +11,29 @@
 ## Setup
 
 ### Configuration
+Create the resources needed to connect to the Cloudify Manager:
 - create `conf/me.json` file basing on `conf/me.json.template` - provide IP of your Cloudify Manager there
 - copy `/opt/cloudify-stage/resources/admin_token` from Cloudify Manager machine to `resources/admin_token` in repo main directory 
 
 ### Installation
+Install the webserver and database components:
```
qooban: Hmm... Maybe sth like "Install application dependencies and initialize database?" would be more accurate? 

As actually `npm run beforebuild` is calling the following commands:
```
npm install
cd backend
npm install
```
which means it just installs all dependencies in right places.

`npm run db-migrate` is for creating necessary tables in DB 
(it is actually doing more, but we don't need to describe it here, I suppose)

README.md
bmansheim: Implemented

README.md
bmansheim: Implemented

PR #607, opened 2018-06-14
CY-402 - Adding the ability to create a user with a tenant in UI

qooban:
# Before
image

# After
image
--
image
--
image


talisela:
I think that maybe if "is admin" is checked, we shouldn't present the tenants list

PR #652, opened 2018-09-14
CY-674 - Workflow inputs not properly handled

qooban:
Modified workflow execution modal to:
- allow sending all different types (boolean, string, number, object, array)
- present boolean types as toggle instead of dropdown
- allow sending strings explicitly
- allow reverting to default (similarly to what was implemented in Deployment Wizard)
- show error indicators on fields when field is mandatory and not filled upon submission

image

image

image


widgets/common/src/ExecuteDeploymentModal.js
```diff
@@ -92,7 +99,7 @@ export default class ExecuteDeploymentModal extends React.Component {
 
         switch (parameter.type){
             case 'boolean':
-                return GenericField.BOOLEAN_LIST_TYPE;
+                return GenericField.BOOLEAN_TYPE;
```
qooban: Rejoice this method

PR #723, opened 2019-01-07
CY-928 Add execution status to the Filter widget

inbala:
Adding execution status to the Filter widget, it will be shown by default in the dashboard and it will be a dynamic list.

screen shot 2019-01-08 at 11 52 41

screen shot 2019-01-08 at 11 51 55

screen shot 2019-01-08 at 11 52 07

### DIFF
--- a/templates/pages/adminDash.json
+++ b/templates/pages/adminDash.json
@@ -91,7 +91,8 @@
       "x": 0,
       "y": 10,
       "configuration": {
-        "filterByExecutions":false
+        "filterByExecutions":false,
+        "filterByExecutionsStatus":true
       }
     },
     {
--- a/templates/pages/app.json
+++ b/templates/pages/app.json
@@ -102,7 +102,8 @@
       "x": 0,
       "y": 37,
       "configuration": {
-        "filterByExecutions":false
+        "filterByExecutions":false,
+        "filterByExecutionsStatus":true
       }
     },
     {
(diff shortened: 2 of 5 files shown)


qooban:
@inbala I have just checked latest changes and it looks better, but when:

1. I select the specific execution status, then I don't have any other options in that execution status dropdown when I open it once again. I think that should be changed, to have always available all options depending on selected blueprint, deployment, execution.

2. I select single or multiple executions, then execution status dropdown list doesn't change, but probably should, you should see only status for selected execution(s) I suspect, am I right?

I think maybe you need to do that using separate REST API request. But this time you don't need to get all executions. Maybe summary API can be used for the case when execution_id context value is not set and executions API when that value is set?

qooban:
@inbala Executions Summary API: https://docs.cloudify.co/api/v3.1/#summarize-executions

qooban:
@inbala Not sure if you have done all the fixes you wanted to, but I have checked the latest version and 1st looks like solved, but 2nd not. What I also discovered is that when you have the same execution status in different blueprints/deployments and you:
1. Set execution status filter to such status (let it be 'completed' and this status is present for executions done on deployments: 'nodecellar' and 'hello-world')
2. Change deployment value from null to 'nodecellar' or 'hello-world'
3. Execution status is cleared
Why? Should stay if it is present in new context, right?

PR #732, opened 2019-01-28
CY-828 Add executions statuses widget

inbala:


widgets/executionsStatus/src/widget.js
```diff
@@ -0,0 +1,37 @@
...
+    initialHeight: 24,
+    color : 'blue',
+    isReact: true,
+    hasReadme: true,
+    permission: Stage.GenericConfig.WIDGET_PERMISSION('executionNum'),
+    categories: [Stage.GenericConfig.CATEGORY.EXECUTIONS_NODES, Stage.GenericConfig.CATEGORY.CHARTS_AND_STATISTICS],
+    initialConfiguration: [
+        Stage.GenericConfig.POLLING_TIME_CONFIG(10)
+    ],
+    fetchUrl: '[manager]/executions?_include=id,status&_include_system_workflows=True&[params]',
```
qooban: Use executions summary API: https://docs.cloudify.co/api/v3.1/#execution-summary

widgets/executionsStatus/src/widget.js
```diff
@@ -0,0 +1,37 @@
...
+Stage.defineWidget({
+    id: 'executionsStatus',
+    name: 'Executions Statuses Graph',
+    description: 'Shows the number of executions per status',
+    initialWidth: 4,
+    initialHeight: 24,
+    color : 'blue',
+    isReact: true,
+    hasReadme: true,
+    permission: Stage.GenericConfig.WIDGET_PERMISSION('executionNum'),
```
qooban: Create new permission in Manager for that. Update Permissions per role document as well.
Llet me know if you need directions.

widgets/executionsStatus/README.md
```diff
@@ -0,0 +1,7 @@
+### Number of running executions
```
qooban: Leave the README.md file empty. Create widget description in docs.getcloudify.com repo.
Let me know if you need directions.

app/components/basic/graphs/Graph.js
```diff
@@ -301,8 +308,29 @@ export default class Graph extends Component {
...
+                    {
+                        this.props.showTooltip &&
+                        IS_KEY_TIME
+                        ?
+                            <Tooltip isAnimationActive={false}
+                                     formatter={VALUE_FORMATTER}
+                                     labelFormatter={xAxisDataFormatter}
+                                     cursor={false}/>
+                        :
+                            <Tooltip isAnimationActive={false}
```
qooban: image

I think the label should also be formatted in case there's something else than time on X-axis. It's not necessary to present decimal part of the value. So maybe the formatter should be passed as an optional prop to Graph component?

app/components/basic/graphs/Graph.js
```diff
@@ -264,18 +264,25 @@ export default class Graph extends Component {
...
+                               dataKey={chart.name}
+                               yAxisId={chart.name}
+                               width={50}
+                               axisLine={STYLE}
+                               tick={STYLE}
+                               tickLine={STYLE}
+                               tickFormatter={VALUE_FORMATTER}
+                               label={<AxisLabel vertical fill={COLOR}>{chart.axisLabel}</AxisLabel>} />
+                    :
+                        <YAxis key={'yaxis'+chart.name}
```
qooban: Minor thing, but YAxis should be also formatted in our case to avoid showing decimal values.
Currently it shows at the beginning (when there are single executions in the system) ticks like 0.25, 0.5, 0.75. Shouldn't be hard to have simple switch telling graph to present only integers and avoid showing decimals.

templates/pages/adminDash.json
```diff
@@ -97,14 +97,25 @@
...
       "x": 0,
-      "y": 12
+      "y": 14
+    },
+    {
+      "name": "Executions Statuses Graph",
+      "width": 4,
+      "height": 24,
+      "definition": "executionsStatus",
+      "configuration": {
```
qooban: Is `pageSize` part of the executionStatus widget configuration?

templates/pages/app.json
```diff
@@ -108,14 +108,25 @@
...
       },
       "x": 0,
       "y": 42
+    },
+    {
+      "name": "Executions Statuses Graph",
+      "width": 4,
+      "height": 24,
+      "definition": "executionsStatus",
+      "configuration": {
```
qooban: The same comment as above :-)

templates/pages/adminDash.json
```diff
@@ -97,14 +97,25 @@
...
       "configuration": {
         "pageSize": 10
       },
       "x": 0,
-      "y": 12
+      "y": 14
+    },
+    {
+      "name": "Executions Statuses Graph",
+      "width": 4,
```
qooban: I'm wondering as width=4 is not very wide how the bar chart looks like on the smaller screens or with plenty of statuses? Have you checked that?

templates/pages/adminDash.json
inbala: Oops

templates/pages/adminDash.json
inbala: You are right it doesn't look good on small screens 
screen shot 2019-02-05 at 17 47 18

So should it be wider? But then I think that the executions table will be too narrow

qooban: You are right about make it wider - if we made it wider then executions would be too narrow.

Hmm... If we are sticked to that place with the chart, then:
- Maybe it's possible in `recharts` to show X-axis labels vertically?
- If not, then maybe it's possible to add labels on the bars?
I think there must be a way to do that.
If we are open to changes, then maybe we can think of another place? 
I don't have idea right know.

Also the coloring and grouping (described in another comment) could help here.

PR #734, opened 2019-01-31
Enable multiple selection in the dashboard filter by default

inbala:


qooban:
Solution for the issue I mentioned before ready: https://github.com/cloudify-cosmo/cloudify-stage/pull/736

PR #750, opened 2019-02-22
CY-1076 - Update stage backend to listen only on localhost

qooban:


qooban:
@idob-cloudify Yes, good catch, by mistake. Fixed.

PR #754, opened 2019-02-28
Updated dependencies to reduce number of security vulnerabilities

qooban:


package.json
```diff
@@ -35,7 +35,7 @@
     "d3-format": "^1.3.2",
     "diff": "^3.5.0",
     "file-saver": "^1.3.3",
-    "gs-ui-infra": "cloudify-cosmo/gs-ui-infra#4.5.5",
+    "gs-ui-infra": "cloudify-cosmo/gs-ui-infra#4.6.0",
```
idob-cloudify: why 4.6? master is 5.0.
we need to add this file to the bump-version replacement list?

qooban: I suppose that repo is abandonded. Nobody uses that except cloudify-stage, I think. And we actually use just a part of it (style + fonts).

That's why I thought it would be quicker to just manually created tag there. I think in the future we can reduce it to keep just a style and fonts = static assets and remove all the other stuff.

PR #795, opened 2019-05-12
CY-1162: adding support for Component node type

tyacbovi:


package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "better-console": "^0.2.4",
     "blueimp-md5": "^2.7.0",
-    "cloudify-blueprint-topology": "http://repository.cloudifysource.org/cloudify/4.5.5/ga-build/cloudify-blueprint-topology-4.5.5-ga.tar",
+    "cloudify-blueprint-topology": "file:../cloudify-blueprint-topology/cloudify-blueprint-topology.tar",
```
qooban: I assume that it's only for development? :-)

widgets/topology/src/GoToWidgetPageService.js
```diff
@@ -0,0 +1,14 @@
+import 'cloudify-blueprint-topology';
+
+export default class GoToWidgetPageServiceImpl extends cloudifyTopology.GoToWidgetPageService{
```
qooban: Not sure about the name (here and in `cloudify-blueprint-topology` repo) - we actually go to page, not to the widget, am I right?

tyacbovi: Yes :) will remove the "widget" string

widgets/topology/src/GoToPageService.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default class GoToPageServiceImpl extends cloudifyTopology.GoToPageService{
+    constructor(toolbox) {
+        super();
+        this.toolbox = toolbox;
+    }
+
+    goToPage(pageName, contextName, contextInput) {
+        this.toolbox.getContext().setValue('deploymentId', pageName);
+        this.toolbox.goToParentPage();
```
qooban: It really is very specific. I'd like to avoid such hardcoding.
Maybe go to specific page instead of going to parent? 

You can try to use similar solution as in Number of Deployments widget. The page to be opened on click on the widget is configurable - see `widgets/deploymentNum/src/widget.js`:
```
        {id: 'page', name: 'Page to open on click', description: 'Page to open when user clicks on widget content',
         type: Stage.Basic.GenericField.CUSTOM_TYPE, default: 'deployments', component: Stage.Basic.PageFilter}
```
Similarly we could add that parameter to topology widget configuration. Can even be hidden for the user (for that use `hidden: true` in configuration parameter definition).

Then instead of `goToParentPage()` you will use `goToPage(widget.configuration...)`. 
It's still not perfect solution, but less problematic I think.

widgets/topology/src/widget.js
```diff
@@ -24,7 +24,8 @@ Stage.defineWidget({
         {id: 'enableGroupClick', name: 'Enable group click', default:true, type: Stage.Basic.GenericField.BOOLEAN_TYPE},
         {id: 'enableZoom', name: 'Enable zoom', default:true, type: Stage.Basic.GenericField.BOOLEAN_TYPE},
         {id: 'enableDrag', name: 'Enable drag', default:true, type: Stage.Basic.GenericField.BOOLEAN_TYPE},
-        {id: 'showToolbar', name: 'Show toolbar', default:true, type: Stage.Basic.GenericField.BOOLEAN_TYPE}
+        {id: 'showToolbar', name: 'Show toolbar', default:true, type: Stage.Basic.GenericField.BOOLEAN_TYPE},
+        {id: 'showDetailsButtons', name: 'Show details buttons for different node types', default:false, type: Stage.Basic.GenericField.BOOLEAN_TYPE}
```
qooban: It's not in use right now or am I missing something?

widgets/topology/src/Topology.js
```diff
@@ -173,6 +175,15 @@ export default class Topology extends React.Component {
         $(this.glassRef.current).removeClass('unlocked');
     }
 
+    _goToDeploymentPage(nodeDeploymentId) {
+        if (! this.props.widget.configuration.showDetailsButtons){
```
qooban: Can be moved to `cloudify-blueprint-topology` repository. Check if `onDeploymentNodeClick` configuration parameter is defined and basing on that show/hide button.

PR #797, opened 2019-05-23
CY-1265 Add site management widget

inbala:
I added the site management widget in a new page :

site_page

The site management widget is a table similar to the secrets :

sites

We can create, update and delete the sites.

create_site

update_site


widgets/sites/src/SitesTable.js
```diff
@@ -0,0 +1,150 @@
+
+import CreateModal from './CreateModal';
+import UpdateModal from './UpdateModal';
+
+export default class SitesTable extends React.Component {
```
qooban: Can you add propTypes to all components you created explaining what are the expected props to be passed with clear type definition? It's really good practice to do that.

I know we don't have it in all components, but since some time I'll try to implement it for all new components.

widgets/sites/src/SitesTable.js
```diff
@@ -0,0 +1,150 @@
...
+                           totalSize={data.total}
+                           pageSize={this.props.widget.configuration.pageSize}
+                           sortColumn={this.props.widget.configuration.sortColumn}
+                           sortAscending={this.props.widget.configuration.sortAscending}
+                           searchable={true}
+                           className="sitesTable"
+                           noDataMessage={NO_DATA_MESSAGE}>
+
+                    <DataTable.Column label="Name" name="name" width="20%" />
+                    <DataTable.Column label="Location" name="location" width="20%" />
```
qooban: How does sorting by Location works? If it does not work, then let's remove name attribute from here.

widgets/sites/src/SitesTable.js
```diff
@@ -0,0 +1,150 @@
...
+                        data.items.map((site)=>{
+                            return (
+                                <DataTable.Row name={site.name}>
+                                    <DataTable.Data>
+                                        {site.name}
+                                        <ResourceVisibility visibility={site.visibility}
+                                                            onSetVisibility={(visibility) => {this._setSiteVisibility(site.name, visibility)}}
+                                                            allowedSettingTo={['tenant', 'global']} className="rightFloated"/>
+                                    </DataTable.Data>
+                                    <DataTable.Data>{site.location}</DataTable.Data>
```
qooban: I'm thinking of enhancing the presentation of location. Seeing just lat, lng is very raw.

To make it more user friendly I would show small map with pin point on hovering location (you can add some icon near the location for that). What do you think?

widgets/sites/src/UpdateModal.js
```diff
@@ -0,0 +1,93 @@
...
+            });
+        }
+    }
+
+    _updateSite() {
+        // Disable the form
+        this.setState({loading: true});
+
+        let actions = new Stage.Common.SiteActions(this.props.toolbox);
+        actions.doUpdate(this.props.site.name, null, this.state.siteLocation, this.state.siteNewName).then(()=>{
```
qooban: Very minor - let's add spaces:
`()=>{` => `() => {`

widgets/sites/src/UpdateModal.js
```diff
@@ -0,0 +1,93 @@
...
+            this.setState({errors: {}, loading: false});
+            this.props.toolbox.refresh();
+            this.props.onHide();
+        }).catch((err)=> {
```
qooban: Similar to above.

widgets/sites/src/UpdateModal.js
```diff
@@ -0,0 +1,93 @@
...
+            <div>
+                <Modal open={this.props.open} onClose={()=>this.props.onHide()}>
+                    <Modal.Header>
+                        <Icon name='edit' /> Update site {this.props.site.name}
+                    </Modal.Header>
+
+                    <Modal.Content>
+                        <Form loading={this.state.loading} errors={this.state.errors}
+                              onErrorsDismiss={() => this.setState({errors: {}})}>
+                            <Form.Field error={this.state.errors.siteNewName}>
```
qooban: I know it's still like that placeholder is field's label, but we had discussion when Tali was with us with UX specialist and he considered it as bad practice. Label should clearly describe what this field is about and placeholder can be a hint or example of what can be the value.

Ref.: https://cloudifysource.atlassian.net/browse/CYBL-487

Check for example - create user modal - `widgets/userManagement/src/CreateModal.js`:
image

Please update both modals - for site creation and for site update.

widgets/sites/src/widget.js
```diff
@@ -0,0 +1,49 @@
...
+                return Object.assign({}, site, {
+                    created_at: Stage.Utils.Time.formatTimestamp(site.created_at),
+                    deploymentCount: deploymentsPerSite[site.name] || 0
+                })
+            }),
+            total : _.get(sites, 'metadata.pagination.total', 0)
+        };
+
+        return (
+            <SitesTable widget={widget} data={formattedData} toolbox={toolbox}/>
```
qooban: I don't know the plans for the future, but as a user I would like not only to see table, but also to see map with all sites marked there. Is it planned as separate widget or as an extension in the future?

backend/package-lock.json
```diff
@@ -156,7 +156,7 @@
     "@types/geojson": {
       "version": "1.0.6",
       "resolved": "https://registry.npmjs.org/@types/geojson/-/geojson-1.0.6.tgz",
-      "integrity": "sha512-Xqg/lIZMrUd0VRmSRbCAewtwGZiAk3mEUDvV4op1tGl+LvyPcb/MIOSxTl9z+9+J+R4/vpjiCAT4xeKzH9ji1w=="
```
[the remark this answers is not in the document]
inbala: You are right, when I used node v10.15.x and npm v6.9.0 I didn't have those changes.
I'll drop this commit.

templates/main-default.json
```diff
@@ -7,6 +7,7 @@
     "catalog",
     "blueprints",
     "deploy",
+    "sites",
```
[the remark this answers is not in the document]
inbala: 1. Yes.
2. Sounds good. We should talk to Ofer (and Ido) about it, maybe he would like to reorganise the pages and the templates.

widgets/sites/src/widget.js
```diff
@@ -0,0 +1,49 @@
+
```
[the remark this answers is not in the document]
inbala: Sure :)
I'll soon add it to the docs (not sure how I forgot to do it for executionsStatus)

widgets/common/src/SiteActions.js
```diff
@@ -0,0 +1,29 @@
+
+class SiteActions {
```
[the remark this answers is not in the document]
inbala: Right, so I'll move it for now to under sites and when maybe in the future it will be used by different widgets we can move it to common.

widgets/sites/src/SitesTable.js
inbala: We plan to add a map with all the sites to the dashboard.
Maybe after we will have the map, it will easier to add it when hovering the location cell?

widgets/sites/src/widget.js
inbala: Yes it is planned as a separate widget.

widgets/sites/src/SitesTable.js
qooban: OK, let's wait for map widget, then with adding location cell popup map.

Still - I think that we should present lat/lng in more user friendly manner (as described in p. 1).
Another idea which could be useful I think is simply making a link from lat/lng text or adding a clickable icon nearby which will redirect us to e.g. google maps on specified location - https://developers.google.com/maps/documentation/urls/guide#map-action (without pin point it's free of use, there's static version but requires API key - which probably at some point of time will be required or we'll use Open Street Maps).

widgets/sites/src/SitesTable.js
inbala: I removed the name attribute of the location.

scripts/readmesConfig.json
```diff
@@ -44,6 +44,8 @@
     {"widget": "timeFilter", "link": "/content/working_with/console/widgets/timeFilter.md"},
     {"widget": "topology", "link": "/content/working_with/console/widgets/topology.md"},
     {"widget": "userGroups", "link": "/content/working_with/console/widgets/userGroups.md"},
-    {"widget": "userManagement", "link": "/content/working_with/console/widgets/userManagement.md"}
+    {"widget": "userManagement", "link": "/content/working_with/console/widgets/userManagement.md"},
+    {"widget": "executionsStatus", "link": "/content/working_with/console/widgets/executionsStatus.md"},
```
qooban: Let's have it sorted alphabetically. It's cleaner that way.

widgets/sites/src/CreateModal.js
```diff
@@ -0,0 +1,106 @@
...
+                </Modal.Header>
+
+                <Modal.Content>
+                    <Form loading={this.state.loading} errors={this.state.errors}
+                          onErrorsDismiss={() => this.setState({errors: {}})}>
+                        <Form.Field label='Name' error={this.state.errors.siteName} required>
+                            <Form.Input name='siteName' value={this.state.siteName}
+                                        onChange={this._handleInputChange.bind(this)}/>
+                        </Form.Field>
+                        <Form.Field label='Location' error={this.state.errors.siteLocation}>
```
qooban: We should provide help for the user - provide a way to select the place on map rather than setting lat/lng.

I assume that it will also be for the future when we have the other widget showing map, right?
Can we have one story for all updates of Site Management widget (add maps to list, create, update)?

widgets/sites/src/UpdateModal.js
```diff
@@ -0,0 +1,103 @@
...
+                                <Form.Input label='Name' name='siteNewName' value={this.state.siteNewName}
+                                            onChange={this._handleInputChange.bind(this)}/>
+                            </Form.Field>
+                            <Form.Field error={this.state.errors.siteLocation}>
+                                <Form.Input label='Location' name='siteLocation' value={this.state.siteLocation}
```
qooban: Similar comment as above:
We should provide help for the user - provide a way to select the place on map rather than setting lat/lng.

I assume that it will also be for the future when we have the other widget showing map, right?
Can we have one story for all updates of Site Management widget (add maps to list, create, update)?

widgets/sites/src/CreateModal.js
inbala: Sure.
https://cloudifysource.atlassian.net/browse/CY-1323

PR #804, opened 2019-06-04
Fixing tasks graph bugs

barucoh:


widgets/executions/src/backend.js
```diff
@@ -303,8 +304,9 @@ module.exports = (r) => {
             // we need to increase the Node's height accordingly and split the text
             // This process must be here after all the nodes are in the list
             const textSplitCalculation = (nodeWidth, textToCalculate) => {
-                let maximumLength = _.floor((nodeWidth - (paddingLeftRight * 2)) / textSizingFactor);
-                if (textToCalculate.length > maximumLength) {
+                let maximumLength = _.floor((nodeWidth - (paddingLeftRight * 2)) / textSizingFactor) - 1;
+                // Add 1 since arrays start from zero...(duh)
```
tehasdf: the comment is wrong because no matter what arrays start from, length is length :P

widgets/executions/src/backend.js
```diff
@@ -303,8 +304,9 @@ module.exports = (r) => {
...
+                if (textToCalculate.length + 1 > maximumLength) {
```
tehasdf: and maybe instead of `+1` here, it should be a `-2` in the maximumLength calculation, because the current `-1` there means "how much free space we want to leave around the text", and so if we want to increase the free space, probably there

PR #805, opened 2019-06-05
CY-1330 Add the ability to set the deployment's site

inbala:
I added the ability to set the deployment's site.

* Show the site name in the deployments widget

deployments

* Add site name input when creating a deployment

create_deployment

* Add the ability to set the deployment's site

set_site_dropdown

set_site

sites_dropdown


widgets/deployments/src/MenuAction.js
```diff
@@ -36,6 +37,8 @@ export default class MenuAction extends React.Component {
                                onClick={this.actionClick.bind(this)}/>
                     <Menu.Item icon='trash' content='Force Delete' name={MenuAction.FORCE_DELETE_ACTION}
                                onClick={this.actionClick.bind(this)}/>
+                    <Menu.Item icon='building' content='Set Site' name={MenuAction.SET_SITE_ACTION}
```
qooban: What do you think about moving it below Update and above Delete?

widgets/deployments/src/SetSiteModal.js
```diff
@@ -0,0 +1,113 @@
...
+                    </Modal.Header>
+
+                    <Modal.Content>
+                        <Form loading={this.state.loading} errors={this.state.errors}
+                              onErrorsDismiss={() => this.setState({errors: {}})}>
+                            <Form.Field error={this.state.errors.siteName} label='Site name'>
+                                <Form.Dropdown search selection value={this.state.siteName} name='siteName'
+                                               options={site_options} onChange={this._handleInputChange.bind(this)}/>
+                            </Form.Field>
+                            <Form.Field className='detachSite'>
```
qooban: Is that necessary? What if user clears dropdown value? Submitting empty site name wouldn't work as detach from current site?

widgets/deployments/src/MenuAction.js
inbala: I agree :)

PR #823, opened 2019-07-22
CY-1323 Add sites map widget

inbala:


widgets/sitesMap/src/widget.js
```diff
@@ -0,0 +1,115 @@
...
+    color: 'green',
+    isReact: true,
+    hasReadme: true,
+    permission: Stage.GenericConfig.WIDGET_PERMISSION('sites'),
+    categories: [Stage.GenericConfig.CATEGORY.SYSTEM_RESOURCES],
+    initialConfiguration: [
+        Stage.GenericConfig.POLLING_TIME_CONFIG(15)
+    ],
+
+    fetchData: function(widget, toolbox) {
```
qooban: You asked about making `fetchData` smaller. Please take a look [here](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/blueprints/src/widget.js). You can define internal function and reference it using `this`.

widgets/sitesMap/src/widget.js
```diff
@@ -0,0 +1,115 @@
...
+            const sitesData = data[0];
+            const deploymentsData = _.groupBy(data[1].items, 'site_name');
+            const executionsData = _.groupBy(data[3].items, 'deployment_id');
+            let siteStatuses = {};
+            _.forEach(sitesData, (site) => {
+                siteStatuses[site.name] = {
+                    ...site
+                };
+
+                let deploymentStates = {'pending': [], 'in progress': [], 'good': [], 'failed': []};
```
qooban: I think we should use consts for those states. You can create the object of empty arrays from `groupStates` (iterate over all and get just the `name` key value).

widgets/sitesMap/src/widget.js
qooban: But maybe - I don't say yes/no - we can fetch some of the data you are fetching here on demand, maybe not all of the data is necessary to be available at the start. If some data is only used after user interaction with widget, then we may move data fetching to that action event.

widgets/sitesMap/src/SiteControl.js
```diff
@@ -0,0 +1,58 @@
...
+            <span>
+                <h5 className='ui header dividing'>
+                    {site.name}
+                </h5>
+                <Grid columns={4} container>
+                    <Grid.Row>
+                        {
+                            _.map(DeploymentConsts.groupStates, (state) =>
+                                {
+                                    let value = site.deploymentStates[state.name].length;
```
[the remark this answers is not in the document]
inbala: Yes we always define for each site the deploymentStates with all the possible states.

widgets/sitesMap/src/SitesMap.js
```diff
@@ -0,0 +1,97 @@
...
+export default class SitesMap extends React.Component {
+
+    /**
+     * propTypes
+     * @property {object} widget - Widget object
+     * @property {object} toolbox - Toolbox object
+     * @property {object} data - object with sites data
+     */
+    static propTypes = {
+        widget: PropTypes.object.isRequired,
```
qooban: Maybe we can have just `showAllLabels` prop? (taken from the `widget.configuration.showAllLabels` in the parent) 
This component does not need the whole widget object I suspect. 

P.S. I know we used such big props before, but let's not copy bad practices, move on and improve :-)

widgets/sitesMap/src/SitesMap.js
```diff
@@ -0,0 +1,97 @@
...
+    render() {
+        const mapOptions =  {
+            zoom: 2.5,
+            minZoom: 2,
+            maxZoom: 18,
+            position: [50, 0],
+            bounds: [[-90, -180], [90, 180]],
+            viscosity: 0.75
+        };
+        const url='https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}{r}.png?lang=en';
```
qooban: Ah, so that's where the tiles come from :-) 
Yeah, so I think internet connection check at some point would be good.

widgets/sitesMap/src/widget.js
```diff
@@ -0,0 +1,134 @@
...
+    _processSite(site, siteStatuses, deploymentsData, executionsData, nodeInstanceData) {
+        const groupStates = Stage.Common.DeploymentConsts.groupStates;
+        let deploymentStates = _.reduce(groupStates, function(result, value, key) {
+            result[key] = [];
+            return result;
+        }, {});
+        let siteState = 'good';
+
+        _.forEach(deploymentsData[site.name], (deployment) => {
+            let currentState = '';
```
qooban: Maybe extract it to separate function, eg. `getDeploymentState`? 
Probably it would be wise to keep that logic closer the states definition (not to have magic strings like inProgress, failed, ..., referenced here).

app/app.js
```diff
@@ -27,6 +27,10 @@ import 'highlight.js/styles/xcode.css';
 // Import hopscotch
 import 'hopscotch/dist/css/hopscotch.css';
 
+// Import leaflet
+import 'leaflet/dist/leaflet.css';
+import 'leaflet/dist/leaflet.js';
```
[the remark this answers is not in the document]
inbala: I'm going with the first option :) 
Currently it's being used only in `widgets/sitesMap/src/SitesMap.js`.

widgets/common/src/GroupState.js
```diff
@@ -0,0 +1,53 @@
...
+            <Popup header={_.capitalize(state.name)}
+                   content={this.props.description}
+                   trigger={
+                       <Segment.Group className={this.props.className} disabled={disabled} onClick={this.props.onClick}>
+                           <Segment color={color} disabled={disabled} inverted>
+                               <Icon name={state.icon}/>
+                           </Segment>
+                           <Segment color={color} disabled={disabled} tertiary inverted>
+                               {value}
+                               </Segment>
```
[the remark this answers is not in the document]
inbala: Is it something I should install for my dev env?

app/utils/WidgetDefinition.js
```diff
@@ -2,6 +2,7 @@
  * Created by kinneretzin on 16/01/2017.
  */
 
+import Consts from '../utils/consts';
```
qooban: Do we need Consts here?

app/utils/WidgetDefinition.js
```diff
@@ -24,6 +25,9 @@ export default class WidgetDefinition {
         this.showBorder = true;
         this.showHeader = true;
 
+        // By default don't check the supported editions and keep backwards compatibility
```
qooban: OK, so I assume that we are all aware that 'smart' user can just take the code of the widget bundle it by himself without supportedEditions set and get it working.

If all are OK with what you introduced, maybe we can update some of the widgets to be available only in Premium version? (if I remember well we discussed that on Slack)
If so, could you create issue in JIRA for that and let me and Ofer know?

widgets/sitesMap/src/widget.js
```diff
@@ -62,7 +65,9 @@ Stage.defineWidget({
             this._processSite(site, siteStatuses, deploymentsData, executionsData, nodeInstanceData);
         });
 
-        return siteStatuses;
+        const sitesAreDefined = data[5].items.length > 0;
```
qooban: Let's switch the lines to have it done according to data array index (first `isMapAvailable`, second `sitesAreDefined`).

widgets/common/src/GroupState.js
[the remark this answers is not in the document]
qooban: Yes, but the configuration in stage is not ready, yet.
Webstorm and Visual Studio Code have good support for that.

app/utils/GenericConfig.js
```diff
@@ -50,6 +51,7 @@ export default class GenericConfig {
             CHARTS_AND_STATISTICS: 'Charts and Statistics',
             EXECUTIONS_NODES: 'Executions/Nodes',
             SYSTEM_RESOURCES: 'System Resources',
+            SPIRE: 'Spire',
```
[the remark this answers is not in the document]
inbala: No, for non-Spire editions there will be no permitted widgets in this category, it will be empty and therefor not visible.

app/utils/GenericConfig.js
```diff
@@ -3,6 +3,7 @@
  */
 
 
+import Consts from '../utils/consts';
```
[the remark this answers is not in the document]
inbala: No, I removed it

app/utils/WidgetDefinition.js
inbala: No, I removed it

app/utils/WidgetDefinition.js
inbala: Sure, I'll ask Ofer

widgets/sitesMap/src/SitesMap.js
```diff
@@ -1,18 +1,27 @@
 import { Icon } from 'leaflet';
 import PropTypes from 'prop-types';
 import SiteControl from './SiteControl';
+import { Link } from 'react-router-dom';
+import Consts from '../../../app/utils/consts';
```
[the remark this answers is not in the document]
inbala: You are right I don't need that consts, I use it now as a local const `const SITES_PAGE_PATH = '/page/site_management';`

widgets/sitesMap/src/SitesMap.js
```diff
@@ -1,18 +1,27 @@
...
 import { Map, TileLayer, Marker, Popup } from 'react-leaflet';
+import MessageContainer from '../../../app/components/MessageContainer';
```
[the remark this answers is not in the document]
inbala: I moved MessageContainer to basic components, but didn't change its content.
Do you think I should add this comments on how to use it? Because I'm not sure what to write.
Also, I changed it but there is still increase in the application size...

app/components/basic/MessageContainer.js
```diff
@@ -5,8 +5,8 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 
-import { Grid, Segment } from './basic';
-import SplashLoadingScreen from '../utils/SplashLoadingScreen';
+import { Grid, Segment } from './index';
+import SplashLoadingScreen from '../../utils/SplashLoadingScreen';
 
 export default class MessageContainer extends Component {
```
qooban: >Do you think I should add this comments on how to use it? Because I'm not sure what to write.
Yes, add sth simple:
```
/**
 * MessageContainer is a component which uses [Grid](https://react.semantic-ui.com/collections/grid) and [Segment](https://react.semantic-ui.com/elements/segment) components from Semantic-UI-React to display message box. Can be displayed in full screen or inside another container.
 *
 * ## Access
 * `Stage.Basic.MessageContainer`
 *
 * ## Usage
 * ### Full screen message - eg. Error Page
 * ![MessageContainer](manual/asset/MessageContainer_0.png)
 * ```
 * <MessageContainer ... props ... />
 * ```
 *
 * ### In widget message - eg. Site Map notification
 * ![MessageContainer](manual/asset/MessageContainer_1.png)
 * ```
 * <MessageContainer ... props ...  />
 * ```
 *
 */
```
so you also need to create screenshots for the examples, put it to `docs/manual/asset/` folder and update props in the examples to be meaningful.

Also provide description for prop-types as it is done in other basic components.

>Also, I changed it but there is still increase in the application size...

Hmm... That is a bit strange, but maybe the bundle size increase comes from 'leaflet', 'react-leaflet' and 'react-router-dom' imports? Can you try using - https://github.com/webpack-contrib/webpack-bundle-analyzer ? I used it few times before. It's very simple to configure and shows the bundle content in very nice way.

widgets/common/src/Consts.js
```diff
@@ -6,7 +6,11 @@ var consts = {
     defaultVisibility: 'tenant',
     sysAdminRole: 'sys_admin',
     defaultUserRole: 'default',
-    adminUsername: 'admin'
+    adminUsername: 'admin',
+    licenseEdition: {
```
qooban: Reg. slack message:

> I moved the LICENSE_EDITION const to Stage.Common.Consts and there all the consts are in camel case and not upper case. And I thought maybe we want to change it all to upper case?

Yes, good point and I agree with the idea. Generally I think we should follow the most common in JS world (afaik) - AirBnB style guide and it says that exported constants can optionally be uppercased - https://github.com/airbnb/javascript#naming--uppercase . In our case it is not explicitly exported, but exposing it through `Stage.Common`  in Stage architecture has similar role as exporting.

widgets/common/src/Consts.js
```diff
@@ -6,7 +6,11 @@ var consts = {
     defaultVisibility: 'tenant',
```
qooban: You can also change `var` to `const`.

widgets/sitesMap/src/SiteControl.js
```diff
@@ -0,0 +1,80 @@
+import PropTypes from 'prop-types';
```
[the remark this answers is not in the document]
inbala: Great, I wondered about the `react` import.
Thanks I removed it.

PR #828, opened 2019-08-09
CY-1162: Adding the expand button to Component and SharedResource

tyacbovi:
- Adding the expand button to Component and SharedResource.
- Fixing a bug in go-to button that only the last Component or SharedResource had worked properly


widgets/topology/src/Topology.js
```diff
@@ -1,7 +1,3 @@
-/**
```
qooban: I think we can leave that :) Kinneret created that file.

widgets/topology/src/Topology.js
```diff
@@ -62,26 +58,89 @@ export default class Topology extends React.Component {
...
-        if (this.props.data && this.props.data.data) {
-            const topologyData = {
-                data: this.props.data.data,
-                instances: this.props.data.instances,
-                executions: this.props.data.executions
-            };
+        return topology;
+    }
+
+    _create_base_topology(deploymentData){
```
qooban: Let's not mix the conventions for function naming and have consistency within the file - I'm OK for using `_`-prefix as it already is used unfortunately in many places, but snakecase is not for JS :-)

widgets/topology/src/Topology.js
```diff
@@ -62,26 +58,89 @@ export default class Topology extends React.Component {
...
+        };
+        return DataProcessingService.encodeTopologyFromRest(topologyData);
+    }
 
-            result = DataProcessingService.encodeTopologyFromRest(topologyData);
+    _add_expanded_topologies(deploymentsData, currentTopology) {
+    /* Will build all the topologies of the expanded deployments
+     * of Component or SharedResource nodes.
+     */
+        for (let i = 1; i < _.size(deploymentsData); i++) {
```
qooban: Maybe you can use `_.each(deploymentsData, (data, i) => { ... });` as in the https://github.com/cloudify-cosmo/cloudify-blueprint-topology/pull/88 ?

widgets/topology/src/Topology.js
```diff
@@ -181,6 +240,35 @@ export default class Topology extends React.Component {
         this.props.toolbox.drillDown(this.props.toolbox.getWidget(), 'deployment', {deploymentId: nodeDeploymentId}, nodeDeploymentId);
     }
 
+    _markDeploymentsToExpand(deploymentId){
+        let currentExpanded = this.props.toolbox.getContext().getValue('deploymentsToExpand');
```
qooban: Can be simplified to `let currentExpanded = this.props... || [];`

widgets/topology/src/Topology.js
```diff
@@ -181,6 +240,35 @@ export default class Topology extends React.Component {
...
+        this.props.toolbox.refresh();
+    }
+
+    _collapseExpendedDeployments(deploymentId){
+        let currentExpanded = this.props.toolbox.getContext().getValue('deploymentsToExpand');
+        if (!currentExpanded){
+            // Nothing to do
+            return;
+        }
+        _.remove(currentExpanded, (expended)=>{
```
qooban: expended => exp**a**nded :-)

widgets/topology/src/widget.js
```diff
@@ -28,31 +28,40 @@ Stage.defineWidget({
     ],
 
     fetchData: function(widget,toolbox) {
-        var deploymentId = toolbox.getContext().getValue('deploymentId');
-        var blueprintId = toolbox.getContext().getValue('blueprintId');
+        let deploymentId = toolbox.getContext().getValue('deploymentId');
```
qooban: Some or even all can be `const`?

widgets/topology/src/Topology.js
tyacbovi: Good eye :)

widgets/topology/src/Topology.js
tyacbovi: I am looping on two different structures, so I don't know if it's good to combine those

PR #832, opened 2019-08-20
CY-1630 Make stage work with ssl-only DB

geokala:


backend/config.js
```diff
@@ -1,7 +1,4 @@
-/**
```
qooban: Intentionally removed?

backend/config.js
```diff
@@ -44,6 +41,13 @@ module.exports = {
 
         config.managerUrl = manager.protocol + '://' + manager.ip + ':' + manager.port;
 
+        // Postgres client libraries require the cert to be loaded, not just given as a path
+        if(config.app.db.options.dialectOptions.ssl) {
```
qooban: 1. Are we sure the full path to the object will be present? I mean - if we won't specify `dialectOptions`, then we will read `undefined.ssl`, which will... Blow up ;-) 

2. `config.app.db.options.dialectOptions.ssl` is meant to be an object? 
If so, we can call `!_.isEmpty(config.app.db.options.dialectOptions.ssl)` if that is what you wanted to check.

backend/config.js
```diff
@@ -44,6 +41,13 @@ module.exports = {
...
+            config.app.db.options.dialectOptions.ssl.ca = fs.readFileSync(
```
qooban: Probably it would be safer (in case file does not exist) to:
1. try-catch the do `fs.readFileSync(...)` and catch potential errors,
2. get string from the result in `try` block, instead of `<string>.toString()` you can use `_.toString(...)`.

Ref.:
https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_readfilesync_path_options - why there's nothing about throwing error? (very bad docs...)
https://stackoverflow.com/questions/14391690/how-to-capture-no-file-for-fs-readfilesync

backend/config.js
geokala: Yeah. I can add it back in, but it really gives no useful information, and has been modified since the initial creation date.

backend/config.js
qooban: So as we have `lodash` imported here, we can use eg. `_.get(config.app, 'db.options.dialectOptions.ssl')`

Ref.: https://lodash.com/docs/4.17.15#get

backend/config.js
geokala: Hmm... should we catch an error here or just blow up?
Specifically I ask because that is expected to have been populated by our installer, meaning that if the file doesn't exist that is an exceptional circumstance and there's nothing we can really do because ssl is enabled and there is no cert, so an ENOENT bubbling up is probably the best result?

backend/config.js
geokala: 1. It did blow up in the CI test. I've added it to the default config, and it now will always exist as part of the manager install.

2. It looks like (unless I misread something along the way) it's supposed to be either false (if ssl is not on) or a dict containing settings (if ssl is on). Please do correct me if that's wrong.

geokala: You're thinking: _.get(config.app.db.options.dialectOptions.ssl, false)?

backend/config.js
qooban: Got it, I'm OK, can be removed.

backend/config.js
qooban: Then OK, if we expects it to have been populated by our installer, then let's keep it as it is.

backend/config.js
qooban: If you added it to always exist as part of the manager install, then probably we can rely on it.

geokala: Resolved- as noted above this will always be present.

PR #845, opened 2019-09-05
Removed cloudify logo image and used cloudify-ui-common supplied image instead

kubama:


backend/package.json
```diff
@@ -27,6 +27,7 @@
     "archiver": "^2.1.1",
     "body": "^5.1.0",
     "body-parser": "^1.16.1",
+    "cloudify-ui-common": "^1.1.0",
```
qooban: Let's also update `package.json` in root folder to use the same version of `cloudify-ui-common` package.

PR #847, opened 2019-09-07
CY-1710 DeploymentUpdate: add the update_executions param

tehasdf:
Add a frontend for the parameter introduced in cloudify-cosmo/cloudify-manager#1919

This involves an additional checkbox in the deployment update modal,
which will make a deployment update also reevaluate operation inputs.

update-operations

### DIFF
--- a/widgets/common/src/DeploymentActions.js
+++ b/widgets/common/src/DeploymentActions.js
@@ -53,7 +53,8 @@ class DeploymentActions {
         shouldRunReinstall = true,
         reinstallList = [],
         forceUpdate = false,
-        preview = false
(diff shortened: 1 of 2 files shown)


tehasdf:
@qooban sure, moved to after `force` (could well go before `force` as well but I don't think there's a good reason for either)

tehasdf:
And an updated screenshot as well
update-operations

tehasdf:
this is waiting for cloudify-cosmo/cloudify-manager#1919

tehasdf:
this is not required anymore because the feature is just straight enabled, instead of being behind a flag

PR #854, opened 2019-09-13
CY-1770 - Upgrade node-postgres package to the latest version

qooban:
Reverts cloudify-cosmo/cloudify-stage#853


qooban:
Do not merge before v5.0.5 release.

PR #856, opened 2019-09-13
Set Sites Map widget default zoom and focus based on available sites

kubama:


test/cypress/fixtures/license/valid_spire_license.yaml
```diff
@@ -0,0 +1,15 @@
+license:
```
qooban: I think we can't expose such license to the public. We need to obtain (@ofercloudify ?) Spire trial license with short expiration date.

PR #858, opened 2019-09-17
Set Sites Map widget default zoom and focus based on available sites

kubama:


widgets/sitesMap/src/SitesMap.js
```diff
@@ -123,27 +127,32 @@ export default class SitesMap extends React.Component {
...
-            viscosity: 0.75
+            maxBounds: [[-90, -180], [90, 180]],
+            maxBoundsViscosity: 0.75
         };
         const tilesUrl = `${mapUrl}/osm-intl/{z}/{x}/{y}{r}.png?lang=en`;
         const attribution = '<a href="https://wikimediafoundation.org/wiki/Maps_Terms_of_Use">Wikimedia</a>';
 
+        const sites = _.values(this.props.data);
+        const firstSiteLatLng = this._mapToLatLng(sites[0]);
+        if (sites.length > 1) {
```
tyacbovi: Won't it focus on the first site in both cases?

kubama: For case when there is more than one site the map is focused on a LatLngBounds instance, which is a rectangle containing all the sites - please see lines 142-147 where initially created LatLngBounds instance is extended to ensure it encapsulates all the sites.

widgets/sitesMap/src/SitesMap.js
```diff
@@ -123,27 +127,32 @@ export default class SitesMap extends React.Component {
...
+            mapOptions.bounds = L.latLngBounds([firstSiteLatLng, firstSiteLatLng])
```
tyacbovi: maybe this line could be:
L.latLngBounds( _(sites)
                        .map(this._mapToLatLng)
                        .value())
https://leafletjs.com/reference-1.5.0.html#latlngbounds-l-latlngbounds

kubama: You're right - updated

PR #867, opened 2019-09-27
CY-1736 - Update external part of UI developers documentation (cont.)

qooban:
* Added README file for configuration files
* Updated me.json.template file


conf/README.md
```diff
@@ -0,0 +1,92 @@
+# Configuration
+
+The following section describes different types of configuration used in cloudify-stage.
```
kubama: It would be more descriptive and accurate to say "configuration files" rather than "different types of configuration"

conf/README.md
```diff
@@ -0,0 +1,92 @@
...
+    * `delete` - integer, DELETE request timeout (milliseconds)
+    * `blueprintUpload` - integer, blueprint upload request timeout (milliseconds)
+  
+* `influx` - object, configuration of Influx DB connection
+  * `ip` - string, DB IP address
+  * `port` - integer, DB port number 
+  * `user` - string, DB username
+  * `password` - string, DB password
+  * `database` - string, DB name
+  * `timeout` - string, DB connection timeout
```
kubama: Is this a string, not an integer?
It would be good to add information about the unit.

conf/README.md
qooban: Makes sense. Updated.

conf/README.md
qooban: Good catch. Updated.

conf/README.md
```diff
@@ -0,0 +1,92 @@
...
+  * `certPath` - string, SAML certificate path
+  * `ssoUrl` - string, redirect URL to the application at the Single Sign-On identity provider
+  * `portalUrl` - string, redirect URL to the organization portal
+
+## Manager (`manager.json`)
+
+* `ip` - string, Cloudify Manager IP
+* `apiVersion` - string, Cloudify Manager REST API version, eg. "v3.1" 
+* `protocol` - string, Cloudify REST API protocol type, "http" or "https" 
+* `port` - string, Cloudify REST API port number
```
kubama: You may want to double check the type here as well (port string/number?)

qooban: Yeah, that one I checked before:
```
[centos@cloudify conf]$ pwd
/opt/cloudify-stage/conf
[centos@cloudify conf]$ cat manager.json
{
  "ip":"127.0.0.1",
  "apiVersion": "v3.1",
  "protocol" : "https",
  "port": "53333"
}
```

PR #871, opened 2019-10-03
CY-1795 - Update Manager Service Status to use new /status REST API

qooban:
### Before
image
image

### After
image
image

### DIFF
--- a/app/actions/status.js
+++ b/app/actions/status.js
@@ -33,15 +33,15 @@ export function getStatus() {
             .doGet('/status')
             .then(data => {
                 const { services, status } = data;
-                const filteredServices = _.chain(services)
+                const filteredServices = _(services)
                     .keys()
+                    .sort()
                     .map(serviceName => ({
                         name: serviceName,
                         isExternal: services[serviceName].is_external,
                         status: services[serviceName].status,
                         description: _.get(services[serviceName], 'extra_info.systemd.instances[0].Description', '')
                     }))
-                    .sortBy(service => service.name)
                     .value();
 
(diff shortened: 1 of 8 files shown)


app/actions/status.js
```diff
@@ -32,9 +32,19 @@ export function getStatus() {
         return managerAccessor
             .doGet('/status')
             .then(data => {
-                let services = _.filter(data.services, item => !_.isEmpty(item.instances));
-                services = _.sortBy(services, service => service.display_name);
-                dispatch(setStatus(data.status, services));
+                const { services, status } = data;
+                const filteredServices = _.chain(services)
```
kubama: FYI: `_.chain(...)` has a shorter equivalent: `_(...)`

app/actions/status.js
```diff
@@ -32,9 +32,19 @@ export function getStatus() {
...
+                    .keys()
+                    .map(serviceName => ({
+                        name: serviceName,
+                        isExternal: services[serviceName].is_external,
+                        status: services[serviceName].status,
+                        description: _.get(services[serviceName], 'extra_info.systemd.instances[0].Description', '')
+                    }))
+                    .sortBy(service => service.name)
```
kubama: I would do the `.sortBy` directly on service names array, before `.map`. It wouldn't be necessary to pass `service => service.name` as argument.

app/actions/status.js
qooban: 👍

app/actions/status.js
qooban: Good point. Refactored.

PR #877, opened 2019-10-24
CY-1599 - Create Cluster Status widget

qooban:
#### Main view
image
#### Node status details
image
#### External service
image
#### ID popup
image
#### Maintenance Mode
image

### TODO
- [x] Update thumbnail


widgets/highAvailability/src/ClusterService.js
```diff
@@ -0,0 +1,27 @@
...
+        [clusterService.broker]: 'comments'
+    }[name];
+
+    return (
+        <div style={{ verticalAlign: 'middle' }}>
+            {isExternal && (
+                <Popup trigger={<Icon name="external" style={{ float: 'right' }} />} content="External Service" />
+            )}
+            <Header style={{ marginTop: 0 }}>
+                <Icon name={icon} /> {clusterServiceName[name]}{' '}
```
kubama: Does `{' '}` have any purpose here?

widgets/highAvailability/src/ClusterServicesList.js
```diff
@@ -0,0 +1,85 @@
+import ClusterService from './ClusterService';
+import NodeStatus from './NodeStatus';
+import { clusterNodeStatuses, clusterService, clusterServiceStatus, clusterServiceStatuses } from './consts';
+
+export default function ClusterServicesList(props) {
+    const { services, toolbox } = props;
```
kubama: Destructuring assignment can be used directly in function parameter here

widgets/highAvailability/src/ClusterServicesList.js
```diff
@@ -0,0 +1,85 @@
...
+    const { DataTable } = Stage.Basic;
+    const fetchData = fetchParams => toolbox.refresh(fetchParams);
+
+    return (
+        <DataTable fetchData={fetchData}>
```
kubama: Why creating `fetchData` wrapper function instead of referencing `toolbox.refresh` directly here?

widgets/highAvailability/src/NodeStatus.js
```diff
@@ -0,0 +1,49 @@
...
+NodeStatus.propTypes = {
+    name: PropTypes.string.isRequired,
+    type: PropTypes.oneOf(clusterServices).isRequired,
+    status: PropTypes.oneOf(clusterNodeStatuses).isRequired,
+    services: PropTypes.objectOf(
+        PropTypes.shape({
+            is_external: PropTypes.bool,
+            status: PropTypes.oneOf(nodeServiceStatuses),
+            extra_info: PropTypes.shape({
+                systemd: PropTypes.shape({
```
tyacbovi: This area could be dynamic, which could contain systemd record like u wrote but also other structured info

widgets/highAvailability/src/ClusterServicesList.js
```diff
@@ -0,0 +1,85 @@
...
+            <DataTable.Column label="Service Type" width="15%" />
+            <DataTable.Column label="Node Name" width="20%" />
+            <DataTable.Column label="Status" width="5%" />
+            <DataTable.Column label="Private IP" width="15%" />
+            <DataTable.Column label="Public IP / LB IP" width="15%" />
+            <DataTable.Column label="Version" width="15%" />
+
+            {_.map(services, (service, serviceName) => {
+                const numberOfNodes = service.nodes.length;
```
kubama: I would use `_.size` here just to be safe

widgets/highAvailability/src/ClusterServicesList.js
```diff
@@ -0,0 +1,85 @@
...
+                const backgroundColor = {
+                    [clusterServiceStatus.OK]: '#21ba45',
+                    [clusterServiceStatus.DEGRADED]: '#fbbd08',
+                    [clusterServiceStatus.FAIL]: '#db2828'
+                }[service.status];
+
+                return _(service.nodes)
+                    .sortBy(node => node.name)
```
kubama: Shorted equivalent: `.sortBy('name')`

widgets/highAvailability/src/ClusterServicesList.js
```diff
@@ -0,0 +1,85 @@
...
+                                    <ClusterService isExternal={service.is_external} name={serviceName} />
+                                </DataTable.Data>
+                            )}
+                            <DataTable.Data>{node.name}</DataTable.Data>
+                            <DataTable.Data>
+                                <NodeStatus
+                                    name={node.name}
+                                    type={serviceName}
+                                    status={node.status}
+                                    services={node.services}
```
kubama: You may consider adding `services` to `nodes` shape in prop types

widgets/highAvailability/src/ClusterServicesList.js
```diff
@@ -0,0 +1,85 @@
...
+                            <DataTable.Data>{node.version}</DataTable.Data>
+                        </DataTable.Row>
+                    ))
+                    .value();
+            })}
+        </DataTable>
+    );
+}
+
+ClusterServicesList.clusterServiceProps = PropTypes.shape({
```
kubama: Why not just `const clusterServiceProps`?

widgets/highAvailability/src/ClusterServicesList.js
```diff
@@ -0,0 +1,85 @@
...
+        })
+    ).isRequired,
+    is_external: PropTypes.bool.isRequired
+}).isRequired;
+
+ClusterServicesList.propTypes = {
+    services: PropTypes.shape({
+        [clusterService.manager]: ClusterServicesList.clusterServiceProps,
+        [clusterService.db]: ClusterServicesList.clusterServiceProps,
+        [clusterService.broker]: ClusterServicesList.clusterServiceProps
```
kubama: Why not generate this dynamically to always include all `clusterService` keys? Something like this: `_.mapValues(clusterService, () => ClusterServicesList.clusterServiceProps)`

widgets/highAvailability/src/NodeServices.js
```diff
@@ -0,0 +1,141 @@
+import _ from 'lodash';
```
kubama: Unnecessary import?

widgets/highAvailability/src/NodeServices.js
```diff
@@ -0,0 +1,141 @@
...
+    isExternal: PropTypes.bool.isRequired
+};
+ServiceHeader.defaultProps = {
+    description: ''
+};
+
+const ServiceStatus = ({ status }) => {
+    const { Icon } = Stage.Basic;
+
+    const StatusIcon = () => {
```
kubama: Why creating a wrapping `StatusIcon` component instead of directly assigning `<Icon .../>` instance to a variable?

widgets/highAvailability/src/consts.js
```diff
@@ -0,0 +1,33 @@
+import _ from 'lodash';
```
kubama: Unnecessary import?

widgets/highAvailability/src/consts.js
```diff
@@ -0,0 +1,33 @@
...
+    OK: 'OK',
+    FAIL: 'FAIL'
+};
+export const clusterNodeStatuses = _.keys(clusterNodeStatus);
+
+export const nodeServiceStatus = {
+    Active: 'Active',
+    Inactive: 'Inactive'
+};
+export const nodeServiceStatuses = _.keys(nodeServiceStatus);
```
kubama: For `clusterService`, `clusterServiceStatus`, `clusterNodeStatus` and `nodeServiceStatus` keys are (and need to be) equal to values. If so I would simplify and make it less prone to typos by first creating array of values and then creating corresponding objects with `_.keyBy`, like so:
```
export const clusterServices = ['manager', 'db', 'broker'];
export const clusterService = _.keyBy(clusterServices);
```

Also, I would consider renaming the consts. When both `clusterServices` and `clusterService` are imported I find it confusing to tell which is what. Unless you have a better idea I think renaming `clusterService` to `clusterServiceEnum` would help.

widgets/highAvailability/src/ClusterService.js
qooban: Good catch, it was because there was more components under Header previously and I needed separation. Fixed.

widgets/highAvailability/src/ClusterServicesList.js
qooban: Right, fixed.

widgets/highAvailability/src/NodeStatus.js
qooban: Good point, I'll simplify `extra_info` part.

widgets/highAvailability/src/ClusterServicesList.js
qooban: Changed.

widgets/highAvailability/src/ClusterServicesList.js
qooban: Nice, changed.

widgets/highAvailability/src/ClusterServicesList.js
qooban: Yep, done.

widgets/highAvailability/src/ClusterServicesList.js
qooban: Can be. Changed.

widgets/highAvailability/src/ClusterServicesList.js
qooban: Nice method, changed.

widgets/highAvailability/src/NodeServices.js
qooban: Removed.

widgets/highAvailability/src/NodeServices.js
qooban: Simplified.

widgets/highAvailability/src/NodeStatus.js
```diff
@@ -0,0 +1,49 @@
...
+                    instances: PropTypes.arrayOf(
+                        PropTypes.shape({
+                            Description: PropTypes.string
+                        })
+                    )
+                })
+            })
+        })
+    )
+};
```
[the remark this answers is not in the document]
qooban: Done.

widgets/highAvailability/src/consts.js
qooban:
```
export const clusterServices = ['manager', 'db', 'broker'];
export const clusterService = _.keyBy(clusterServices);
```
It's nice solution, but not helping developer IMHO. `clusterService` is calculated in runtime, so you cannot get benefits of IDE's auto-complete features during development. Values here are not important, could be 1, 2, 3, instead of manager, db, broker. As in JS we don't have pure enums we have to deal somehow with it.

Regarding naming - I will change them to have `Enum` suffix.

I decided after doing quick resarch to add Object.freeze function to make enums immutable.

widgets/highAvailability/src/consts.js
qooban: Yes, removed.

PR #878, opened 2019-10-24
Updating stage and composer to work with cfyuser instead of stageuser…

barucoh:
… composeruser


qooban:
@tehasdf Same question as in https://github.com/cloudify-cosmo/cloudify-blueprint-composer/pull/759#issuecomment-598594062

barucoh:
@qooban @tehasdf Can you guys please either merge or close this?

qooban:
@barucoh I'm closing that one as I'm introducing RPM for stage (#972) and there will be more changes regarding user/group when we do the RPM for CY-1859. 

@tehasdf Do you know if CY-1859 in scope of 5.1?

PR #882, opened 2019-10-30
CY-1644 Display missing tasks graph as a note instead of an error

davidginzbourg:


.circleci/config.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           command: npm run prodtest
       - run:
           name: Check code style
-          command: npm run lint
+          command: npm run lint -- --quiet
```
tyacbovi: We want to see this

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -3,6 +3,7 @@
  */
 import GraphNodes from './GraphNodes';
 import GraphEdges from './GraphEdges';
+import { Header, Message } from 'semantic-ui-react';
```
qooban: Do not import from external libraries in widgets source code (it will make widget bundle size large).
Update L72 to get all necessary components: `const { Header, Loading, Message } = Stage.Basic;`

davidginzbourg: Import removed and imported from `Stage.Basic` instead.

PR #884, opened 2019-11-14
CY-1489 - Update cloudify-stage to use new repository

qooban:


qooban:
Do not merge before v5.0.5 release.

PR #887, opened 2019-12-06
CY-1527 - Update for Spire Manager widget to handle bulk status update

qooban:


test/cypress/support/users.js
```diff
@@ -23,7 +23,7 @@ Cypress.Commands.add('addUserToTenant', (username, tenant, role) =>
 );
 
 Cypress.Commands.add('removeUserFromTenant', (username, tenant) => {
-    if (username !== 'admin' && tenant !== 'default_tenant') {
+    if (username !== 'admin' && username !== 'manager_status_reporter' && tenant !== 'default_tenant') {
```
tyacbovi: there are another two
DB_STATUS_REPORTER = 'db_status_reporter'
BROKER_STATUS_REPORTER = 'broker_status_reporter'
MANAGER_STATUS_REPORTER = 'manager_status_reporter'
well do we want to test that we can not do this or maybe hide it from the UI?

qooban: I'll add them as well. Thanks!

qooban: System Tests (Stage-SystemTests # 963) passed after that change. Please re-review.

widgets/managers/src/ManagersTable.js
```diff
@@ -50,13 +51,10 @@ export default class ManagersTable extends React.Component {
...
         this.props.toolbox.getEventBus().on('managers:refresh', this.refreshData, this);
 
-        _.forEach(this.props.data.items, manager =>
-            this.actions.getClusterStatus(
-                manager.id,
-                this.handleStatusFetching,
-                this.handleStatusUpdate,
-                this.handleStatusError
-            )
+        const managerIds = _.map(this.props.data.items, manager => manager.id);
```
kubama: Simpler: `_.map(this.props.data.items, 'id')`

widgets/managers/src/ManagersTable.js
```diff
@@ -50,13 +51,10 @@ export default class ManagersTable extends React.Component {
...
+        this.handleStatusBulkFetching(managerIds);
+        _.forEach(managerIds, managerId =>
+            this.actions.getClusterStatus(managerId, () => {}, this.handleStatusUpdate, this.handleStatusError)
```
kubama: I believe using `_.noop` instead of `() => {}` would save from creating a new function in each iteration

widgets/managers/src/RefreshButton.js
```diff
@@ -35,8 +35,9 @@ export default class RefreshButton extends React.Component {
         const actions = new Actions(this.props.toolbox);
 
         const clusterStatusPromise = managerId =>
-            actions.getClusterStatus(managerId, this.props.onStart, this.props.onSuccess, this.props.onFail);
+            actions.getClusterStatus(managerId, () => {}, this.props.onSuccess, this.props.onFail);
```
kubama: I believe using `_.noop` instead of `() => {}` would save from creating a new function each time this line is executed

widgets/managers/src/ManagersTable.js
qooban: Done.

widgets/managers/src/ManagersTable.js
qooban: Done.

widgets/managers/src/RefreshButton.js
qooban: Done.

PR #888, opened 2019-12-09
cloudify-ui-components updated to 0.0.7

kubama:
As cloudify-ui-components 0.0.7 included some breaking changes to UrlOrFileInput component cloudify-stage had to be updated accordingly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cloudify-stage",
-  "version": "5.0.5-.dev1",
+  "version": "5.0.5--.dev1",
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cloudify-stage",
-  "version": "5.0.5-.dev1",
+  "version": "5.0.5--.dev1",
```
qooban: Probably by mistake?

kubama: No idea why as the file is auto generated, but corrected

PR #891, opened 2019-12-12
cloudify-ui-components updated to 0.0.8

kubama:


app/styles/style.scss
```diff
@@ -105,10 +105,52 @@ a:hover {
       top: 2rem;
```
qooban: Are all those changes additions or just you moved some parts of styling from one place to another?

My point here is not to extend that file as it is planned to move styles from here to be closer to components.

kubama: Actually both moved and added some styling to existing rules

PR #892, opened 2019-12-13
CY-1527 & CY-1599 & CY-1514 - Cluster Status in UI - tests and improvements

qooban:
- Improved WidgetBackend Manager Service to allow getting all data (doGetFull method)
- Fixed SystemServicesStatus to be clickable
- Moved spire deployments fetching to widget backend
- Added system tests for:
  - CY-1527 - Create Spire Manager widget (formerly Cloudify MoM widget)
  - CY-1599 - Cluster status widget
  - CY-1514 - UI - Manager service status replaced with Cluster service status

### DIFF
--- a/backend/handler/services/ManagerService.js
+++ b/backend/handler/services/ManagerService.js
@@ -20,6 +20,27 @@ module.exports = (function() {
         return call(consts.ALLOWED_METHODS_OBJECT.get, url, params, null, headers);
     }
 
+    function doGetFull(url, params, headers, fullData = { items: [] }, size = 0) {
+        const p = params;
+        p._size = 1000;
+        p._offset = size;
+        let s = size;
+
+        const pr = this.doGet(url, p, headers);
+
+        return pr.then(data => {
+            s += data.items.length;
+            const fd = fullData;
+            fd.items = _.concat(fullData.items, data.items);
+            const total = _.get(data, 'metadata.pagination.total');
+
+            if (total > size) {
+                return this.doGetFull(url, p, headers, fullData, s);
+            }
+            return fd;
+        });
+    }
+
     function doPost(url, params, data, headers) {
         return call(consts.ALLOWED_METHODS_OBJECT.post, url, params, data, headers);
     }
@@ -39,6 +60,7 @@ module.exports = (function() {
     return {
         call,
         doGet,
+        doGetFull,
         doPost,
         doDelete,
         doPut,
--- a/backend/handler/services/ManagerService.js
+++ b/backend/handler/services/ManagerService.js
@@ -21,23 +21,20 @@ module.exports = (function() {
     }
 
     function doGetFull(url, params, headers, fullData = { items: [] }, size = 0) {
-        const p = params;
-        p._size = 1000;
-        p._offset = size;
-        let s = size;
+        params._size = 1000;
+        params._offset = size;
 
-        const pr = this.doGet(url, p, headers);
+        const promise = this.doGet(url, params, headers);
 
-        return pr.then(data => {
-            s += data.items.length;
-            const fd = fullData;
-            fd.items = _.concat(fullData.items, data.items);
+        return promise.then(data => {
+            size += data.items.length;
+            fullData.items = _.concat(fullData.items, data.items);
             const total = _.get(data, 'metadata.pagination.total');
 
             if (total > size) {
-                return this.doGetFull(url, p, headers, fullData, s);
+                return this.doGetFull(url, params, headers, fullData, size);
             }
-            return fd;
+            return fullData;
         });
     }
 
--- a/test/cypress/integration/spire_manager_spec.js
+++ b/test/cypress/integration/spire_manager_spec.js
@@ -1,7 +1,23 @@
+import { styles } from '../support/cluster_status_commons';
+
 describe('Spire Manager widget', () => {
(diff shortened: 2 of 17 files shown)


backend/handler/services/ManagerService.js
```diff
@@ -20,6 +20,27 @@ module.exports = (function() {
         return call(consts.ALLOWED_METHODS_OBJECT.get, url, params, null, headers);
     }
 
+    function doGetFull(url, params, headers, fullData = { items: [] }, size = 0) {
+        const p = params;
```
kubama: Is this reassign enforced by no-param-reassign eslint rule? If yes you may consider adding `"no-param-reassign": ["warn", { "props": false }]` to .eslintrc file as it is already done in composer - there is really no justification for preventing parameter properties reassignment, especially if it can be still done with such workaround.

backend/handler/services/ManagerService.js
```diff
@@ -20,6 +20,27 @@ module.exports = (function() {
...
+        p._size = 1000;
+        p._offset = size;
+        let s = size;
+
+        const pr = this.doGet(url, p, headers);
```
kubama: `s` and `pr` are very indescriptive names for variables

test/cypress/integration/spire_manager_spec.js
```diff
@@ -0,0 +1,196 @@
...
+        cy.get('table.servicesData').should('be.visible');
+        cy.get('table.servicesData').within(() => {
+            cy.get('tbody tr:nth-child(1)').should('have.text', ' Manager');
+            cy.get('tbody tr:nth-child(1)').should('have.attr', 'style', styles.ok);
+            cy.get('tbody tr:nth-child(2)').should('have.text', ' Database');
+            cy.get('tbody tr:nth-child(2)').should('have.attr', 'style', styles.ok);
+            cy.get('tbody tr:nth-child(3)').should('have.text', ' Message Broker');
+            cy.get('tbody tr:nth-child(3)').should('have.attr', 'style', styles.ok);
+        });
+        cy.get('tbody > :nth-child(1) > :nth-child(5) i.statusIcon').trigger('mouseout');
```
kubama: There is a lot of code duplication in this file, for example lines 100-110 could be extracted into a function and easily reused in 112-122 and 124-134

test/cypress/support/index.js
```diff
@@ -16,13 +16,11 @@
 // Import commands.js using ES2015 syntax:
 import './commands';
 
-// Alternatively you can use CommonJS syntax:
-// require('./commands')
-
 // use `Cypress` instead of `cy` so this persists across all tests
 Cypress.on('window:before:load', window => {
     // Uncomment for testing HTTP Requests from client-side
```
kubama: I don't think this comment is valid anymore

backend/handler/services/ManagerService.js
```diff
@@ -20,6 +20,27 @@ module.exports = (function() {
         return call(consts.ALLOWED_METHODS_OBJECT.get, url, params, null, headers);
     }
 
+    function doGetFull(url, params, headers, fullData = { items: [] }, size = 0) {
```
kubama: Looking at how this function is used I don't see any reason to wrap `items` array in `fullData` object

backend/handler/services/ManagerService.js
```diff
@@ -20,6 +20,27 @@ module.exports = (function() {
...
+        return pr.then(data => {
+            s += data.items.length;
+            const fd = fullData;
+            fd.items = _.concat(fullData.items, data.items);
+            const total = _.get(data, 'metadata.pagination.total');
+
+            if (total > size) {
+                return this.doGetFull(url, p, headers, fullData, s);
+            }
+            return fd;
```
kubama: Could return `items` array directly here and update caller of this function accordingly

backend/handler/services/ManagerService.js
qooban: I just learned about the rule a bit more. Still don't have strict verdict what is the best. I like the idea of immutability, but for sure my quick implementation (it was actually partial copy-paste from client-side code: see `Manager.doGetFull`) was not perfect. For now I extended eslint rule in stage, because it will be updated anyway (as we need to update whole stage code to follow new rules), but I still am not convinced we should avoid that rule for parameter properties.

backend/handler/services/ManagerService.js
qooban: Updated.

backend/handler/services/ManagerService.js
qooban: It could be optimized, but I think it is reasonable to leave it. We want to return the object with the same construction as single request.

test/cypress/support/index.js
qooban: Removed.

test/cypress/integration/spire_manager_spec.js
qooban: Generally in tests I don't bother about duplications as it helps you to spot the problematic place faster.

But in case of Cypress test (where errors can easily be spotted with help of test runner) it makes sense to improve it slightly. Updated.

backend/handler/services/ManagerService.js
qooban: See previous comment.

PR #894, opened 2019-12-19
CY-1889: adding pre-push hook for checking lint issues

tyacbovi:


package.json
```diff
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "better-console": "^0.2.4",
-    "blueimp-md5": "^2.7.0",
+    "blueimp-md5": "^2.7G.0",
```
qooban: Typo?

package.json
```diff
@@ -149,5 +150,15 @@
     "webpack-cli": "^3.3.9",
     "webpack-dev-middleware": "^3.7.2",
     "webpack-dev-server": "^3.8.2"
+  },
+  "husky": {
```
qooban: Husky shouldn't be added to `devDependencies`?

tyacbovi: That's the git hook settings, not the package

qooban:
Cleaning open PRs, not merging solution for now.

PR #901, opened 2020-01-10
CY-2184 - System status icon not up to date

qooban:
Implemented approach:
* cluster status is fetched periodically, every 10 seconds, response from REST API is saved in redux store, status fetching is optimized this way:
  * when Cluster Status widget is on the active page, full cluster status is fetched
  * when there is no Cluster Status widget on the active page, only summary cluster status is fetched
* Cluster Status widget is not polling data at all, it is reusing data from redux store


app/actions/clusterStatus.js
```diff
@@ -22,15 +22,28 @@ export function errorClusterStatus(error) {
     };
 }
 
-export function getClusterStatus() {
+function isClusterStatusWidgetOnPage(pageId, pages) {
+    const currentPage = _.find(pages, page => page.id === pageId);
```
kubama: If `pageId` is string or number it could be simplified to `const currentPage = _.find(pages, {id: pageId});`

app/actions/clusterStatus.js
```diff
@@ -22,15 +22,28 @@ export function errorClusterStatus(error) {
...
+    const clusterStatusWidgetDefinitionName = 'highAvailability';
+    const clusterStatusWidgets = _.find(
+        _.get(currentPage, 'widgets'),
+        widget => widget.definition === clusterStatusWidgetDefinitionName
```
kubama: Assuming `clusterStatusWidgetDefinitionName` is a string line 30 could be simplified to:
`{definition: clusterStatusWidgetDefinitionName}`

app/actions/clusterStatus.js
```diff
@@ -22,15 +22,28 @@ export function errorClusterStatus(error) {
...
+    );
+
+    return !_.isUndefined(clusterStatusWidgets);
```
kubama: To omit `!` operator I would change this line to `return _.isObject(clusterStatusWidgets);`

app/reducers/clusterStatusReducer.js
```diff
@@ -9,7 +9,7 @@ const clusterStatus = (state = {}, action) => {
                 isFetching: false,
                 error: undefined,
                 status: action.status,
-                services: action.services
+                services: action.services !== undefined ? action.services : state.services
```
kubama: Could it be simplified to `services: action.services || state.services`?

app/utils/StatusPoller.js
```diff
@@ -12,6 +13,7 @@ export default class StatusPoller {
         this._store = store;
         this._pollerTimer = null;
         this._fetchStatusPromise = null;
+        this._fetchClusterStatusPromise = null;
```
kubama: This violates `no-underscore-dangle` eslint rule. This currently gives a warning, but is listed in `.eslintrc` as a candidate for error once all occurrences are fixed. If so I think we shouldn't introduce more violations of this rule.

package.json
```diff
@@ -82,14 +82,14 @@
...
-    "@babel/register": "^7.5.5",
-    "@babel/runtime": "^7.5.5",
+    "@babel/core": "^7.7.7",
+    "@babel/plugin-proposal-class-properties": "^7.7.4",
+    "@babel/plugin-transform-runtime": "^7.7.6",
+    "@babel/polyfill": "^7.7.0",
+    "@babel/preset-env": "^7.7.7",
+    "@babel/preset-react": "^7.7.4",
+    "@babel/register": "^7.7.7",
+    "@babel/runtime": "^7.7.7",
```
kubama: Are these upgrades related to the code changes? If not I don't think it is a good idea to make such upgrades in a pre-release patch. I'm aware these are dev dependencies but as I understand they are used in project build and as such influence the whole application providing some level of risk.

app/actions/clusterStatus.js
```diff
@@ -22,15 +22,28 @@ export function errorClusterStatus(error) {
     };
 }
 
-export function getClusterStatus() {
+function isClusterStatusWidgetOnPage(pageId, pages) {
+    const currentPage = _.find(pages, page => page.id === pageId);
+    const clusterStatusWidgetDefinitionName = 'highAvailability';
+    const clusterStatusWidgets = _.find(
```
kubama: Since `_.find` returns a single object I would rather use singular form for this const: `clusterStatusWidget`

app/actions/clusterStatus.js
```diff
@@ -22,15 +22,28 @@ export function errorClusterStatus(error) {
...
+export function getClusterStatus(summaryOnly = false) {
     return (dispatch, getState) => {
-        const managerAccessor = new Manager(getState().manager);
+        const { app, manager, pages } = getState();
+        const managerAccessor = new Manager(manager);
+        const fetchOnlySummary = summaryOnly && !isClusterStatusWidgetOnPage(app.currentPageId, pages);
         dispatch(requestClusterStatus());
         return managerAccessor
-            .doGet('/cluster-status')
+            .doGet(`/cluster-status${fetchOnlySummary ? '?summary=true' : ''}`)
```
kubama: I'm not sure if it would work equally, but maybe it could be simplified to:
``.doGet(`/cluster-status?summary=${fetchOnlySummary}`)``

qooban:
@kubama Thanks for review and comments, I see the point in most of them. As for now I merged the PR in current shape, but I'm going to reply to your comments and implement requested changes after weekend, as well as update cluster status system tests.

app/actions/clusterStatus.js
qooban: Yes, `pageId` is string. Thanks for keeping me informed by that shortcuts in lodash.
Optimized according to suggestions.

app/actions/clusterStatus.js
qooban: Can be. Changed.

app/actions/clusterStatus.js
qooban: Changed.

app/reducers/clusterStatusReducer.js
qooban: I originally thought of such solution, don't remember why I changed it.
Fixed.

package.json
qooban: Yes, I had a problem with tests. After short investigation I found that the reason was in Babel's old version. I am aware of the risk.

app/actions/clusterStatus.js
qooban: Fixed.

app/actions/clusterStatus.js
qooban: Yes, you're right, it turned out that it works equally.

app/utils/StatusPoller.js
qooban: Yes, I was aware of that, but didn't want to refactor whole file. Will fix it in new PR.

PR #912, opened 2020-01-22
Save topology elements locations

kubama:
Please note the redesigned topology toolbar:
obraz

The following changes were made to the toolbar:
- updated look'n'feel to conform to overall stage styling
- removed three right-most buttons as they were not applicable to stage
- introduced buttons grouping by their function
- added three buttons related to layout handling (save, revert, do auto layout)

Minor composer update is required due to introduced topology changes.


.size-limit
```diff
@@ -29,7 +29,7 @@
     },
     {
         "name": "Application package",
-        "limit": "10 MB",
+        "limit": "10.22 MB",
```
qooban: I assume that the increase is mostly due to moving some parts of the topology into stage, right?
Did cloudify-blueprint-topology package size decrease?

package-lock.json
```diff
@@ -2916,8 +2916,7 @@
       "version": "2.7.0",
       "resolved": "https://registry.npmjs.org/acorn/-/acorn-2.7.0.tgz",
       "integrity": "sha1-q259nYhqrKiwhbwzEreaGYQz8Oc=",
-      "dev": true,
```
qooban: Which version of npm are you using?
I think we add/remove `optional` field in loop ;-) I'm not sure if that is because of `npm` version, but this is my first suspection. I'm using v6.13.6.

webpack.config.js
```diff
@@ -16,6 +16,11 @@ const getWidgetEntries = () => {
 };
 
 const rules = [
+    {
+        test: /\.js$/,
+        use: ['source-map-loader'],
```
qooban: This is to allow debugging cloudify-blueprint-topology by extracting source code from source maps?

widgets/topology/src/DataFetcher.js
```diff
@@ -1,31 +1,57 @@
...
         if (_.isEmpty(deploymentId) && _.isEmpty(blueprintId)) {
             return Promise.resolve({ data: {} });
         }
 
+        function getLayoutPromise(layoutBlueprintId) {
+            return toolbox
+                .getInternal()
+                .doGet(`/bud/layout/${layoutBlueprintId}`)
+                .catch(() =>
+                    toolbox
```
qooban: What do you think about moving that logic to the backend?
(either as a separate endpoint in `/bud` route or by enhancing `/bud/layout/:blueprintId`)

widgets/topology/src/Topology.js
```diff
@@ -210,13 +197,16 @@ export default class Topology extends React.Component {
             this._topology.setLoading(true);
             this._topologyData = null;
             this.props.toolbox.refresh();
-        } else {
+        } else if (!_.isEmpty(this.props.data.deploymentsData)) {
+            if (!this._topology) {
+                this._startTopology();
```
qooban: Will it be big effort to reduce no-underscore-dangle warning by updating properties and methods of Topology component?

widgets/topology/src/NodeStatusService.js
```diff
@@ -0,0 +1,113 @@
+import { consts, icons } from 'cloudify-ui-common';
+
+const states = [
```
qooban: I'd like we clean-up node instances status definitions. Not necessarily withing this activity.
Currently there are some definitions already in `widgets/common/src/NodeInstancesConsts.js`
I think it might be higher effort, so could you please create issue in JIRA for that? (R&D UI Improvements 5.0.5, labels: UI, stage, common) Are node instances states defined as well somewhere in cloudify-blueprint-topology? Please check.

We should probably align Node Instance Status and Execution Status handling in stage to provide similar developer experience.

backend/routes/BlueprintUserData.js
```diff
@@ -0,0 +1,29 @@
+const router = require('express').Router();
+const passport = require('passport');
+const bodyParser = require('body-parser');
+
+const db = require('../db/Connection');
+
+router.use(passport.authenticate('token', { session: false }));
+router.use(bodyParser.json());
+
+router.get('/layout/:blueprint', (req, res) => {
```
qooban: Can you review `doc/diagrams/architecture-backend.png` and `doc/diagrams/architecture-backend.png`?
At least new route (BlueprintUserData) should be added. But maybe there are some more gaps.

Check `doc/diagrams/architecture-backend.png` for details on how to update it.

test/widgets/topology/DataProcessorTest.js
```diff
@@ -0,0 +1,77 @@
+import { expect } from 'chai';
```
qooban: First widget unit tests. Great! :-)

widgets/topology/src/ExecutionsService.js
```diff
@@ -0,0 +1,12 @@
+export default class ExecutionsService {
```
qooban: There's `app/utils/shared/ExecutionUtils.js` aiming to provide utility functions for Execution resources. Can we use/enhance that one?

widgets/topology/src/NodeStatusService.js
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    static isCompleted(instance) {
+        return NodeStatusService.isStarted(instance);
+    }
+
+    static isInProgress(instance) {
+        // todo: consider using !uninitialized.. single source of truth
+        const stateNum = stateToNum(instance);
+        return stateNum > 0 && stateNum <= 7;
```
qooban: I know it's copied from cloudify-blueprint-topology, but let's fix that at least not to base on the index.

Please check eg. `app/utils/shared/ExecutionUtils.js` for possible solution.

widgets/topology/src/DataProcessor.js
```diff
@@ -0,0 +1,127 @@
...
+    const { plan } = data.data;
+
+    if (!plan) {
+        return {};
+    }
+
+    return DataProcessingService.encodeTopology(createBlueprintData(data), createHierarchy(plan));
+}
+
+let componentNodeId = 0;
```
qooban: Is it intentionally outside the `createExpandedTopology` function?
Are there any bad implications of keeping it like that? (having global index)

widgets/topology/src/DataProcessor.js
```diff
@@ -0,0 +1,127 @@
...
+        _.each(node.relationships, rel => {
+            rel.target = rel.target_id;
+        });
+    });
+
+    return { node_templates: _.keyBy(plan.nodes, 'name'), groups: plan.groups };
+}
+
+/**
+ * @description
```
qooban: Using JSDoc for exported functions is a good practice. I think we need to use it more in UI projects. Let's have some basic description for other exported functions in this file.

.size-limit
kubama: The increase is mostly due to fontawesome introduction in cloudify-blueprint-topology. So cloudify-blueprint-topology increased in size as well.

package-lock.json
kubama: I'm using 6.4.1. I'll try to update to latest node to stay in line.

qooban: For now just npm (`npm i -g npm`). 

Btw - we are planning to upgrade to Node v12 (see: https://cloudifysource.atlassian.net/browse/CY-1694)

webpack.config.js
kubama: Yes, this was of great help during development so I think it's worth having it.

widgets/topology/src/DataProcessor.js
kubama: Done

widgets/topology/src/DataProcessor.js
kubama: Yes it is intentionally outside as it is intended to be a global counter. This solution was already present in original code moved from cloudify-blueprint-topology. I can't think of any bad implications.

widgets/topology/src/ExecutionsService.js
kubama: Actually it turned out logic in this class is redundant and can be simply removed

widgets/topology/src/NodeStatusService.js
kubama: Done

widgets/topology/src/Topology.js
kubama: Not a big deal - refactored

widgets/topology/src/DataFetcher.js
kubama: Done

backend/routes/BlueprintUserData.js
kubama: Done

widgets/topology/src/NodeStatusService.js
kubama: I'm not sure if status calculation as implemented in this file can be simply replaced with map lookup. Also these statuses serve a specific purpose and doesn't seem to be compatible with what's defined in `widgets/common/src/NodeInstancesConsts.js`.
Please see refactored version of this file as it is now easier to understand what's going on.

PR #914, opened 2020-01-28
CY-1485 - Preparation before updating components documentation deployment

qooban:
- Reused styles defined in cloudify-ui-common repository
- Moved not-really basic components from Stage.Basic into Stage.Shared (new) or Stage.Common
- Removed old components' documentation
- Removed Toolbox usage from application components (MaintenanceModePageMessage, ClusterServicesList, LicensePage, GenericField, SiteLocationMap)
- Removed dependencies cycles and set up import/no-cycle to be reported as error
- Moved GenericField to cloudify-ui-components
- Removed internal components documentation and Jenkins mechanism uploading it to S3
- Updated cloudify-ui-components version 
- Updated style.scss


qooban:
@kubama Only fyi: there will be major change in components organization in cloudify-stage. PR is not ready yet, I'm still working on it.

PR #924, opened 2020-02-14
Fixes for Internet disconnected environments

kubama:
* Replaced semantic-ui css with offline version that doesn't import fts from googleapis
* Updated blueprint catalog, plugins catalog and deployment wizard widgets to handle lack of Internet connection gracefully

### DIFF
--- a/.size-limit
+++ b/.size-limit
@@ -1,14 +1,14 @@
 [
     {
         "name": "Application (gzip)",
-        "limit": "3.02 MB",
+        "limit": "8.12 MB",
         "gzip": false,
         "webpack": false,
         "path": ["dist/static/js/*.gz", "dist/appData/widgets/*/widget.js.gz", "dist/appData/widgets/*/widget.css", "dist/appData/widgets/*/widget.png", "dist/appData/widgets/*/widget.html", "dist/appData/widgets/*/README.md", "dist/appData/widgets/*/backend.js", "dist/appData/templates/**", "dist/static/images/*", "dist/static/fonts/*", "dist/static/*" ]
     },
     {
         "name": "Application (brotli)",
-        "limit": "2.71 MB",
+        "limit": "7.9 MB",
         "gzip": false,
         "webpack": false,
         "path": ["dist/static/js/*.br", "dist/appData/widgets/*/widget.js.br", "dist/appData/widgets/*/widget.css", "dist/appData/widgets/*/widget.png", "dist/appData/widgets/*/widget.html", "dist/appData/widgets/*/README.md", "dist/appData/widgets/*/backend.js", "dist/appData/templates/**", "dist/static/images/*", "dist/static/fonts/*", "dist/static/*" ]
@@ -29,7 +29,7 @@
     },
     {
         "name": "Application package",
-        "limit": "10.22 MB",
+        "limit": "15.2 MB",
         "gzip": false,
         "webpack": false,
         "path": "dist/**"
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.3.0",
     "request": "^2.88.0",
-    "semantic-ui-css": "^2.4.1",
+    "semantic-ui-css-offline": "^1.0.3-2.4.1",
     "semantic-ui-react": "0.84.0",
     "uuid": "^3.3.2"
   },
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -77,7 +77,7 @@ const rules = [
 
                 options: {
                     limit: 100000,
-                    name: '/static/fonts/[name].[ext]'
+                    name: './static/fonts/[name].[ext]'
                 }
             }
         ]
--- a/widgets/blueprintCatalog/src/widget.js
+++ b/widgets/blueprintCatalog/src/widget.js
@@ -26,7 +26,7 @@ Stage.defineWidget({
             name: 'Blueprints Examples URL',
             placeHolder: 'Type URL to blueprint examples JSON file',
             description: 'If set, then GitHub options are not used for fetching data.',
-            default: '//repository.cloudifysource.org/cloudify/blueprints/5.0.5/examples.json',
+            default: Stage.Common.Consts.externalUrls.blueprintsCatalog,
             type: Stage.Basic.GenericField.STRING_TYPE
         },
         {
--- a/widgets/blueprintCatalog/src/widget.js
+++ b/widgets/blueprintCatalog/src/widget.js
@@ -80,42 +80,60 @@ Stage.defineWidget({
             widget.configuration.jsonPath
         );
 
-        return actions.doGetRepos(params).then(data => {
-            const defaultImagePath = Stage.Utils.Url.widgetResourceUrl(
-                'blueprintCatalog',
-                Consts.DEFAULT_IMAGE,
-                false,
-                false
-            );
-            let repos = data.items;
-            const { source } = data;
-            const total = data.total_count;
-            if (data.source === Consts.GITHUB_DATA_SOURCE) {
-                const isAuthenticated = data.isAuth;
+        return actions
+            .doGetRepos(params)
+            .then(data => {
+                const defaultImagePath = Stage.Utils.Url.widgetResourceUrl(
+                    'blueprintCatalog',
+                    Consts.DEFAULT_IMAGE,
+                    false,
+                    false
+                );
+                let repos = data.items;
+                const { source } = data;
+                const total = data.total_count;
+                if (data.source === Consts.GITHUB_DATA_SOURCE) {
+                    const isAuthenticated = data.isAuth;
 
-                const fetches = _.map(repos, repo =>
-                    actions
-                        .doFindImage(repo.name, defaultImagePath)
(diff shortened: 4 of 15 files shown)


.size-limit
```diff
@@ -1,14 +1,14 @@
 [
     {
         "name": "Application (gzip)",
-        "limit": "3.02 MB",
+        "limit": "8.12 MB",
```
qooban: 5 MB, big increase. Can you provide me a list with new files? Are there just font files? Lato font?

package.json
```diff
@@ -68,7 +68,7 @@
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.3.0",
     "request": "^2.88.0",
-    "semantic-ui-css": "^2.4.1",
+    "semantic-ui-css-offline": "^1.0.3-2.4.1",
```
qooban: If we go with such change, we'll need to update cloudify-ui-components as well.

widgets/blueprintCatalog/src/widget.js
```diff
@@ -80,42 +80,60 @@ Stage.defineWidget({
...
+            })
+            .catch(_.identity);
     },
 
     render(widget, data, error, toolbox) {
+        if (data instanceof Error) {
+            const { Message } = Stage.Basic;
+            const { MessageContainer } = Stage.Shared;
+            return (
+                <MessageContainer wide margin="30px auto">
```
qooban: Maybe we can have common component for all (sites, blueprints catalog, plugins catalog, deployment wizard) of that "external connection required" cases?

The message is very similar, so maybe we can have kind of template instead of passing whole string as a prop?

As it will be used in widgets we can create it in `Stage.Common` probably.

widgets/deploymentWizardButtons/src/steps/PluginsStep.js
```diff
@@ -149,9 +150,7 @@ class PluginsStepContent extends React.Component {
                     this.props.toolbox
                         .getManager()
                         .doGet('/plugins?_include=distribution,package_name,package_version,visibility'),
-                    this.props.toolbox
-                        .getExternal()
-                        .doGet('http://repository.cloudifysource.org/cloudify/wagons/plugins.json')
+                    this.props.toolbox.getExternal().doGet(PLUGINS_CATALOG_URL)
```
qooban: 1. Could we move all the external URLs called from widgets (hello world blueprint URL, plugins catalog URL and blueprints catalog URL) into `Stage.Common`? That way they will be in one place.

2. What do you think about changing fetching plugins and blueprints in respectively Plugins Catalog and Blueprints Catalog widgets to be done using `/external/content` endpoint instead of direct call from the browser? (see https://cloudifysource.atlassian.net/browse/CY-2293 , specifically my last comment with solution suggestion in that issue)

widgets/deploymentWizardButtons/src/urls.js
```diff
@@ -0,0 +1,3 @@
+export const PLUGINS_CATALOG_URL = 'http://repository.cloudifysource.org/cloudify/wagons/plugins.json';
```
qooban: Can we have both using https?

qooban:
One more general comment - what would be the path if at some point we would like to use some specific font (not default Lato)? Probably regenerate the output using sth like https://github.com/mjasnikovs/semantic-ui-css-offline#do-you-want-newer-version-of-semantic-ui ?

.size-limit
kubama: Please see `node_modules/semantic-ui-css-offline/fonts`

kubama:
@qooban I believe so.

widgets/deploymentWizardButtons/src/urls.js
kubama: We can't due to `ERR_CERT_COMMON_NAME_INVALID`

widgets/deploymentWizardButtons/src/steps/PluginsStep.js
kubama: Ad 1) Addressed.
Ad 2) I think best solution for CY-2293 would be to enable HTTPS requests to `repository.cloudifysource.org` by fixing cert issue on server side. Proxying requests through backend to bypass broken HTTPS (by using HTTP or disabling cert validation) seems like a workaround that should be avoided.

package.json
kubama: Addressed in https://github.com/cloudify-cosmo/cloudify-ui-components/pull/28

widgets/blueprintCatalog/src/widget.js
kubama: Addressed

.size-limit
qooban: Yes, I get it https://github.com/mjasnikovs/semantic-ui-css-offline/tree/master/src/fonts.

I'm wondering if they are all loaded (I mean all different flavours - bold, italic, etc.) when you open our app? (I suspect no, but let's verify and be sure)
I'm also wondering if there is maybe a webpack plugin for compressing fonts?

widgets/common/src/Consts.js
```diff
@@ -19,6 +19,11 @@ const consts = {
             maxBoundsViscosity: 0.75
         },
         initialZoom: 2.5
+    },
+    externalUrls: {
+        pluginsCatalog: '//repository.cloudifysource.org/cloudify/wagons/plugins.json',
```
qooban: Good to have it in one place. 

I saw you already implemented https://cloudifysource.atlassian.net/browse/CY-2293 . Great!
Have you verified everything works fine when Cloudify Manager is in SSL mode?

Regarding:

> I think best solution for CY-2293 would be to enable HTTPS requests to `repository.cloudifysource.org` by fixing cert issue on server side. Proxying requests through backend to bypass broken HTTPS (by using HTTP or disabling cert validation) seems like a workaround that should be avoided.

My intention of proxying requests was to avoid CORS problem, not HTTPS related. I agree with your comment that we should have cert issue fixed first on server side. I'm not sure how quick will it happen. We can also use the URL provided by Trammell in the last comment of https://cloudifysource.atlassian.net/browse/CY-2293 . But that would require us to differentiate between HTTP/HTTPS. I'm not sure that's the right way to go.

widgets/deploymentWizardButtons/src/widget.js
```diff
@@ -59,7 +59,32 @@ Stage.defineWidget({
     ],
     permission: Stage.GenericConfig.WIDGET_PERMISSION('deploymentWizardButtons'),
 
+    fetchData(widget, toolbox) {
+        return Promise.all([
+            toolbox.getExternal().isReachable(Stage.Common.Consts.externalUrls.pluginsCatalog),
```
qooban: Shouldn't be `Stage.Common.Consts.externalUrls.helloWorldBlueprint`, here?

kubama: No, plugins catalog URL is also accessed by this widget. Hello World blueprint URL is covered by the next line.

widgets/common/src/Consts.js
kubama: > I saw you already implemented https://cloudifysource.atlassian.net/browse/CY-2293

Why do you think it is implemented?

widgets/deploymentWizardButtons/src/widget.js
qooban: Oh, I see right now, so actually https://cloudifysource.atlassian.net/browse/CY-2293 is not yet implemented. OK, good, let's finish CYBL-718 first and then update the code to fix CY-2293.

widgets/common/src/Consts.js
qooban: My bad, it is not implemented. Let's finish this one and continue CY-2293 afterwards.

.size-limit
kubama: Only regular and bold flavors are loaded on app startup.

Chrome chooses to load woff2 format which is already compressed, so no need for extra compression plugin.

webpack.config.js
```diff
@@ -77,7 +77,7 @@ const rules = [
 
                 options: {
                     limit: 100000,
-                    name: '/static/fonts/[name].[ext]'
+                    name: './static/fonts/[name].[ext]'
```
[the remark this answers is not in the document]
kubama: > Have you checked production webpack configuration?
> There's also reference to `/static/fonts/[name].[ext]`

It's a workaround for webpack dev server only, no need to touch production configuration.

> It would be worth to test production version (`npm run build` + run backend server + open in browser the content of /dist folder) if you didn't check that, yet.

Done with `npm run build && npm run zip && npm run upload` and checking with a browser - it's working correctly.

kubama:
I believe size-limit does its job as a guard against uncontrolled application size growth. When a limit is exceeded it's very easy to manually check what's the reason behind it. I don't see a need for a better tool.

qooban:
@kubama Yes, it guards size limit, but the size is not sth user is interested in after all. It was useful for a long time. Until we had only small increases (like few bytes or kilobytes) it made sense to monitor it and basing on that predict how it can affect performance. When we increased app size with 5MB such size-limit is not very useful.
I think it would be better to have performance monitoring based on more reliable metrics and metrics provided by Lighthouse looks better for me. I'll create JIRA for the future.

PR #928, opened 2020-02-28
Use /external backend endpoint as a proxy for fetching Internet resources

kubama:


backend/routes/External.js
```diff
@@ -11,7 +11,7 @@ function _pipeRequest(req, res, next, url, queryString) {
     logger.debug(`Piping get request to url: ${url} with query string: ${queryString}`);
 
     req.pipe(
-        request.get({ url, qs: queryString }).on('error', function(err) {
+        request.get({ url: url.startsWith('//') ? `http:${url}` : url, qs: queryString }).on('error', function(err) {
```
qooban: I think that, when string is started with `//` and Cloudify is running with SSL enabled, user expects that https is called.

As we moved the call to the backend, maybe we could cut `//` from the configured jsonPath string in Blueprints Catalog and Plugins Catalog widgets?

kubama: We could, but we still need to handle URLs starting with `//` - I think it's extremelly unlikely that any user changed the default URL, and it's even less likely that it was changed to anything starting with `//`, but it's quite possible that some users saved widget configuration with default URLs so now they are stored in users' DBs, so we need to handle them.

qooban: OK, I get your point and you're right about default URLs stored in users' DBs.
Still, I think that my second question can be address within this PR. WDYT?

kubama: Simply removing the slashes is not an option as the URL would lack required protocol part. I ended up by prepending the default URLs with `http` as `https` is not usable due to unresolved cert issue (which was the reason for this PR).

qooban: Yes, but hopefully cert issue will be solved (see mail thread) and I think we should have solution assuming the cert is valid.

kubama: As you wish

qooban:
@kubama Let's test https://raw.githubusercontent.com/ofercloudify/test/master/plugins.json for Plugins Catalog.

kubama:
@qooban It doesn't work as raw.githubusercontent.com doesn't set content-type correctly.
Since plugins catalog URL should be pointing to a JSON anyway it would be reasonable to force JSON parsing in such cases.

qooban:
@kubama Agree. Let's change that to force JSON parsing in both - Blueprints and Plugins Catalogs.

kubama:
"There are no Plugins available in catalog. Check widget's configuration."

qooban:
@kubama Good 👍 Let's merge.

kubama:
Actually, after further checking it can be:
"The widget content cannot be displayed because there is no connection to plugins repository. Please check network connection and widget's configuration." - for 2XX response
"There are no Plugins available in catalog. Check widget's configuration." - for other responses

qooban:
@kubama Hmm... Looks like that current errors messages can be not very accurate. What do you suggest to do with that?

kubama:
I customized the message - please review

PR #930, opened 2020-03-02
Added confirmation popup for layout save operation

kubama:
obraz

### DIFF
--- a/widgets/topology/src/Topology.js
+++ b/widgets/topology/src/Topology.js
@@ -30,7 +30,7 @@ export default class Topology extends React.Component {
         this.processedTopologyData = null;
         this.scrollerGlassHandler = new ScrollerGlassHandler(this.glassRef);
 
-        this.state = {};
+        this.state = { saveConfirmationOpen: false };
     }
 
     shouldComponentUpdate(nextProps, nextState) {
@@ -270,7 +270,6 @@ export default class Topology extends React.Component {
                     content="Topology layout saved"
                     position="top center"
                     style={{ left: 'unset', right: 65 }}
-                    on=""
                     trigger={<div id="topologyContainer" />}
                 />
             </div>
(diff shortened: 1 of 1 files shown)


widgets/topology/src/Topology.js
```diff
@@ -85,6 +87,10 @@ export default class Topology extends React.Component {
                         null,
                         layout
                     )
+                    .then(() => {
+                        this.setState({ saveConfirmationOpen: true });
+                        setTimeout(() => this.setState({ saveConfirmationOpen: false }), 2500);
```
qooban: Let's have const for `2500`. I think in the future we may want to move it a bit "higher" and reuse in other quick notification solutions like here.

widgets/topology/src/Topology.js
```diff
@@ -259,7 +265,13 @@ export default class Topology extends React.Component {
                 <div className="scrollGlass" ref={this.glassRef}>
                     <span className="message">Click to release scroller</span>
                 </div>
-                <div id="topologyContainer" />
+                <Stage.Basic.Popup
```
qooban: Cosmetic issue. Can we follow the convention from most of the other widgets to first grab component from `Stage.*` and then use with the exact name?

widgets/topology/src/Topology.js
```diff
@@ -259,7 +265,13 @@ export default class Topology extends React.Component {
...
+                    open={this.state.saveConfirmationOpen}
+                    content="Topology layout saved"
+                    position="top center"
+                    style={{ left: 'unset', right: 65 }}
```
qooban: `65` is the pixel-distance from the right side of the buttons pane (`topologyContainer` div)?
I suspect it would be hard to "associate" the popup with the proper button from the toolbar and remove additional styling, right?

widgets/topology/src/Topology.js
kubama: Done

widgets/topology/src/Topology.js
kubama: Done

widgets/topology/src/Topology.js
kubama: Yes, it would be very difficult since topology is an independent non-react component

PR #932, opened 2020-03-03
Updated execution graph look-and-feel

kubama:
image

### DIFF
--- a/widgets/executions/src/backend.js
+++ b/widgets/executions/src/backend.js
@@ -158,7 +158,7 @@ module.exports = r => {
             const allSubgraphs = {};
             _.map(operationsList, task => {
                 let taskName = _.split(task.name, 'cloudify.interfaces.');
-                taskName = taskName.length > 1 ? taskName[1] : taskName[0];
+                taskName = taskName.length > 1 ? taskName[1] : _.upperFirst(taskName[0]);
 
                 let taskOperation = '';
                 if (
--- a/widgets/executions/src/backend.js
+++ b/widgets/executions/src/backend.js
@@ -179,7 +179,7 @@ module.exports = r => {
                             text: taskName,
                             retry: 0,
                             type: task.type,
-                            state: task.state,
+                            state: _.upperFirst(task.state),
                             operation: taskOperation,
                             display_text: ''
                         }
--- a/widgets/executions/src/backend.js
+++ b/widgets/executions/src/backend.js
@@ -91,9 +91,9 @@ module.exports = r => {
             children: [],
             edges: []
         };
-        const paddingLeftRight = 12;
-        const paddingTop = 20;
-        const paddingBottom = 16;
+        const paddingLeftRight = 24;
+        const paddingTop = 46;
+        const paddingBottom = 28;
         const subGraphLayoutOptions = {
             'elk.padding': `[top=${paddingTop},left=${paddingLeftRight},bottom=${paddingBottom},right=${paddingLeftRight}]`
         };
@@ -102,7 +102,7 @@ module.exports = r => {
         // This will be used when a text needs to be displayed inside a node (rectangle)
         // and exceeds its width, resulting in increasing the rectangle's height and
         // breaking the string into 2 (and so forth...)
-        const textSizingFactor = 5.6;
+        const textSizingFactor = 5.8;
         const textHeight = 18;
 
         const runGraphCreation = () => {
@@ -378,6 +378,7 @@ module.exports = r => {
                     if (numberOfSplits > 0) {
                         subGraph.height += textHeight * numberOfSplits;
                     }
+                    subGraph.height += 10;
                 }
             });
             return allSubgraphs;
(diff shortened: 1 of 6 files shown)


widgets/executions/widget.css
```diff
@@ -18,29 +18,27 @@
...
 }
 
 .rect-tasks-graph-succeeded {
-    fill: rgba(25, 200, 54, 0.836);
+    fill: rgb(39, 230, 109);
 }
 
 .rect-tasks-graph-failed {
-    fill: rgba(204, 50, 50, 0.822)
+    fill: rgb(204, 50, 50);
```
qooban: From where did you take the colors?

widgets/executions/src/backend.js
```diff
@@ -102,7 +102,7 @@ module.exports = r => {
         // This will be used when a text needs to be displayed inside a node (rectangle)
         // and exceeds its width, resulting in increasing the rectangle's height and
         // breaking the string into 2 (and so forth...)
-        const textSizingFactor = 5.6;
+        const textSizingFactor = 5.8;
```
qooban: That change is to prevent overlapping the rectangle borders by the text, right? Will it cover all cases?

kubama: It won't cover all cases, it just reduces the possibility.
To cover all cases we would need to set this const to the width of broadest character. But then in an average case we would split the lines very early, in about half of the available line length.

widgets/executions/widget.css
kubama: From WebStorm color picker :-)

qooban: Can we adjust it to be aligned with Semantic-UI colors (main green, yellow, red, ...)?

kubama: I find the colors that I used a little bit better suited, for example "official" red and green are to bright and intense and don't look that good when used as a fill for rects that might be much bigger than Semantic components like buttons.

qooban: OK, then let's keep it.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -114,7 +114,7 @@ export default class ExecutionWorkflowGraph extends React.Component {
     scrollToInProgress() {
         const focusNode = _.find(this.state.graphResult.children, containerNode =>
             _.find(containerNode.children, subGraphNode =>
-                _.includes(['sent', 'rescheduled'], subGraphNode.labels[0].state)
+                _.includes(['Sent', 'Rescheduled'], subGraphNode.labels[0].state)
```
qooban: Let's have an array for all of that task states.
AFAIR we should add 'started' here as well.

kubama: > Let's have an array for all of that task states.

We already have an array.

> AFAIR we should add 'started' here as well.

Added.

widgets/executions/src/tasksGraph/GraphNode.js
```diff
@@ -4,64 +4,87 @@
...
+    let currentTextPlacement_Y = 0;
+
+    const title = labels.display_title || [labels.text];
+    const displayText = labels.display_text;
+
+    const { state } = labels;
+    let stateColor;
+    switch (state) {
+        case 'Rescheduled':
+        case 'Sent':
```
qooban: What about having `started` here as well?

widgets/executions/src/tasksGraph/GraphNode.js
```diff
@@ -4,64 +4,87 @@
...
+            stateColor = 'rgb(215,227,45)';
```
qooban: I would suggest to assign those strings to some `const`'s and name it after color it represents.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -115,7 +116,7 @@ export default class ExecutionWorkflowGraph extends React.Component {
     scrollToInProgress() {
         const focusNode = _.find(this.state.graphResult.children, containerNode =>
             _.find(containerNode.children, subGraphNode =>
-                _.includes(['sent', 'rescheduled'], subGraphNode.labels[0].state)
+                _.includes(['Started', 'Sent', 'Rescheduled'], subGraphNode.labels[0].state)
```
qooban: By array I meant having `const` with those values.

widgets/executions/src/tasksGraph/GraphNode.js
kubama: Of course - good catch

widgets/executions/src/tasksGraph/GraphNode.js
kubama: Addressed

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
kubama: Addressed

widgets/executions/src/tasksGraph/States.js
```diff
@@ -0,0 +1,5 @@
+export default {
+    inProgress: ['Started', 'Sent', 'Rescheduled'],
+    succeeded: ['Succeeded'],
+    failed: ['Failed']
+};
```
qooban: That makes it clear now. :+1:

kubama:
> Could you please add some recent screenshots?

Done

PR #934, opened 2020-03-06
Enhanced task graph to allow automatic scrolling and focusing on tasks in progress

kubama:
New icon was placed near existing "open in window" icon:

obraz

### DIFF
--- a/widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
+++ b/widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
@@ -1,13 +1,16 @@
 /**
  * Created by barucoh on 23/1/2019.
  */
-import { UncontrolledReactSVGPanZoom } from 'react-svg-pan-zoom';
+import { ReactSVGPanZoom } from 'react-svg-pan-zoom';
 import GraphNodes from './GraphNodes';
 import GraphEdges from './GraphEdges';
 
 const POLLING_INTERVAL = 5000;
 const MAX_GRAPH_HEIGHT = 380;
-const GRAPH_VERTICAL_MARGIN = 15;
+const GRAPH_MARGIN = 15;
+
+const AUTO_FOCUS_ANIMATION_FRAMES = 30;
+const AUTO_FOCUS_ANIMATION_FRAME_DURATION = 15;
 
 export default class ExecutionWorkflowGraph extends React.Component {
     /**
@@ -24,7 +27,9 @@ export default class ExecutionWorkflowGraph extends React.Component {
             error: '',
             maximized: false,
             containerWidth: 0,
-            modalWidth: 0
+            modalWidth: 0,
+            position: {},
+            modalPosition: {}
         };
         this.timer = null;
         this.cancelablePromise = null;
@@ -60,6 +65,9 @@ export default class ExecutionWorkflowGraph extends React.Component {
                     this.setState({
                         graphResult: tasksGraph
                     });
+                    if (this.state.autoFocus) {
+                        this.scrollToInProgress();
+                    }
                 }
             })
             .catch(error => {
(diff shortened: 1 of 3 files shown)


widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -86,53 +94,114 @@ export default class ExecutionWorkflowGraph extends React.Component {
...
+            [maximized ? 'modalPosition' : 'position']: { ...currentPosition, ...positionToFocusOn }
+        });
+        if (frame !== AUTO_FOCUS_ANIMATION_FRAMES && this.state.autoFocus)
+            setTimeout(() => this.scrollTo(x, y, frame + 1), AUTO_FOCUS_ANIMATION_FRAME_DURATION);
+    }
+
+    scrollToInProgress() {
+        const focusNode = _.find(this.state.graphResult.children, containerNode =>
+            _.find(containerNode.children, subGraphNode =>
+                _.includes(['sent', 'rescheduled'], subGraphNode.labels[0].state)
```
qooban: These are "starting" states for a task in the graph?
Have you consulted the states with someone?

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -86,53 +94,114 @@ export default class ExecutionWorkflowGraph extends React.Component {
...
-                            background: 'white',
-                            opacity: 1,
-                            display: 'inline-table',
-                            padding: '0 2px'
+                            padding: '0 2px',
+                            marginRight: 0
                         }}
-                        onClick={() => this.setState({ maximized: true })}
+                        onClick={() => this.setState({ autoFocus: true }, this.scrollToInProgress)}
+                        title="Focus on tasks in progress"
```
qooban: I'm thinking of having the same approach for tooltips in the whole project. Right now we are mixing browser-rendered tooltips (eg. provided by `title` prop to icon) and browser-independant SUIR Popup wrapping.

1. If we go with the first one - I'd like us to style it (there's example: https://jsfiddle.net/8out2fey/)
2. If we go with SUIR - as Popup is very generic, the second approach can be a bit more developer-friendly if we have Tooltip component which just implements Popup with necessary stuff to behave like tooltip.

What do you think?

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -86,53 +94,114 @@ export default class ExecutionWorkflowGraph extends React.Component {
...
+                    width={width}
+                    height={height}
+                    background="#fff"
+                    tool="pan"
+                    miniatureProps={{ position: 'none' }}
+                    toolbarProps={{ position: 'none' }}
+                    value={this.state[positionStateProp]}
+                    onChangeValue={position => {
+                        this.setState({ [positionStateProp]: position });
+                        if (!_.isMatch(this.state[positionStateProp], _.pick(position, ['a', 'b', 'c', 'd', 'e', 'f'])))
```
qooban: I'd like to get more information on how that positioning works. Let's discuss f2f.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
kubama: These are "in progress" states

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
kubama: I haven't investigated it deeply but my first impression was that `title` was more commonly used across the project so I chose it.
I'm fine with the first approach - it's simpler to use and more generic - can be less problematic in some non-standard situations. Whichever we choose we should address the change in a separate JIRA ticket.

PR #936, opened 2020-03-11
Added minimap and "fit to view" toolbar button

kubama:
Updated toolbar look:
obraz

### DIFF
--- a/widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
+++ b/widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
@@ -6,7 +6,8 @@ import GraphNodes from './GraphNodes';
 import GraphEdges from './GraphEdges';
 
 const POLLING_INTERVAL = 5000;
-const MAX_GRAPH_HEIGHT = 380;
+const MAX_TABLE_GRAPH_HEIGHT = 380;
+const MIN_MODAL_GRAPH_HEIGHT = 300;
 const GRAPH_MARGIN = 15;
 
 const AUTO_FOCUS_ANIMATION_FRAMES = 30;
@@ -94,21 +95,21 @@ export default class ExecutionWorkflowGraph extends React.Component {
         return this.props.widgetBackend.doGet('get_tasks_graph', { ...tasksGraphParams });
     }
 
-    scrollTo(x, y, frame = 1) {
+    scrollTo(x, y, zoom = 1, autoFocusOnly = true, frame = 1) {
         const { maximized, modalPosition, position } = this.state;
         const currentPosition = maximized ? modalPosition : position;
         const positionToFocusOn = {
             // See https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#Matrix
             e: currentPosition.e - ((currentPosition.e - x) * frame) / AUTO_FOCUS_ANIMATION_FRAMES,
             f: currentPosition.f - ((currentPosition.f - y) * frame) / AUTO_FOCUS_ANIMATION_FRAMES,
-            a: currentPosition.a - ((currentPosition.a - 1) * frame) / AUTO_FOCUS_ANIMATION_FRAMES,
-            d: currentPosition.d - ((currentPosition.d - 1) * frame) / AUTO_FOCUS_ANIMATION_FRAMES
+            a: currentPosition.a - ((currentPosition.a - zoom) * frame) / AUTO_FOCUS_ANIMATION_FRAMES,
+            d: currentPosition.d - ((currentPosition.d - zoom) * frame) / AUTO_FOCUS_ANIMATION_FRAMES
         };
         this.setState({
             [maximized ? 'modalPosition' : 'position']: { ...currentPosition, ...positionToFocusOn }
         });
-        if (frame !== AUTO_FOCUS_ANIMATION_FRAMES && this.state.autoFocus)
-            setTimeout(() => this.scrollTo(x, y, frame + 1), AUTO_FOCUS_ANIMATION_FRAME_DURATION);
+        if (frame !== AUTO_FOCUS_ANIMATION_FRAMES && (this.state.autoFocus || !autoFocusOnly))
+            setTimeout(() => this.scrollTo(x, y, zoom, autoFocusOnly, frame + 1), AUTO_FOCUS_ANIMATION_FRAME_DURATION);
     }
 
     scrollToInProgress() {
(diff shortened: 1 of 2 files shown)


widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -94,21 +95,21 @@ export default class ExecutionWorkflowGraph extends React.Component {
...
-            d: currentPosition.d - ((currentPosition.d - 1) * frame) / AUTO_FOCUS_ANIMATION_FRAMES
+            a: currentPosition.a - ((currentPosition.a - zoom) * frame) / AUTO_FOCUS_ANIMATION_FRAMES,
+            d: currentPosition.d - ((currentPosition.d - zoom) * frame) / AUTO_FOCUS_ANIMATION_FRAMES
         };
         this.setState({
             [maximized ? 'modalPosition' : 'position']: { ...currentPosition, ...positionToFocusOn }
         });
-        if (frame !== AUTO_FOCUS_ANIMATION_FRAMES && this.state.autoFocus)
-            setTimeout(() => this.scrollTo(x, y, frame + 1), AUTO_FOCUS_ANIMATION_FRAME_DURATION);
+        if (frame !== AUTO_FOCUS_ANIMATION_FRAMES && (this.state.autoFocus || !autoFocusOnly))
```
qooban: You can probably destructure `autoFocus` from `this.state` at the beggining of `scrollTo` method.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -122,39 +123,38 @@ export default class ExecutionWorkflowGraph extends React.Component {
...
+                                    height / (this.state.graphResult.height + GRAPH_MARGIN)
+                                ),
+                                false
+                            )
+                        }
+                        title="Fit to view"
+                    />
                     {openInModalIcon && (
                         <Icon
                             name="expand"
```
qooban: I'm not sure it is the right icon now.
1. We are using the same icon for maximizing widgets. 
2. When we introduced "Fit to view" icon I think users can get confused.

Maybe for "Open in window" we could use sth like: 
image
from SUIR set of icons? It's not perfect, I know, can be associated with external links. Perfect would to have "copy"-like icon with use of "window" and "external".

For an inspiration - check e.g. http://shorturl.at/oswKS

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -192,16 +192,23 @@ export default class ExecutionWorkflowGraph extends React.Component {
                 <div ref={this.wrapper} style={{ position: 'relative' }}>
                     {this.renderGraph(
                         Math.max(0, this.state.containerWidth - 1),
-                        Math.min(MAX_GRAPH_HEIGHT, height),
-                        'position',
-                        true
+                        Math.min(MAX_TABLE_GRAPH_HEIGHT, height),
+                        'position'
                     )}
                     <Modal
```
qooban: 1. Maybe it would be good that add "X" close icon (see `closeIcon` prop) as we have in some places. There is no control button in this modal and from the accessibility and UX perspective IMHO we should have clear way to close the modal instead of just clicking on the black background or pressing Esc.

2. When you start dragging the graph and you end up with mouse outside the modal it get closed.
I know it is issue in all modals, but can you check if it is possible to solve it quickly?
Probably if you turn off `closeOnDimmerClick` and add `closeIcon` it solves the issue.

kubama: Done - added close icon and disabled `closeOnDimmerClick` - dragging issue is solved

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
kubama: Done

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
kubama: I'm not sure about this one. I intentionally chose it to be same as widget maximize because this is what the button does - it maximizes the view.
The "external" icon suggests that the view will open in a separate browser tab or window.
I wouldn't consider the modal view as "external" as users can't switch between the modal and the "base" view without closing the modal. The way it works is closer to the idea of maximizing.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -122,50 +123,57 @@ export default class ExecutionWorkflowGraph extends React.Component {
...
                             name="expand"
                             link
-                            style={{
-                                marginRight: 0
-                            }}
                             onClick={() => this.setState({ maximized: true })}
                             title="Open in window"
                         />
+                    ) : (
+                        <Icon
```
qooban: You didn't want to use built-in `closeIcon` Modal prop?

It would be simpler to maintain and closer to other modals we have in stage.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
```diff
@@ -122,50 +123,57 @@ export default class ExecutionWorkflowGraph extends React.Component {
...
+                            name="close"
+                            link
+                            onClick={() => this.setState({ maximized: false })}
+                            style={{ fontSize: '1.25em', marginTop: -2 }}
```
qooban: On my laptop - Firefox on Linux it looks a bit too close to top border:
image

I would also avoid manipulating `fontSize` for icons. Browsers can render it in a different way.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
qooban: OK, let's leave it as it is.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
kubama: It would be much harder to style it to be in line with rest of the icons (auto focus and fit to view)

qooban: I see the problem right now.

I thought that adding `closeIcon` will give you the following:
image
but it looks like that when `size='fullscreen'` we get:
image
which indeed would be much harder to style.

Let's keep your solution. Just make sure it is displayed properly in common browsers.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.js
kubama: > On my laptop - Firefox on Linux it looks a bit too close to top border

Updated widget.css doesn't seem to be loaded into your browser. Try to restart the backend and flush browser cache.

> I would also avoid manipulating `fontSize` for icons. Browsers can render it in a different way.

The close icon is noticeably smaller than the other icons when using same font size. So the font size needs to be increased for this one to look proportionally.

qooban: No, cache is not a problem. It just get rendered differently on different browsers/systems.
But... OK, let's leave it as it is. It's minor issue we can live with.

PR #939, opened 2020-03-16
Added "skip plugins validation" hint

kubama:


widgets/common/src/BlueprintActions.js
```diff
@@ -43,7 +43,19 @@ class BlueprintActions {
...
-        return this.toolbox.getManager().doPut(`/deployments/${deploymentId}`, null, data);
+        return this.toolbox
+            .getManager()
+            .doPut(`/deployments/${deploymentId}`, null, data)
+            .catch(err =>
+                Promise.reject(
+                    err.code === 'deployment_plugin_not_found'
+                        ? {
+                              ...err,
+                              message: `${err.message}. Install the plugin or use "Skip plugins validation" option.`
```
qooban: I see you are concatenating strings, so the error message will be:
_Use 'cfy plugins upload' to upload the missing plugins to the Manager, or use 'cfy deployments create' with the '--skip-plugins-validation' flag to skip this validation. Install the plugin or use "Skip plugins validation" option._ 
?

If so, I think we can leave just the part you added. As we use in UI/CLI the word "Upload" in relation to installing plugins and there can be multiple plugins missing, I think we can have something like: "Upload missing plugins to the Manager or use "Skip plugins validation" option."

Or maybe there's information about which plugins are missing? That we can expose to the user as well.

kubama: The actual message after implementing this change is:
Required plugin cloudify-aws-plugin, version 2.3.0 is not installed on the manager. Install the plugin or use "Skip plugins validation" option.

qooban: Summing up - message looks fine. Just check how multiple plugins are handled and adapt message.

kubama: No change is needed - in case multiple plugins are missing manager returns error for single plugin only.

qooban: Got. Let's merge.

PR #941, opened 2020-03-18
CY-2291 Tours fixed and covered with cypress tests

kubama:


test/cypress/integration/tours_spec.js
```diff
@@ -0,0 +1,45 @@
+import _ from 'lodash';
+
+describe('Tours', () => {
```
qooban: Can you add similar test for non-admin user?

I know the tours for `default` user role are subset of `adimin` user role tours, but they are in separate files, so still this is error prone and adding a similar test should be quite easy.

test/cypress/integration/tours_spec.js
```diff
@@ -0,0 +1,45 @@
...
+        cy.get('#toursButton').click();
+    });
+
+    function checkTourSteps(tourName, numberOfSteps) {
+        cy.contains(tourName).click();
+        _.times(numberOfSteps - 1, () => cy.contains('button:not(.loading)', 'Next').click());
+        cy.contains('Done');
+    }
+
+    it('Cloudify Console Overview', () => {
```
qooban: Probably the more accurate name instead of just "Cloudify Console Overview" would be "provides Cloudify Console Overview tour for admin user". It refers to all test cases.

test/cypress/integration/tours_spec.js
kubama: Done

test/cypress/integration/tours_spec.js
kubama: Done

PR #945, opened 2020-03-25
CY-2485 Update jenkins stage

mabuaisha:


Jenkinsfile
```diff
@@ -1,7 +1,11 @@
 pipeline {
     agent { label 'web-ui' }
+    parameters {
+        choice(choices: "v12.16.1\nv10.14.0", description: 'NodeJS Version', name: 'NODEJS_VERSION')
```
qooban: I think that we used 10.15.x. We used such version in CircleCI as well.
Can we have that one as an option instead of 10.14.0?

Jenkinsfile
```diff
@@ -17,6 +21,14 @@ pipeline {
                 step([$class: 'WsCleanup'])
             }
         }
+        stage('Set NodeJS Version') {
+            steps {
+                sh '''#!/usr/bin/env bash
+                    source ${JENKINS_HOME}/.profile
+                    nvm use ${NODEJS_VERSION}
```
qooban: I'm going to add `.nvmrc` file to every UI project (see eg. https://github.com/cloudify-cosmo/cloudify-stage/pull/944). 
Maybe we can just execute:
`nvm install` in the main project directory? That way nvm:
* if version is already installed just switch to the specified version
* if version is not installed yet install it and switch to it.

What do you think?

Jenkinsfile
mabuaisha: @qooban when checked the version of the installed nodejs before installing then nvm it was `10.14.0` so that the reason why I update this to `10.14.0` but its ok I can update this to  `10.15.x`

Jenkinsfile
mabuaisha: If this is what you are going to do, then not sure if  `NODEJS_VERSION` will be needed as the version will be updated on the repo itself inside the `.nvmrc`

mabuaisha: then I think this can be updated as the following 
```
        stage('Set NodeJS Version') {
            steps {
                sh '''#!/usr/bin/env bash
                    source ${JENKINS_HOME}/.profile
                    nvm install
                '''
            }
        }
```

That way we just invoke the nvm and it parse the version from `.nvmrc`

Jenkinsfile
mabuaisha: and this section will be removed 
```
    parameters {
        choice(choices: "v12.16.1\nv10.14.0", description: 'NodeJS Version', name: 'NODEJS_VERSION')
    }
```

Jenkinsfile
mabuaisha: and this section will be removed 
```
    parameters {
        choice(choices: "v12.16.1\nv10.14.0", description: 'NodeJS Version', name: 'NODEJS_VERSION')
    }
```

qooban: Yes, makes sense.
Right, dropdown will not be needed. Sorry for confusion. Originally I wasn't aware of `.nvmrc`, just learned about it today.

Jenkinsfile
qooban: OK, got it. No issue then. So there was small mismatch between Jenkins and CircleCI.

PR #948, opened 2020-04-03
Added support for displaying plugin icons

kubama:


widgets/topology/src/Topology.js
```diff
@@ -84,63 +87,87 @@ export default class Topology extends React.Component {
...
+                topology.groups.push.apply(topology.groups, expandedTopology.groups);
+                topology.nodes.push.apply(topology.nodes, expandedTopology.nodes);
+
+                _.pull(deploymentsToExpand, deploymentId);
+            });
+
+            return deploymentsToExpand.length;
+        };
+
+        const deployemntsToExpand = [...this.state.expandedDeployments];
```
qooban, line 160: Minor typo - should be `deploymentsToExpand`.

widgets/topology/src/pluginsData.json
```diff
@@ -0,0 +1,82 @@
...
+    {
+        "icon": "https://cloudify.co/wp-content/uploads/2019/08/oslogo.png",
+        "name": "cloudify-openstack-plugin",
+        "title": "OpenStack v3"
+    },
+    {
+        "icon": "https://cloudify.co/wp-content/uploads/2019/08/oslogo.png",
+        "name": "cloudify-openstack-plugin",
+        "title": "OpenStack"
+    },
```
qooban, line 11: I know this file is temporary, but we can probably remove `OpenStack v3` and keep only `OpenStack`.

PR #952, opened 2020-04-16
CY-2588 Added plugin icon to plugins list widget

kubama:
image

### DIFF
--- a/widgets/common/src/PluginIcon.js
+++ b/widgets/common/src/PluginIcon.js
@@ -1,0 +1,19 @@
+import RoleModal from './RoleModal';
+
+export default function PluginIcon({ src }) {
+    const { Image, Icon } = Stage.Basic;
+    return src ? <Image src={src} width="25" /> : <Icon name="plug" size="large" />;
+}
+
+PluginIcon.propTypes = {
+    src: PropTypes.string
+};
+
+PluginIcon.defaultProps = {
+    src: null
+};
+
+Stage.defineCommon({
+    name: 'PluginIcon',
+    common: PluginIcon
+});
--- a/widgets/common/src/PluginIcon.js
+++ b/widgets/common/src/PluginIcon.js
@@ -1,5 +1,3 @@
-import RoleModal from './RoleModal';
-
 export default function PluginIcon({ src }) {
     const { Image, Icon } = Stage.Basic;
     return src ? <Image src={src} width="25" /> : <Icon name="plug" size="large" />;
--- a/widgets/plugins/src/PluginsTable.js
+++ b/widgets/plugins/src/PluginsTable.js
@@ -119,9 +119,9 @@ export default class extends React.Component {
 
     render() {
         const NO_DATA_MESSAGE = 'There are no Plugins available. Click "Upload" to add Plugins.';
-        const { Button, DataTable, ErrorMessage, ResourceVisibility, Image, Icon } = Stage.Basic;
+        const { Button, DataTable, ErrorMessage, ResourceVisibility } = Stage.Basic;
         const { IdPopup } = Stage.Shared;
-        const { DeleteConfirm, UploadPluginModal } = Stage.Common;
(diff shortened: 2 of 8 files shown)


widgets/plugins/src/widget.js
```diff
@@ -17,12 +17,43 @@ Stage.defineWidget({
     categories: [Stage.GenericConfig.CATEGORY.SYSTEM_RESOURCES],
 
     initialConfiguration: [Stage.GenericConfig.POLLING_TIME_CONFIG(30), Stage.GenericConfig.PAGE_SIZE_CONFIG()],
-    fetchUrl:
-        '[manager]/plugins?_include=id,package_name,package_version,supported_platform,distribution,distribution_release,uploaded_at,created_by,visibility[params]',
-    fetchParams: (widget, toolbox) =>
```
qooban, line 22: We can still use `fetchParams` (no matter if you use `fetchUrl` or `fetchData`). 

Just add `params` to `fetchData` as the last argument.
Take a look at eg. `blueprintInfo` or `blueprintSources` widgets or [Widget Definition section in Writing Widgets docs](https://docs.cloudify.co/5.0.5/developer/writing_widgets/widget-definition/#fetchdatawidget-toolbox-params) if you need more details.

widgets/plugins/src/PluginsTable.js
```diff
@@ -164,6 +165,13 @@ export default class extends React.Component {
                                 <DataTable.Data>
                                     <IdPopup selected={item.id === this.state.hoveredPlugin} id={item.id} />
                                 </DataTable.Data>
+                                <DataTable.Data>
+                                    {item.icon ? (
+                                        <Image src={item.icon} width="25" />
+                                    ) : (
+                                        <Icon name="plug" size="large" />
+                                    )}
```
qooban, line 173: Maybe we should share some code with Plugins Catalog widget for displaying the icon?

kubama: This code is not common with Plugins Catalog

widgets/plugins/src/widget.js
kubama: What would be the benefit over the way it is implemented right now?

widgets/plugins/src/PluginsTable.js
qooban: So what is displayed when there is no icon provided in Plugins Catalog?

widgets/plugins/src/widget.js
qooban: Clear and logical separation of the parameters passed to fetching function.

widgets/plugins/src/PluginsTable.js
kubama: There was never any special handling for such case - I believe it was not necessary because catalog data contains icons for all the plugins.

qooban: Catalog data is not hardcoded, you can provide your JSON in widgets' configuration, so I assume it is worth to update Plugins Catalog to handle it properly.

kubama: Done

widgets/plugins/src/widget.js
kubama: Hm... it's hard to agree.
`fetchData` makes an explicit API call. Having the call params implicitly passed makes the code no clearer - actually it's the opposite - it makes it harder to understand and maintain.
`fetchParams` makes sense when used with `fetchUrl`, but not with `fetchData`.

qooban: Let's leave it then.

widgets/common/src/PluginIcon.js
```diff
@@ -0,0 +1,19 @@
+import RoleModal from './RoleModal';
```
qooban, line 1: Not used, right?

kubama: No idea where it came from 😃. Fixed.

PR #954, opened 2020-04-17
CY-2630 - Added support for displaying Terraform module details

kubama:
New modal look:
image


widgets/topology/src/Topology.js
```diff
@@ -82,8 +82,9 @@ export default class Topology extends React.Component {
             onNodeSelected: node => this.setSelectedNode(node),
             onDataProcessed: data => (this.processedTopologyData = data),
             onDeploymentNodeClick: deploymentId => this.goToDeploymentPage(deploymentId),
-            onExpandClick: (deploymentId, nodeId) => this.markDeploymentsToExpand(deploymentId, nodeId),
+            onExpandClick: deploymentId => this.markDeploymentsToExpand(deploymentId),
             onCollapseClick: deploymentId => this.collapseExpendedDeployments(deploymentId),
+            onTerraformDetailsClick: node => this.setState({ terraformDetails: node.templateData.terraformResources }),
```
qooban, line 87: We are adding more and more specific handlers here. I think at some point (possibly during migration to React with cloudify-blueprint-topology) maybe we should consider refactoring that to be more configurable from the topology user perspective. For example if c-b-t will expose Topology component we could provide a prop containing list of actions with icons, its placement and callback function.

widgets/topology/src/Topology.js
```diff
@@ -295,6 +280,34 @@ export default class Topology extends React.Component {
                     style={{ left: 'unset', right: 65 }}
                     trigger={<div id="topologyContainer" />}
                 />
+                <Modal open={this.state.terraformDetails} onClose={() => this.setState({ terraformDetails: null })}>
```
qooban, line 283: Hmm... It's very specific for Terraform node inside very generic code. Can we create separate component for the modal ?

widgets/topology/src/Topology.js
kubama: Looking at what we have at the moment the idea seems good as all the "action icons" are specific to deployment topology used only in stage - this way we would move some stage specific stuff to component configuration made on stage side.

widgets/topology/src/Topology.js
kubama: Done

test/cypress/integration/topology_spec.js
```diff
@@ -63,9 +66,50 @@ describe('Topology', () => {
...
+
+        // Install the deployment
+        cy.contains('Execute workflow').click();
+        cy.contains('Default workflows').click();
+        cy.contains('.menu a', 'Install').click();
+        cy.contains('.modal .button', 'Execute').click();
+        cy.get('.executionsTable tr:eq(2)');
+        cy.get('.executionsTable tr:eq(1)').contains('completed');
+
+        cy.reload();
```
qooban, line 80: Wouldn't it be better to use REST API for that? Checkout `executeWorkflow` in `support/executions.js`. Not used yet, but should simplify this part. Regarding waiting for completion, you can create something similar to `waitUntilEmpty` from `support/resource_commons.js`.

Nightwatch tests are written this way (do all stuff in UI), but I don't think that it was good approach.

kubama: > Wouldn't it be better to use REST API for that? Checkout `executeWorkflow` in `support/executions.js`. Not used yet, but should simplify this part.

Yes, haven't noticed it was there. Updated.
> Regarding waiting for completion, you can create something similar to `waitUntilEmpty` from `support/resource_commons.js`.

In this case execution completion is reported right on the page on which the test performs, so it's super easy and straightforward to check as it requires no navigation or clicks.

PR #955, opened 2020-04-17
CY-2577 Added field in plugin upload form to provide an icon file

kubama:
image

image


kubama:
> I see you moved props related to file resources (eg. `wagonFile`, `yamlFile`) fully to state.

Not really. I only removed props that were not used, nothing more.

PR #956, opened 2020-04-20
CY-2235 Extended execution tasks graph to include retry count and logs refinement button

kubama:
See https://cloudifysource.atlassian.net/browse/CY-2235 for screenshots

### DIFF
--- a/.size-limit
+++ b/.size-limit
@@ -1,7 +1,7 @@
 [
     {
         "name": "Application (gzip)",
-        "limit": "8.5 MB",
+        "limit": "8.6 MB",
         "gzip": false,
         "webpack": false,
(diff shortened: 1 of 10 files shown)


kubama:
> Make context changing better by improving Resource Filter (Node Instance ID) and Events/Logs Filter (Operation string) widgets. They are changing the context, so I think they should also show the value from the context. When values in either Resource Filter or Events/Logs Filter are changed then context should be updated as well.

That's reasonable - will do.

> What do you think about marking the selected part in execution task graph? (similarly to what we do in Topology widget by changing the node's border color) It should be reset when node_instance_id or operation_type is changed.

We decided to go with a clickable icon to stress out what action is available - so here we are rather performing an action over the node rather then selecting a node. So I don't think this is a right scenario for adding node highlighting.

PR #962, opened 2020-04-30
CY-2651 Switched topology widget to use manager provided plugin icons instead of static data

kubama:


package.json
```diff
@@ -36,7 +36,7 @@
   "dependencies": {
     "better-console": "^0.2.4",
     "blueimp-md5": "^2.7.0",
-    "cloudify-blueprint-topology": "2.1.0",
+    "cloudify-blueprint-topology": "^2.2.0",
```
qooban, line 39: Please create release with changes description for v2.2.0 in https://github.com/cloudify-cosmo/cloudify-blueprint-topology/releases .

widgets/topology/src/widget.js
```diff
@@ -70,17 +70,72 @@ Stage.defineWidget({
...
+                    .map(plugin => _.pick(plugin, 'package_name', 'package_version'))
+                    .uniqWith(_.isEqual)
+                    .filter(plugin => plugin.package_name && plugin.package_version)
+                    .value();
+
+                if (_.isEmpty(plugins)) return data;
+
+                return toolbox
+                    .getManager()
+                    .doGet('/plugins?_include=id,package_name,package_version')
```
qooban, line 99: So REST API is not providing `title` yet, right?

widgets/topology/src/widget.js
```diff
@@ -70,17 +70,72 @@ Stage.defineWidget({
...
+                                                    reader.addEventListener('error', () => resolve());
+                                                    reader.addEventListener('load', () => {
+                                                        resolve({ [plugin.package_name]: reader.result });
+                                                    });
+                                                    reader.readAsDataURL(blob);
+                                                } else {
+                                                    resolve();
+                                                }
+                                            })
+                                    )
```
qooban, line 130: Just wondering - maybe it's worth to move it all to the stage backend? Or `GET /plugins/icons` is used already in different way in other place?

package.json
kubama: Done

widgets/topology/src/widget.js
kubama: No, not yet

widgets/topology/src/widget.js
kubama: A call to `/plugins/icons` is used in another place in a slightly different way. I considered a refactoring to extract a common part responsible for converting an image to data URL (`FileReader`), but it turned out it would be complex and possibly not worth the effort.

PR #968, opened 2020-05-08
CY-2756 - Improved system tests runner

qooban:
* Improved nightwatch configuration
* Added new script for running Nightwatch and Cypress tests properly (not fail if Nightwatch fails before running Cypress tests)
* Added default resetting templates request on activation
* Changed bash scripts access rights (644 -> 755) and removed explicit 'bash' call from npm scripts
* Removed unused scripts


scripts/runE2ETests.sh
```diff
@@ -0,0 +1,22 @@
...
+echo "Finished Cypress tests."
+
+if [[ $NIGHTWATCH_TESTS_EXIT_CODE -eq 0 ]] && [[ CYPRESS_TESTS_EXIT_CODE -eq 0 ]];
+then
+  echo "All E2E tests passed successfully."
+  exit 0
+else
+  echo "Some E2E tests failed."
+  exit 1
+fi
```
kubama, line 22:
```suggestion
#!/usr/bin/env bash
set -xo pipefail

npm run --silent e2e:old
npm run --silent e2e:new
```
The `-x` will print the executed commands, which makes the echos not necessary.
The `-o pipefail` eliminates the need for the logic that starts at line 15.

package.json
```diff
@@ -20,7 +20,7 @@
     "devtest": "mocha --require @babel/register --require esm --require ./test/setup.js -R spec",
     "docCheck": "bash scripts/checkDocs.sh",
     "docWidgets": "node scripts/updateReadmes.js",
-    "e2e": "npm run e2e:old && npm run e2e:new",
+    "e2e": "bash scripts/runE2ETests.sh",
```
kubama, line 23: I think `bash` here is kind of redundant with line 1 of the script where script interpreter is specified. So is `bash` really necessary here?

qooban: Didn't think about it. Just used the same approach as in the other npm scripts. 
What do you suggest? If we change to `./scripts/runE2ETests.sh` will it work properly on Windows (GIT Bash)? Are we sure it has always proper permissions set (x for all)?

kubama: Git is capable of holding `x` flag for files. The file just needs to be correctly committed. Since this is part of git functionality I would expect it to be working correctly in every environment. So my suggestion is to remove `bash` from here and verify if `scripts/runE2ETests.sh` is committed with correct access rights.

qooban: Branch updated. Please verify.

PR #970, opened 2020-05-08
CY-2643 - Extended Outputs/Capabilities widget with export button and Type column

kubama:
image


widgets/outputs/src/OutputsTable.js
```diff
@@ -1,8 +1,12 @@
+import { saveAs } from 'file-saver';
```
qooban, line 1: That way `file-saver` will be included in both `outputs/widget.js` bundle and in `main.bundle.js` file (as it is already used in `app/utils/External.js`).

The way we optimized such issues till now was exposing things through `Stage` global, but I'm wondering if webpack could do that for us somehow. Do you have any idea?

widgets/outputs/src/OutputsTable.js
```diff
@@ -38,12 +42,10 @@ export default class extends React.Component {
     render() {
         const NO_DATA_MESSAGE =
             "There are no Outputs/Capabilities available. Probably there's no deployment created, yet.";
-        const { DataTable, ErrorMessage, Header, Label } = Stage.Basic;
+        const { DataTable, ErrorMessage, Header, Button } = Stage.Basic;
```
qooban, line 45: Let's keep it in alphabetical order.

```suggestion
        const { Button, DataTable, ErrorMessage, Header } = Stage.Basic;
```

widgets/outputs/src/OutputsTable.js
```diff
@@ -38,12 +42,10 @@ export default class extends React.Component {
...
         const { ParameterValue, ParameterValueDescription } = Stage.Common;
 
-        const { outputsAndCapabilities } = this.props.data;
-
-        const compareNames = (a, b) => (a.name > b.name ? 1 : b.name > a.name ? -1 : 0);
+        const { outputsAndCapabilities, blueprintId, deploymentId } = this.props.data;
```
qooban, line 48:
```suggestion
        const { blueprintId, deploymentId, outputsAndCapabilities } = this.props.data;
```

widgets/outputs/src/OutputsTable.js
kubama: There is https://webpack.js.org/plugins/dll-plugin/ which would work but it has it's downsides - in order for tree shaking and dead code elimination to work the 'dll' needs to have its entry points explicitly specified. Please have a look but I doubt it would be useful in our case - we could create a common bundle that could be shared by main and widget bundles but we would need to keep entry points in sync with what's actually used by widgets - not very handy. I'd rather add `saveAs` to `stageUtils.js`.

qooban: Agree, let's add it to `stageUtils.js`.

PR #971, opened 2020-05-11
CY-1720 Enabled npm audit in Circle CI build

kubama:
- updated `decompress` and `pm2` to fix vulnerability issues
- added audit job to Circle CI build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,6 +95,15 @@ jobs:
           name: Check if cloudify-ui-components version is the same as in official documentation
           command: npm run docCheck
 
+  audit:
+    <<: *default-env
+
+    steps:
+      -   attach_workspace:
+            at: ~/
+      -   run:
+            name: Run npm production dependencies security audit
+            command: npm run audit
 
 workflows:
   version: 2
@@ -110,3 +119,9 @@ workflows:
       - checkDocs:
           requires:
             - build
+      - audit:
+          requires:
+            - build
+          filters:
+            branches:
+              only: master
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -110,3 +119,9 @@ workflows:
       - checkDocs:
           requires:
             - build
+      - audit:
+          requires:
+            - build
```
qooban, line 124: Does it require `build`?

backend/package.json
```diff
@@ -7,6 +7,7 @@
   "author": "Cloudify",
   "main": "server.js",
   "scripts": {
+    "audit": "npm audit --production --audit-level=moderate",
```
qooban, line 10: Why we explicitly set audit-level here?

kubama: This sets the minimum vulnerability level on which the build will fail. In this case the build will fail if there is moderate, high or critical vulnerability.
At the moment there are some low vulnerabilities reported which I didn't consider to be something worth addressing. If you think low level should also fail the build I can try to resolve them.

qooban: I think that we should work towards having consistent setting. If we are expecting low level vulnerabilities to be reported on front-end part, let's also have the same level set for back-end part. Let's invest some more time and try to resolve low vulnerabilities. If there will be time-consuming blockers we can create separate issue to fix that, but I would like that we will end with consistent setting across UI projects within CY-1720.

.circleci/config.yml
```diff
@@ -95,6 +95,15 @@ jobs:
...
 
+  audit:
+    <<: *default-env
+
+    steps:
+      -   attach_workspace:
+            at: ~/
+      -   run:
+            name: Run npm production dependencies security audit
+            command: npm run audit
```
qooban, line 106: Probably we can reduce the indent here.

PR #973, opened 2020-05-14
CY-2739 - Added more detailed description about handling types of deployment inputs

qooban:
image

### DIFF
--- a/widgets/common/src/InputsHeader.js
+++ b/widgets/common/src/InputsHeader.js
@@ -59,7 +59,7 @@ class InputsHeader extends React.Component {
                                     Surround value with <strong>"</strong> (double quotes) to explicitly declare it as a
                                     string, e.g.:
                                     <br />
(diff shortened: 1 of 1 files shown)


widgets/common/src/InputsHeader.js
```diff
@@ -36,25 +36,34 @@ class InputsHeader extends React.Component {
...
+                                        <strong>null</strong> will be casted to a null value
                                     </List.Item>
                                 </List>
-                                Use <strong>""</strong> for an empty string.
+                                <p>Value is treated as string if it is not possible to cast it to a different type.</p>
+                                <p>
+                                    Surround value with <strong>"</strong> (double quotes) to explicitly declare it as a
+                                    string, e.g.:
+                                    <br />
+                                    <strong>"true"</strong> will be treated as string not a boolean value.
```
kubama, line 62:
```suggestion
                                    <strong>"true"</strong> will be treated as a string, not a boolean value.
```

PR #975, opened 2020-05-15
CY-2650 Added 'Edit a copy in Composer' action to Blueprints and Blueprint Actions widgets

kubama:
Blueprints:
image

Blueprint actions:
image


widgets/blueprintActionButtons/src/widget.js
```diff
@@ -18,9 +18,22 @@ Stage.defineWidget({
     permission: Stage.GenericConfig.WIDGET_PERMISSION('blueprintActionButtons'),
     categories: [Stage.GenericConfig.CATEGORY.BLUEPRINTS, Stage.GenericConfig.CATEGORY.BUTTONS_AND_FILTERS],
 
+    fetchUrl: '[manager]/blueprints?_include=main_file_name[params]',
```
qooban, line 21: This fetch makes sense only if `blueprintId` is in the context, because we do not operate on multiple blueprints in Blueprint Action Buttons widget, right?

If so, then I would suggest either do `main_file_name` field fetch when user clicks the "Edit a copy..." button in BlueprintActionButtons component or limit the fetch here (using `fetchData`) to the situation when `blueprintId` is set. The first option is probably optimal solution.

kubama: Yes, I agree, updated

PR #976, opened 2020-05-18
CY-2598 - Fix react/destructuring-assignment warnings

qooban:
Reduced number of ESLint issues from:
✖ 7018 problems (0 errors, 7018 warnings)
to:
✖ 3702 problems (0 errors, 3702 warnings)

### Changes
- Most of the react/destructuring-assignment warnings fixed with codemod
- The rest react/destructuring-assignment issues fixed manually
- Fixed some security issues detected by NPM audit

### DIFF
--- a/app/components/AddWidgetModal.js
+++ b/app/components/AddWidgetModal.js
@@ -182,8 +182,8 @@ export default class AddWidgetModal extends Component {
     }
 
     doFilterWidgets(field, isCategoryChange = false) {
-        const { selectedCategory } = this.state;
-        const search = isCategoryChange ? search : field.value;
+        const { search: stateSearch, selectedCategory } = this.state;
+        const search = isCategoryChange ? stateSearch : field.value;
         const category = isCategoryChange ? field.name : selectedCategory;
 
         const { widgetDefinitions } = this.props;
--- a/app/components/AddWidgetModal.js
+++ b/app/components/AddWidgetModal.js
@@ -62,7 +62,8 @@ export default class AddWidgetModal extends Component {
     };
 
     componentDidUpdate(prevProps) {
-        if (!_.isEqual(prevProps.widgetDefinitions, this.props.widgetDefinitions)) {
+        const { widgetDefinitions } = this.props;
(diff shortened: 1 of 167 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "better-console": "^0.2.4",
     "blueimp-md5": "^2.7.0",
     "cloudify-blueprint-topology": "^2.3.0",
-    "cloudify-ui-common": "1.6.0",
```
kubama, line 41: Is this version upgrade (and others done in this file) somehow related to the topic of this PR?

app/components/AddWidgetModal.js
```diff
@@ -116,22 +120,27 @@ export default class AddWidgetModal extends Component {
...
+        const { widgetsToAdd } = this.state;
+        return _.without(widgetsToAdd, widgetId);
     }
 
     uninstallWidget() {
-        const widgetId = this.state.widget.id;
+        const { onWidgetUninstalled } = this.props;
+        const {
+            widget: { id: widgetId }
+        } = this.state;
```
kubama, line 131: This is a general comment, just using this code fragment for reference.
I understand the benefit of using destructuring assignments, but for nested cases, or for cases where const name is changed over the property name I'm in doubt. Looking at this case it seems neither easy to read nor intuitive. Personally I would stick with `const { widget } = this.state;` and then reference `widget.id`. What do you think?

package.json
qooban: 1. cloudify-ui-common - this was done 'by the way' as 1.6.1 and 1.6.2 introduced some ESLint changes (see: https://github.com/cloudify-cosmo/cloudify-ui-common/releases/tag/v1.6.1 and https://github.com/cloudify-cosmo/cloudify-ui-common/releases/tag/v1.6.2)
2. the rest devDependencies - upgraded after doing `npm audit`

So the first one is related somehow. The second one is not related, but I'm not sure it is worth to divide it. If you think so, I can create separate PR.

app/components/AddWidgetModal.js
qooban: Yes, make sense. I'll go over such cases and try to simplify it.

PR #977, opened 2020-05-19
CY-2650 Additional changes for "Edit a copy in Composer" functionality

kubama:
- Added action button to blueprints widget catalog view:
image

- Moved common logic to `BlueprintsActions.js`
- Fixed cypress test


widgets/blueprints/src/BlueprintsCatalog.js
```diff
@@ -127,6 +129,21 @@ export default class BlueprintsCatalog extends React.Component {
                                 this.props.onCreateDeployment(item);
                             }}
                         />
+
```
qooban, line 132: Refers to the lines 111-130 (above).

Let's fix "Delete" button at once as the styling looks bad. Probably just need to remove `basic` prop and play with 'labeled' class (should probably be `label` prop instead of explicit class..

kubama: My impression is that different styling in intentional here. Since "Delete" is a destructive action it's good that it's not stressed out as much as the other buttons.

qooban: To be clear - we have two differences here:
1. basic vs. no-basic
2. label vs. no-label

I can agree that basic may be intentional, but I'm almost sure that the second difference is not and it just looks bad.

qooban: Should be:
```
                        <Button
                            icon="trash"
                            content="Delete"
                            labelPosition="left"
                            basic
                            onClick={event => {
                                event.stopPropagation();
                                onDeleteBlueprint(item);
                            }}
                        />

                        <Button
                            icon="rocket"
                            content="Deploy"
                            labelPosition="left"
                            onClick={event => {
                                event.stopPropagation();
                                onCreateDeployment(item);
                            }}
                        />
```

kubama: Right, fixed

PR #986, opened 2020-05-26
CY-2814 Added read-only support for tabs

kubama:
Adds read-only support for `tab` attribute in page JSON files:
```
{
  "name": "Admin Dashboard",
  "widgets": [...],
  "tabs": [
    { "name": "Tab 1", "widgets": [...]},
    ...
  ]
}
```

Tabs are displayed below page level widgets:

image

Includes `PageManagement` refactoring (moving from redux to internal component state).

Cypress test will be provided in separate PR.


app/components/AddWidgetModal.js
```diff
@@ -285,7 +288,7 @@ export default class AddWidgetModal extends Component {
             StageUtils.Url.url(LoaderUtils.getResourceUrl(`widgets/${widget.id}/widget.png`, widget.isCustom));
 
         return (
-            <div className={className}>
+            <div className="compactBlock">
```
qooban, line 291: Probably we can even go further and replace `className="compactBlock"` with `style={{display: 'inline-block'}}` here and remove `compactBlock` from `styles.scss`.

app/components/EditModeBubble.js
```diff
@@ -4,69 +4,25 @@
...
+                    <Message.Content className="alignCenter">{children}</Message.Content>
                 </Message>
             </Sidebar>
-        );
-    }
+        </>
+    );
 }
+
+EditModeBubble.propTypes = {
```
qooban, line 24: Great :+1: One component less to add propTypes.

app/components/PageContent.js
```diff
@@ -0,0 +1,41 @@
...
+        <>
+            {createWidgetList(page.widgets)}
+            {!_.isEmpty(page.tabs) && (
+                <Tab
+                    panes={_.map(page.tabs, (tab, tabIndex) => ({
+                        menuItem: tab.name,
+                        render: () => createWidgetList(tab.widgets, tabIndex)
+                    }))}
+                />
+            )}
```
qooban, line 28: I assume you confirmed that limitation (tabs only below widgets and no widgets below tabs) with Ofer, if not it's just a mere formality to do so.

app/components/Widget.js
```diff
@@ -119,10 +111,7 @@ export default class Widget extends Component {
                     <div className="widgetButtons" onMouseDown={e => e.stopPropagation()}>
                         {isEditMode && (
                             <div className="widgetEditButtons">
-                                <i
-                                    className="remove link icon small"
-                                    onClick={() => onWidgetRemoved(pageId, widget.id)}
-                                />
+                                <i className="remove link icon small" onClick={() => onWidgetRemoved(widget.id)} />
```
qooban, line 114: Let's migrate it to proper `<Icon />` component (from `semantic-ui-react`).

app/components/layout/Layout.js
```diff
@@ -85,8 +84,19 @@ export default class Layout extends Component {
...
+                            path="/page_preview/:pageId"
+                            render={({ match }) => <PageManagement pageId={match.params.pageId} />}
+                        />
+                    )}
+                    {isUserAuthorizedForTemplateManagement && (
+                        <Route
+                            exact
+                            path="/page_edit/:pageId"
+                            render={({ match }) => <PageManagement pageId={match.params.pageId} isEditMode />}
+                        />
```
qooban, line 99: Nice that you added distinction on page preview/edit on URL level. I'm just wondering if we should not have both URLs prefixed with /template_management. I'm not for or against. Just asking for your opinion?

app/components/templates/PageManagement.js
```diff
@@ -2,124 +2,208 @@
...
+    const isWidgetMaximized = findWidget({ maximized: true });
+
+    $('body')
+        .css({ overflow: isWidgetMaximized ? 'hidden' : 'inherit' })
+        .scrollTop(0);
+
+    return (
+        <div className="main">
+            <div className="sidebarContainer">
+                <div className="ui visible left vertical sidebar menu small basic">
```
qooban, line 140: Looks like leftover from the time where we used `semantic-ui` and not `semantic-ui-react`.
Let's migrate it to proper `semantic-ui-react` component (`<Menu />`? `<Sidebar />`?).

app/components/templates/PageManagement.js
```diff
@@ -2,124 +2,208 @@
...
+                    <div className="pages">
+                        <div className="item link pageMenuItem">{page.name}</div>
+                        <div className="item link pageMenuItem" />
```
qooban, line 143: `div.item.link` can probably be also migrated to `semantic-ui-react` component.

app/components/PageContent.js
kubama: This is clearly stated in story description:
> a page may contain a widget and below it a tab (or set of tabs) containing other widgets

app/components/layout/Layout.js
kubama: Both options seem equally good to me. I don't see any real benefit but also no issue from prefixing it with /template_management.
I would leave it as it is.

app/components/templates/PageManagement.js
kubama: Done

app/components/templates/PageManagement.js
kubama: Done

app/components/Widget.js
kubama: Done

app/components/AddWidgetModal.js
kubama: Done

PR #987, opened 2020-05-27
CY-2541 - Fixed path to npm

qooban:


scripts/uploadPackage.sh
```diff
@@ -9,9 +9,9 @@ PRE_COMMANDS="
 
 POST_COMMANDS="
   cd /opt/cloudify-stage/backend;
-  sudo /bin/npm run db-migrate;
+  sudo /usr/bin/npm run db-migrate;
   sudo service cloudify-stage restart;
-  sudo /bin/npm run wait-on-server;"
+  sudo /usr/bin/npm run wait-on-server;"
```
kubama, line 14: Why not just `sudo npm run ...`? Is there a reason for specifying a full path to the executable?

qooban: You're somehow right, but I wanted to be aligned with what is used in other repositories https://github.com/cloudify-cosmo/cloudify-manager/search?q=npm&unscoped_q=npm
https://github.com/cloudify-cosmo/cloudify-manager-install/search?q=npm&unscoped_q=npm

PR #988, opened 2020-05-28
CY-2814 Added tabs cypress test

kubama:


qooban:
@kubama I think we need to start making more order in our cypress `integration` folder.
We can have separate folders inside for widget (eg. blueprint_actions, cluster_status), page (eg. sites), app framework (templates, tours) tests. Wdyt?

kubama:
Yes, definitely a good idea

PR #992, opened 2020-06-01
CY-2873 - Additional improvements to Blueprints Catalog widget

qooban:
- Fixed pagination removal in Blueprints Catalog for JSON source
- Removed pagination in Blueprints Catalog for JSON source in catalog view
- Added classname to each row in table view to help adding tour selectors


tours/main-sys_admin.json
```diff
@@ -203,7 +203,7 @@
                 "title": "Upload a Blueprint",
                 "content": "When you click <strong>Upload</strong>, the blueprint is uploaded to your Cloudify Manager. The blueprint is available in Local Blueprints, where you can also upload other blueprints.",
                 "onNextRedirectTo": ["/page/local_blueprints", "Local Blueprints", "No Blueprints", "You need to have <strong>Blueprints</strong> widget present on current page with at least one blueprint to proceed to the next step."],
-                "target": "div.blueprintCatalogWidget div.ui.grid > div.row > div:first-child button.uploadButton",
+                "target": "div.blueprintCatalogWidget table.ui tr:first-child > td.center.aligned.rowActions > i.upload",
```
kubama, line 206: `center ` and `aligned` are purely presentation classes so can we omit them here? It shouldn't be required to update this file if any of these classes are removed.

qooban: Good catch, I agree, fixing.

PR #996, opened 2020-06-02
CY-2894 Added ability to add widgets to predefined tabs

kubama:
Updated edit mode look:
image

### DIFF
--- a/app/components/AddWidgetModal.js
+++ b/app/components/AddWidgetModal.js
@@ -27,6 +27,7 @@ import {
 } from './basic/index';
 import InstallWidgetModal from './InstallWidgetModal';
 import { addWidget } from '../actions/widgets';
+import EditModeButton from './EditModeButton';
 
 let nameIndex = 0;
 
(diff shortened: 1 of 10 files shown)


app/components/AddWidgetModal.js
```diff
@@ -215,7 +216,14 @@ export default class AddWidgetModal extends Component {
...
         const { canInstallWidgets, onWidgetInstalled } = this.props;
         const addWidgetBtn = (
-            <Button icon="bar chart" labelPosition="left" basic content="Add Widget" className="addWidgetBtn" />
+            <EditModeButton
+                icon="add"
+                labelPosition="left"
+                basic
+                content="Add Widget"
+                className="addWidgetBtn"
+                style={{ marginBottom: 15 }}
```
qooban, line 225: As we already adding `style` here, I would move everything from `.addWidgetBtn` class in `styles.scss` here as well.
Class itself could still be useful (in Cypress tests), so we can keep it.

app/components/EditModeButton.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from './basic';
+import './EditModeButton.css';
+
+export default function EditModeButton({ className, ...rest }) {
+    return <Button className={`editModeButton ${className}`} {...rest} />;
```
qooban, line 7: If I didn't miss something, we can move more `Button` props here. Looks like `icon="add" basic labelPosition="left"` are in common.

kubama: Yes, I planned to do it later as this component is going to be reused in tabs management as well, so the set of common props may be eventually different.

app/components/AddWidgetModal.js
kubama: It turned out the styling was not applicable, so I just removed it.

PR #1001, opened 2020-06-04
CY-2645 - Added support for managing page tabs (add/remove/rename/set as default)

kubama:
Tabs in edit mode:
image

Clicking the remove action icon on a tab containing any widgets displays the following confirmation dialog:
image

Clicking the edit action icon opens a dialog, which allows changing the tab name as well as setting it as default tab:
image

When no tabs are defined the following button is displayed in place:
image

TODO: Add system tests


app/components/EditTabModal.js
```diff
@@ -0,0 +1,52 @@
...
+                    }}
+                />
+                <CancelButton onClick={() => setOpen(false)} />
+            </Modal.Actions>
+        </Modal>
+    );
+}
+
+EditTabModal.propTypes = {
+    tab: PropTypes.shape({}).isRequired,
```
qooban, line 49: I see that you are using `tab.name` and `tab.isDefault` in this component, so I would suggest to add it to the propTypes checking here.

webpack.config.js
```diff
@@ -185,10 +185,6 @@ module.exports = (env, argv) => {
                             {
                                 from: 'tours',
                                 to: 'appData/tours'
-                            },
-                            !isProduction && {
```
qooban, line 189: 100% sure that we don't need it in development environment? If so, then fine, let's remove it.

kubama:
> 1. Are we allowing to switch the order of the tabs? (best UX would be probably using drag&drop) It's not critical feature anyway, so only if you see a way to implement that in reasonable time.

Tabs reordering is specified as a requirement in story description.
Since `Tab` component doesn't support this out of the box the implementation may be tricky, so I was planning to work on it separately, as the last tabs implementation phase.

> 2. Possibly it would be more user friendly to allow tab name inline edit rather than showing modal, but the problem is with the default tab toggle. Or maybe you have an idea how we can also make default toggle to be in-lined in easy way? (maybe shown only on tab item hover) It's only suggestion, but maybe worth to consider if not very hard to implement.

Since clicking on a tab activates it, it shouldn't start in-line edit at the same time. So an action icon for in-line edit would be still needed, I think.
An alternative approach for setting default tab could be to have a "settings" icon next to "add tab" icon. It would open a modal with a dropdown for selecting the default tab.
Lets wait for @ofercloudify 's comment on this.

webpack.config.js
kubama: It's not needed, userData files are served by the backend.
Having userData files as part of webpack bundle was causing issues - after editing page in the UI the dev server had to be restarted for the changes to be served. Deleting this entry solved the issue.

app/components/EditTabModal.js
kubama: Done

kubama:
Done

PR #1006, opened 2020-06-09
CY-2958 Added support for changing tabs order

kubama:
NOTE 1: No system tests. Some time ago I spent quite a lot of time trying to cover stencils drag'n'drop with cypress tests (in composer) and I failed. Since tabs reordering is not a crucial functionality I'd leave it without tests for now.

NOTE 2: I'm aware that `PageList.js` and `CreateTemplateModal.js` already implement reordering, but they are using a jquery based solution that I didn't want to copy, thus I decided to use a pure react solution. I assume `PageList.js` and `CreateTemplateModal.js` should be refactored at some point in the future to move away from jquery and consistently use the same solution as `PageContent.js` does.

### DIFF
--- a/app/components/PageContent.css
+++ b/app/components/PageContent.css
@@ -1,0 +1,20 @@
+.draggedTab {
+    border: 1px solid #d4d4d5;
+    border-bottom: none;
+    border-top-left-radius: 4px;
+    border-top-right-radius: 4px;
+    margin-bottom: 4px;
+    padding-top:13px;
+    text-align: center;
+    color:rgba(0, 0, 0, 0.95);
+    background-color: white;
+    opacity: 0.9;
+}
+
+.draggedTab.active {
+    font-weight: bold;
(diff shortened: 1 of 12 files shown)


app/components/PageContent.css
```diff
@@ -0,0 +1,20 @@
+.draggedTab {
+    border: 1px solid #d4d4d5;
+    border-bottom: none;
+    border-top-left-radius: 4px;
+    border-top-right-radius: 4px;
+    margin-bottom: 4px;
+    padding-top:13px;
+    text-align: center;
+    color:rgba(0, 0, 0, 0.95);
```
qooban, line 9:
```suggestion
    padding-top: 13px;
    text-align: center;
    color: rgba(0, 0, 0, 0.95);
```

kubama:
> Let me know the ID when you create it.

CY-2975

PR #1022, opened 2020-06-17
Fix package upload permissions

qooban:
* Fixed upload script
* Added `nginx` to `stage_group` to provide access to static resources


packaging/cloudify-stage.spec
```diff
@@ -58,6 +58,7 @@ groupadd -fr stage_group
 getent passwd stage_user >/dev/null || useradd -r -g stage_group -d /opt/cloudify-stage -s /sbin/nologin stage_user
 usermod -aG cfyuser stage_user
 usermod -aG stage_group cfyuser
+usermod -aG stage_group nginx
```
geokala, line 61: Same user and group management commands needed here as for the composer one.

qooban: Fixed.

PR #1023, opened 2020-06-17
CY-2200 - Change default logging level

qooban:
- Added logLevel option to user configuration
- Changed default log level to WARN from DEBUG

### DIFF
--- a/backend/handler/LoggerHandler.js
+++ b/backend/handler/LoggerHandler.js
@@ -4,6 +4,9 @@
 
 const winston = require('winston');
 const _ = require('lodash');
+const config = require('../config').get();
+
+const { logLevel } = config.app;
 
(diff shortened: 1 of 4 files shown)


backend/handler/LoggerHandler.js
```diff
@@ -30,7 +33,7 @@ module.exports = (() => {
         return logger;
     }
 
-    function getLogger(category, level = 'debug') {
+    function getLogger(category, level = logLevel) {
```
kubama, line 36: I think the second argument to `getLogger` and `getStream` functions should be removed. As log level is application level configuration option it shouldn't be possible to bypass this setting.

qooban: Agree. Changed.

PR #1025, opened 2020-06-18
CY-2947 update map key

warolv:
CY-2947 update map key

### DIFF
--- a/conf/app.json
+++ b/conf/app.json
@@ -48,6 +48,6 @@
     "portalUrl": ""
   },
   "maps": {
-    "accessToken": "<TOKEN>"
+    "accessToken": ""
   }
 }
--- a/conf/app.json
+++ b/conf/app.json
@@ -46,5 +46,8 @@
     "certPath": "",
     "ssoUrl": "",
     "portalUrl": ""
+  },
+  "maps": {
+    "accessToken": "<TOKEN>"
   }
 }
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -32,9 +32,13 @@ pipeline {
   stages {
     stage('Prepare'){
       steps {
-        dir("${stage_dir/conf}") {
-          withVault([configuration: configuration, vaultSecrets: secrets]) {
-            jq '".maps.accessToken = ${env.WIKIMEDIA_ACCESS_TOKEN}" app.json > app.json.tmp && cp app.json.tmp app.json'
+        container('node'){
+          dir("${stage_dir}/conf") {
+            withVault([configuration: configuration, vaultSecrets: secrets]) {
+              sh """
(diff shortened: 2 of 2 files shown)


warolv:
@qooban Need real token for vault

jenkins/Jenkinsfile
```diff
@@ -1,3 +1,10 @@
+def secrets = [
+  [path: 'secret/jenkins/wikimedia', engineVersion: 2, secretValues: [
+    [envVar: 'WIKIMEDIA_ACCESS_TOKEN', vaultKey: 'access_token']]],
```
qooban, line 3: I would change the name to be more generic. Right now we are using WikiMedia, but we are moving to StadiaMaps (that's why access token is needed).

Maybe
```suggestion
  [path: 'secret/jenkins/maps', engineVersion: 2, secretValues: [
    [envVar: 'MAPS_ACCESS_TOKEN', vaultKey: 'access_token']]],
```
?

jenkins/Jenkinsfile
```diff
@@ -23,6 +30,19 @@ pipeline {
   }
   
   stages {
+    stage('Prepare'){
+      steps {
+        container('node'){
+          dir("${stage_dir}/conf") {
+            withVault([configuration: configuration, vaultSecrets: secrets]) {
+              sh """
+                jq '.maps.accessToken = "${env.WIKIMEDIA_ACCESS_TOKEN}"' app.json > app.json.tmp && cp app.json.tmp app.json
```
qooban, line 39: 1. We need to adapt to the previous change (whatever you choose as a name of env var)
2. I think it would be better to use `mv` instead of `cp` not to keep the `app.json.tmp` file in the RPM package (and then after installation on the Cloudify Manager VM)

```suggestion
                jq '.maps.accessToken = "${env.MAPS_ACCESS_TOKEN}"' app.json > app.json.tmp && mv app.json.tmp app.json
```

conf/app.json
```diff
@@ -46,5 +46,8 @@
     "certPath": "",
     "ssoUrl": "",
     "portalUrl": ""
+  },
+  "maps": {
+    "accessToken": "<TOKEN>"
```
qooban, line 51: Let's keep it empty before modifications:
```suggestion
    "accessToken": ""
```
I will treat such case as a way to access maps without authentication.

jenkins/Jenkinsfile
warolv: ok, sounds good

jenkins/Jenkinsfile
warolv: sounds good

conf/app.json
warolv: Ok

PR #1045, opened 2020-07-21
CY-3159 Fixed plugin upload with default title

kubama:


test/cypress/integration/widgets/plugins_spec.js
```diff
@@ -1,12 +1,9 @@
...
-    before(() =>
-        cy
-            .activate('valid_trial_license')
-            .deletePlugins()
-            .login()
-    );
+    before(() => cy.activate('valid_trial_license'));
+
+    beforeEach(() => {
+        cy.deletePlugins().login();
```
qooban, line 5: `login()` is quite time consuming. I suspect you used that one to avoid the situation when plugins are not yet deleted. I suggest better way to ensure that - use `waitUntilEmpty` function. Take a look at `deleteBlueprints` and `deleteDeployments` support commands.

kubama: Actually in this situation it is not necessary to wait for the plugins to be deleted. I moved `login()` back to `before` and aligned rest of the code.

PR #1048, opened 2020-07-23
CY-3210 Fixed terraform topology features

kubama:


test/cypress/integration/widgets/topology_spec.js
```diff
@@ -61,22 +62,23 @@ describe('Topology', () => {
...
         cy.contains('#gridContainer > #gridSvg > #gridContent > .nodeContainer > .title', 'terraform');
         cy.contains('#gridContainer > #gridSvg > #gridContent > .nodeContainer > .title', 'cloud_resources');
 
-        // Install the deployment
+        cy.log('Install the deployment');
         cy.executeWorkflow(deploymentId, 'install');
-        cy.get('.executionsTable tr:eq(2)');
-        cy.get('.executionsTable tr:eq(1)').contains('completed');
+        cy.reload().waitUntilLoaded();
+        cy.contains('install completed');
```
qooban, line 77: Maybe we can have something similar to `ExecutionActions.waitUntilFinished` or `DeploymentActions.waitUntilCreated` in Cypress tests instead?

kubama: I did the other way around - I fixed the topology module so it shows/hides terraform action icons dynamically as the deployment/execution state changes without the need to reload the page. So now it's only necessary to wait for the button to show up.

PR #1050, opened 2020-07-24
CY-3169 - Execution task graph - no-task graph error

qooban:
### Main changes:
- Fixed fetching tasks graph (fetching is not stopped on tasks graph not present as it was before)
- Improved backend part of tasks graph poller (it fails quickly right now without trying to fetch not existing data)
- Improved fetching tasks graph (now we fetch tasks graph with interval depending on execution status - for active executions we poll data every 3 seconds, for inactive executions we poll data every 6 seconds)
- Updated layout for single execution view (to include status no matter if the graph is present)
- Reviewed and updated tests

### Chore:
- Fixed `makeCancelable` function to allow proper handling of failed promises
- Fixed GraphNode prop types
- Improved logging in backend part of tasks graph poller (using internal logger instead of console calls)

### DIFF
--- a/widgets/executions/src/backend.js
+++ b/widgets/executions/src/backend.js
@@ -71,6 +71,7 @@ module.exports = r => {
         const _ = require('lodash');
 
         const elk = new ELK();
+        const logger = helper.Logger();
 
         const tasksGraphsFetchUrl = '/tasks_graphs';
         const operationsFetchUrl = '/operations';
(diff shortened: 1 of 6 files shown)


widgets/executions/src/tasksGraph/GraphNode.js
```diff
@@ -123,7 +123,7 @@ GraphNode.propTypes = {
     graphNode: PropTypes.shape({
         labels: PropTypes.arrayOf(
             PropTypes.shape({
-                display_text: PropTypes.arrayOf(PropTypes.string),
+                display_text: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
```
kubama, line 126: There is `StringOrArray` prop type deifned in common, might be reused here

widgets/executions/src/backend.js
```diff
@@ -107,15 +108,25 @@ module.exports = r => {
 
         const runGraphCreation = () => {
             const tasksGraphParams = { ...req.query };
+            const headers = _.pick(req.headers, ['authentication-token', 'tenant']);
```
kubama, line 111:
```suggestion
            const headers = _.pick(req.headers, 'authentication-token', 'tenant');
```

widgets/executions/src/backend.js
```diff
@@ -107,15 +108,25 @@ module.exports = r => {
...
-            let { headers } = req;
             const operationsList = [];
             helper.Manager.doGet(tasksGraphsFetchUrl, tasksGraphParams, headers)
                 .then(data => {
-                    headers = _.omit(headers, 'accept-encoding'); // Required otherwise the data is returned in bytes and screwed up
-                    const operationsPromises = _.map(data.items, graph =>
+                    const { items } = data;
+
+                    if (_.isEmpty(items)) {
+                        const { execution_id: id } = tasksGraphParams;
```
kubama, line 119: Nothing important but I find this notation easier to read, applies to other places as well.
```suggestion
                        const id = tasksGraphParams.execution_id;
```

widgets/executions/src/tasksGraph/GraphNode.js
qooban: Right, forgot about it, will update.

widgets/executions/src/backend.js
qooban: Done.

widgets/executions/src/backend.js
qooban: Done.

PR #1056, opened 2020-07-30
CY-3243 Fixed terraform nodes expansion logic

kubama:


kubama:
But this has nothing to do with plugin changes, it is a regular bug on source code level.

PR #1058, opened 2020-07-31
CY-3245 Replaced deployment wizard with "getting started" link button

kubama:
Modified dashboard section:
image

### DIFF
--- a/widgets/buttonLink/src/widget.js
+++ b/widgets/buttonLink/src/widget.js
@@ -14,7 +14,8 @@ Stage.defineWidget({
 
     initialConfiguration: [
         { id: 'label', name: 'Button label', default: 'Button Link', type: Stage.Basic.GenericField.STRING },
-        { id: 'url', name: 'URL address', default: '', type: Stage.Basic.GenericField.STRING }
+        { id: 'url', name: 'URL address', default: '', type: Stage.Basic.GenericField.STRING },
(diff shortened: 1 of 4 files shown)


widgets/buttonLink/src/widget.js
```diff
@@ -14,7 +14,8 @@ Stage.defineWidget({
 
     initialConfiguration: [
         { id: 'label', name: 'Button label', default: 'Button Link', type: Stage.Basic.GenericField.STRING },
-        { id: 'url', name: 'URL address', default: '', type: Stage.Basic.GenericField.STRING }
+        { id: 'url', name: 'URL address', default: '', type: Stage.Basic.GenericField.STRING },
+        { id: 'fullHeight', name: 'Full height', default: false, type: Stage.Basic.GenericField.BOOLEAN_TYPE }
```
qooban, line 18: During original discussion with @ofercloudify about this issue we were also thinking about providing a way to define the button color. It could be easy to add and make the widget more flexible. Take a look at Text box widget configuration. I'm not sure it will work fine withe Semantic UI Button, so maybe a dropdown list with allowed colors is also an option.

kubama: I reused color picker as utilized by Text box - it works perfectly.

qooban:
@kubama Would be perfect to have some basic tests for that widget, but as it is not critical one, I let you decide.

PR #1063, opened 2020-08-03
CY-3268 Replaced markdown with marked

kubama:


app/utils/widgetDefinitionsLoader.js
```diff
@@ -48,7 +47,6 @@ export default class WidgetDefinitionsLoader {
         };
 
         window.moment = momentImport;
-        window.markdown = markdownImport;
```
qooban, line 51: Hopefully nobody (customers in custom widgets) uses that :-)
I'm almost sure they don't, so there should be no issues.

PR #1071, opened 2020-08-26
CY-3386 - Positioning improvement for node services status popup in Cluster Status widget

qooban:
Peek 2020-08-26 15-03


kubama (approved):
I'm just wondering about `>` selector operator usage in `cluster_status_spec.js`. Unless we care about actual HTML structure generated I think it should be avoided if possible. Is it required to use it in each case?

qooban:
Yeah, in some cases it was necessary, not in all. But I see it like that: caring about HTML structure is not the goal itself, but by caring about it we can detect visual errors.

PR #1078, opened 2020-08-31
CY-2765 Custom widgets test migration

kubama:


app/components/InstallWidgetModal.js
```diff
@@ -31,7 +31,7 @@ export default class InstallWidgetModal extends Component {
     }
 
     shouldComponentUpdate(nextProps, nextState) {
-        return !_.isEqual(nextState, this.state) || !_.isEqual(nextProps, this.props);
```
qooban, line 34: Why remove this?

Looking at the props definition looks like the modal should be updated on its change. If there are props changing to often (and without real impact on InstallWidgetModal), then we should probably fix it on higher level.

qooban (commented):
Added one comment.

I'm on `widgetBackend` test migration. Looks like we are using the same additional library - `cypress-file-upload`, so I'll wait for you when you merge your part and then update `edit_mode_spec.js` (originally I thought about separate file, but maybe it would be more accurate to extend 'should open widget install modal and').

app/components/InstallWidgetModal.js
kubama: Looking at how the modal is used the props never change.
However, on each parent update a new instance of callback function is created and passed as a prop - I guess even though the function is virtually the same `isEqual` is not able to determine that. This was effectively leading to an update on each parent update causing issues with file input and test instability. Hence the fix.

qooban: Which callback function?

I investigated that and the actual reason why `InstallWidgetModal` is re-rendering is that `trigger` prop is changing, which is understandable when you look at `AddWidgetModal` render function. `installWidgetBtn` is recreated whenever `InstallWidgetModal` is re-rendered. 

In my opinion we should fix it and removing equality check is not the solution. IMO whenever its possible the widget should be prepared for props change even if in current use-cases we do not change the props. 

One of the simplest solutions is to move `installWidgetBtn` outside of the `AddWidgetModal` class (possibly all of the components defined inside `render()` should be externalized). If we would like to go deeper then we would investigate why `AddWidgetModal` re-render is triggered, but it's outside of this PR's scope.

kubama: Solved by adding `shouldComponentUpdate` to `AddWidgetModal`

qooban (approved):

PR #1083, opened 2020-09-02
CY-3424 Added overflow: auto to deployment description field

kubama:
image

### DIFF
--- a/widgets/common/src/DeploymentDetails.js
+++ b/widgets/common/src/DeploymentDetails.js
@@ -49,18 +49,20 @@ export default function DeploymentDetails({
         <Grid stackable={stackable}>
             <Grid.Row>
                 <Grid.Column width={4}>
-                    <ResourceVisibility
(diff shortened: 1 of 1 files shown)


qooban (commented):
I like the idea with scroller, but I'm not sure about the style for the deployment name and visibility icon.

Take a look at this example:
image
There are two things here:
1. Icon positioning - Previously visibility icon was floated to the right and deployment name position was relative to the icon, so it went below the icon (which stayed in the upper right corner).
2. Word wrapping - I've seen much longer names for deployments in customer environments, so we probably need something like `word-wrap`.

I know it's hard with our current layout architecture (fixed height of the widget), but maybe you'll find a way to have dynamic height for the description box with optional scroller?

kubama:
I brought back `float: right` and introduced `display: flex` to address variable deployment name label height

qooban (approved):
Looks good! Optionally you can also do that for blueprint name.

qooban (approved):

PR #1085, opened 2020-09-03
CY-2333 Added db migration tests

kubama:


backend/test/migration.test.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * Created by jakub.niezgoda on 07/05/2019.
+ */
+
```
qooban, line 4: You can remove that :-)

backend/migrations/20190401111659-5_0-MovePageResourcesToFiles.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
                             updatedAt: moment(pageRow.updatedAt).format(),
                             widgets: []
                         };
+                        fs.mkdirSync(path.dirname(pageFilePath), { recursive: true });
```
qooban, line 40: Hmm... If we need that, then probably we should provide a patch for 5.0.5, am I right?

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
```diff
@@ -34,6 +34,7 @@ module.exports = {
                     } catch (error) {
                         pages = [];
                         logger.info('File does not exist. Creating new one...');
+                        fs.mkdirSync(path.dirname(templateFilePath), { recursive: true });
```
qooban, line 37: The same comment as above.

backend/migration.js
```diff
@@ -184,6 +184,7 @@ function handleCommand(cmd) {
             logger.error('='.repeat(errorStr.length));
             logger.error(err);
             logger.error('='.repeat(errorStr.length));
+            process.exit(1);
```
qooban, line 187: I assume that it returned zero code before your change?

qooban (dismissed):
This is exactly what we needed! :+1: 

I'm only concerned about changing already existing migrations for 5.0.
I think - if they are necessary to keep migrations work properly - then we need to create patch for 5.0.5 as well (of course can be done as a separate issue/PR).

backend/test/migration.test.js
kubama: OK, done.

backend/migrations/20190401111659-5_0-MovePageResourcesToFiles.js
kubama: This is required so the test could run migration, as in a clean test environment the whole directory structure under `dist` does not exist. However in a prod snapshot restoration scenario the structure should already exist as it is created during 5.0.5 (or 5.1) initialization.

qooban: Then, if it is only for test, why not create the whole `dist` structure just before the test?

kubama: This is specific to 5.0 migration. I'd rather keep the test generic and independent of logic specific to any particular migration.
On the other hand I don't think there is anything wrong in ensuring a directory used by migration exists, even if it is expected to exist.

qooban: I don't think there is anything wrong with ensuring a directory exists, too. The problem is that - without patch - there will be inconsistency in migration code between 5.0.5 and 5.1. 

If the specific directory inside `dist` (the one you ensure exists in migration code) is really present on clean installation before first user-Cloudify interaction, so that upgrade is working fine, then let's keep it as it is.

qooban (approved):

backend/migrations/20190401111659-5_0-MovePageResourcesToFiles.js
kubama: I see the migration scripts as an internal part of the application. We change application code frequently in a variety of ways, very often without changing the actual functionality. Is there anything special about the migration scripts that the code can't change? Why would we want to patch it in 5.0.5?

qooban: I was only concern about somebody doing upgrade to 5.0.5 (they will not have that part of the code in the migration scripts, then). 

Anyway, it will probably not happen, as users will go directly to 5.1 (hopefully).

PR #1095, opened 2020-09-22
CY-3594 - Change isomorphic-fetch to maintained equivalent

qooban:


qooban:
It appears that after long time without updates there was an update of isomorphic-fetch, so maybe it'll be maintained => not changing it for now.

PR #1099, opened 2020-09-25
CY-3518 - Backport PM2 drop to 5.0.5

qooban:


backend/server.js
```diff
@@ -104,17 +103,15 @@ app.use(
...
     `${contextPath}/userData`,
     passport.authenticate('cookie', { session: false }),
-    expressStaticGzip(path.resolve(__dirname, '../dist/userData'), { enableBrotli: true, indexFromEmptyFile: false })
+    expressStaticGzip(
+        path.resolve(__dirname, process.env.NODE_ENV === 'development' ? '../userData' : '../dist/userData'),
+        {
+            enableBrotli: true,
+            indexFromEmptyFile: false
+        }
+    )
```
kubama, line 112: How is this change related to pm2 drop?

qooban: Good catch. Added by mistake during cherry-picking. Fixed.

kubama: I guess this should be addressed in https://github.com/cloudify-cosmo/patchify/pull/193 as well, right?

kubama (approved):

backend/server.js
qooban: Yes. Working on it.

PR #1103, opened 2020-09-30
CY-3568 - Fix for Deployments widget

qooban:


widgets/deploymentInfo/src/widget.js
```diff
@@ -87,13 +87,14 @@ Stage.defineWidget({
...
-                .then(deploymentItem =>
-                    _.mapValues(deploymentItem, (fieldValue, fieldName) =>
+                .then(deploymentItem => ({
+                    ..._.mapValues(deploymentItem, (fieldValue, fieldName) =>
                         fieldName === 'created_at' || fieldName === 'updated_at'
                             ? Stage.Utils.Time.formatTimestamp(fieldValue)
                             : fieldValue
-                    )
-                );
+                    ),
```
kubama, line 95: After adding `isUpdated` I think it would be more readable to explicitly set `created_at` and `updated_at`, as it is done in widgets/deployments/src/widget.js, lines 122-123

kubama (approved):

widgets/deploymentInfo/src/widget.js
qooban: Makes sense, fixed.

PR #1107, opened 2020-10-07
CY-2876 Enabled coverage checks for frontend and backend

kubama:


package.json
```diff
@@ -21,14 +21,18 @@
...
     "devtest": "jest",
     "docCheck": "scripts/checkDocs.sh",
     "docWidgets": "node scripts/updateReadmes.js",
     "e2e": "cypress run --reporter cypress-multi-reporters --reporter-options configFile=test/cypress/reporter-config.json",
     "e2e:open": "cypress open",
+    "jest": "jest test/jest",
+    "jest:coverage": "jest test/jest --coverage",
     "lint": "eslint --cache --quiet app backend widgets test scripts",
-    "prodtest": "jest test/jest && cd backend && npm test",
+    "prodtest": "npm run jest && cd backend && npm test",
```
qooban, line 35: Please update test documentation: `test/README.md` with information about how to configure build and check coverage - for both jest and cypress tests.

package.json
```diff
@@ -21,14 +21,18 @@
     "build:analyse:main": "npm run build -- --env.analyse=main",
     "build:analyse:widgets": "npm run build -- --env.analyse=widgets",
     "build:analyse:widgets-common": "npm run build -- --env.analyse=widgets-common",
+    "build:coverage": "webpack --config webpack.config.js --mode=production --bail --env.coverage",
```
qooban, line 24: `npm run build -- --env.coverage` does not work?

backend/jest.config.js
```diff
@@ -0,0 +1,32 @@
+// eslint-disable-next-line security/detect-child-process
+const { execSync } = require('child_process');
+
+let changedFiles;
+try {
+    // Collect coverage only for files changed after this check was first introduced
+    changedFiles = execSync(
+        'git diff --name-only --relative 4347094038309750af37c8990389a360bf480e07 -- "*.js" "**/*.js"'
```
qooban, line 8: This commit is part of the branch, right? Will this work properly after squash merge?

qooban (commented):
Added a few comments.

.nycrc
```diff
@@ -0,0 +1,6 @@
+{
+    "branches": 65,
+    "lines": 71,
+    "functions": 65,
+    "statements": 71
```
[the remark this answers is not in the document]
kubama: My thinking was the thresholds should be manually reviewed after each release or every couple of months. They could be also manually updated as part of any merge that actually increases the coverage.

Another option - at the moment there are some widgets that are not covered at all. With Cypress it would be super fast and easy to add some basic but still valuable tests. It would allow to increase coverage quickly - I didn't calculate it but I think 80% coverage should be easy to achieve. I think it could be done as part of this JIRA ticket.

package.json
kubama: Done

package.json
kubama: Should work - updated

backend/jest.config.js
kubama: I'm not sure, I guess it might be necessary to update this file after squash merge

.nycrc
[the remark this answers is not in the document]
qooban: OK, great, but I would suggest adding new tests as part of separate ticket.
I think It would be good to move to master soon.

backend/jest.config.js
qooban: OK, will see.

qooban (approved):

PR #1112, opened 2020-10-13
CY-2601 Fixed agents widget

kubama:


widgets/common/src/NodeFilter.jsx
```diff
@@ -1,84 +1,116 @@
...
-function toId(resourceName) {
-    return `${resourceName}Id`;
-}
-
 /**
  * NodeFilter  - a component showing dropdowns for filtering blueprints, deployments, nodes and nodes instances.
  * Data (list of blueprints, deployments, nodes and node instances) is dynamically fetched from manager.
  *
+ * @param event
+ * @param field
```
qooban, line 6: Was that added intentionally? Possibly can be removed?

widgets/common/src/NodeFilter.jsx
```diff
@@ -1,84 +1,116 @@
...
-    allowMultipleNodes,
-    allowMultipleNodeInstances,
-    allowedBlueprints,
-    allowedDeployments,
-    allowedNodes,
-    allowedNodeInstances
-}) {
-    const { useState, useEffect } = React;
-    const { useErrors } = Stage.Hooks;
+export default class NodeFilter extends React.Component {
```
qooban, line 9: I assume it was too complex to convert to functional component and you reverted it back to class component, right?

qooban (commented):
Added some comments.

widgets/common/src/NodeFilter.jsx
kubama: Yes, this is the case

widgets/common/src/NodeFilter.jsx
kubama: Removed

qooban (approved):

PR #1127, opened 2020-11-12
RD-235 Added basic i18n support in stage core framework

kubama:
TODO: Add `overrides.json` usage description to cloudify docs


qooban (commented):
Two general comment:
1. Do we need this default value for all translations in the code? I mean, it is not a big problem, we can have it, but can introduce inconsistency when we update `en.json` file, but not the default.
2. From the main items in RD-235 we still have left: 1. All "cloudify" word occurences and 2, 3. Links. Am I right?

kubama:
> 1. Do we need this default value for all translations in the code? I mean, it is not a big problem, we can have it, but can introduce inconsistency when we update `en.json` file, but not the default.

The defaults allowed for automatic generation of en.json file. At this point we don't need them anymore (I doubt anyone would want to recreate this file from scratch) but I also don't think they are an issue - I can remove them if you think so.

> 2. From the main items in RD-235 we still have left: 1. All "cloudify" word occurences and 2, 3. Links. Am I right?

I believe items 1. and 2. are done (did I miss something?). Item 3. is widgets specific, but we decided to leave out widgets at the moment, right?

qooban:
Let's leave the defaults.

> I believe items 1. and 2. are done (did I miss something?). Item 3. is widgets specific, but we decided to leave out widgets at the moment, right?

Yes, we decided to leave out widgets (in general) at the moment, but we need to cover all points from the RD-235 list (this was consulted with Ofer and that is required for November release). So we need just a few parts in our widgets to be updated. All (I hope) are listed in "List of the strings to be moved" section.

qooban (commented):
Changes are fine, but I still see in the widget's code occurrences of "cloudify" word in the strings (Cluster Status widget, Events/Logs widget, Spire Manager widget, Composer Link widget). I think in most case we can just remove Cloudify word and not necessarily provide translation feature right now. 

As a reference you can also take a look on Ofer's work: https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1546

qooban:
and regarding:

> Gap: /app/components/banner not covered, I assumed these are going to be migrated to ui-components

and

> Note: whiteLabel attributes from userConfig.json are still there - they are being used in login page, which is not affected by overrides.json, as fetching files from userData requires authorization.

Let's discuss that on Monday's meeting. We need to synchronize our PRs to cover the first (gap) and discuss solutions to avoid the second limitation.

kubama:
> Changes are fine, but I still see in the widget's code occurrences of "cloudify" word in the strings (Cluster Status widget, Events/Logs widget, Spire Manager widget, Composer Link widget). I think in most case we can just remove Cloudify word and not necessarily provide translation feature right now.

I removed "Cloudify" occurrences. I think it works fine with one exception - I changed "Cloudify Composer" to "Blueprint Composer"

kubama:
> Let's discuss that on Monday's meeting. We need to [...] discuss solutions to avoid the second limitation.

Let's discuss - if this limitation is not acceptable then I think the easiest solution would be to add an exception to the `userData` backend endpoint so that this single `overrides.json` file could be served without authorization.

qooban (approved):

PR #1133, opened 2020-11-17
RD-441 Added ability to customize widgets' README.md files in runtime

kubama:


scripts/updateReadmes.js
```diff
@@ -49,44 +48,6 @@ function updateTitle(widget, content) {
...
-
-        Array.from(newContent.matchAll(paramRegex)).forEach(match => {
-            const paramName = match[1];
-            const paramValue = params[paramName];
-            log(widget, match[0], match[1]);
-            if (paramValue !== undefined) {
-                log(widget, `Converting "${match[0]}" into "${paramValue}".`);
-                newContent = newContent.replace(match[0], paramValue);
-            } else {
-                reject(new Error(`No ${paramName} parameter found.`));
```
qooban, line 82: That check is vital not to forget about adding parameter to our internal file. 
I think we need such check here or system test checking all README.md files for missing parameters.
I think it would be quicker and easier to add it in updateReadmes script.

kubama: Done

qooban (approved):
LGTM. 

I would consider extending [documentation page Label Overrides](https://docs.cloudify.co/staging/dev/working_with/console/customization/overrides/) you created in official documentation site. You can add few words about that functionality.

And, btw, I think that p. 3 in that page need to be re-checked. As far as I remember to get changes from `userConfig.json` stage backend needs to be restarted. I'm not sure that just  browser reload allows you to see the changes.

See [User Configuration page](https://docs.cloudify.co/staging/dev/working_with/console/customization/user-configuration/) - in p. 3 there's a sentence about restarting stage backend.

kubama:
> I would consider extending [documentation page Label Overrides](https://docs.cloudify.co/staging/dev/working_with/console/customization/overrides/) you created in official documentation site. You can add few words about that functionality.

Actually it's exactly the same functionality - this PR only extends the set of labels/URLs that can be customized. I'm not sure if we want to go deeper and describe individual sections of `overrides.json`?

> And, btw, I think that p. 3 in that page need to be re-checked. As far as I remember to get changes from `userConfig.json` stage backend needs to be restarted. I'm not sure that just browser reload allows you to see the changes.

Here browser reload is sufficient - `overrides.json` is served as static resource, so any change is visible immediately after the file is refetched.

qooban:
Yes, it's the same functionality. But from the user perspective it may not be as obvious as for us. Let's leave it for now and maybe update during [RD-232](https://cloudifysource.atlassian.net/browse/RD-232).

PR #1134, opened 2020-11-17
RD-494 Moved React components styling out of style.ejs

kubama:


app/components/SideBar.jsx
```diff
@@ -3,18 +3,32 @@
...
+`;
+
 export default function SideBar({ homePageId, isEditMode, isOpen, pageId }) {
     const className = isOpen ? 'open' : '';
+    const theme = useContext(ThemeContext);
 
     return (
         <div className="sidebarContainer">
-            <div className={`ui visible left vertical sidebar menu small basic ${className}`}>
+            <Sidebar className={`ui visible left vertical sidebar menu small basic ${className}`} theme={theme}>
```
qooban, line 29: We can possibly remove some of the classes here, e.g. `visible` and `vertical` are props, `ui` is added by SUIR to most of the components automatically, `sidebar` is possibly added automatically for this component.

Please check [Sidebar in SUIR documentation](https://react.semantic-ui.com/modules/sidebar).

app/components/SideBar.jsx
```diff
@@ -3,18 +3,32 @@
...
 
+const Sidebar = styled.div`
+    background-color: ${props => props.theme.sidebarColor} !important;
+    .item {
+        color: ${props => props.theme.sidebarTextColor} !important;
+    }
+    .item.active,
+    .item:hover {
+        background-color: ${props => props.theme.sideBarHoverActiveColor} !important;
+        color: ${props => props.theme.sidebarTextColor} !important;
```
qooban, line 19: I'm not sure if we need to make it safer (check if theme is set, first). I know it complicates it, so I'm not saying that here it is a must. Just wondering. 

The best would be to use [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) which would simplify the code in many places.

qooban (commented):
Commented. More important is the second comment about cleaning `className`.

app/components/SideBar.jsx
kubama: Addressed by adding `{}` as default theme. In the future I think the best approach would be to create a custom hook returning the theme.

app/components/SideBar.jsx
kubama: The `Sidebar` here was not a SUIR Sidebar, but I get your intention - addressed

app/components/SideBar.jsx
```diff
@@ -3,18 +3,33 @@
...
+import { Sidebar } from './basic';
+
+const ThemedSidebar = styled(Sidebar)`
+    background-color: ${props => props.theme.sidebarColor} !important;
+    .item {
+        color: ${props => props.theme.sidebarTextColor} !important;
+    }
+    .item.active,
+    .item:hover {
+        background-color: ${props => props.theme.sideBarHoverActiveColor} !important;
```
qooban, line 19: We missed one thing. Please take a look at `/style` endpoint (`backend/routes/Style.js`). We need to:
1. Remove irrelevant parts from this endpoint
1. Add `sideBarHoverActiveColor` to userConfig

qooban: Let's remove dynamic shading. It is not working great.

Example with `sidebarColor` set to `'red'`:
image

kubama: > Let's remove dynamic shading. It is not working great.

Could you please elaborate on this?

qooban: I meant `shadeColor` function from `backend/routes/Style.js` file. It calculates the color for `sidebarHoverActiveColor` parameter automatically basing on `sidebarColor` parameter. We don't need it I think. It is better to explicitly specify the color for hover/active case.

kubama: Done

conf/README.md
```diff
@@ -65,18 +65,15 @@ This configuration can be overridden by: `/dist/userData/userConfig.json`.
...
-  * `showVersionDetails` - boolean, if set to true, then version will be shown in page header 
-  * `loginPageHeader` - string, login page header
-  * `loginPageHeaderColor` - string, login page header color (CSS color)  
-  * `loginPageText` - string, login page text under header
+  * `showVersionDetails` - boolean, if set to true, then version will be shown in page header
+  * `loginPageHeaderColor` - string, login page header color (CSS color)
   * `loginPageTextColor` - string, login page text under header color (CSS color) 
   * `sidebarColor` - string, menu sidebar color (CSS color)
   * `sidebarTextColor` - string, menu sidebar text color (CSS color) 
+  * `sideBarHoverActiveColor` - string, menu sidebar color (CSS color) for active and hovered items
```
qooban, line 76: Let's align it to the rest of the parameters (`sideBar...` -> `sidebar...`).
It was not used anywhere else afaik, so that should not be a problem.

kubama: Done

qooban (dismissed):

qooban (approved):

PR #1136, opened 2020-11-18
RD-235 Stage customization support minor fixes

kubama:
* Moved `overrides.json` file loading to earlier stage
* Removed deprecated node API usage
* Removed 'Cloudify' occurrences from backend


qooban (dismissed):
Only small linter issues to fix.

qooban (approved):

PR #1144, opened 2020-11-27
RD-239 Added support for reading log level from logging.conf

kubama:


conf/README.md
```diff
@@ -42,9 +42,9 @@ The following section describes different configuration files used in cloudify-s
   * `portalUrl` - string, redirect URL to the organization portal
 
 * `logsFile` - string, logs file path
-
 * `errorsFile` - string, errors file path
-
+* `logLevelConf` - string, path to Cloudify Manager log levels configuration file, or blank string if not applicable
```
qooban, line 46: This `README.md` file is part of the package, so I would avoid using "Cloudify" word here. 
Please clean the whole file from such references.

backend/handler/LoggerHandler.js
```diff
@@ -1,12 +1,38 @@
...
  * Created by jakubniezgoda on 17/07/2017.
  */
-
+const fs = require('fs');
 const winston = require('winston');
 const _ = require('lodash');
 const config = require('../config').get();
 
-const { logLevel, logsFile, errorsFile } = config.app;
+const { logsFile, errorsFile } = config.app;
```
qooban, line 9: We don't need to set `EventEmitter.defaultMaxListeners` in Stage?

qooban (commented):
Commented.

backend/handler/LoggerHandler.js
kubama: It seems we didn't need until RD-492 was implemented. I think it should be added as part of RD-492, for sure it should be part of 5.1 patch.

conf/README.md
kubama: Done

qooban (approved):

backend/handler/LoggerHandler.js
qooban: It'd be faster for you to add it, but OK, I created: https://github.com/cloudify-cosmo/cloudify-stage/pull/1147

Patch creation is separate thing. I hope it will not be necessary.

PR #1146, opened 2020-11-27
System Tests - Update spire license to work properly on 2-days system test executions

qooban:


kubama (approved):
Are you sure the issue happens only for test executions spanning across days? I'm almost sure I have seen it failing in consecutive builds ran at the same date. Maybe it's worth to escalate it to the backend team?

qooban:
Don't know, I have spotted that only this week. I will merge it and we'll see.

When did you see that, before?

kubama:
> Don't know, I have spotted that only this week. I will merge it and we'll see.

If we just care about our system tests then fine, however if we're considering manager issue then this is of no help for issue investigation - even worse, if it's indeed a manager issue (I'm not 100% sure it is, just discussing) then we are only working around it without reporting it.

> When did you see that, before?

I've seen it weeks before, happening for consecutive builds. However on the next day the issue was gone so I didn't bother.

PR #1151, opened 2020-12-07
RD-385 Use cloudify-ui-common in the backend

kubama:


qooban (commented):
LGTM. The same ask here - let's upgrade cloudify-ui-common for frontend part.

kubama:
> let's upgrade cloudify-ui-common for frontend part.

Done

qooban (approved):

PR #1153, opened 2020-12-11
RD-217 Mocked login process where possible

kubama:
System tests are green: https://jenkins.eks.cloudify.co/view/UI/job/Stage-UI-System-Test/62/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9117,6 +9117,12 @@
         }
       }
     },
+    "cypress-localstorage-commands": {
+      "version": "1.2.5",
+      "resolved": "https://registry.npmjs.org/cypress-localstorage-commands/-/cypress-localstorage-commands-1.2.5.tgz",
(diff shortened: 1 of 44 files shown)


qooban (approved):
Of course that is just part of RD-217. Login endpoint alone is not the most time consuming part of the whole logging in process. There are a lot of fetches (widgets, templates, pages, readmes) just after we get response from the backend.

kubama:
> Of course that is just part of RD-217. Login endpoint alone is not the most time consuming part of the whole logging in process. There are a lot of fetches (widgets, templates, pages, readmes) just after we get response from the backend.

Widgets, templates, pages and readmes are already covered by RD-188 - they are reduced to a minimum required set. In the majority of test suites we are fetching one template, two pages and three widgets - it's not time consuming for the backend to serve these few files. After mocking the log in process itself I don't think there is anything left that could be mocked for any relevant speed up.

PR #1156, opened 2020-12-16
RD-336 Updated project to use common e2e launch script

kubama:
CI build for the branch: https://jenkins.eks.cloudify.co/view/UI/job/Stage-UI-System-Test/69/


scripts/createPackage.sh
```diff
@@ -8,8 +8,8 @@ rm -Rf ${TMP_DIR} ${STAGE_PACKAGE}
 mkdir -p ${TMP_DIR}/cloudify-stage/dist ${TMP_DIR}/cloudify-stage/backend ${TMP_DIR}/cloudify-stage/conf
 
 cp -r dist/** ${TMP_DIR}/cloudify-stage/dist
-rsync -avr --exclude='test/' --exclude='package-lock.json' backend ${TMP_DIR}/cloudify-stage
-rsync -avr --exclude='me.json*' conf ${TMP_DIR}/cloudify-stage
+rsync -avrq --exclude='test/' --exclude='package-lock.json' backend ${TMP_DIR}/cloudify-stage
+rsync -avrq --exclude='me.json*' --exclude='manager.json' conf ${TMP_DIR}/cloudify-stage
```
qooban, line 12: Intentionally including `package-lock.json`?

.size-limit
```diff
@@ -7,7 +7,7 @@
     },
     {
         "name": "Vendor JS bundle",
-        "limit": "4.15 MB",
+        "limit": "4.25 MB",
```
qooban, line 10: Just wondering how that happened?
New additions in cloudify-ui-common? But that's just backend part...

qooban (approved):
Added two questions.

.size-limit
kubama: I think it's not cloudify-ui-common, at least not due to backend addition. On the `next` branch (https://github.com/cloudify-cosmo/cloudify-stage/pull/1151) we already used the backend-enabled 2.1.0 version and it didn't require size limit increase. Adding single sh script also shouldn't increase the size. No idea what it might be.

scripts/createPackage.sh
kubama: Not sure what you mean - `package-lock.json` is not part of the package

qooban: My bad, thought you removed `package-lock.json` exclusion. All good.

PR #1157, opened 2020-12-18
RD-751 - Add Labels Input component

qooban:
### Video
Peek 2021-01-29 09-08

NOTE: Icon in upper right corner is not part of the component. For the testing/debugging/video making purposes I'm using new component directly in the Button Link widget, so that's why it appears there.

### Features
* Input providing (mouse selection with click and keyboard arrows selection or manual type-in)
* Labels key/value auto-completion
* Labels key/value keyboard input validation
* Long labels truncated and popup triggered on hover with full label
* New labels in the system marked using blue color
* Option to revert labels to initial set

`LabelsInput` component further changes (long set of labels handling, permission restrictions, tests) are planned as part of [RD-1217 Enhance Labels Input component](https://cloudifysource.atlassian.net/browse/RD-1217).


widgets/common/src/labels/LabelsList.jsx
```diff
@@ -0,0 +1,55 @@
...
+                            >
+                                {truncatedKey} <span style={{ fontWeight: 'lighter' }}>{truncatedValue}</span>
+                                <Icon
+                                    name="delete"
+                                    onClick={() =>
+                                        onChange(
+                                            _.differenceBy(
+                                                labels,
+                                                [{ key, value }],
+                                                label => label.key === key && label.value === value
```
kubama, line 33:
```suggestion
                                                {key, value}
```

widgets/common/src/labels/common.js
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
+export function addSearchToUrl(url, search) {
+    return search ? `${url}?_search=${search}` : url;
```
kubama, line 3: What if `url` already has a query parameter? I'd reuse `appendQueryParam` from `Filter.jsx` here.

kubama (commented):
Question regarding keyboard interaction - is it possible to add a label with keyboard only?
I think adding multiple labels is a common scenario, do you think it would be a good idea if adding a label would result in focusing the label key field? Users could then start typing next label key right away.

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,123 @@
...
+        isLabelInSystem().then(isInSystem => {
+            const newLabels = [...labels, { key: newLabelKey, value: newLabelValue, isInSystem }];
+            setLabels(newLabels);
+            resetNewLabelKey();
+            resetNewLabelValue();
+        });
+    }
+
+    return (
+        <Segment className={`dropdown selection fluid ${open && 'active'}`} style={{ padding: 0 }}>
```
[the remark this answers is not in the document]
Gelio: I've just merged #1181 so this is doable :slightly_smiling_face: Good point

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -33,6 +34,14 @@ function DynamicDropdown({
     const [loaderVisible, setLoaderVisible] = useState(false);
     const [isLoading, setLoading] = useState(false);
 
+    function normalizeItems(items = []) {
+        return valueProp === '' ? items.map(item => ({ id: item })) : items;
+    }
+
+    function getValueProp() {
+        return valueProp || 'id';
+    }
```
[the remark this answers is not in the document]
qooban: Right. Fixed.

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,123 @@
...
+    function onChangeLabelValue(labelValue) {
+        setNewLabelValue(labelValue);
+    }
+
+    function isAddAllowed() {
+        const isLabelInList = _.findIndex(labels, { key: newLabelKey, value: newLabelValue }) >= 0;
+        const isNewKeyProvided = newLabelKey !== '';
+        const isNewValueProvided = newLabelValue !== '';
+
+        return !isLabelInList && isNewKeyProvided && isNewValueProvided;
```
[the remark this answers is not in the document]
qooban: Makes sense, simplified it.

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    function onAddLabel() {
+        function isLabelInSystem() {
+            const fetchUrl = addSearchToUrl(`/labels/deployments/${newLabelKey}`, newLabelValue);
+            return toolbox
+                .getManager()
+                .doGet(fetchUrl)
+                .then(({ items }) => !_.isEmpty(items))
+                .catch(() => false);
```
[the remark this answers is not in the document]
qooban: I added a bit more error handling, but as it is not critical check, I think it is enough to have it in the browser console. At least for now.

widgets/common/src/LabelsInput.jsx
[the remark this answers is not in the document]
qooban: Yes, you're right. Mentioned PR not merged, yet, so I just fixed it not to return false. Will add a note to RD-1217 in which I'm planning to enhance Labels Input component.

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,123 @@
...
+            const newLabels = [...labels, { key: newLabelKey, value: newLabelValue, isInSystem }];
+            setLabels(newLabels);
+            resetNewLabelKey();
+            resetNewLabelValue();
+        });
+    }
+
+    return (
+        <Segment className={`dropdown selection fluid ${open && 'active'}`} style={{ padding: 0 }}>
+            <div role="presentation" onClick={open ? unsetOpen : setOpen} style={{ cursor: 'pointer' }}>
```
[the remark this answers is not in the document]
qooban: Good point. Tried that, but I found that extending `useBoolean` might not be the best idea. I think that it is not very common that you need both set/unset functions and toggle function. So toggle-like uses of `useBoolean` would look like:
```const [value, , , toogle] = useBoolean(false);
```, 
so I decided to create simple `useToggle` hook.

File: widgets/common/src/labels/CommonDropdown.jsx [hunk 8]
```diff
@@ -0,0 +1,89 @@
...
+            <DynamicDropdown
+                clearable={false}
+                fetchAll
+                fetchUrl={fetchUrl}
+                onChange={onValueChange}
+                onSearchChange={onSearchChange}
+                searchQuery={typedValue}
+                toolbox={toolbox}
+                value={selectedValue}
+                valueProp=""
```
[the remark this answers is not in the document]
qooban: Yes, it is necessary with the current implementation. 

By knowing that `valueProp` is empty string `DynamicDropdown` is aware that data fetched will be in the format of array of strings and not array of objects with a specific key inside (like `id`).

widgets/common/src/labels/CommonDropdown.jsx
```diff
@@ -0,0 +1,89 @@
...
+        unsetShowError();
+    }
+
+    useDebouncedSetValue(baseFetchUrl ? addSearchToUrl(baseFetchUrl, typedValue) : '', setFetchUrl, [
+        baseFetchUrl,
+        typedValue
+    ]);
+
+    useEffect(() => {
+        if (value === '') {
```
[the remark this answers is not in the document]
qooban: Actually in that case it should not matter as we always use strings as values, but I changed that.

widgets/common/src/labels/LabelsList.jsx
qooban: Nice, changed.

widgets/common/src/labels/common.js
qooban: Thanks, I was not aware of `appendQueryParam`. Actually found out that we have multiple functions for appending query params in both frontend and backend part. Both are based on jquery param function. 

For now I externalized `appendQueryParam` function and reused it in `labels/common.js` and in `Filter.jsx`.

I'll add an issue in technical debt epic to go even further and reuse it in cloudify-stage.

widgets/common/src/labels/LabelsList.jsx
```diff
@@ -0,0 +1,55 @@
...
+export default function LabelsList({ labels, onChange }) {
+    const { Label, Icon, Popup } = Stage.Basic;
+    const maxLength = 20;
+    const newLabelColor = 'blue';
+
+    return (
+        <div className="ui multiple dropdown" style={{ paddingRight: '4.1em', minHeight: '2em' }}>
+            {_.map(labels, ({ key, value, isInSystem = true }) => {
+                const truncatedKey = _.truncate(key, { length: maxLength });
+                const truncatedValue = _.truncate(value, { length: maxLength });
```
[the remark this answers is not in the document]
qooban: No, but that is good point to consider in RD-1217. I'll add a point there.

widgets/common/src/labels/LabelsList.jsx
```diff
@@ -0,0 +1,55 @@
...
+export const LabelsPropType = PropTypes.arrayOf(LabelPropType);
+
+export default function LabelsList({ labels, onChange }) {
+    const { Label, Icon, Popup } = Stage.Basic;
+    const maxLength = 20;
+    const newLabelColor = 'blue';
+
+    return (
+        <div className="ui multiple dropdown" style={{ paddingRight: '4.1em', minHeight: '2em' }}>
+            {_.map(labels, ({ key, value, isInSystem = true }) => {
```
Gelio, line 12: What's the benefit of lodash's `map` over `labels.map`?

widgets/common/src/labels/LabelsList.jsx
```diff
@@ -0,0 +1,55 @@
...
+                const truncatedKey = _.truncate(key, { length: maxLength });
+                const truncatedValue = _.truncate(value, { length: maxLength });
+                const allowPopup = truncatedKey !== key || truncatedValue !== value;
+
+                return (
+                    <Popup key={`${key}:${value}`} open={allowPopup ? undefined : false} wide>
+                        <Popup.Trigger>
+                            <Label
+                                as="a"
```
Gelio, line 21: I appreciate using the `as` pattern :smile: Helps with being semantically correct

widgets/common/src/labels/LabelsList.jsx
```diff
@@ -0,0 +1,55 @@
...
+                                color={!isInSystem ? newLabelColor : undefined}
```
Gelio, line 22: I reckon the negation here is unnecessary, as we still need to provide values for both cases :smile: 

```suggestion
                                color={isInSystem ? undefined : newLabelColor}
```

widgets/.eslintrc.json
```diff
@@ -17,7 +17,8 @@
         "moment": true,
         "PropTypes": true,
         "React": true,
-        "Stage": true
+        "Stage": true,
+        "styled": true
```
Gelio, line 21: I did not notice this being used anywhere :thinking: Is it necessary?

widgets/common/src/LabelsInput.jsx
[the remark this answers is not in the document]
qooban: Haven't seen the previous comment. Used `combineClassNames`.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -33,6 +34,12 @@ function DynamicDropdown({
     const [loaderVisible, setLoaderVisible] = useState(false);
     const [isLoading, setLoading] = useState(false);
 
+    const normalizedValueProp = valueProp || 'id';
+
+    function normalizeItems(items = []) {
```
Gelio, line 39: Should we not be passing `normalizeItems` (a new optional prop) and the `valueProp` in the parent components, instead of adding some defaults here?

At the moment it looks like we are tailoring the general component (`DynamicDropdown`) for some specific use-cases that are only used in the Labels case.

If we provide such props from parents, we're using dependency injection to customize the behavior and not make the general component (`DynamicDropdown`) more complex :slightly_smiling_face:

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,126 @@
...
+        Utils: { combineClassNames },
+        i18n
+    } = Stage;
+
+    const [labels, setLabels, resetLabels] = useResettableState(initialValue);
+    const [open, toggleOpen] = useToggle();
+    const [newLabelKey, setNewLabelKey, resetNewLabelKey] = useResettableState('');
+    const [newLabelValue, setNewLabelValue, resetNewLabelValue] = useResettableState('');
+
+    const iconStyle = {
```
Gelio, line 22: How about extracting this object to the module scope, to avoid recreating it on every render?

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,126 @@
+import AddButton from './labels/AddButton';
+import LabelsList, { LabelsPropType } from './labels/LabelsList';
+import KeyDropdown from './labels/KeyDropdown';
+import ValueDropdown from './labels/ValueDropdown';
+import { addSearchToUrl } from './labels/common';
+
+export default function LabelsInput({ initialValue, onChange, toolbox }) {
```
Gelio, line 7: Can we change `initialValue` to `initialLabels`? This way the variable name provides a bit more contextual information (we know the values are labels, not arbitrary "values")

```suggestion
export default function LabelsInput({ initialLabels, onChange, toolbox }) {
```

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,126 @@
...
+        onChange(labels);
+    }, [labels]);
+
+    useEffect(() => {
+        setLabels(initialValue);
+    }, [initialValue]);
+
+    function onChangeLabelKey(labelKey) {
+        setNewLabelKey(labelKey);
+        resetNewLabelValue();
```
Gelio, line 38: I'm not sure about the UX here. Fixing a typo in the label key should not cause the value to be erased, IMO :smile: I would leave it up to the user to reset the value, if they want

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    function onChangeLabelValue(labelValue) {
```
Gelio, line 41: How about using `setNewLabelValue` in the place that `onChangeLabelValue` was used? This way we avoid a function that only calls another function with the same parameter :slightly_smiling_face:

widgets/common/src/LabelsInput.jsx
```diff
@@ -0,0 +1,126 @@
...
+        setNewLabelValue(labelValue);
+    }
+
+    function isAddAllowed() {
+        return newLabelKey && newLabelValue && !_.find(labels, { key: newLabelKey, value: newLabelValue });
+    }
+
+    function onAddLabel() {
```
Gelio, line 49: I would consider making the add button disabled when the request is in flight :slightly_smiling_face: This avoids creating multiple labels when the user spams the "Add" button :smile:

app/utils/hooks/useToggle.js
```diff
@@ -0,0 +1,13 @@
+import { useState } from 'react';
+
+/**
+ * Returns a stateful boolean value and one function to toggle it
+ *
+ * @param initialValue
+ * @returns {(boolean|(function(): void))[]}
```
Gelio, line 7: This is a bit more precise type (also a correct TypeScript type :smile: ) that works well in VSCode:
```suggestion
 * @returns {[boolean, () => void]}
```

I'm not sure if WebStorm parses it correctly too

app/utils/hooks/useToggle.js
```diff
@@ -0,0 +1,13 @@
...
+ */
+export default function useToggle(initialValue = false) {
```
Gelio, line 9: I like the idea of adding this hook instead of modifying `useBoolean` :+1:

app/translations/en.json
```diff
@@ -369,6 +369,14 @@
     },
     "widgets": {
         "common": {
+            "labels": {
+                "keyPlaceholder": "Select or type in a key",
+                "newKey": "New key",
+                "newValue": "New value",
+                "valuePlaceholder": "Select or type in a value",
+                "validationError": "Only letters, digits and the characters \"-\", \".\" and \"_\" are allowed and it must begin with a letter.",
+                "isLabelInSystemCheckError": "Cannot check is label already present in the system",
```
Gelio, line 378: A trailing comma here is invalid JSON :worried: 

```suggestion
                "isLabelInSystemCheckError": "Cannot check is label already present in the system"
```

Gelio (commented):
Very nice component decomposition :+1: I really like it

I did not test this PR in the UI

A general note about GIFs - I prefer using MP4 instead of GIFs when recording videos and uploading them to GitHub, because they take less space, and you can pause/stop/seek them, whereas with gifs you just get a single loop, and have no idea how long the video is :slightly_smiling_face:

widgets/common/src/labels/CommonDropdown.jsx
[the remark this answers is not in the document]
Gelio: I'm actually for being as precise with code as possible :smile: value === ''` shows that we handle only strings, thus we have to be precise on other places to pass this function a string, otherwise it will break. Using `isEmpty` and not using TypeScript does not allow any contextual information about the possible values of `value`. It gives a bit more safety, since this function now silently handles every data type, but should it? In the case that it receives something unexpected (e.g. an object :smile: ) I would rather it fail then and there, rather than go on and treat it as valid input :slightly_smiling_face: 

Also, that's 1 fewer use of lodash :smile: 

I do not expect any changes, just giving my opinion :slightly_smiling_face:

widgets/common/src/LabelsInput.jsx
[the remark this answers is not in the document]
kubama: I think I'd rather modify `useBoolean` to return an object instead of an array, and include toggle function in it - would be much more flexible and equally elegant.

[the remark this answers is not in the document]
kubama: Hm, actually I think an array is better suited for boolean/toggle hooks, since it prevents constructs like: `const {value: open, setValue: doOpen, unsetValue: doClose} = useBoolean()`. So I think your solution is optimal.

widgets/common/src/RevertToDefaultIcon.jsx
```diff
@@ -0,0 +1,44 @@
+function RevertToDefaultIcon({ value, defaultValue, onClick, style }) {
+    const { i18n } = Stage;
+    const { Icon, Popup } = Stage.Basic;
+
+    return !_.isUndefined(defaultValue) && !_.isEqual(value, defaultValue) ? (
```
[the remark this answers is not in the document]
qooban: The whole component was just moved out from `InputUtils` and I did not focus on improving it.
Changed according to your suggestions.

widgets/common/src/RevertToDefaultIcon.jsx
```diff
@@ -0,0 +1,44 @@
...
+        <Popup trigger={<Icon name="undo" link onClick={onClick} style={style} />}>
+            {i18n.t('widgets.common.revertToDefault')}
+        </Popup>
+    ) : null;
+}
+
+RevertToDefaultIcon.propTypes = {
+    /**
+     * value typed field value
```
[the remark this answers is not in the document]
qooban: This kind of comments were used differently before, that's why it is not a sentenc. First word was a name of the prop, it was used by one of the documentation frameworks AFAIR. Updated.

widgets/common/src/labels/CommonDropdown.jsx
```diff
@@ -0,0 +1,91 @@
+import { addSearchToUrl } from './common';
+import ValidationErrorPopup from './ValidationErrorPopup';
+
+function useDebouncedSetValue(value, setValue, deps) {
+    const { useCallback, useEffect } = React;
+    const delay = 500; // ms
```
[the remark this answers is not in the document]
qooban: Updated.

widgets/common/src/labels/CommonDropdown.jsx
```diff
@@ -0,0 +1,91 @@
...
+    } = Stage;
+
+    const [selectedValue, setSelectedValue, resetSelectedValue] = useResettableState('');
+    const [typedValue, setTypedValue, resetTypedValue] = useResettableState('');
+    const [showError, setShowError, unsetShowError] = useBoolean();
+    const [fetchUrl, setFetchUrl] = useState(baseFetchUrl);
+
+    function onSearchChange(event, { searchQuery: newTypedValue }) {
+        const allowedCharacters = /^[a-z][a-z0-9._-]*$/i;
+        const lowercasedNewTypedValue = _.toLower(newTypedValue);
```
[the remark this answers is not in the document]
qooban: Yes, it should do that same. The benefit of using lodash (as I see it) is that it prevents from unexpected function calls on `undefined` or `null`. Also, lodash is not very big dep and it is already included.

IMHO we can keep it for now, at least until we get support for optional chaining / nullish coalescing.

widgets/common/src/labels/LabelsList.jsx
qooban: See my previous answer.

widgets/.eslintrc.json
qooban: Good catch. No, it is not used now (I used it in draft implementation only).

widgets/common/src/LabelsInput.jsx
qooban: Makes sense. Done.

widgets/common/src/LabelsInput.jsx
qooban: Yep, changed.

widgets/common/src/LabelsInput.jsx
qooban: Good point, changed. Also removed `onChangeLabelKey` in favor of just `setNewLabelKey`.

widgets/common/src/LabelsInput.jsx
qooban: Of course, changed.

app/utils/hooks/useToggle.js
qooban: image

I'm not sure how it should be for WebStorm, but neither the original part (taken from `useBoolean`) nor the one you suggested is visible in the popup provided by the WebStorm. Don't know if there is any other way to trigger it.

Anyways, I can change it according to your suggestion if that is the valid way (similiarly I can change `useBoolean`). @kubama do you see it differently?

qooban: Works fine in WebStorm. Changed.

Still there are other places using that syntax - `useResettableState`, `useBoolean`, `useErrors`. That can be handled in a separate issue, I think.

webpack.config.js
```diff
@@ -249,9 +249,13 @@ module.exports = (env, argv) => {
             mode,
             context,
             devtool,
+            resolve: {
+                extensions: ['.js', '.jsx']
+            },
             entry: glob
                 .sync('./widgets/common/src/props/*.js')
                 .concat(glob.sync('./widgets/common/src/hooks/*.js'))
+                .concat(glob.sync('./widgets/common/src/!(props|hooks)/*.js*'))
```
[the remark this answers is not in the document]
qooban: I agree with you, it's not perfect, but I just extended what was already. I spent some short time looking for better solution before I created that PR, but I ended up with that solution. Don't want to spend more time on that. 

If you find another way, I'd be more than happy to see the solution.

widgets/common/src/LabelsInput.jsx
qooban: Make sense, I've added your suggestion to RD-1217, which will be implemented later during Labels development.

widgets/common/src/labels/common.js
qooban: Issue created [RD-1237](https://cloudifysource.atlassian.net/browse/RD-1237).

widgets/common/src/DynamicDropdown.jsx
qooban: Very good point. Simplified it a bit differently than you suggested, but at the end I think the result is better than it was.
Removed `normalizedValueProp` const and created `itemsFormatter` prop.

qooban:
Regarding general comments

@kubama 

> Question regarding keyboard interaction - is it possible to add a label with keyboard only?

Yes, it is possible to add a label with keyboard only:
0. Focus on key dropdown using `TAB` key
1. Type in key, press `TAB` to focus on value dropdown
2. Type in value, press `TAB` to focus on add button
3. Press `ENTER` to add label

From this point you can press `SHIFT` + `TAB` to get back to the key dropdown.

> I think adding multiple labels is a common scenario, do you think it would be a good idea if adding a label would result in focusing the label key field? Users could then start typing next label key right away.

Yes, in my draft implementation (before asking you for review) I had that (the focus was automatically changed after adding new label to key dropdown). But after some testing I found it quite annoying, so I removed that. I can consider it again in RD-1217.

@Gelio

> A general note about GIFs - I prefer using MP4 instead of GIFs when recording videos and uploading them to GitHub, because they take less space, and you can pause/stop/seek them, whereas with gifs you just get a single loop, and have no idea how long the video is

Yes, makes sense. I already changed my default settings in [Peek](https://github.com/phw/peek).

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -218,7 +214,8 @@ DynamicDropdown.defaultProps = {
     prefetch: false,
     multiple: false,
     className: '',
-    refreshEvent: null
+    refreshEvent: null,
+    itemsFormatter: options => options
```
Gelio, line 218: Having it named `options` may be misleading, as it's not options, but the items from the request. How about

```suggestion
    itemsFormatter: items => items
```

widgets/common/src/RevertToDefaultIcon.jsx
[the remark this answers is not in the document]
Gelio: I noticed that after posting comments :smile: Thanks for making the changes :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
qooban: My bad, fixed.

widgets/common/src/labels/CommonDropdown.jsx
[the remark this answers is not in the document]
Gelio: Sure thing :slightly_smiling_face: 

btw. in the future, if we use TypeScript, I always suggest using `strictNullChecks`, which forces us to explicitly tell when a variable can be `null` or `undefined`. This may seem like a bit of overhead, but has many benefits - you precisely know when a variable can be null, and only use optional chaining or some other way to get around it :slightly_smiling_face: More certainty about what possible values you can have. `strictNullChecks` have been adopted by some TS codebases that documented that process, e.g. [Figma](https://www.figma.com/blog/inside-figma-a-case-study-on-strict-null-checks/) or [VSCode](https://code.visualstudio.com/blogs/2019/05/23/strict-null).

I'm saying that just FYI, to give you some insights about what TS allows for and what I will recommend in the future :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
Gelio: I'm glad you liked it :smile:

widgets/common/src/LabelsInput.jsx
Gelio: Cool, thanks for acting on it :slightly_smiling_face:

app/utils/hooks/useToggle.js
Gelio: Yup, definitely :smile: We can refactor it as we go :slightly_smiling_face:

Gelio (commented):
Just to be on the safe side - does `DynamicDropdown` get correct `valueProp`s now? I failed to spot where it's set to `id` (which I assume it should be set to :smile: )

btw. I've also used Peek, but it's not available on Ubuntu 20.10 :worried: If you were to upgrade and look for an alternative, I'm using [SimpleScreenRecorder](https://www.maartenbaert.be/simplescreenrecorder/) now. It isn't as nice as Peek, but gets the job done :slightly_smiling_face:

webpack.config.js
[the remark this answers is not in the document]
Gelio: Would it not suffice to go with a single `./widgets/common/src/**.*.js`? I am actually not sure what is the point of this last part of the webpack config. Is it to produce a single chunk with all items inside `widgets/common/src`?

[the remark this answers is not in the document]
qooban: > Would it not suffice to go with a single ./widgets/common/src/**.*.js?

No, unfortunately, tried that at the beginning. The order matters here. I think the reason for that is that there is no single entry point for that `common.js` chunk, so it is created by getting all files one by one. And in some cases it doesn't get the order right (for example `hooks` and `props` folders which are used by the components have to be added to the chunk before components to make it work in runtime).

> I am actually not sure what is the point of this last part of the webpack config. Is it to produce a single chunk with all items inside `widgets/common/src`?

Yes, that was I think the original intention. It was flat at the beginning (just components, no inter-component dependencies).
It should be loaded before all other `widget.js` bundles is loaded.

qooban:
@Gelio Yes, by default `valueProp` is set to `id` in `DynamicDropdown` (through `defaultProps`). 
So as we are now not passing it from `CommonDropdown` it is set to default automagically.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -202,7 +214,8 @@ DynamicDropdown.defaultProps = {
     prefetch: false,
     multiple: false,
     className: '',
-    refreshEvent: null
+    refreshEvent: null,
+    itemsFormatter: items => items
```
kubama, line 218: I'd reuse already existing function:
```suggestion
    itemsFormatter: _.identity
```

kubama (approved):

webpack.config.js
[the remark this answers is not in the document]
Gelio: I see. Then, maybe it would make sense to have a single entry-point that just imports everything in those folders, in correct order?

That may be a task for the future, though :slightly_smiling_face:

Gelio (approved):
Looks good! :+1:

webpack.config.js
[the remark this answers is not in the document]
qooban: Yep. That was also one of my ideas 🙂  
Created [RD-1280](https://cloudifysource.atlassian.net/browse/RD-1280) to handle that in the future.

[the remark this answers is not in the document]
Gelio: Cool, thanks :+1: I hooked it up to the UI Tech Debt epic

PR #1158, opened 2020-12-22
RD-833 Split app.json into 3 smaller files, updated RPM spec

kubama:


conf/app.json
```diff
@@ -5,12 +5,6 @@
   "db": {
     "url": "postgres://cloudify:cloudify@localhost:5432/stage",
     "options": {
-      "dialect": "postgres",
```
qooban, line 8: Why extract just part of the options? 
As far as I know `db.url` is the only part modified by the manager.
With that change `app.json` will still be modified by both - stage and manager - and that we want to avoid, right?

conf/README.md
```diff
@@ -41,13 +43,22 @@ The following section describes different configuration files used in cloudify-s
   * `ssoUrl` - string, redirect URL to the application at the Single Sign-On identity provider
   * `portalUrl` - string, redirect URL to the organization portal
 
+## DB Options (`db.options.json`)
```
qooban, line 46: I think we need DB URL extracted, the rest could still be part of `app.json`.

conf/app.json
kubama: `db.url` is not the only part modified by the manager installer, some `db.options` are also being set. The idea is to make the `app.json` the 'external' file, so it could be freely modified by the installer - thus configuration portions that were altered in 5.1.1 were moved to separate files to avoid the conflict. As agreed with @alexmolev this is the simplest solution that doesn't require changes in manager installer.

conf/README.md
kubama: Please see my previous comment

qooban (approved):

qooban:
Please check if any updates needed in https://docs.cloudify.co/ in relation to this PR.

PR #1162, opened 2020-12-30
RD-869 Added migration script for updating UserApp.managerIp

kubama:


YoniItzhak:
@kubama, looks fine to me, but I guess we should let @alexmolev review it as he's more familiar with the syntax.

alexmolev (approved):

PR #1165, opened 2021-01-05
RD-725 run tests on master

yuvlar:


jenkins/Jenkinsfile
```diff
@@ -161,13 +161,24 @@ pipeline {
...
         }
       }
     }
+    stage('Run Stage-UI System-tests') {
+      when {
+        branch "master"
+      }
+      steps {
+        echo 'Trigger Stage-UI System-tests'
+        build(job: 'Stage-UI-System-Test')
```
warolv, line 180: @yuval, you need to send params there I think ? like branch name ...

warolv (changes requested):

jenkins/Jenkinsfile
yuvlar: Fixed

warolv (approved):

PR #1166, opened 2021-01-07
RD-625 Asynchronous blueprints upload

kubama:
New column in blueprints widget (error details are shown on icon hover):
image

In catalog mode:
image


ofercloudify:
We should try to avoid breaking to the next line in the majority of the use cases.
If there is enough real estate and we can expand that column a bit it will be best. If not, perhaps show the icon + Failed, and share the exact issue in the tooltip.
Other alternatives?

kubama:
I don't think we have room to expand this column - to save space and prevent wrapping I introduced a short 'Failed' label common to all failed states and moved full state label to popup header. I think it works quite well - please see the updated screenshot.

ofercloudify:
Indeed much better. Works for me.

widgets/common/src/UploadBlueprintBasicForm.jsx
```diff
@@ -0,0 +1,97 @@
...
+            </Form>
+        </>
+    );
+}
+
+UploadBlueprintBasicForm.propTypes = {
+    blueprintName: PropTypes.string,
+    blueprintUploading: PropTypes.bool,
+    blueprintYamlFile: PropTypes.string,
+    children: PropTypes.arrayOf(PropTypes.element),
```
qooban, line 73: I understand from the code construction (L41 and L63) is that we actually accept max. 2 elements - one to be added  before the common blueprint form fields and one element to be added after the common blueprint form fields. In case there is no element or 2 elements, it's fine.

The case when just one is provided could result in runtime error as there will be no `children[1]`.
So I'd suggest to handle that case in the code (just improve the check) and maybe add a short note here in the props (if not possible to make it obvious from the prop definition what kind of array - in terms of length - we expect).

widgets/common/src/BlueprintActions.js
```diff
@@ -85,7 +110,42 @@ class BlueprintActions {
...
+
+            // eslint-disable-next-line no-await-in-loop
+            const blueprint = await this.doGetFullBlueprintData({ id: blueprintName });
+
+            if (blueprint.state === BlueprintActions.CompletedBlueprintStates.Uploaded) {
+                return;
+            }
+
+            if (BlueprintActions.CompletedBlueprintStates[blueprint.state]) {
+                const error = Error(blueprint.error);
```
qooban, line 135: I know that it works the same, but in all other places (except one I found which should be fixed) we use `new Error` construction instead of just `Error`. I think for the sake of consistency, let's use the same in all places.

widgets/blueprints/src/BlueprintsTable.jsx
```diff
@@ -62,45 +64,64 @@ export default function BlueprintsTable({
...
-                                    event.stopPropagation();
-                                    new Stage.Common.BlueprintActions(toolbox).doEditInComposer(
-                                        item.id,
-                                        item.main_file_name
-                                    );
-                                }}
-                            />
+                        {BlueprintActions.CompletedBlueprintStates[item.state] && (
+                            <>
+                                {item.state === BlueprintActions.CompletedBlueprintStates.Uploaded && (
```
qooban, line 85: This check is present in different places: 
```item.state === BlueprintActions.CompletedBlueprintStates.Uploaded``` 
Maybe add common function for that? In BlueprintActions?

File: widgets/blueprints/src/BlueprintsCatalog.jsx [hunk 4]
```diff
@@ -51,35 +52,64 @@ export default function BlueprintsCatalog({
...
+                            <Grid.Column width="9">
+                                {_.words(_.startCase(item.state))[0]}
+                                {item.error && (
+                                    <Popup
+                                        offset={[-11, 0]}
+                                        trigger={<Icon link name="warning circle" color="red" />}
+                                        content={item.error}
+                                        header={_.capitalize(_.startCase(item.state))}
+                                    />
+                                )}
```
qooban, line 104: As this is used in both type of views (table, catalog), this could be extracted to a separate component, I think.

qooban (commented):
Added some comments.

I did not test how it works, only reviewed the code. If I find some time, I'll also do some user tests. 

Probably not within this issue, but in the future - as we are having more and more parameters here - I would suggest to make visibility of the columns configurable (like it is done in Executions widget).

kubama:
> I did not test how it works, only reviewed the code. If I find some time, I'll also do some user tests.

AFAIK changes on manager side are not yet completed, so I don't think you can do user tests yet.
I'll submit another review request once it's done and I verify the implementation.

widgets/common/src/UploadBlueprintForm.jsx
```diff
@@ -134,16 +134,30 @@ class UploadBlueprintForm extends React.Component {
 
     render() {
         const { loading: loadingState, yamlFiles } = this.state;
-        const { blueprintFileName, blueprintName, errors, loading: loadingProp, showErrorsSummary } = this.props;
+        const {
+            blueprintFileName,
```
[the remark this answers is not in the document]
kubama: Done

widgets/common/src/UploadBlueprintForm.jsx
```diff
@@ -134,16 +134,30 @@ class UploadBlueprintForm extends React.Component {
...
-                errors={showErrorsSummary ? errors : null}
+            <UploadBlueprintBasicForm
+                errors={showErrorsSummary ? errors : {}}
+                blueprintYamlFile={blueprintFileName}
+                blueprintName={blueprintName}
+                yamlFiles={yamlFiles}
+                uploadState={uploadState}
+                formLoading={loadingState}
+                blueprintUploading={loadingProp}
+                yamlFileHelp="If you choose archive as blueprint package, you must specify
```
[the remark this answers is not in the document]
kubama: Done

widgets/common/src/UploadBlueprintBasicForm.jsx
```diff
@@ -0,0 +1,97 @@
+const { InProgressBlueprintStates, CompletedBlueprintStates } = Stage.Common.BlueprintActions;
```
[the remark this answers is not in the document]
kubama: Done

widgets/common/src/UploadBlueprintBasicForm.jsx
```diff
@@ -0,0 +1,97 @@
...
+    onErrorsDismiss,
+    onInputChange,
+    uploadState,
+    yamlFileHelp,
+    yamlFiles
+}) {
+    const { Form, LoadingOverlay } = Stage.Basic;
+
+    return (
+        <>
```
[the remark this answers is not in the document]
kubama: I don't think so - removed

widgets/common/src/UploadBlueprintBasicForm.jsx
kubama: Solved by adding support for variable length of `children` array

widgets/common/src/BlueprintActions.js
kubama: If `new` is optional then I'd rather go the other way around and try to avoid it. But since this is a cosmetic thing I'd leave existing occurrences untouched for now.

widgets/blueprints/src/BlueprintsTable.jsx
kubama: Done

widgets/blueprints/src/BlueprintsCatalog.jsx
kubama: Done

widgets/blueprints/src/BlueprintState.jsx
```diff
@@ -0,0 +1,23 @@
+export default function BlueprintState({ blueprint }) {
+    const { Icon, Popup } = Stage.Basic;
+    return (
+        <>
+            {_.words(_.startCase(blueprint.state))[0]}
+            {blueprint.error && (
+                <Popup
+                    offset={[-11, 0]}
+                    trigger={<Icon link name="warning circle" color="red" />}
```
qooban, line 9: Right now just FYI: Probably we need to align it with other places. I can't find the exact place right now, but I'm almost sure that  we are not consistent in reporting errors this way (icon + popup).

app/translations/en.json
```diff
@@ -384,6 +384,39 @@
...
                 "helloWorldBlueprint": "https://github.com/cloudify-cosmo/cloudify-hello-world-example/archive/master.zip",
                 "linkParameterExample": "https://cloudify.co"
+            },
+            "blueprintUpload": {
+                "uploadLabels": {
+                    "pending": "0/4: Waiting for blueprint upload to start...",
+                    "uploading": "1/4: Uploading blueprint...",
+                    "extracting": "2/4: Extracting blueprint...",
+                    "parsing": "3/4: Parsing blueprint...",
+                    "uploadingImage": "4/4: Uploading image..."
```
qooban, line 394: AFAIR it was Ofer who pointed out that we should start from 1 and not 0 here. I'm leaning towards his opinion as this will be shown to the user. What do you think?

widgets/blueprints/src/BlueprintsTable.jsx
```diff
@@ -62,45 +65,56 @@ export default function BlueprintsTable({
                     <DataTable.Data>{item.updated_at}</DataTable.Data>
                     <DataTable.Data>{item.created_by}</DataTable.Data>
                     <DataTable.Data>{item.main_file_name}</DataTable.Data>
+                    <DataTable.Data>
+                        <BlueprintState blueprint={item} />
```
qooban, line 69: I see that blueprint in failed state provides no action click (on uploaded correctly user can drill down), but probably for even better UX we should change the cursor not to be `pointer` which suggests that it can be clicked on such blueprints:

image

What do you think?

qooban (commented):
Added some comments. I saw that system tests are passing - great. 

Did quick user tests. I uploaded few blueprints and have never seen stage higher than 1 (only 0 and 1). Are 2, 3, 4 so short or we are polling not so frequently?

kubama:
> Did quick user tests. I uploaded few blueprints and have never seen stage higher than 1 (only 0 and 1). Are 2, 3, 4 so short or we are polling not so frequently?

When testing myself I was constantly getting phase 0 and then 3, it heavily depends on the actual blueprint you are using. BTW phase 4 is image upload, so it's empty if no image is used.
We are polling with a constant one second interval - ideally I think we could start polling with a shorter interval (200ms for example), then after each poll gradually extend the interval by, lets say 100ms until we reach 1 second. So the polling scheme could look like: 200ms, 300ms, ..., 900ms, 1s, 1s, 1s, ...
I guess it could improve experience for simple, fast uploading blueprints. What do you think?

app/translations/en.json
kubama: Definitely - updated

widgets/blueprints/src/BlueprintsTable.jsx
kubama: Right, done

widgets/common/src/UploadBlueprintBasicForm.jsx
```diff
@@ -0,0 +1,99 @@
+const { InProgressBlueprintStates, CompletedBlueprintStates } = Stage.Common.BlueprintActions;
+
+const UploadLabels = _(InProgressBlueprintStates)
+    .keyBy()
+    .mapValues(value => Stage.i18n.t(`widgets.common.blueprintUpload.uploadLabels.${_.camelCase(value)}`))
```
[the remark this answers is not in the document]
kubama: Done

qooban:
> So the polling scheme could look like: 200ms, 300ms, ..., 900ms, 1s, 1s, 1s, ... I guess it could improve experience for simple, fast uploading blueprints. What do you think?

Yes, seems like a good idea 👍🏻

widgets/common/src/BlueprintActions.js
```diff
@@ -80,12 +113,47 @@ class BlueprintActions {
...
         }
 
-        return promise.then(() => this.doUploadImage(blueprintName, imageUrl, image));
+        return promise
+            .then(() => this.waitUntilUploaded(blueprintName, onStateChanged))
+            .then(() => onStateChanged(BlueprintActions.InProgressBlueprintStates.UploadingImage))
+            .then(() => this.doUploadImage(blueprintName, imageUrl, image));
+    }
+
+    async waitUntilUploaded(blueprintName, onStateChanged, maxNumberOfRetries = 60, waitingInterval = 1000 /* ms */) {
```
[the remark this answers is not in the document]
kubama: Created `PollHelper` class to accommodate the common part, it also covers increasing polling interval.
`waitUntilEmpty` works in a different way, it uses `cy.wait` instead of `setTimeout` among other differences, so I don't think it's a candidate for fitting into the common solution.

backend/routes/ServerProxy.js
```diff
@@ -43,18 +42,13 @@ function buildManagerUrl(req, res, next) {
 
 async function proxyRequest(req, res) {
     const options = {};
-    let timeout;
 
-    // if is a blueprint upload request = set higher timeout
```
Gelio, line 48: Nice unhack :smile:

widgets/common/src/UploadBlueprintBasicForm.jsx
```diff
@@ -0,0 +1,101 @@
...
+
+    return (
+        <Form
+            loading={formLoading}
+            errorMessageHeader={!_.isEmpty(errors) ? UploadErrorHeaders[uploadState] : null}
+            errors={errors}
+            onErrorsDismiss={onErrorsDismiss}
+        >
+            {blueprintUploading && <LoadingOverlay message={UploadLabels[uploadState]} />}
+            {_.head(children)}
```
Gelio, line 42: Instead of expecting a very specific form of `children` (one that is an array), maybe we could expect this element as a separate prop, and the rest as another prop?

Then, it would be clear that there are 2 "slots" in this component - one that goes in the header and one that goes at the bottom. When passing an array of children to this component I would not expect for them to be broken up, as they are now :slightly_smiling_face:

Gelio (commented):
This is a partial review. I will continue the review tomorrow :slightly_smiling_face:

widgets/filter/src/Filter.jsx
```diff
@@ -159,13 +159,12 @@ export default class Filter extends React.Component {
...
                 const { [stateProp || `${camelCaseEntityName}Id`]: value } = this.state;
+                const url = `/${fetchManagerEndpoint || `${entityName.replace(' ', '-').toLowerCase()}s`}`;
                 return (
                     <Form.Field key={entityName}>
                         <DynamicDropdown
                             multiple={configuration.allowMultipleSelection}
-                            fetchUrl={`/${
-                                fetchManagerEndpoint || `${entityName.replace(' ', '-').toLowerCase()}s`
-                            }?_include=${_(filter)
+                            fetchUrl={`${url}${url.indexOf('?') > 0 ? '&' : '?'}_include=${_(filter)
```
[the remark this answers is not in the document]
kubama: If I understand correctly the `URL` class works with absolute URLs only, while here we have a relative URL. A workaround might be to use kind of dummy base URL just to create an instance, but this would be ugly as well.

widgets/blueprintCatalog/src/UploadModal.jsx
```diff
@@ -102,39 +111,18 @@ export default function UploadModal({
...
-                                onChange={handleInputChange}
-                            />
-                        </Form.Field>
-                        <Form.Field
-                            label="Blueprint YAML file"
-                            required
-                            error={errors.blueprintYamlFile}
-                            help="As you can have more than one yaml file in the archive,
+                    <UploadBlueprintBasicForm
+                        yamlFileHelp="As you can have more than one yaml file in the archive,
```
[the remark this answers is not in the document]
kubama: We use `en.json` for strings used in core application as well as strings used in components common to many widgets (widgets/common directory). For strings used in specific widgets only we were planning to introduce separate per-widget translation files, but this is not yet implemented.

widgets/common/src/UploadBlueprintBasicForm.jsx
```diff
@@ -0,0 +1,101 @@
+const { InProgressBlueprintStates, CompletedBlueprintStates } = Stage.Common.BlueprintActions;
+
+const { i18n } = Stage;
+
+const UploadLabels = _(InProgressBlueprintStates)
+    .keyBy()
+    .mapValues(value => i18n.t(`widgets.common.blueprintUpload.uploadLabels.${_.camelCase(value)}`))
+    .value();
```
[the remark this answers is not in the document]
kubama: I don't think these two implementations are semantically equal. Original implementation produces an object, while the one suggested produces an array.

widgets/common/src/UploadBlueprintBasicForm.jsx
kubama: Agree - updated

widgets/filter/src/Filter.jsx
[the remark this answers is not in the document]
Gelio: You are right! I did not know that :smile: Today I learned :slightly_smiling_face: 

Then, in this case, to keep the level of abstractness on the same value in the whole block, I would encourage extracting this "low-level" string manipulation to a contained function, e.g. `appendQueryParam`

widgets/common/src/UploadBlueprintBasicForm.jsx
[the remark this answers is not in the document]
Gelio: You are right, my bad :slightly_smiling_face:

widgets/blueprintCatalog/src/UploadModal.jsx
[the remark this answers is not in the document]
Gelio: I see, thanks for the explanation

widgets/filter/src/Filter.jsx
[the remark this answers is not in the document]
kubama: Why not - done

widgets/common/src/DeploymentActions.js
```diff
@@ -94,28 +94,22 @@ class DeploymentActions {
         return this.toolbox.getManager().doGet('/sites?_include=name&_sort=name');
     }
 
-    async waitUntilCreated(deploymentId, maxNumberOfRetries = 60, waitingInterval = 1000 /* ms */) {
-        const { ExecutionActions } = Stage.Common;
+    async waitUntilCreated(deploymentId, maxNumberOfRetries = 60) {
```
Gelio, line 97: We are never using this function in a way that needs to provide its own `maxNumberOfRetries`. Can we make it a local variable to shrink the API surface?

widgets/common/src/PollHelper.js
```diff
@@ -0,0 +1,27 @@
...
+    constructor(maxAttempts) {
+        this.maxAttempts = maxAttempts;
+        this.attempts = 0;
+        this.waitInterval = 100;
+    }
+
+    wait() {
+        this.attempts += 1;
+        if (this.attempts > this.maxAttempts) {
+            return Promise.reject(Error('Timeout exceeded'));
```
Gelio, line 11: I would appreciate if we created instances using the `new` keyword for constructors :slightly_smiling_face: This looks quite odd to me and frankly, I've never seen errors constructed this way

widgets/common/src/UploadBlueprintBasicForm.jsx
```diff
@@ -0,0 +1,104 @@
...
+    blueprintYamlFile: '',
+    errors: {},
+    firstFormField: null,
+    formLoading: false,
+    lastFormField: null,
+    uploadState: null,
+    yamlFiles: null
+};
+
+Stage.defineCommon({
```
Gelio, line 101: This component is only used internally by 2 other forms. Could we just export it (via a regular ES export), instead of defining it globally? IMO that would make it a bit easier to navigate, and we would not expose "internal" implementation details

widgets/blueprints/widget.css
```diff
@@ -15,5 +16,6 @@
 
 .uploadBlueprintButton {
     z-index: 5;
-    float: right;
-}
\ No newline at end of file
+    position: absolute !important;
```
Gelio, line 19: Can you document why `important` is necessary?

Gelio (commented):
I very like the global `blueprintYamlFile` rename :+1: Makes things clearer

This is a partial review, again. I reviewed everything aside from cypress tests :slightly_smiling_face: The change seems complex, and I'm not sure about its correctness overall, but I did not notice any problems (maybe aside from not being able to cancel uploading, but that's a larger topic)

I did not test it in the UI

widgets/common/src/UploadBlueprintBasicForm.jsx
Gelio: Nevermind, I see it is used in the `blueprintCatalog` widget

widgets/common/src/BlueprintActions.js
```diff
@@ -80,12 +113,45 @@ class BlueprintActions {
...
+            .then(() => onStateChanged(BlueprintActions.InProgressBlueprintStates.UploadingImage))
+            .then(() => this.doUploadImage(blueprintName, imageUrl, image));
+    }
+
+    async waitUntilUploaded(blueprintName, onStateChanged, maxNumberOfRetries = 60) {
+        const { PollHelper } = Stage.Common;
+        const pollHelper = new PollHelper(maxNumberOfRetries);
+
+        let previousState = BlueprintActions.InProgressBlueprintStates.Pending;
+        for (;;) {
```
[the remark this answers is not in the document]
kubama: Cancelling the upload was never taken into account, I don't think we have similar functionality anywhere.
Cancelling the polling (by closing the modal and letting the upload to continue in the background) was discussed but eventually it was decided not to include such functionality.
Generally, design for this ticket is modelled after existing functionality of blueprint deployment modal and is in line with it.

Gelio:
Stuff I have noticed during the demo:

1. Missing whitespace between "Failed" state and the ! icon

    image

2. Maybe it would be nice not to automatically close the Upload blueprint modal after the upload is done, but let the see that it is done and close it themselves (I do not have 100% confidence that this is a problem, you might have just quickly closed the modal and I did not notice)

widgets/common/src/BlueprintActions.js
```diff
@@ -80,12 +113,45 @@ class BlueprintActions {
...
         }
 
-        return promise.then(() => this.doUploadImage(blueprintName, imageUrl, image));
+        return promise
+            .then(() => this.waitUntilUploaded(blueprintName, onStateChanged))
+            .then(() => onStateChanged(BlueprintActions.InProgressBlueprintStates.UploadingImage))
+            .then(() => this.doUploadImage(blueprintName, imageUrl, image));
+    }
+
+    async waitUntilUploaded(blueprintName, onStateChanged, maxNumberOfRetries = 60) {
```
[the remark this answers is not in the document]
kubama: We don't need it - removed

widgets/common/src/DeploymentActions.js
kubama: Right, removed the function argument

widgets/common/src/PollHelper.js
kubama: This is an explicit feature of `Error` API: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#function_call_or_new_construction - nothing odd to me. However, @qooban made a similar point - it's 2:1 so I won't insist and change it.

kubama:
> 1. Missing whitespace between "Failed" state and the ! icon

Addressed.

> 2. Maybe it would be nice not to automatically close the Upload blueprint modal after the upload is done, but let the see that it is done and close it themselves (I do not have 100% confidence that this is a problem, you might have just quickly closed the modal and I did not notice)

The modal closes automatically. After all, it is existing functionality, as mentioned somewhere else it is also in line with blueprint deployment modal. There was no such requirement raised during solution design and I don't think it's the right moment to introduce such changes. If you think it's worth to push this idea then I think it should be applied consistently (at least blueprint upload and blueprint deployment modals, possibly more). Nevertheless it would seem to be a candidate for a separate JIRA.

qooban (approved):

widgets/common/src/PollHelper.js
Gelio: Thanks :slightly_smiling_face:

widgets/blueprints/widget.css
```diff
@@ -1,8 +1,13 @@
+.blueprintsTable th:nth-child(6) {
```
Gelio, line 1: This seems brittle and harder to maintain than using a class. Do you think we could add the ability to customize the class for the `th` (`TableColumn`)?

I do not expect it to be added here. A task in the Technical Debt epic is equally fine :slightly_smiling_face:

test/cypress/support/resource_commons.js
```diff
@@ -6,7 +6,9 @@ export function waitUntilEmpty(resource, search, numberOfRetriesLeft = 60, waiti
         throw new Error(`Number of retries exceeded for resource=${resource}, search=${search}.`);
     }
 
-    cy.cfyRequest(`/${resource}?_search=${search}`, 'GET').then(response => {
+    let url = `/${resource}`;
+    if (search) url += `?_search=${search}`;
```
Gelio, line 10: Do you think we could use a similar `appendToSearchParams` function here as we do in the application's code? If the URL will contain a query param already, this basic concatenation will not work :worried:

test/cypress/support/blueprints.js
```diff
@@ -2,17 +2,16 @@ import { waitUntilEmpty } from './resource_commons';
 
 Cypress.Commands.add('uploadBlueprint', (pathOrUrl, id, yamlFile = 'blueprint.yaml', visibility = 'tenant') => {
     if (pathOrUrl.startsWith('http')) {
-        return cy.cfyRequest(
+        cy.cfyRequest(
```
Gelio, line 5: Are the changes in this file in any way meaningful? Does returning here change anything? :thinking: 

I presume it may have to do with the chainability of cypress commands, may it not?

test/cypress/integration/widgets/blueprint_catalog_spec.js
```diff
@@ -31,5 +32,25 @@ describe('Blueprints catalog widget', () => {
         cy.contains(
             "Invalid blueprint - Plugin cloudify-aws-plugin (query: {'package_name': 'cloudify-aws-plugin'}) not found"
         );
+
+        cy.route('PUT', RegExp(`/console/sp\\?su=/blueprints/${blueprintName}`), {});
```
Gelio, line 36: `RegExp` used as a factory function :smile: I learned something new, I have never seen such a thing before :slightly_smiling_face: Anyway, I will not press strongly here to add `new`, just FYI, I'm surprised :slightly_smiling_face:

test/cypress/integration/widgets/blueprints_spec.js
```diff
@@ -5,145 +5,348 @@ describe('Blueprints widget', () => {
...
+        it('should do nothing on click when drill down is disabled', () => {
+            cy.editWidgetConfiguration('blueprints', () => cy.get('input[name=clickToDrillDown]').parent().click());
 
-            const blueprintName = `${blueprintNamePrefix}_default_file`;
-            cy.get('input[name=blueprintName]').clear().type(blueprintName);
-            cy.get('.button.ok').click();
+            cy.contains(emptyBlueprintName).click();
 
-            getBlueprintRow(blueprintName).contains('read-secret-blueprint.yaml');
+            cy.get('div.blueprintsWidget table.blueprintsTable');
```
Gelio, line 76: IMO the lone `get` looks line an unfinished line :smile: How about showing a bit more intent? Maybe I misunderstood something? :thinking: 

```suggestion
            cy.get('div.blueprintsWidget table.blueprintsTable').should('exist');
```

Gelio: The same applies to other bare `get`s without `should`

test/cypress/integration/widgets/blueprints_spec.js
```diff
@@ -5,145 +5,348 @@ describe('Blueprints widget', () => {
...
+                cy.get('.blueprintName').click();
+                cy.get('.warning').trigger('mouseover');
+            });
+            cy.contains('invalid error');
 
-    describe('configuration', () => {
-        it('should allow to disable drill down on click', () => {
-            cy.editWidgetConfiguration('blueprints', () => cy.get('.checkbox').click());
+            cy.route(RegExp(`/console/sp\\?su=/blueprints.*&state=uploaded`)).as('filteredBlueprints');
+            cy.editWidgetConfiguration('blueprints', () => cy.get('input[name=hideFailedBlueprints]').parent().click());
```
Gelio, line 154: I'm wondering whether we should not test that there are no failed blueprints visible in the table/catalog (in those 2 describe blocks).

Right now I only see we test that there is one that is uploaded, which was also present before, so if `hideFailedBlueprints` does not work correctly and shows all blueprints, we may not catch it.

Do you think it's reasonable?

test/cypress/integration/widgets/blueprints_spec.js
```diff
@@ -5,145 +5,348 @@ describe('Blueprints widget', () => {
...
+                    cy.get('.warning').should('not.exist');
+                    cy.get('a').click({ force: true });
+                });
+
+            cy.get('div.blueprintsWidget .segmentList').should('not.exist');
+            cy.contains('.pageTitle', 'uploaded');
+        });
+    });
+
+    describe('should open upload modal and', () => {
```
Gelio, line 277: Nice naming structure, I like it :+1:

Gelio (commented):
Thanks for responding to all the comments :slightly_smiling_face: I've finally reviewed all the code :eyes: 

> The modal closes automatically
Sure, that's ok then, we can maintain consistency here

btw. I like the way the tests are written - they are easy to follow

widgets/blueprints/widget.css
kubama: Created RD-1151 to address this.

test/cypress/support/blueprints.js
kubama: Actually my intention was to revert all changes made to this file, looks like I failed to do it properly. Updated.

test/cypress/support/resource_commons.js
kubama: Done

test/cypress/integration/widgets/blueprints_spec.js
kubama: We are checking if `state=uploaded` query param is used in data fetch request after enabling `hideFailedBlueprints` configuration option. The widget does not perform any filtering on its own, it just shows whatever it gets. It's the manager's responsibility to supply properly filtered list.

test/cypress/integration/widgets/blueprints_spec.js
kubama: As for me the intention is clear, after all we are using the same approach with `cy.contains`. Adding `should('exist')` seems a redundancy to me, but I have nothing against it after all - applied.

widgets/blueprints/widget.css
Gelio: Great, thanks :+1:

Gelio (approved):
The code looks good to me :+1: I have not tested it in the UI

PR #1167, opened 2021-01-12
RD-976 - Login procedure refactoring and Okta integration fix

qooban:
I did major refactoring of login procedure to allow Okta authentication to work.

#### Normal authentication
1. User opens `/console/login` page (automatically redirected to that page from any URL on that domain when token is not set in the cookies).
2. User provides credentials and click LOGIN button.
3. User gets the token from the manager (and we save the token and user role from the backend response on the frontend side)
4. Application loading is started and behind the scenes:
  1. Manager data fetching - tenants, RBAC config, version, license. If all fine, then go next, if not show Manager data fetching error (`/console/error`). If license not active user ends up on License page (`/console/license`). 
  2. User data fetching - user role, system role, user tenants, etc. If all fine, then go next, if not show User data fetching error (` (`/console/error`). If user is not assigned to any tenant, show NoTenant page (`/console/noTenants`)
  3. App data fetching - templates, pages, widgets, etc. If all fine, then user will see main page, if not user should end up with User app data fetching error (`/console/error`).

#### Okta authentication
_Assuming proper configuration is done on both - Stage and Okta side_

1. It can start two different ways:
   1. User logs in to Okta and from the list of applications user selects Cloudify app
   2. User opens `/console/login` page and clicks LOGIN WITH SSO button, user is redirected to Okta and log in there.
2. Okta is sending HTTP POST request to `/auth/saml/callback` endpoint on manager VM with `SAMLResponse` in the payload
3. Stage backend sends SAMLResponse to the manager API asking for token
4. User gets the token from the stage backend (there are 3 cookies sent: token, username and role)
5. Application loading is done the same way as in normal authentication with small difference: just before manager data fetching cookies are used to initialize state management with proper login time, username and user role (which is eg. necessary to determine if user have access to license page)

### Changes description

1. New login page when SAML authentication is enabled (`saml.enabled === true` in the configuration file)
It uses the same code as normal login with two differences:
* there are no username, password input fields
* login button text is set to "LOGIN WITH SSO" instead of simple "LOGIN"
image
Login page (`/console/login`) is being loaded when the user is logged out or at any point the authentication fails - e.g. token expires - during normal application runtime. 
In the original Okta authentication implementation the user once reached `/console/login` page was automatically redirected to a page provided in the configuration file (`saml.ssoUrl`). That lead often to redirection loops, so I decided not to do that redirection automatically, but just provide one click action to do it. It helped in error showing (no-tenants error, license expired error, user error, manager error) for which usually we do logout before. So now the proper error is shown in both cases - normal authentication and Okta authentication. 
Logout works fine right now as just after clicking Logout option from the top right menu, the token is invalidated and the user is being redirected to the portal (Okta).

2. Fixed setting user role gather from the token
In the other PR (https://github.com/cloudify-cosmo/cloudify-stage/pull/1168) I fixed system role calculation so that templates/tours fetching provided relevant data for the specific user. In this PR I fixed also a way how we set the role taken from the token in case of external authentication is turned on. Now in case of Okta authentication user role is passed from the backend (`/auth/saml/callback` endpoint) to the frontend with cookie. This cookie once process on the frontend side (role and username added to redux store) is cleared when calling `/auth/manager` endpoint, which is done on first time page loading.

3. Fixed checking if user has any tenants assigned (to get proper pages loaded)
Before we used wrong endpoint to check whether user has tenants attached (listing all tenants in the system instead of tenants assigned to the user). In case of Okta authentication it will probably be common when user gets tenant from the group assigned to him instead of tenant assigned directly to the user, so it was important to fix it in this PR.

4. Updated configuration of store initial state
There was inconsistency in how manager reducer is providing state changes on login/logout, so this is fixed as well. It was important from testing perspective to have clear state anytime we logout and log in again.

5. Removed error details from being passed to client side, only logging on backend side.
That was pointed out by Sandy. I did not aim to fix all the places with this issue, but focused only on authentication endpoints (`/auth` route in stage backend).

6. Added backend tests for authentication endpoints, added/updated tests for manager and license reducers, added system tests for Authentication (auth_spec.js), updated tests for community, license and login specs


app/actions/app.js
```diff
@@ -4,40 +4,25 @@ import { setAppError, setAppLoading } from './appState';
...
 import { getClientConfig } from './clientConfig';
 import { loadOrCreateUserAppData } from './userApp';
-import { getLdap, getUserData } from './managers';
+import { getLdap } from './managers';
 import { getClusterStatus } from './clusterStatus';
-import { NO_TENANTS_ERR } from '../utils/ErrorCodes';
 
 export function intialPageLoad() {
-    return (dispatch, getState) => {
+    return dispatch => {
```
kubama, line 14: Not a big deal, but personally I like the idea of commenting out unused parameters - it documents what's being passed to the function in case it needs to be used again.

app/actions/tenants.js
```diff
@@ -33,14 +33,15 @@ function errorTenants(err) {
...
-    return dispatch => {
+export function getTenants() {
+    return (dispatch, getState) => {
         dispatch(requestTenants());
+        const managerAccessor = new Manager(getState().manager);
         return managerAccessor
             .doGet('/tenants', { _include: 'name', _get_all_results: true })
             .then(tenants => {
                 dispatch(recieveTenants(tenants));
+                return Promise.resolve(tenants);
```
kubama, line 44:
```suggestion
                return tenants;
```

app/components/AuthRoutes.jsx
```diff
@@ -0,0 +1,74 @@
...
+import Auth from '../utils/auth';
+import Consts from '../utils/consts';
+import { getManagerData, getUserData, logout } from '../actions/managers';
+import Layout from '../containers/layout/Layout';
+import LicensePage from '../containers/LicensePage';
+import MaintenanceMode from '../containers/maintenance/MaintenanceModePageMessage';
+import SplashLoadingScreen from '../utils/SplashLoadingScreen';
+
+export default function AuthRoutes() {
+    const [isInitialized, setInitialized] = useState(false);
```
kubama, line 19: `useBoolean` could be used here

backend/routes/Auth.js
```diff
@@ -60,23 +35,42 @@ router.post('/login', (req, res) =>
 );
 
 router.post('/saml/callback', passport.authenticate('saml', { session: false }), (req, res) => {
+    logger.error(JSON.stringify(req.body), JSON.stringify(req.user));
```
kubama, line 38: Debugging statement? :-)

backend/routes/Auth.js
```diff
@@ -60,23 +35,42 @@ router.post('/login', (req, res) =>
...
-        res.status(401).send('Invalid Request');
+        res.status(401).send({ message: 'Invalid Request' });
+    } else {
+        AuthHandler.getTokenViaSamlResponse(req.body.SAMLResponse)
+            .then(token => {
+                res.cookie(Consts.TOKEN_COOKIE_NAME, token.value);
+                res.redirect(Consts.CONTEXT_PATH);
+            })
+            .catch(err => {
+                logger.error(err);
```
kubama, line 48: In other places you used `logger.error(JSON.stringify(err))`. Is it necessary? If yes, maybe it's a good idea to override `logger.error` so it calls `JSON.stringify` internally for `Error` instances?

backend/routes/Auth.js
```diff
@@ -60,23 +35,42 @@ router.post('/login', (req, res) =>
...
+                ? AuthHandler.getLicense(token).then(data => ({
+                      license: _.get(data, 'items[0]', {}),
+                      version,
+                      rbac
+                  }))
+                : Promise.resolve({
+                      license: null,
+                      version,
+                      rbac
+                  })
```
kubama, line 68:
```suggestion
                : {
                      license: null,
                      version,
                      rbac
                  }
```

backend/test/routes/Auth.test.js
```diff
@@ -1,15 +1,138 @@
...
+        it('invalid credentials', () => {
+            jest.doMock('handler/AuthHandler', () => ({
+                getToken: () => Promise.reject({ error_code: 'unauthorized_error' })
             }));
+            return request(require('app'))
+                .post('/console/auth/login')
+                .then(response => {
+                    expect(response.body.message).toStrictEqual('Invalid credentials');
+                });
+        });
```
kubama, line 29: Would be perfect if one of these tests checked if `getToken` is called with correct parameter

test/jest/reducers/managerReducer.test.js
```diff
@@ -0,0 +1,420 @@
+/**
+ * Created by jakub.niezgoda on 20/03/2019.
```
kubama, line 2: The date isn't correct, anyway I'm not sure if such comments are of any use

app/actions/app.js
qooban: I'm OK with that, changed.

app/actions/tenants.js
qooban: Changed.

app/components/AuthRoutes.jsx
qooban: Changed.

backend/routes/Auth.js
qooban: Yep :-) Removed.

backend/routes/Auth.js
qooban: Created https://github.com/cloudify-cosmo/cloudify-ui-common/pull/45 to handle that.
Generally winston logger does not handle printing objects/arrays in user-friendly way out of the box. It prints: `[object Object]`. So I used `JSON.stringify` in some places, but I agree that it would be better to improve parsing arguments within the logger we have in cloudify-ui-common.

For now, I suggest to keep `logger.error(JSON.stringify(err))` as planned patch (5.1.2) will base on 5.1.1 which does not have cloudify-ui-common v2.0.0 (on master we have cloudify-ui-common v2.1.0), so it will make patching more complex.

On master we can soon remove `JSON.stringify` from `logger.error` calls after upgrading to new cloudify-ui-common (to be released after https://github.com/cloudify-cosmo/cloudify-ui-common/pull/45 merge).

What do you think?

backend/routes/Auth.js
qooban: Changed.

backend/test/routes/Auth.test.js
qooban: Rigth, added.

test/jest/reducers/managerReducer.test.js
qooban: Of course, not intentional, copy-paste issue :-) (as this file was created from `licenseReducer.test.js`)

backend/routes/Auth.js
kubama: 5.1.1 already handles `Error` instances by replacing them with their `stack` property. https://github.com/cloudify-cosmo/cloudify-ui-common/pull/45 doesn't change that. So calling `JSON.stringify` on `Error` instances doesn't make sense. If you think in any particular cases caught errors may be not `Error` instances then it should be fixed at its source if possible. If not, I would still avoid using `JSON.stringify` - except one place we are not using it on caught errors and I guess we are fine. So introducing `JSON.stringify` in 5.1.2 to handle non `Error` instances is not a must, especially that in 5.2 this change will become redundant and should be reverted.

qooban: For the debugging purposes I see value in that.

> If you think in any particular cases caught errors may be not Error instances then it should be fixed at its source if possible.

Exactly that's the case. They are not always `Error` instances, unfortunatelly, but fixing it at source is not trivial IMHO. See eg. https://github.com/cloudify-cosmo/cloudify-stage/blob/a8ecbd4b303813f614f87d46e88dea6e38332a21/backend/handler/ManagerHandler.js#L73 - this type of rejecting promises is I think since the beginning of Cloudify 4.x. That is why I created [RD-257](https://cloudifysource.atlassian.net/browse/RD-257) long time ago.

ofercloudify:
Good summary, very helpful!
Two tiny comments:
1. Can we add the extra string of the new login page added to the white-labeling file? (or is it already there?)
2. Okta and other IdP services are very similar. In the future we may get requests to support other Identity providers, and so I rather avoid any "hardcoded" references to "Okta". We could either add a property of "idp_name" to the config file and use it in that  string or go with something more generic such as "LOGIN WITH SSO".
This will save us some later code changes...

qooban:
@ofercloudify 
Thanks for your comment. 

> Can we add the extra string of the new login page added to the white-labeling file? (or is it already there?)

This extra string is already part of the login section in translation file (https://github.com/cloudify-cosmo/cloudify-stage/blob/RD-976/app/translations/en.json#L169), I changed only the name from `oktaSubmit` to `ssoSubmit`. 

We removed strings from white-labeling file when doing refactoring in November last year (user configuration file: https://github.com/cloudify-cosmo/cloudify-stage/blob/master/conf/userConfig.json) and all white-label strings are now kept in translation file (and can be overridden using `overrides.json` file).

> Okta and other IdP services are very similar. In the future we may get requests to support other Identity providers, and so I rather avoid any "hardcoded" references to "Okta". We could either add a property of "idp_name" to the config file and use it in that string or go with something more generic such as "LOGIN WITH SSO".
This will save us some later code changes...

Yes, I actually thought the same way about that, but was not sure we want to show that we support something not tested. I have changed the button name to "LOGIN WITH SSO". I think that in the documentation we should still keep the information that we support only Okta. User can use any SAML 2.0 solution, but we do not guarantee it will work.

ofercloudify:
Agree with all your points. Thanks for the details.

backend/routes/Auth.js
kubama: So lets go with this but lets create another JIRA for removing unnecessary `JSON.stringify` calls in 5.2. Or, if 5.1.2 is going to be cherry-picked from master lets create another PR for removing it right away.

kubama (dismissed):

backend/routes/Auth.js
qooban: Yes, I'm going to cherry-pick few commits from master to 5.1.2 once all are merged.
So I'll create another PR for removing unnecessary `JSON.stringify` calls it once this is merged.

Gelio (commented):
Most comments are code-style and code-quality related :smile: None of them are blocking.

This is a partial review. I have yet to review `components`, `containers`, and cypress tests

app/components/LoginPage.jsx
```diff
@@ -25,9 +25,13 @@ export default class LoginPage extends Component {
 
     onSubmit = () => {
         const { password, username } = this.state;
-        const { location, onLogin } = this.props;
+        const { isSamlEnabled, location, onLogin, samlSsoUrl } = this.props;
         const errors = {};
 
+        if (isSamlEnabled) {
+            window.open(samlSsoUrl, '_self');
```
Gelio, line 32: Why not 

```suggestion
            window.location.href = samlSsoUrl;
```

which seems to be the community-preferred option https://stackoverflow.com/questions/4813879/window-open-target-self-v-window-location-href

backend/routes/Auth.js
```diff
@@ -9,74 +9,77 @@ const _ = require('lodash');
...
-            }
-            return Promise.reject({ message: 'User has no tenants', error_code: 'no_tenants' });
+        .then(token => {
+            res.cookie(Consts.TOKEN_COOKIE_NAME, token.value);
+            res.send({ role: token.role });
         })
-        .then(({ license, version, role, rbac }) => res.send({ license, version, role, rbac }))
         .catch(err => {
-            logger.error(err);
+            logger.error(JSON.stringify(err));
```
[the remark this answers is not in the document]
qooban: To make it short - it'll be removed in 5.2.

But this is a bit longer story. Checkout this discussion https://github.com/cloudify-cosmo/cloudify-stage/pull/1167#discussion_r555801411 and this PR: https://github.com/cloudify-cosmo/cloudify-ui-common/pull/45 if you're interested.

backend/test/routes/Auth.test.js
```diff
@@ -1,15 +1,207 @@
...
+                    expect(response.body).toStrictEqual({ role: 'default' });
+                });
+        });
+
+        it('manager maintenance mode', () => {
+            jest.doMock('handler/AuthHandler', () => ({
+                getToken: () => Promise.reject({ error_code: 'maintenance_mode_active' })
+            }));
+            return request(require('app'))
+                .post('/console/auth/login')
```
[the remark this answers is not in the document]
qooban: Good idea, I'll add such assertions.

app/actions/app.js
```diff
@@ -4,40 +4,25 @@ import { setAppError, setAppLoading } from './appState';
...
 import { getClientConfig } from './clientConfig';
 import { loadOrCreateUserAppData } from './userApp';
-import { getLdap, getUserData } from './managers';
+import { getLdap } from './managers';
 import { getClusterStatus } from './clusterStatus';
-import { NO_TENANTS_ERR } from '../utils/ErrorCodes';
 
 export function intialPageLoad() {
-    return (dispatch, getState) => {
+    return (dispatch /* , getState */) => {
```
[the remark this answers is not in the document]
qooban: I think this is a matter of preference. I'm OK with both approaches (having it and not having it) as I see value in pros and cons of both. I'll keep it as it is for now as we use similar approach in the other places (both in stage and composer).

app/components/LoginPage.jsx
```diff
@@ -133,7 +141,7 @@ export default class LoginPage extends Component {
                             color="yellow"
                             size="large"
                             type="submit"
-                            content={i18n.t('login.submit')}
+                            content={isSamlEnabled ? i18n.t('login.ssoSubmit') : i18n.t('login.submit')}
```
Gelio, line 144: How about 

```suggestion
                            content={i18n.t(isSamlEnabled ? 'login.ssoSubmit' : 'login.submit')}
```

app/actions/managers.js
```diff
@@ -93,10 +77,21 @@ function responseUserData(username, systemRole, groupSystemRoles, tenantsRoles)
     };
 }
 
+export function getManagerData() {
```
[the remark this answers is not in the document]
qooban: I prefer consistency (at least) within the file.
If we would like to do change it here, then we would need to change it in all other places (see `getUserData`, `getMaintenanceStatus`, etc.). That is not time and place to do so.

app/components/AuthRoutes.jsx
```diff
@@ -0,0 +1,89 @@
...
+import Auth from '../utils/auth';
+import Consts from '../utils/consts';
+import { getManagerData, getUserData, logout, receiveLogin } from '../actions/managers';
+import Layout from '../containers/layout/Layout';
+import LicensePage from '../containers/LicensePage';
+import MaintenanceMode from '../containers/maintenance/MaintenanceModePageMessage';
+import SplashLoadingScreen from '../utils/SplashLoadingScreen';
+
+export default function AuthRoutes({ isSamlEnabled }) {
+    const [isManagerDataFetched, setManagerDataFetched /* , unsetManagerDataFetched */] = useBoolean();
```
Gelio, line 22: What's the point of having the commented out parts?

app/reducers/managerReducer.js
```diff
@@ -79,6 +69,7 @@ const manager = (state = {}, action) => {
         case types.SET_MANAGER_VERSION:
             return { ...state, version: action.version };
         case types.SET_MANAGER_LICENSE:
+        case types.SET_LICENSE_REQUIRED:
             return { ...state, license: license(state.license, action) };
```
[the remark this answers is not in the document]
qooban: Agree. Generally, IMHO our store needs refactoring at some point, I saw few issues there in the past. But that's not the right time to do so. You can add something to `UI Technical Debt` epic.

test/jest/reducers/managerReducer.test.js
```diff
@@ -0,0 +1,416 @@
...
+                fetchMock.post('/console/auth/login', {
+                    body: { role },
+                    headers: { 'content-type': 'application/json' }
+                });
+            });
+
+            it('triggers actions', () => {
+                store = mockStore({});
+
+                return store.dispatch(login(username, password)).then(() => {
```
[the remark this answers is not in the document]
qooban: We test the application state after the action is done in the next test case within that describe section.

app/components/AuthRoutes.jsx
```diff
@@ -0,0 +1,89 @@
...
+    const [isUserDataFetched, setUserDataFetched /* , unsetUserDataFetched */] = useBoolean();
+    const isInMaintenanceMode = useSelector(
+        state => _.get(state, 'manager.maintenance') === Consts.MAINTENANCE_ACTIVATED
+    );
+    const isLicenseRequired = useSelector(state => _.get(state, 'manager.license.isRequired', false));
+    const isProductOperational = useSelector(state => Auth.isProductOperational(_.get(state, 'manager.license', {})));
+    const isSamlLogin = isSamlEnabled && !!Cookies.get(Consts.ROLE_COOKIE_NAME);
+    const dispatch = useDispatch();
+
+    useEffect(() => {
```
Gelio, line 32: I believe we could make this component a bit shorter by extracting out custom hooks.

1. `useManagerData` that returns `{ fetched, licenseRequired, productOperational }`. It would have a part of this `useEffect` inside (the `dispatch` from this `useEffect`)
2. `useUserData` that returns `userDataFetched` and contains the 2nd `useEffect` from this component

So this component would look like so:

```tsx
export default function AuthRoutes({ isSamlEnabled }) {
    const managerData = useManagerData();
    const userDataFetched = useUserData(managerData.productOperational, managerData.fetched);
    const isSamlLogin = isSamlEnabled && !!Cookies.get(Consts.ROLE_COOKIE_NAME);
    const dispatch = useDispatch();

    useEffect(() => {
        SplashLoadingScreen.turnOn();

        if (isSamlLogin) {
            const role = Cookies.get(Consts.ROLE_COOKIE_NAME);
            const username = Cookies.get(Consts.USERNAME_COOKIE_NAME);
            dispatch(receiveLogin(username, role));
        }
    }, []);

    return managerData.isManagerDataFetched ? (
        <Switch>
            {managerData.isLicenseRequired && <Route exact path={Consts.LICENSE_PAGE_PATH} component={LicensePage} />}
            <Route exact path={Consts.MAINTENANCE_PAGE_PATH} component={MaintenanceMode} />
            {managerData.isInMaintenanceMode && <Redirect to={Consts.MAINTENANCE_PAGE_PATH} />}
            <Route
                render={() =>
                    managerData.isProductOperational ? (
                        userDataFetched && <Layout />
                    ) : (
                        <Redirect to={Consts.LICENSE_PAGE_PATH} />
                    )
                }
            />
        </Switch>
    ) : null;
}
```

Much easier to grok, if you asked me :slightly_smiling_face:

app/translations/en.json
```diff
@@ -179,7 +179,8 @@
         },
         "password": "Password",
         "username": "Username",
-        "submit": "LOGIN"
+        "submit": "LOGIN",
+        "ssoSubmit": "LOGIN WITH SSO"
```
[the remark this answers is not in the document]
qooban: We allow customization and overriding strings, that's the reason why I intentionally don't want to use `text-transform: uppercase`.

I see more value and better user experience in providing it to the user like that. We don't use this string anywhere else than in the login page and for user not very familiar with CSS and just wanting to change single string it is easier to just change the text and don't be surprised with uppercase output when using lowercase in the translation file.

app/containers/LoginPage.js
```diff
@@ -2,19 +2,23 @@
...
 
 const mapStateToProps = state => {
+    const { config, manager } = state;
     return {
-        username: state.manager ? state.manager.username : '',
-        isLoggingIn: state.manager.isLoggingIn,
-        loginError: state.manager ? state.manager.err : '',
-        mode: state.config.mode,
-        whiteLabel: state.config.app.whiteLabel
+        username: _.get(manager, 'username', ''),
```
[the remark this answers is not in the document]
qooban: I agree, there's already issue about it: https://cloudifysource.atlassian.net/browse/RD-600 . You can take it if you want and ensure all packages have support for the latest ES.

app/utils/consts.js
```diff
@@ -18,6 +18,8 @@ export default {
     APP_DATA_PATH: '/appData',
 
     TOKEN_COOKIE_NAME: 'XSRF-TOKEN',
+    ROLE_COOKIE_NAME: 'ROLE',
```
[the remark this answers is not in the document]
qooban: Definitely. That's not something new, we already thought about it, but this is not something trivial, I think. How would you do that? `backend` is not bundled, it is served in a package as is. `app` is bundled. Extending package would probably be necessary, so we would need to extract common consts, update webpack, update RPM spec, etc. 

It's definitely worth to solve that in the future, but not quick action, so I suggest to create an issue in `UI Technical Debt` epic about that.

Gelio (commented):
I'm done with the review :slightly_smiling_face: I only focused on code quality. I tried to read other parts too, but I have too little expertise in the system to confirm its correctness :slightly_smiling_face: 

I did not test it in the UI, I only read the code

app/components/LoginPage.jsx
qooban: I would do so, that was my original implementation, but had problem with Cypress test and changed it to `window.open`. If you provide me a way to test it with Cypress, I'll change it.

See: https://github.com/cypress-io/cypress/issues/3994
FYI I was testing `experimentalSourceRewriting` without success.

app/components/AuthRoutes.jsx
qooban: See one of the previous comment and have a duel with Kuba M. :-) 

From my perspective both approaches have pros and cons.

app/actions/managers.js
```diff
@@ -93,10 +77,21 @@ function responseUserData(username, systemRole, groupSystemRoles, tenantsRoles)
...
+    return (dispatch, getState) =>
+        Auth.getManagerData(getState().manager).then(({ version, license, rbac }) => {
+            dispatch(setVersion(version));
+            dispatch(setLicenseRequired(!_.isEqual(version.edition, Consts.EDITION.COMMUNITY)));
```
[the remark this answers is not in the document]
qooban: Could be, I'll refactor it.

app/utils/consts.js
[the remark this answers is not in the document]
Gelio: That does not require another package :smile: We could have a top-level directory called `shared` in the repository, that could contain shared code relevant only to the stage repo - cookie names, for instance.

Anyway, here's a ticket: [RD-1103](https://cloudifysource.atlassian.net/browse/RD-1103)

app/reducers/managerReducer.js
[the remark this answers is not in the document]
Gelio: I'm not sure what the scope of a ticket you expected. I created a ticket only for moving those 2 reducers to their own directory, creating a self-contained module. [RD-1104](https://cloudifysource.atlassian.net/browse/RD-1104)

test/jest/reducers/managerReducer.test.js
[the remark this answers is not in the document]
Gelio: So why do we care whether the thunk dispatched correct actions in that exact order? Should we not only observe the end results?

app/containers/LoginPage.js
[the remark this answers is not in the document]
Gelio: Thanks for pointing me that way :slightly_smiling_face:

app/components/LoginPage.jsx
Gelio: I read about the problem for a bit and IMO that's more closely related to https://docs.cypress.io/guides/guides/web-security.html#JavaScript-Redirects rather than some URL rewriting. The problem stems from going cross-origin, which looks not to be supported in Cypress.

To use `window.location.href`, a change is necessary in all tests to make `ssoUrl` be in the same origin as the initial page visited. Thus, for example in the `provides SSO login button when SAML is enabled` test in `login_spec.js`:
1. Change `ssoUrl`, e.g. to `/sso-redirect`
2. Instead of asserting that `window.open` was called, do: `cy.url().should('include', ssoUrl);`

Worked on my end:
image

app/components/AuthRoutes.jsx
Gelio: I see :smile: Let's leave this be and I'll talk with Kuba M

app/utils/consts.js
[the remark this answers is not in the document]
qooban: I did not say anything about another package :-) I was talking about need of extending cloudify-stage package definition and content, which is out of the scope of this PR (as it is going to be cherry-picked to 5.1.2 branch once merged to master). Anyway, thanks for creating the ticket.

app/actions/managers.js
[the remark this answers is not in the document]
qooban: Let's agree to disagree on this point and discuss another time :-)

test/jest/reducers/licenseReducer.test.js
```diff
@@ -4,319 +4,15 @@
...
-                status: 'no_license'
-            };
-
-            return store.dispatch(login(username, password)).then(() => {
-                expect(store.getState()).toEqual(expectedLicenseState);
-            });
-        });
-    });
 
     describe('Set license action', () => {
```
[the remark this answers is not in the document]
qooban: Could be, but I think this is out of the scope of this PR. Let's keep it for the future.

test/jest/reducers/managerReducer.test.js
```diff
@@ -0,0 +1,416 @@
...
+                    expect(actualActions).toEqual(expectedActions);
+                });
+            });
+
+            it('initializes state', () => {
+                store = createStore(managerReducer, {}, applyMiddleware(thunk));
+
+                return store.dispatch(login(username, password)).then(() => {
+                    expect(store.getState()).toEqual({
+                        auth: {
```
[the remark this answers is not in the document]
qooban: Make sense. Changed.

test/jest/reducers/managerReducer.test.js
```diff
@@ -0,0 +1,416 @@
...
+                    const expectedActions = [
+                        { type: types.REQ_LOGIN },
+                        { type: types.RES_LOGIN, username, role, receivedAt: Date.now() },
+                        {
+                            type: '@@router/CALL_HISTORY_METHOD',
+                            payload: { args: ['/'], method: 'push' }
+                        }
+                    ];
+
+                    expect(actualActions).toHaveLength(expectedActions.length);
```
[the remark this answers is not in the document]
qooban: Seems reasonable. Removed.

test/jest/reducers/managerReducer.test.js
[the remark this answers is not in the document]
qooban: From my perspective it has a value - it shows you how our store works. I remember from my previous job such kind of traces helped a lot in understanding what is going on behind the scenes and it was much easier than tracking the code.

app/components/LoginPage.jsx
qooban: Good idea. Implemented. 
I used `window.location` instead of `window.location.href` to keep it consistent with other places in the code.

From my perspective in the future we could consider moving to `location.assign(...)` as the location change is more like an action ( => function call) and not just a state change ( => variable assignment).

test/jest/reducers/managerReducer.test.js
```diff
@@ -0,0 +1,416 @@
...
+
+    afterAll(() => {
+        timeKeeper.reset();
+    });
+
+    afterEach(() => {
+        fetchMock.restore();
+    });
+
+    describe('Receive login action', () => {
```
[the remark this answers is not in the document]
qooban: Maybe better, but definitely I would consider it as out of the scope of this PR.

Moreover, these thunks are related to `managerReducer`, so I don't see it as a big sin :-)

app/components/AuthRoutes.jsx
qooban: Like the idea of introducing custom hooks, not sure I would do it the same way, but definitely I would consider it.

But... I'd like to leave it for the future and don't change too much at this stage (have to provide the solution for the RC build today).

test/jest/reducers/managerReducer.test.js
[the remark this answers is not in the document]
Gelio: I don't fully get it, as I see it as repeated code from the thunk definition :smile: Let's leave it be, thanks for sharing

app/components/LoginPage.jsx
Gelio: Sure thing :smile: Today I learned! Thanks for sharing about `location.assign`, I did not know such a thing existed. I definitely agree with your reasoning here :slightly_smiling_face:

app/components/AuthRoutes.jsx
Gelio: Sounds good, let's leave it be

Gelio (approved):
The code looks good to me from quality-perspective. I have too little knowledge to reason about its correctness. I did not test in the UI

Thanks for all the nice discussions :slightly_smiling_face:

PR #1168, opened 2021-01-12
RD-1043 - Fix system role calculation in template management

qooban:


backend/handler/TemplateHandler.js
```diff
@@ -136,6 +135,30 @@ module.exports = (() => {
...
+
+        const systemRoles = [userSystemRole, ..._.keys(groupSystemRoles)];
+        let systemRole = null;
+        for (let i = 0; i < roles.length; i += 1) {
+            const role = roles[i];
+            if (_.includes(systemRoles, role.name)) {
+                systemRole = role.name;
+                break;
+            }
+        }
```
kubama, line 156: Can't we use `_.find` here?

backend/test/handlers/TemplateHandler.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+describe('TemplateHandler', () => {
+    describe('allows to select built-in template', () => {
+        const userTemplatesFolder = Utils.getResourcePath('templates', true);
+        const userPagesFolder = path.resolve(userTemplatesFolder, 'pages');
+        if (fs.existsSync(userTemplatesFolder)) {
+            fs.rmdirSync(userTemplatesFolder, { recursive: true });
+        }
+        mkdirp.sync(userTemplatesFolder);
+        mkdirp.sync(userPagesFolder);
```
kubama, line 44: Since unit tests are intended to be run in local development environments as well I don't think it's a good idea to allow a test to do any permanent changes in the environment, especially if it involves removing any files. If this is necessary, then the test should operate on files or directories created in temporary location, which is discarded after test completion. This tool https://www.npmjs.com/package/tmp seems to be handy as documentation states it removes created resources on process exit.
Also, this logic should be placed in `before`.

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -89,8 +89,8 @@ export default class UsersTable extends React.Component {
             user: UsersTable.EMPTY_USER,
             tenants: {},
             groups: {},
-            activateLoading: false,
-            settingUserRoleLoading: false
+            activateLoading: '',
+            settingUserRoleLoading: ''
```
kubama, line 93: I think these state properties names are very misleading - they suggest a boolean value. I'd go further and rename them.

backend/handler/TemplateHandler.js
qooban: Yes, simplified it and extracted similar code to a separate function.

backend/test/handlers/TemplateHandler.test.js
qooban: Right. Simplified it by mocking function from `fs` library.

widgets/userManagement/src/UsersTable.jsx
qooban: Agree. Changed it to `usernameDuringActivation` and `usernameDuringRoleSetting`.
I'm not saying they are the best, but a bit more accurate I think. If you have better names, I'm open to change it.

kubama (approved):

PR #1177, opened 2021-01-25
Don't kill stage config on upgrade

geokala:


tehasdf (approved):

qooban:
@geokala Let's wait with it. 
This problem was solved already in master: https://github.com/cloudify-cosmo/cloudify-stage/pull/1158
I thought it was part of 5.1.1, but it is not, so probably wiser it would be to port it back to 5.1.2.

PR #1179, opened 2021-01-25
RD-1150 Fixed buggy migrations logic

kubama:
backend/migrations/20201030110539-5_1_1-MigratePageLayouts.js invoked async `save()` operation in a loop - this led to an instant connections allocation in an amount equal to the number of `UserApp` table rows. The fix makes the `save()` operations to be invoked sequentially.
backend/migrations/20201230113319-5_1_1-UpdateManagerIp.js was not an issue in case of this particular ticket, however it might face the same problem at some point, so it was updated to use single UPDATE statement (possible since the new column value is static).

Other migrations where also investigated, but it seems they are not prone to the same issue.


qooban (commented):
Changes look good 👍🏻 

Can we add one test which will test migration with large number of entries in UserApps table?

kubama:
> Can we add one test which will test migration with large number of entries in UserApps table?

Extended 4.5.5.sql with 160 UserApps rows, so now 4.5.5 -> 5.1.1 migration test covers the scenario.

qooban (approved):

PR #1181, opened 2021-01-26
RD-1073 Enlarge widget help popup triggers

Gelio:
Use default size when there is a header
Use small size when the widget is headerless
Use the same size whether the component is maximized or not

Also, for widgets with headers there is now a small gap so the icons are not directly in the corner of the widget, on the edge. I've done it using CSS custom properties, which I think is new in the project :smile: at least I have not seen it used in this repo
    
A new 'iconSize" prop has been added to EditWidget for consistency.

This PR also fixes the missing `indent_size` setting in `.editorconfig` for json, scss and md files.

## Screenshots

Headerless widget
image

Widget with header
image

Maximized widget
image

In edit mode (I forced buttons to show just for the purpose of the screenshot and the comparison. Normally they are only shown on hover)
image


.editorconfig
```diff
@@ -19,3 +19,10 @@ insert_final_newline = true
 
 [*.md]
 trim_trailing_whitespace = false
+indent_size = 2
+
+[*.scss]
+indent_size = 2
+
+[*.json]
+indent_size = 2
```
qooban, line 28: If you add this to cloudify-stage, let's add the same to other repositories as well. Or at least create a JIRA issue to align it.

Another thing is that real (used now) indent sizes differ inside cloudify-stage and across UI packages (not only in terms of `.scss` and `.json` files), so that's bigger thing to make it aligned. It's minor from my perspective, but if we are changing anything, let's agree on the direction at least (probably a topic for discussion at UI weekly meeting).

app/components/Widget.jsx
```diff
@@ -168,37 +174,33 @@ export default class Widget extends Component {
...
-                                                  onClick={() => onWidgetUpdated(widget.id, { maximized: true })}
-                                              />
-                                          ]}
+                                    {helpIcon()}
+                                    <Icon
+                                        name={widget.maximized ? 'compress' : 'expand'}
+                                        link
+                                        size={widgetIconButtonSize}
+                                        onClick={() => onWidgetUpdated(widget.id, { maximized: !widget.maximized })}
+                                    />
```
qooban, line 200: I like this kind of refactoring 👍🏻

app/containers/EditWidget.jsx
```diff
@@ -42,12 +42,12 @@ class EditWidgetComponent extends React.Component {
     }
 
     render() {
-        const { configDef, configuration, onWidgetEdited, widget } = this.props;
+        const { configDef, configuration, onWidgetEdited, widget, iconSize } = this.props;
```
qooban, line 45: Let's try to use alphabetic order. In most of the cases it makes it easier to read.

app/containers/EditWidget.jsx
```diff
@@ -65,7 +65,10 @@ EditWidgetComponent.propTypes = {
     widget: PropTypes.shape({}).isRequired,
     onWidgetEdited: PropTypes.func.isRequired,
     configDef: PropTypes.arrayOf(PropTypes.shape({})).isRequired,
-    configuration: PropTypes.shape({}).isRequired
+    configuration: PropTypes.shape({}).isRequired,
+    // iconSize is an optional forwarded prop
+    // eslint-disable-next-line react/require-default-props
+    iconSize: PropTypes.string
```
qooban, line 71: Why don't define default for the new prop (`undefined`) the same way as you did in `app/components/EditWidgetIcon.jsx`?

qooban (commented):
Added few comments.

I would also suggest to get feedback from @ofercloudify on that change.

app/containers/EditWidget.jsx
```diff
@@ -42,12 +42,12 @@ class EditWidgetComponent extends React.Component {
     }
 
     render() {
```
qooban, line 44: One more comment to the last review - maybe worth to refactor/move this container (to `components`)?

app/components/Widget.jsx
```diff
@@ -150,7 +156,7 @@ export default class Widget extends Component {
                     }
                     className={`widgetItem ${
                         isEditMode && widget.definition && !widget.definition.showBorder ? 'borderOnHover ' : ''
-                    }`}
+                    } ${widget.definition.showHeader ? '' : 'headerless'}`}
```
kubama, line 159: I'm not a fan of such style of `className` attribute value composition, especially when there is more than onc conditional. After all, we are creating a space separated list of names. If so, I'd rather create an array of names and then used `join(' ')`. Would be much cleaner and easier to read and understand. Additionally, it would prevent rendering of unnecessary spaces into the generated `class` attribute.

app/styles/style.scss
```diff
@@ -226,14 +226,19 @@ a:hover {
 }
 
 .widget {
+  --widget-buttons-whitespace: 5px;
+
+  .headerless {
+    --widget-buttons-whitespace: 0;
```
kubama, line 232: Minor thing, but wouldn't `--widget-buttons-margin` be a better suited name?

Gelio:
@ofercloudify Could you take a look at the updated appearance of the icons in the top-right of each widget and see if they look good to you? As it says in the PR description, I have made them a bit larger by default, so their size visually matches nearby elements

.editorconfig
Gelio: I have added the indent sizes here based on what I observed in multiple files in this repo :slightly_smiling_face: Sure, I'll add a ticket to make it consistent across repos: https://cloudifysource.atlassian.net/browse/RD-1213

app/components/Widget.jsx
Gelio: Thanks :slightly_smiling_face:

app/containers/EditWidget.jsx
Gelio: Sure, I don't mind it here :slightly_smiling_face:

app/containers/EditWidget.jsx
Gelio: I don't see the benefit in saying that some prop is not required, and then having to manually say that it's default value is `undefined`, which is the default value for unspecified properties :smile: Thus, IMO this rule does not make sense in this case.

To me, it makes sense in `EditWidgetIcon`, as there is also a bit of commentary that explains that the `undefied` there has meaning :slightly_smiling_face:

app/containers/EditWidget.jsx
Gelio: Great idea :slightly_smiling_face: I'll do that

app/components/Widget.jsx
Gelio: Cool point :slightly_smiling_face: I agree. I have always used the `classnames` library before, and I did not think much here. I'll change it

app/styles/style.scss
Gelio: I was thinking about it and decided not to name it `margin`, since it's not the `margin` CSS property that is set, but `top`/`right`, and IMO using a misleading name is worse than a more general, but correct name

kubama: OK, sounds reasonable

app/containers/EditWidget.jsx
Gelio: Done :slightly_smiling_face: While moving the file I have also noticed that this component does not use anything from the store, yet it is connected, so I removed the connection and refactored it to a function component :slightly_smiling_face: I hope you do not mind

I have verified that it's still working fine and the configs are retrieved correctly when editing widgets:

image

app/containers/EditWidget.jsx
Gelio: Done :slightly_smiling_face: Although I do not see much of a benefit

app/components/Widget.jsx
Gelio: Done :slightly_smiling_face:

app/components/Widget.jsx
```diff
@@ -154,9 +154,13 @@ export default class Widget extends Component {
...
-                    className={`widgetItem ${
-                        isEditMode && widget.definition && !widget.definition.showBorder ? 'borderOnHover ' : ''
-                    } ${widget.definition.showHeader ? '' : 'headerless'}`}
+                    className={[
+                        'widgetItem',
+                        isEditMode && widget.definition && !widget.definition.showBorder ? 'borderOnHover' : '',
+                        widget.definition.showHeader ? '' : 'headerless'
+                    ]
+                        .filter(Boolean)
+                        .join(' ')}
```
kubama: It's not exactly what I meant. I was thinking about using regular `if`s for adding elements to the array conditionally. Here we are still using the elvis operator which is hard to read and not really suitable (it's an `if...else` used where just `if` is needed).
Alternatively, the classnames library that you are referring to seems to solve this issue in an elegant and simple way. Maybe we could start using it?

Gelio: Now I see :slightly_smiling_face: I'm not a fan of extracting the class names to a variable in the function scope, because it increases the distance between a simple variable that is only used in 1 place, meaning you have to scroll/jump around the file more than necessary :slightly_smiling_face: Besides, doing `push`es sounds more imperative and would take up more lines of code

I'm not a fan of using ternary operators here too, but I was thinking maybe we could use just the `&&`?

```ts
            className={[
                        'widgetItem',
                        isEditMode && widget.definition && !widget.definition.showBorder && 'borderOnHover',
                        !widget.definition.showHeader && 'headerless'
                    ]
                        .filter(Boolean)
                        .join(' ')}
```

That looks reasonable to me and is declarative :slightly_smiling_face: What do you think?

app/containers/EditWidget.jsx
qooban: Great 👍🏻

app/containers/EditWidget.jsx
qooban: OK, let's keep it.

app/components/Widget.jsx
kubama: > Now I see 🙂 I'm not a fan of extracting the class names to a variable in the function scope, because it increases the distance between a simple variable that is only used in 1 place, meaning you have to scroll/jump around the file more than necessary 🙂

You could extract the `className` construction logic into separate function.
> Besides, doing `push`es sounds more imperative and would take up more lines of code

Thanks to this it would be clear and readable.
> 
> I'm not a fan of using ternary operators here too, but I was thinking maybe we could use just the `&&`?
> 
> ```ts
>             className={[
>                         'widgetItem',
>                         isEditMode && widget.definition && !widget.definition.showBorder && 'borderOnHover',
>                         !widget.definition.showHeader && 'headerless'
>                     ]
>                         .filter(Boolean)
>                         .join(' ')}
> ```
> 
> That looks reasonable to me and is declarative 🙂 What do you think?

It's certainly much better.
I think the pattern with ternary operator in `className` construction is used in other places as well, maybe we could create a common util function that could be reused and which would apply the `.filter(Boolean).join(' ')` part?

.editorconfig
qooban: > I have added the indent sizes here based on what I observed in multiple files in this repo
Hmm... Actually, I see it differently as we've got:
* one .scss file - `style.scss` - in this repository and there are both - 2 and 4 spaces indents
* multiple .json files using 4 spaces indents (see `.eslintrc.json` files or fixtures in cypress tests)
so suggested `indent_size = 2` describes more "where you want to go" and not what is the most common indent size in the repo. 

> Sure, I'll add a ticket to make it consistent across repos: https://cloudifysource.atlassian.net/browse/RD-1213
Thanks.

app/components/Widget.jsx
Gelio: Extracting the className construction into a separate function would still mean jumping around the file, which I'm not a fan of :slightly_smiling_face: 

I will extract that `.filter.join` part to an util (`app/utils/shared/combineClassNames.js`) and expose it in `stageUtil.js`, so it is accessible to both widgets and components. I will do that tomorrow, so if you have a better name/place suggestion, let me know :slightly_smiling_face:

.editorconfig
Gelio: > Hmm... Actually, I see it differently as we've got:

I did not inspect those files in their entirety :smile: You're right. If you want to handle that all in 1 ticket, let me know, and I'll revert the changes to this editorconfig.

I don't think it hurts to have them here and start fixing it part by part, but I don't press strongly :slightly_smiling_face:

app/components/Widget.jsx
kubama: > I will extract that `.filter.join` part to an util (`app/utils/shared/combineClassNames.js`) and expose it in `stageUtil.js`, so it is accessible to both widgets and components. I will do that tomorrow, so if you have a better name/place suggestion, let me know 🙂

I'm perfectly fine with your proposal.

Gelio: Done in e57c4bd0b6a087d6ee47e1e1ccb3deadbcdaf630, along with tests :slightly_smiling_face:

test/jest/utils/shared/combineClassNames.test.js
```diff
@@ -0,0 +1,33 @@
...
+        },
+        {
+            description: 'multiple strings with falsy values',
+            classNames: ['a', false, 'b', 0, '', 'c'],
+            expectedResult: 'a b c'
+        },
+        {
+            description: 'all falsy values',
+            classNames: [false, 0, ''],
+            expectedResult: ''
```
kubama, line 23: Great test covering multiple scenarios 👍.
However, looking at this 'all falsy values' case, shouldn't the expected value be rather `undefined`? Depending on component internals passing `''` to `className` could result in rendering `class=""` attribute in generated HTML. I guess having `undefined` in this case would equal to not specyfing `className` at all, and as such would be more appropriate. What do you think?

Gelio: Thanks :smile: 

Good point :slightly_smiling_face: I can see a problem with trying to use `combineClassNames` that returns `undefined` in a string interpolation, which would cause `undefined` to be inserted in the string, as opposed to an empty string:

image

The classnames library in this case returns an empty string (https://www.npmjs.com/package/classnames#usage):
image

Personally, I would leave it as is now and conform to the classnames' library behavior, to not potentially introduce unexpected `undefined` in some HTML element's `class` :slightly_smiling_face: 

I am open for discussion, though, if you think we should still return `undefined` here

kubama: I wouldn't expect anyone to use string interpolation, in fact `combineClassNames` purpose is not to use it.
But in the end it's a cosmetic thing, so I'm OK with empty string.

kubama (approved):

.editorconfig
qooban: OK, let's leave it as you did. 
I have that topic on my list for tomorrow's UI weekly meeting.

qooban (approved):

Gelio:
As agreed with both Kubas, I'm merging this PR

PR #1182, opened 2021-01-26
RD-1205 Remove HSTS header from proxied Stadia response

Gelio:
Stadia sends an HSTS header that was proxied to the browser, which forced subsequent connections to use SSL, regardless of whether it was configured or not.

A confirmation that there is no `strict-transport-security` header in the response for `maps`:
image


backend/routes/Maps.js
```diff
@@ -27,11 +27,16 @@ router.get('/:z/:x/:y/:r?', (req, res) => {
...
+        request(url)
+            .on('error', err => {
+                const message = 'Cannot fetch map tiles.';
+                logger.error(message, err);
+                res.status(500).send({ message });
+            })
+            .on('response', proxiedResponse => {
+                // NOTE: Stadia enforces HSTS, but SSL is not required when using Cloudify Manager
+                delete proxiedResponse.headers['strict-transport-security'];
+            })
```
qooban, line 39: One notice here: Right now we are using Stadia Maps, but it happened already that we switched the maps provider (OSM used before). Also, customers can change the provider using configuration (see `userConfig.json` file). So just to be sure, I assume that it will work properly also for the cases when there is no `strict-transport-security` header set, right?

backend/test/routes/Maps.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+        return proxiedRequest;
+    });
+
+    mockRequest.defaults = () => mockRequest;
+
+    return mockRequest;
+});
+
+describe('/maps endpoint', () => {
```
qooban, line 62: BTW Maybe we can also have a single test using the real service (Stadia Maps)? Just to verify we get 200 OK status and map image in response. That way we could detect potential problems with that service earlier (than in customers environment).

Stadia Maps access is secured by the token (which is added during build time), but there should also be some limited free access.

qooban (commented):
Added 2 comments.

kubama (approved):

backend/routes/Maps.js
Gelio: That's correct, `delete` on a non-existing property works fine :slightly_smiling_face: 

I've removed the HSTS header from the mock response and the test passed
image

backend/test/routes/Maps.test.js
Gelio: We could, but I have not found a way to execute a request without an API key. There is [a free tier](https://stadiamaps.com/pricing/), yet it requires signing up, or, if I understand correctly, making requests directly from the browser (hence they mention 127.0.0.1 or localhost).

It seems out of scope for this ticket, which is about handling HSTS, but I reckon I could create a new ticket for adding a test that runs only when the access token is filled in, and uses the real service. Let me know if that's what you mean

Out of curiosity, why is the token injected in Jenkins? Do we interact with the map somehow? If yes, where is it tested?

qooban: Thanks for checking that, I think we can leave it, then.
It should have already been tested within cypress tests (`sites_map_spec.js`).

Regarding the token - it is injected in Jenkins for the security reasons: not to keep it in the public repository.

qooban (approved):

backend/test/routes/Maps.test.js
Gelio: I see, I did not spot the `sites_map_spec` file :slightly_smiling_face: Thanks

Gelio:
I tried to run Stage system tests for this branch to make sure nothing is broken, but they failed twice on some connectivity error, not a problem in the code :worried: https://jenkins.eks.cloudify.co/job/Stage-UI-System-Test/130/console 

I'll merge the PR without the system tests passing

Gelio:
Thanks for swift reviews @qooban, @kubama

PR #1184, opened 2021-02-01
RD-1230 Add TypeScript

Gelio:
This PR adds TypeScript support to this repository. It builds on https://github.com/cloudify-cosmo/cloudify-ui-common/pull/46 and https://github.com/cloudify-cosmo/cloudify-ui-components/pull/56

It is also waiting for https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1715 to be merged before CI passes

Bundle sizes

File | Size before TS | Size after TS
-- | -- | --
main.bundle.js | 271 | 281.1074219
vendor.bundle.js | 4352 | 4270.605469
widgets | ? | ?

The bundle size can be greatly decreased if we enable `loose` mode and set `browserslist`: https://cloudifysource.atlassian.net/browse/RD-1312

See https://docs.google.com/spreadsheets/d/1rP41QOnJ2clBBikEKy_M3oiPEbzODYsSBGRVNXJgawc/edit#gid=0 for a complete comparison

## Changes
1. ESLint

    1. Install TS plugin and parser
    2. Add overrides that allow a bit looser typing (`.eslintrc.temp-overrides.json`)
2. Babel - add TS preset. Upgrade packages to allow using optional chaining (`a?.b`) and nulling coalescing (`a ?? b`)
3. Webpack
    1. Add `fork-ts-checker-webpack-plugin` that allows reporting ESLint and TypeScript errors in webpack
    2. Expose lodash as external package. This means widgets can use `import _ from 'lodash';` which will be transpiled to using the lodash package :slightly_smiling_face: IMO it's a bit nicer API than using global variables. The same goes for using `React`.
    3. Use `babel.config.js` as the default config for `babel-loader`. Do not override it
4. `package.json`
    1. Add scripts for checking types.
        Only TS files report errors. Errors in JS files are fine.
5. Add tsconfigs for backend, backend's tests, frontend, frontend's tests, cypress tests, widgets
6. Add checking types in Jenkins
7. Adjust jest config for TypeScript

### Changes in the code

1. Add a very simple interface for the `Stage` API (`StageAPI` interface). Add necessary typings to expose it as a global variable in the widgets.
2. Fix some TS/ESLint errors
3. Add some module imports to a few widgets. They can import lodash and React freely, which IMO is better than using global variables. Maybe we could also import the `Stage` API? :thinking: 
4. Move a few jest tests to TS
5. Add typings for a custom Cypress command as an example
6. Move 1 Cypress test to TS
7. Add a bit more details to `StageAPI` so it is usable

## Won't do

1. Add support for TS in backend (`ts-node` or similar) ->https://cloudifysource.atlassian.net/browse/RD-1290
2. Emit type declarations for `StageAPI` for other widget developers to use (outside of `cloudify-stage` repo) -> https://cloudifysource.atlassian.net/browse/RD-1292
3. ~~Try to allow using `import StageAPI from 'cloudify-stage';` in widgets~~ I did not manage to do it quickly
4.  ~~Check if `@babel/polyfill` is necessary (it says it's deprecated) and if not - remove it~~ I want to avoid breaking changes. https://cloudifysource.atlassian.net/browse/RD-1300


test/jest/widgets/common/DeploymentsStates.test.ts
```diff
@@ -7,12 +9,12 @@ describe('(Widgets common) DeploymentStates', () => {
 
     it('returns in progress state', () => {
         Stage.Utils.Execution.isActiveExecution = _.constant(true);
-        expect(DeploymentStates.getDeploymentState(null, {})).toBe(DeploymentStates.IN_PROGRESS_STATE);
+        expect(DeploymentStates.getDeploymentState('', {})).toBe(DeploymentStates.IN_PROGRESS_STATE);
```
kubama, line 12: Why this change - is `null` invalid where `string` is expected?

widgets/pluginsNum/src/widget.jsx
```diff
@@ -2,6 +2,9 @@
  * Created by pawelposel on 04/11/2016.
  */
 
+import * as _ from 'lodash';
+import React from 'react';
```
kubama, line 6: Are these no longer available in global scope?

Gelio: They are available :slightly_smiling_face: Other widgets remain unchanged. However, IMO, importing them is less _magical_ and the more standard way. Webpack will transform those imports to using the global variables.

So, widget authors can choose whether to use variables globally, or import them (which I prefer)

kubama: > They are available 🙂 Other widgets remain unchanged. However, IMO, importing them is less _magical_ and the more standard way. Webpack will transform those imports to using the global variables.

How would it happen? Won't it lead to including imported dependencies in widget bundle (which is something we were avoiding by using the global scope)?

backend/app.js
```diff
@@ -140,7 +140,7 @@ app.get('*', (request, response) => {
 
 // Error handling
 // eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
Gelio: Good point :smile: Actually, it looks like it was not needed before, because according to https://eslint.org/docs/rules/no-unused-vars#options only unused parameters that appear after the last used parameter are considered errors. Meaning that `req`, which is unused in the line below, is ok, since it appears in the middle.

I will remove it :slightly_smiling_face: Good catch, thanks

test/jest/widgets/common/DeploymentsStates.test.ts
Gelio: Yes, your reasoning is correct. This is thanks to [strictNullChecks](https://basarat.gitbook.io/typescript/intro/strictnullchecks) enabled in `tsconfig.json`.

When enabled, `null` or `undefined` can be provided as an argument only when the function specifies that it's possible

e.g.
```ts
const fn = (param: string) => {}
fn(null) // invalid

const fn2 = (param: string | null) => {}
fn(null); // valid
```

This enforces that you handle null/undefined values explicitly, rather than let them flow through the code silently. I like it, as it provides better guarantees about the possible types of values.

kubama: Got it 👍

test/cypress/support/commands.ts
```diff
@@ -32,20 +32,58 @@ const getCommonHeaders = () => ({
...
+            setBlueprintContext(field: string, value: string): Chainable;
+            clearBlueprintContext(field: string): Chainable;
+            setDeploymentContext(field: string, value: string): Chainable;
+            clearDeploymentContext(field: string): Chainable;
+        }
+    }
+}
 
-Cypress.Commands.add('waitUntilPageLoaded', () => {
+Cypress.Commands.add('restoreState', (() => cy.restoreLocalStorage()) as Cypress.Chainable['restoreState']);
```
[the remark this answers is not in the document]
Gelio: Good observation :smile: I did not think about a better solution initially, but there is a better solution indeed! One that removes this need for repetition, and improves the ability of jumping to the source code for a given command :smile: I'll change it shortly

It involves some advanced TypeScript, though :slightly_smiling_face: I hope you'll like it

widgets/pluginsNum/src/widget.jsx
Gelio: That's due to the `externals` part defined in `webpack.config.js`:
```ts
    const externals = {
        react: 'React',
        'react-dom': 'ReactDOM',
        lodash: '_'
    };
```

https://webpack.js.org/configuration/externals/

I'm pretty sure `lodash` is not bundled twice, but we will have the final confirmation when `size-limit` runs for this PR :slightly_smiling_face: Anyway, I've checked the dist directory and here's the output for `pluginsNums` widget:

image

There does not seem to be any lodash/React here :slightly_smiling_face: When I looked at a formatted version of thee above, it looks as follows:
image

`i` (the variable for `_`) comes from `n(19)`, which looks like referencing lodash from somewhere else

kubama: OK, thanks for checking out and confirming. I was concerned as I thought global scope was a workaround to not include dependencies with each widget bundle. But if added webpack configuration does the trick, then it's great 👍.

widgets/pluginsNum/src/widget.jsx
```diff
@@ -2,6 +2,9 @@
  * Created by pawelposel on 04/11/2016.
  */
 
+import * as _ from 'lodash';
```
kubama, line 5:
```suggestion
import _ from 'lodash';
```

backend/app.js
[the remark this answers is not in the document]
Gelio: Removed :slightly_smiling_face:

test/cypress/support/commands.ts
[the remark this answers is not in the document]
Gelio: I've modified it, and I like the new solution better :smile: It allows jumping straight to the function definition when doing "Go to source"

Check it out :slightly_smiling_face:

widgets/pluginsNum/src/widget.jsx
Gelio: Done :slightly_smiling_face:

test/cypress/support/commands.ts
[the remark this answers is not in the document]
kubama: I can't say I understand how it works, but if it does it's more than great 😃.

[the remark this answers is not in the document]
Gelio: Good point, I will try to explain in on Friday, to show the capabilities of TS :smile:

widgets/topology/src/DataFetcher.js
```diff
@@ -1,3 +1,5 @@
+import * as _ from 'lodash';
```
kubama, line 1:
```suggestion
import _ from 'lodash';
```

Gelio:
I have updated this PR with `cloudify-ui-common`@3 (with TS). I have also merged `master` to this PR to resolve conflicts (in `webpack.config.js` and `commands.js` in Cypress).

This PR is still waiting for https://github.com/cloudify-cosmo/cloudify-ui-components/pull/56, so I'm keeping it as a draft.

Gelio:
I have installed the new version of https://github.com/cloudify-cosmo/cloudify-ui-components/ (the one with TypeScript declarations).

Thus, this PR is ready for reviewing & merging :slightly_smiling_face:

Gelio:
The CI failed as the PR is waiting for https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1715 to be merged. It can still be reviewed in the meantime

test/cypress/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "extends": "../../node_modules/cloudify-ui-common/configs/tsconfig.base.json",
```
kubama, line 2: The extended configuration file sets `target` to `ESNext`. Cypress tests are run in node, so I guess this target is not supported, or at least not fully. Or is it?

test/cypress/tsconfig.json
```diff
@@ -0,0 +1,11 @@
...
+    "compilerOptions": {
+        // Some files do not import/export anything, so cannot use isolatedModules
```
kubama, line 4: I can see you use comments in JSON files extensively. AFAIK comments in JSON are not valid, at least not per the specification (or are they?). I assume it all works correctly, but doesn't it put us at risk of parse failures?

backend/app.js
```diff
@@ -139,8 +139,7 @@ app.get('*', (request, response) => {
 });
 
 // Error handling
-// eslint-disable-next-line no-unused-vars
-app.use((err, req, res, next) => {
+app.use((err, req, res) => {
```
qooban, line 142: Why did you remove it?

According to https://expressjs.com/en/guide/error-handling.html - "Writing error handlers" it is necessary to have 4 arguments instead of 3 for ExpressJS to recognize this as error handler. Even if you don't use the latest (`next`) argument.

app/utils/scriptLoader.js
```diff
@@ -13,17 +13,16 @@ export default class ScriptLoader {
     load(id, rejectOnError) {
         log.log(`Loading javascript from ${this.path}...`);
 
-        const scriptLoader = this;
```
qooban, line 16: Is it related to TypeScript changes or just improved by the way? 
Same comment for `app/utils/StyleLoader.js`.

app/utils/Context.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Created by kinneretzin on 14/11/2016.
+ */
+
```
qooban, line 4: You can remove it. 

When we did refactoring we usually remove this type of comments.

app/utils/Context.ts
```diff
@@ -0,0 +1,29 @@
...
+    private context: any;
+
+    constructor(store: MiddlewareAPI) {
+        this.store = store;
+        this.context = store.getState().context;
+    }
+
+    public setValue(key: any, value: any) {
+        this.store.dispatch(setContextValue(key, value));
+        // NOTE: should we not set this.context[key] = value too?
```
qooban, line 23: I suspect the reason why it is not set here is that in `Toolbox` we have:
```
        this.unsubscribe = store.subscribe(() => {
            this.initFromStore();
        });
```
and
```
    initFromStore() {
        const state = this.store.getState();
        // (...)
        this.context = new Context(this.store);
    }
```
so `Context` is re-created after any change in the store.

app/typings/stage-api.d.ts
```diff
@@ -0,0 +1,20 @@
...
+    export const Stage: StageAPI;
+
+    // External libraries
+    // @see https://docs.cloudify.co/developer/writing_widgets/widget-apis/#external-libraries
+    export const PropTypes: typeof import('prop-types');
+    export const moment: typeof import('moment');
+    export const React: typeof import('react');
+    export const _: typeof import('lodash');
+
+    interface Window {
```
qooban, line 13: Can't we add all libs attached to the window?

In `app/app.jsx` there are also:
```
window.L = Leaflet;
window.log = log;
window.connectToStore = connect;
window.moment = moment;
```

.eslintrc.temp-overrides.json
```diff
@@ -0,0 +1,7 @@
+{
```
qooban, line 1: Can we reference an issue in JIRA which will be about removing this file?

app/utils/widgetDefinitionsLoader.ts
```diff
@@ -61,7 +65,7 @@ export default class WidgetDefinitionsLoader {
             GenericConfig,
             Utils: StageUtils,
 
-            Common: [],
+            Common: {},
```
qooban, line 68: :-) Good catch.

widgets/common/src/DeploymentStates.ts
```diff
@@ -1,3 +1,5 @@
+import _ from 'lodash';
```
qooban, line 1: Is it necessary here after introducing TS? Does it extend `common.js` bundle file with `lodash` lib?

widgets/deploymentNum/src/widget.tsx
```diff
@@ -2,7 +2,18 @@
  * Created by pawelposel on 03/11/2016.
  */
 
-Stage.defineWidget({
+import { isEmpty } from 'lodash';
+import React from 'react';
```
qooban, line 6: Is it necessary?

widgets/deploymentNum/src/widget.tsx
```diff
@@ -2,7 +2,18 @@
  * Created by pawelposel on 03/11/2016.
  */
 
-Stage.defineWidget({
+import { isEmpty } from 'lodash';
```
qooban, line 5: We have full lodash in `_`, so I think it is not necessary here.

widgets/topology/src/DataFetcher.js
qooban: I think this one is not even necessary to include.

widgets/pluginsNum/src/widget.jsx
```diff
@@ -2,6 +2,9 @@
  * Created by pawelposel on 04/11/2016.
  */
 
+import _ from 'lodash';
+import React from 'react';
+
```
qooban, line 7: Is it necessary to include?

qooban (commented):
Added some comments. 

I see that you managed to make file extension change + file content change to be visible properly (e.g. `PollHelper.ts`) - that's very good! 👍🏻 

Unfortunately not for all, see files from `test/cypress/support` directory. It makes quite hard to review them.

test/cypress/tsconfig.json
Gelio: I'm not certainly sure here :slightly_smiling_face: `target` changes which ES feature are available (https://www.typescriptlang.org/tsconfig#target), so yeah, going with `ESNext` may yield some false-negatives, when TS will not transpile some feature, and then NodeJS will not understand it.

What we could do is change `target` to `ES2019`, as is done in [the recommended config for Node 12](https://github.com/tsconfig/bases/blob/master/bases/node12.json), and also mimic `lib`.#1184 

Actually, after thinking about it for a while, let's do that :smile: It's probably worth it to consider fine-tuning `target` and `lib` in other `tsconfigs`

test/cypress/tsconfig.json
Gelio: Yup, it works well. Comments are generally welcome in `tsconfig.json` files (e.g. [the recommended tsconfig for Svelte](https://github.com/tsconfig/bases/blob/master/bases/svelte.json)), and looks like it was added a while back :slightly_smiling_face: https://github.com/microsoft/TypeScript/pull/5450

Despite those comments rendering improperly in GitHub diff editor, they render correctly e.g. in VSCode when working with `tsconfig.json`, and in my TS career I have not seen a tool fail because of a comment in tsconfig.json :slightly_smiling_face: 

As an aside note, there is [a proposal for `jsonc`](https://komkom.github.io/) which is JSON with comments :smile: There is even such a file type recognized by VSCode

Gelio:
@qooban 

> Unfortunately not for all, see files from test/cypress/support directory. It makes quite hard to review them.

Sorry for that, but I'm not sure there's much I can do, as we have seen in https://github.com/cloudify-cosmo/cloudify-ui-components/pull/56#discussion_r572763573. I'm not sure if it's a general comment, or a silent suggestion to improve it in some way (which I do not have an idea about :smile: )

You can see the individual commits that make the changes:
1. https://github.com/cloudify-cosmo/cloudify-stage/pull/1184/commits/49e78ff5baf62bcf8710354f3d509ce313c0afe4 - here `commands.ts` is detected as a rename and not much is changed
2. https://github.com/cloudify-cosmo/cloudify-stage/pull/1184/commits/641225e3c1bffebac823b514bb9248fc41283b4d - here is a refactor after [@kubama's comment to see if there is a way to avoid duplicating command name](https://github.com/cloudify-cosmo/cloudify-stage/pull/1184#discussion_r569431968).

I have not modified the logic in the functions, just the first and last line of each command, since instead of calling `Command.add` the command is assigned to an object property.

qooban:
@Gelio 
Got it 👍🏻 All good.
It wasn't silent suggestion, just general comment as I'm aware of the problems in that case 🙂

backend/app.js
Gelio: Today I learned that Express has some additional error handling, and that there are magic rules related to the number of parameters :smile: Thanks!

I'll revert the change :smile: 

I reckon it would be helpful to have an explanation as to why we are disabling the `no-unused-vars` rule, and to signal that it's not someone being lazy and leaving the last parameter :smile: I'll add it too

app/utils/scriptLoader.js
Gelio: It was picked up by [a rule](https://github.com/typescript-eslint/typescript-eslint/blob/v4.14.1/packages/eslint-plugin/docs/rules/no-this-alias.md) in `@typescript-eslint` recommended config that is extended in `cloudify-ui-common`'s eslint config :slightly_smiling_face: I decided to fix it instead of adding eslint ignore comments, since I believe the rule makes sense

app/utils/Context.ts
Gelio: Ok, good to know :smile: I removed it in the few files that GitHub marks as added (to make the diff that much smaller :smile: )

widgets/topology/src/DataFetcher.js
Gelio: Yup, it's not :smile: lodash can be imported either way

I added it anyway since we do `import _ from 'lodash'` most often in the project :slightly_smiling_face:

test/cypress/tsconfig.json
kubama: It seems even WebStorm doesn't complain about comments in ts configuration files, so 👍.
However, looks like you used comments in other JSON files as well, for example `.eslintrc.temp-overrides.json`. In this case WebStorm marks this file as invalid, which is not really welcome.

app/utils/Context.ts
Gelio: I see :smile: Seems kind of wasteful, but cool

I'll add such an explanation there

Gelio: But wait a second. Is it not a problem if we do not change `this.context`, and `getValue` reads from it?

Imagine:

```ts
const v1 = context.getValue('a'); // imagine it returns 1
context.setValue('a', 2);
const v2 = context.getValue('a'); // still returns 1

console.log(v1 === v2); // false
```

IMO that's unintuitive :smile: I'd rather fix it by assigning `this.context[key] = value`, but since `context` is the same object that is in redux, it will modify the store in an impure way.

I can make a note here and add a tech debt item to fix this behavior and make the snippet above return true. What do you say?

app/typings/stage-api.d.ts
Gelio: Yup, I must have forgotten about the rest when adding them initially :sleeping: Added `ReactDOM`, `L`, `log`, `connectToStore`

.eslintrc.temp-overrides.json
Gelio: Good idea :+1:

app/utils/widgetDefinitionsLoader.ts
Gelio: Thanks :smile:

widgets/common/src/DeploymentStates.ts
Gelio: @kubama brought it up in https://github.com/cloudify-cosmo/cloudify-stage/pull/1184#discussion_r569438035 

I encourage you to read the whole discussion there :slightly_smiling_face: 

Short answer:
1. It's not necessary, but IMO it's less magical and it's good style to use explicit imports
2. The bundle size is safe thanks for using `externals` in webpack config :slightly_smiling_face:

widgets/deploymentNum/src/widget.tsx
Gelio: That's a personal opinion, but I prefer explicit imports (hence `import` instead of using the global `_`), and also, when using imports, I prefer being precise. In this case it does not matter, since we always load the whole lodash, but if we didn't, we could probably yield some kBs thanks to tree-shaking, which is easier done when using named imports :slightly_smiling_face:

widgets/pluginsNum/src/widget.jsx
Gelio: I assume the answer will be the same as https://github.com/cloudify-cosmo/cloudify-stage/pull/1184#discussion_r574444620

Gelio:
@qooban Alright, cool, thanks for explaining :smile:

package.json
```diff
@@ -22,14 +22,19 @@
     "build:analyse:widgets": "npm run build -- --env.analyse=widgets",
     "build:analyse:widgets-common": "npm run build -- --env.analyse=widgets-common",
     "build:coverage": "npm run build -- --env.coverage",
+    "check-types": "npm run check-types:backend && npm run check-types:frontend && npm run check-types:widgets && npm run check-types:cypress",
+    "check-types:backend": "tsc -p backend --checkJs false && tsc -p backend/test --checkJs false",
+    "check-types:frontend": "tsc -p app --checkJs false && tsc -p test/jest --checkJs false",
```
kubama, line 27: I already asked this question in context of cloudify-ui-common, but I'll ask it here as well so it's also brought to others attention. Since we are building the app anyway, what's the purpose of running tsc on app folder explicitly? It only extends build time as the same task is performed twice during the build. Since we are all working with IDEs I can't imagine anyone running this script by hand.

test/cypress/tsconfig.json
Gelio: If that would suit you, I can convert that `json` file to a `js` file, which ESLint will also handle, and where comments are allowed. That will be something new, though, as we have been using JSON files for ESLint configs.

Or, we can just ignore it, since it works :smile: And IMO explanatory comments for rules are a worthy addition

Let me know what you'd prefer. Personally, I would leave it as is

app/components/LicensePage.jsx
```diff
@@ -51,6 +51,7 @@ function DescriptionMessage({ canUploadLicense, isTrial, isEditLicenseActive, on
                         </Message.Header>
                         {canUploadLicense ? (
                             <span
+                                // eslint-disable-next-line react/no-danger
```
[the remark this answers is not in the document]
Gelio: It was a warning

image

I'm not sure why it was not picked up before. Perhaps changing something in the common's eslint config helped it surface

app/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "../node_modules/cloudify-ui-common/configs/tsconfig.base.json",
+    "compilerOptions": {
+        "types": []
```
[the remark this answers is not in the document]
Gelio: By default, all types from `@types/*` are included automatically. Here we are saying that we do not want anything to be explicitly allowed as a global variable, since we are manually controlling what's allowed globally in `stage-api.d.ts`

For example, without this line you could write `jest.fn()` somewhere in the `app` and TypeScript would allow it :slightly_smiling_face:

I know that ESLint has a rule that would detect it, but by letting TS know that it should not handle such global exports, we probably reduce the amount of memory it uses, since it does not need to remember about those global types

package.json
Gelio: That's true that we are compiling `app` and `backend` twice, since `backend/test` and `test/jest` already include the corresponding directories.

Here I specified it just for completeness, and to make it look as expected (you probably would expect to compile both `app` and `test/jest` when compiling frontend)

Since I agree that the compilation can take longer here than in `cloudify-ui-common`, we can change it. Just FYI:

```ts
13:12 $ time npm run check-types

> cloudify-stage@5.3.0--.dev1 check-types /home/grzegorz/projects/cloudify/cloudify-stage
> npm run check-types:backend && npm run check-types:frontend && npm run check-types:widgets && npm run check-types:cypress


> cloudify-stage@5.3.0--.dev1 check-types:backend /home/grzegorz/projects/cloudify/cloudify-stage
> tsc -p backend --checkJs false && tsc -p backend/test --checkJs false


> cloudify-stage@5.3.0--.dev1 check-types:frontend /home/grzegorz/projects/cloudify/cloudify-stage
> tsc -p app --checkJs false && tsc -p test/jest --checkJs false


> cloudify-stage@5.3.0--.dev1 check-types:widgets /home/grzegorz/projects/cloudify/cloudify-stage
> tsc -p widgets --checkJs false


> cloudify-stage@5.3.0--.dev1 check-types:cypress /home/grzegorz/projects/cloudify/cloudify-stage
> tsc -p test/cypress --checkJs false


real    0m17,837s
user    0m32,329s
sys     0m1,047s
```

and when skipping `app` and `backend`, the result is:

```sh
13:18 $ time npm run check-types

> cloudify-stage@5.3.0--.dev1 check-types /home/grzegorz/projects/cloudify/cloudify-stage
> npm run check-types:backend && npm run check-types:frontend && npm run check-types:widgets && npm run check-types:cypress


> cloudify-stage@5.3.0--.dev1 check-types:backend /home/grzegorz/projects/cloudify/cloudify-stage
> tsc -p backend/test --checkJs false


> cloudify-stage@5.3.0--.dev1 check-types:frontend /home/grzegorz/projects/cloudify/cloudify-stage
> tsc -p test/jest --checkJs false


> cloudify-stage@5.3.0--.dev1 check-types:widgets /home/grzegorz/projects/cloudify/cloudify-stage
> tsc -p widgets --checkJs false


> cloudify-stage@5.3.0--.dev1 check-types:cypress /home/grzegorz/projects/cloudify/cloudify-stage
> tsc -p test/cypress --checkJs false


real    0m16,527s
user    0m29,206s
sys     0m0,877s
```

So, it's 17.837s vs 16.527s - the difference is negligible to me :smile: 

What I can suggest in this case is:
1. Using [RD-1295 Project references](https://cloudifysource.atlassian.net/browse/RD-1295) that will speed up the compilation, but require some configuration work
2. Skipping compiling `app` and `backend` and adding some explanation comments since that's unexpected to me
3. Not doing anything for now, since the time overhead is negligible

I recommend going with option 3 :smile:

app/tsconfig.json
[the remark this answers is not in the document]
kubama: OK, thanks for explaining

app/utils/Context.ts
```diff
@@ -0,0 +1,25 @@
+import { MiddlewareAPI } from 'redux';
+import { setValue as setContextValue } from '../actions/context';
+
+/** @see https://docs.cloudify.co/developer/writing_widgets/widget-apis/#getcontext */
+export default class Context {
+    private store: MiddlewareAPI;
```
gregcfy, line 6: it is good practice to use `readonly` for private properties assigned inside the constructor
e.g.
private readonly store: MiddlewareAPI;

app/utils/EventBus.ts
```diff
@@ -1,18 +1,16 @@
...
+type Callback = (...args: any[]) => void;
+
 export default class EventBus {
-    static events = {};
+    private static events: Record<string, [Callback, any][]> = {};
 
-    static on(event, callback, context) {
+    public static on(event: string, callback: Callback, context?: any) {
         this.events[event] = this.events[event] || [];
         this.events[event].push([callback, context]);
```
gregcfy, line 10: In the form of curiosity:
```
const events = this.events[event] ?? (this.events[event] = []);
events.push([callback, context]);
```

app/utils/StyleLoader.js
```diff
@@ -11,14 +11,13 @@ export default class StyleLoader {
     }
 
     load() {
```
gregcfy, line 13: maybe in future we should use:
```
load = () => {
//...
};
```

app/utils/Context.ts
qooban: `getValue` just after `setValue` is unlikely IMHO. Let's leave a note and do not change the behavior now.

widgets/deploymentNum/src/widget.tsx
qooban: I understand your point perfectly. 

I just don't like the idea of changing the whole approach to handling external dependencies (like `lodash`) within this PR, but let it be as I'm all for merging this PR as soon as possible.

qooban (dismissed):

widgets/deploymentNum/src/widget.tsx
Gelio: Well, no rush here, I can extract those pieces and move them to a separate PR, or remove them entirely and postpone them. I'm happy as long as we make this change relatively soon, or you convince me that such imports are a bad idea :smile: 

I'm not sure if you agree with the motion, but prefer not to combine it with TS, or disagree with the motion :sweat_smile:

app/utils/Context.ts
Gelio: Yup, good point :smile: I'll change it here

app/utils/EventBus.ts
Gelio: Good point, but I would rather avoid making such functional changes as I don't have certainty it won't break :smile: I'd rather pass on this one

app/utils/StyleLoader.js
Gelio: Yup, we often do in new files or when refactoring :smile: I'd rather not make the change here

app/utils/Context.ts
Gelio: I guess you're right, we can assume that `context` will be used similarly to `this.state` in class components, where `this.state` does not immediately change after `this.setState` (actually, a similar thing is true for `useState` :smile: )

I'll add some explanatory comments

app/utils/EventBus.ts
gregcfy: yes, let's stay without functional changes

test/cypress/tsconfig.json
kubama: I just found out that WebStorm allows JSON comments in eslint configuration files (good to know!), it just doesn't recognise this particular file as eslint config. Since this is a temporary solution anyway - lets leave it.

app/utils/Context.ts
Gelio: Done :slightly_smiling_face:

app/utils/Context.ts
Gelio: Done :slightly_smiling_face: I've added the exact same description as we agreed on on Slack :slightly_smiling_face:

test/cypress/tsconfig.json
Gelio: Good to know :smile: thanks for sharing

package.json
kubama: You did the time measurement in a state where only few files in `app` where migrated to TS. As we proceed with the migration I expect this time to rise by orders of magnitude. We could try to estimate it based on number of TS and JS files in the `app`, but I don't think it's necessary - for sure the final time difference would be noticeable.
And still there is no real benefit. Also, if we consider something unexpected or unusual a proper comment is the most common way to address such situation.
Summing up, I'd go with option 2 for now.

qooban: For me 3rd option is fine, but I leave it to @kubama to decide.

test/cypress/tsconfig.json
qooban: So we can expect one more PR to `cloudify-ui-common`, right? How it relates to https://github.com/cloudify-cosmo/cloudify-ui-common/pull/47 ?

package.json
qooban: I commented in the same time and did not spotted what @kubama said 🙂 
Then let's go with option 2.

qooban:
@Gelio Looks like you need to update size limits:
https://jenkins.cloudify.co/blue/organizations/jenkins/cloudify-stage/detail/RD-1230-add-typescript/20/pipeline
```
[2021-02-11T13:43:09.313Z]   Main JS bundle
[2021-02-11T13:43:09.313Z]   Package size limit has exceeded by 9.67 KB
[2021-02-11T13:43:09.313Z]   Size limit:   272 KB
[2021-02-11T13:43:09.313Z]   Size:         281.67 KB
[2021-02-11T13:43:09.313Z]   Loading time: 5.7 s     on slow 3G
[2021-02-11T13:43:09.313Z]   Running time: 160 ms    on Snapdragon 410
[2021-02-11T13:43:09.313Z]   Total time:   5.8 s
[2021-02-11T13:43:09.313Z]   
[2021-02-11T13:43:09.313Z]   Vendor JS bundle
[2021-02-11T13:43:09.313Z]   Size limit:   4.25 MB
[2021-02-11T13:43:09.313Z]   Size:         4.1 MB
[2021-02-11T13:43:09.313Z]   Loading time: 84 s    on slow 3G
[2021-02-11T13:43:09.313Z]   Running time: 858 ms  on Snapdragon 410
[2021-02-11T13:43:09.313Z]   Total time:   84.9 s
[2021-02-11T13:43:09.313Z]   
[2021-02-11T13:43:09.313Z]   Widgets
[2021-02-11T13:43:09.313Z]   Size limit:   2.5 MB
[2021-02-11T13:43:09.313Z]   Size:         2.45 MB
[2021-02-11T13:43:09.313Z]   Loading time: 50.3 s  on slow 3G
[2021-02-11T13:43:09.313Z]   
[2021-02-11T13:43:09.313Z]   Application package
[2021-02-11T13:43:09.313Z]   Size limit:   13.05 MB
[2021-02-11T13:43:09.313Z]   Size:         13 MB
[2021-02-11T13:43:09.313Z]   Loading time: 266.4 s  on slow 3G
```

Please adjust both - Main (increase) and Vendor (decrease, that we will notice the increase easily in the future) JS Bundles.

test/cypress/tsconfig.json
Gelio: I was thinking about only adding `target: "ES2019"` in this `tsconfig.json`, which would override the `ESNext` value set in the main `tsconfig.base.json`. Then, @kubama would probably need to introduce the same change in `cypress/tsconfig.json` https://github.com/cloudify-cosmo/cloudify-ui-common/pull/47

We could also fine-tune the `target` value for each project individually now, but I would rather do it just in time to avoid spending time trying to define the correct `target` at the moment :slightly_smiling_face:

package.json
Gelio: Ok cool, let's do it :slightly_smiling_face:

test/cypress/tsconfig.json
Gelio: I decided to copy the values from https://github.com/tsconfig/bases/blob/master/bases/node12.json in the end

Gelio:
@qooban Thanks for mentioning decreasing `vendor` bundle size - I would not have thought about that.

I have modified the limits, although I am not sure if setting the vendor limit to the value shown by `npm run size` (4.1 MB) will be allowed :smile: Let's see what the CI will tell us

Gelio:
@qooban @kubama There is no way to re-request a review, but I would do it if it was possible :slightly_smiling_face:

kubama (approved):

qooban (approved):

Gelio:
I'm running Stage UI System tests for this branch: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/156/

I'll post another comment with its results.

Gelio:
Update about UI System tests: https://jenkins.cloudify.co/job/Stage-UI-System-Test/156/console

All except the known `topology_spec.js` have passed. Thus, I reckon it's safe to merge the PR, which I'll do right after posting this comment :slightly_smiling_face: 

Thank you @qooban @kubama @gregcfy for all the comments and the time spent reviewing the PR :slightly_smiling_face:

PR #1186, opened 2021-02-03
RD-752 - Add Manage Labels modal

qooban:
https://user-images.githubusercontent.com/5202105/106861440-0cd86c80-66c6-11eb-90fa-a1f2f4ba799b.mp4

Added new component available through `Stage.Common`. It is not used anywhere yet. Manage labels button was created just for testing and video recording. Deployment labels fetch is done when modal gets opened. On Apply button click, deployment labels are updated according to changes user did. I'm planning to add Cypress tests in the next tasks when it would be accessible from the widgets.

### DIFF
--- a/widgets/common/src/labels/common.js
+++ b/widgets/common/src/labels/common.js
@@ -1,5 +1,5 @@
 // eslint-disable-next-line import/prefer-default-export
 export function addSearchToUrl(url, search) {
     const { appendQueryParam } = Stage.Utils.Url;
-    return search ? appendQueryParam(url, 'search', search) : url;
+    return search ? appendQueryParam(url, '_search', search) : url;
 }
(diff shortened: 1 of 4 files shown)


widgets/common/src/labels/common.js
```diff
@@ -1,5 +1,5 @@
 // eslint-disable-next-line import/prefer-default-export
 export function addSearchToUrl(url, search) {
     const { appendQueryParam } = Stage.Utils.Url;
-    return search ? appendQueryParam(url, 'search', search) : url;
+    return search ? appendQueryParam(url, '_search', search) : url;
```
kubama, line 4: Just wondering - maybe it would be worth to extend `appendQueryParam` so it ignored given param if it's `null` or `undefined`?

qooban: I actually thought about that previously when extracting `appendQueryParam` to `Stage.Utils.Url`, but didn't want to get too far in the refactoring. I would prefer to leave it for the deeper change within [RD-1237](https://cloudifysource.atlassian.net/browse/RD-1237).

kubama (approved):

PR #1190, opened 2021-02-08
RD-1354 RD-1355 (5.2.0) Fix deployments widget system tests

Gelio:
Cherry-pick of https://github.com/cloudify-cosmo/cloudify-stage/pull/1189 to 5.2.0

### DIFF
--- a/test/cypress/integration/widgets/deployments_spec.js
+++ b/test/cypress/integration/widgets/deployments_spec.js
@@ -48,7 +48,7 @@ describe('Deployments widget', () => {
 
         it('clickToDrillDown option', () => {
             searchForDeployment(deploymentName);
-            cy.contains(deploymentName).click();
+            cy.get('.deploymentsWidget').contains(deploymentName).click();
             cy.location('pathname').should('contain', '_deployment/deployments_test_hw_dep');
             cy.contains('.breadcrumb', 'deployments_test_hw_dep');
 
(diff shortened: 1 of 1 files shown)


kubama (approved):
This is of course welcome and approved, however it's not really required - I don't think system tests on release branches are ever verified by anyone :).

Gelio:
@kubama Thanks for reviewing :slightly_smiling_face: 

I was under the impression that we use the build branch (`5.2.0-build` in this case) when running UI system tests. This is in order not to use `master` to do the testing, which could have a different feature set. I know it's pretty similar at the moment, but still :smile: If we were to release 5.3, and then 5.2.1 as a patch, we probably should use individual build branches to do the testing :slightly_smiling_face: 

This means that if this fix is included only on `master`, the tests will keep failing from time to time on the build branch :slightly_smiling_face: 

Anyway, maybe I misunderstood the process :smile: Let me know if it differs or some of what I said does not make sense

qooban:
@Gelio @kubama 

As we are starting to release more frequently it makes sense to fix system tests on build branches, so good that we have this one merged to `5.2-build`. 

History showed that it is sometimes quite difficult to run system tests on a branch (some parts of the Jenkins-based test runner is using `master` branches). Adding to it Nightwatch (test framework we used before Cypress) instability, Jenkins environment issues, Rackspace environment issues, etc it's very painful topic :-) But let's be optimistic, we are going towards real and stable environment as there are moves on DevOps side.

Summing it up, sooner or later we have to be prepared for running tests on release branches, so good that we have it here 👍🏻

PR #1198, opened 2021-02-12
RD-220 Use common cypress resources from cloudify-ui-common

kubama:
System tests build: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/159/


package.json
```diff
@@ -32,8 +32,8 @@
     "devServer": "node devServer.js",
     "docCheck": "scripts/checkDocs.sh",
     "docWidgets": "node scripts/updateReadmes.js",
-    "e2e": "cypress run --reporter cypress-multi-reporters --reporter-options configFile=test/cypress/reporter-config.json",
-    "e2e:open": "cypress open",
+    "e2e": "cypress run -C node_modules/cloudify-ui-common/cypress/cypress.json --reporter cypress-multi-reporters --reporter-options configFile=test/cypress/reporter-config.json",
+    "e2e:open": "cypress open -C node_modules/cloudify-ui-common/cypress/cypress.json",
```
qooban, line 36: So in case we need to differentiate some parts of the configuration between Composer and Stage in the future what is your suggested way to do that?

By adding overrides into `setupPluginsAndConfig` from `test/cypress/plugins/index.ts`? If so, I think that is good enough.

qooban (dismissed):
Looks good 👍🏻

package.json
kubama: Yes, this is the idea

kubama:
As you can see even though cypress tests execute correctly the build is failing on type checking cypress sources with tsc, as it tries to compile cloudify-ui-common ts sources from node_modules (why?).

I'm pretty sure that cypress runner uses compiled js files when importing cypress commons from cloudify-ui-common - so why does tsc compilation work differently and picks up ts source files instead of their js compiled counterparts?

Maybe I should have excluded source ts files from cloudify-ui-common? I'm not sure what's the best practice here.

@Gelio Do you have any thoughts on this?

Gelio:
@kubama Good observation, you are correct that TS files are resolved instead of just their JS counterparts.

This is because of [the module resolution strategy that TypeScript uses](https://www.typescriptlang.org/docs/handbook/module-resolution.html#how-typescript-resolves-modules). When importing a file, it will first try the `ts`, `tsx` extensions before trying `d.ts`, which contains the types in our case.

I confirmed that by running `cd test/cypress && npx typescript@beta --explainFiles` and seeing:

```
../../node_modules/cloudify-ui-common/cypress/plugins/index.ts
  Imported via 'cloudify-ui-common/cypress/plugins' from file 'plugins/index.ts' with packageId 'cloudify-ui-common/cypress/plugins/index.ts@3.1.0'
```

After removing `index.ts` from `node_modules/cloudify-ui-common/cypress/plugins`, type checking passed.

Thus, looks like having output JS files at the same level as source TS files does not play well (something we discussed in https://github.com/cloudify-cosmo/cloudify-ui-common/pull/47#discussion_r574593374).

What I would suggest in this case is add `outDir` to the `tsconfig.json` in `cloudify-ui-common`. That will make it so that the output files will land in some other directory (e.g. `node_modules/cloudify-ui-common/cypress/out/plugins`), and there will be no TS files there, meaning that the type-checking in this repo will pass. Then we could also simplify the `npmignore` and `gitignore` patterns because we will only need to ignore that one `out` directory. The downside is a slightly obscure path (that has `out` in the path).

I discourage removing TS files from the `cloudify-ui-common` bundle, as then we will not be able to jump to the source code of `cloudify-ui-common` since the declaration map will point to a non-existing file :slightly_smiling_face:

Gelio (commented):
Aside from that type-checking problem, the changes look good :+1:

qooban (approved):

test/cypress/support/index.ts
```diff
@@ -19,6 +19,8 @@ import '@cypress/code-coverage/support';
 import './commands';
 // use `Cypress` instead of `cy` so this persists across all tests
 Cypress.on('window:before:load', window => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore: The operand of a 'delete' operator must be optional
     delete window.fetch;
```
Gelio, line 24: I'm curious about this line... Why do we need to delete browser's `fetch`?

Gelio (approved):
Looks good :+1: Just one comment out of curiosity :slightly_smiling_face:

test/cypress/support/index.ts
kubama: If I remember correctly it was added so `fetch` polyfill based on XHR was used. And this was due to `cy.route` not working with `fetch`.
But this was introduced when much older Cypress version was in use. I guess it's very likely we no longer need it after moving to `cy.intercept` - would need some verification though.

qooban (dismissed):

qooban (approved):

test/cypress/support/index.ts
Gelio: I see, thanks for explaining :slightly_smiling_face:

PR #1199, opened 2021-02-15
RD-753 - Add DeploymentActionsMenu component

qooban:
This PR adds new common component - `DepoymentActionsMenu` - which groups all the actions possible to run on the specific deployment. It was based on the list of actions available in `DeploymentsList` and `DeploymentActionButtons` components.
All the actions require additional input from the user, so all of them just after selection from the menu are continued in the modal windows.

The most recent system tests execution on this branch: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/178/

## Video

https://user-images.githubusercontent.com/5202105/107920572-7b8eb300-6f6d-11eb-9df0-fad6fa892cfd.mp4

## Changes

* `DeploymentActionsMenu` - new components available in `Stage.Common`. Similarly to `LabelsInput` and `ManageLabelsModal` it is not attached to any part of the production code, yet. It will be used in the production code after implementation of [RD-755](https://cloudifysource.atlassian.net/browse/RD-755) and [RD-756](https://cloudifysource.atlassian.net/browse/RD-756).
* Deployment modals changes
  * General idea behind most of the changes in deployment modals was to 1. Make them consistent, 2. Make them fetch additional deployment data on open instead of requiring it to be passed as a prop.
  * `RemoveDeploymentModal` - new component was created to handle deployment removal (normal and forced).
  * `SetSiteModal` - component was moved from Sites Management widget to `Stage.Common`. It was refactored to be functional component, fetch site_name on open, use translations, get `deploymentId` instead of `deployment` as a prop
    * (bug-fix) errors handling fixed
  * `UpdateDeploymentModal` - component was refactored to fetch deployment inputs on open and get `deploymentId` instead of `deployment` as a prop.
    * (bug-fix) `dataTypes` handling fixed
  * `ExecuteDeploymentModal` - component was refactored to allow fetching deployment workflow details on open and get `deploymentId` instead of `deployment` as a prop


widgets/common/src/DeploymentActions.js
```diff
@@ -82,14 +82,18 @@ export default class DeploymentActions {
     }
 
     doSetSite(deploymentId, siteName, detachSite) {
-        const data = { detach_site: detachSite };
-        if (siteName) {
-            data.site_name = siteName;
-        }
+        const data = detachSite ? { detach_site: detachSite } : { site_name: siteName };
```
qooban, line 85: This is a bug-fix. `/sites` API does not allow to send `detach_site` set to `true` and `site_name` with string value at the same time.

widgets/common/src/UpdateDeploymentModal.jsx
```diff
@@ -36,12 +33,16 @@ function UpdateDeploymentModal({ deployment, open, onHide, toolbox }) {
...
+                    _.forEach(plannedDeploymentInputs, (inputObj, inputName) => {
+                        const { type } = inputObj;
+                        const dataType = !_.isEmpty(dataTypes) && type ? dataTypes[type] : undefined;
+
+                        newDeploymentInputs[inputName] = InputsUtils.getInputFieldInitialValue(
                             currentDeploymentInputs[inputName],
-                            inputObj.type,
-                            fetchedBlueprint.plan.data_types
+                            type,
+                            dataType
```
qooban, line 45: This is bug-fix. All inputs were treated as having type defined as data-type (custom type). 
Now `dataType` value is only set and passed to `getInputFieldInitialValue` when there is type defined in `dataTypes` object for that specific input type.

widgets/common/src/DeploymentActionsMenu.jsx
```diff
@@ -0,0 +1,108 @@
...
+            ExecuteDeploymentModal,
+            ManageLabelsModal,
+            UpdateDeploymentModal,
+            RemoveDeploymentModal,
+            SetSiteModal
+        },
+        Hooks: { useResettableState }
+    } = Stage;
+
+    const [openModal, setOpenModal, resetOpenModal] = useResettableState('');
```
kubama, line 35: I think this state variable could be named in a more accurate way - if I understand correctly this is an action/workflow name.

widgets/common/src/DeploymentActionsMenu.jsx
```diff
@@ -0,0 +1,108 @@
...
+                open={openModal === actions.delete || openModal === actions.forceDelete}
+                toolbox={toolbox}
+            />
+
+            <SetSiteModal
+                deploymentId={deploymentId}
+                onHide={resetOpenModal}
+                open={openModal === actions.setSite}
+                toolbox={toolbox}
+            />
```
kubama, line 90: Could we reduce some duplication here? `deploymentId`, `onHide` and `toolbox` attributes are same for all the modals.

widgets/common/src/ExecuteDeploymentModal.jsx
```diff
@@ -84,25 +127,28 @@ export default function ExecuteDeploymentModal({
...
-
         let deploymentsList = deployments;
         if (_.isEmpty(deployments)) {
-            deploymentsList = [deployment.id];
+            deploymentsList = _.compact([deploymentId]);
         }
 
-        const executePromises = _.map(deploymentsList, deploymentId => {
+        if (_.isEmpty(deploymentsList)) {
+            setErrors({ error: i18n.t('widgets.common.deployments.executeModal.missingDeployment') });
```
kubama, line 136: Looks like it is a common pattern to use `error` or `errors` as a key holding generic error message (not related to any particular form field). Maybe it would be worth to extend `useErrors` hook to handle it internally?

widgets/common/src/ExecuteDeploymentModal.jsx
```diff
@@ -265,18 +307,33 @@ export default function ExecuteDeploymentModal({
     );
 }
 
+function isDeploymentIdValid(props, propName, componentName) {
```
kubama, line 310: I would name this function in a way more specific to a prop type - `deploymentIdPropType` or similar.

widgets/common/src/RemoveDeploymentModal.jsx
```diff
@@ -0,0 +1,61 @@
...
+        .split('\n')
+        // eslint-disable-next-line react/no-array-index-key
+        .map((line, index) => <p key={index}>{line}</p>);
+
+    function deleteDeployment() {
+        const actions = new DeploymentActions(toolbox);
+        const deleteAction = force ? 'doForceDelete' : 'doDelete';
+
+        clearErrors();
+        actions[deleteAction]({ id: deploymentId })
```
kubama, line 24: Few characters more, but it will be easier to trace function calls:
```suggestion
        const deleteAction = force ? actions.doForceDelete : actions.doDelete;

        clearErrors();
        deleteAction({ id: deploymentId })
```

widgets/common/src/RemoveDeploymentModal.jsx
```diff
@@ -0,0 +1,61 @@
...
+    force: PropTypes.bool.isRequired,
+    open: PropTypes.bool.isRequired,
+    onHide: PropTypes.func.isRequired,
+    toolbox: Stage.PropTypes.Toolbox.isRequired
+};
+
+Stage.defineCommon({
+    name: 'RemoveDeploymentModal',
+    common: RemoveDeploymentModal
+});
```
kubama, line 61: This is of course out of scope of this PR, but what if at some point we changed `defineCommon` so it could be called like this: `defineCommon({RemoveDeploymentModal})`?

widgets/common/src/SetSiteModal.jsx
```diff
@@ -0,0 +1,112 @@
...
+    const {
+        Basic: { Modal, Icon, Form, ApproveButton, CancelButton },
+        Common: { DeploymentActions },
+        Hooks: { useBoolean, useErrors, useInput, useOpenProp, useResettableState },
+        i18n
+    } = Stage;
+
+    const [detachSite, setDetachSite, clearDetachSite] = useInput(false);
+    const { errors, clearErrors, setMessageAsError } = useErrors();
+    const [loading, setLoading, unsetLoading] = useBoolean(false);
```
kubama, line 11:
```suggestion
    const [loading, setLoading, unsetLoading] = useBoolean();
```

Gelio:
Just to clarify, this will be the component that will be used as the _Deployment_ dropdown in Environments? https://pe8xby.axshare.com/#id=tmxf30&p=env__views&g=1

qooban:
@Gelio Yes, this is one place where it will be used. It has `trigger` prop, so we can provide any component to trigger the menu open.

widgets/common/src/RemoveDeploymentModal.jsx
qooban: Good idea. We can have something like that in the future as in most (all?) of the cases we use the same name.

widgets/common/src/RemoveDeploymentModal.jsx
qooban: Are you sure it will work? 

If I remember correctly I tried the same solution and it was not possible to use it this way.

widgets/common/src/DeploymentActionsMenu.jsx
qooban: What solution to avoid duplication would you suggest here?

widgets/common/src/DeploymentActionsMenu.jsx
qooban: Yeah, could be better. What do you think about `activeAction`?

widgets/common/src/RemoveDeploymentModal.jsx
kubama: > Are you sure it will work?

If you are asking then I'm almost sure it won't 🙂.

OK, lets leave it.

widgets/common/src/ExecuteDeploymentModal.jsx
qooban: Sounds reasonable. 

On the other hand - I've got almost no experience with custom prop types, so when I searched the net I couldn't find there any strict naming convention for it. Either people use unnamed function directly assigned to the prop or make the name function like (using verbs).

widgets/common/src/ExecuteDeploymentModal.jsx
qooban: Sounds like a valuable change, I'll check it.

widgets/common/src/DeploymentActionsMenu.jsx
kubama: Maybe it's a case where props spreading would be justified?

widgets/common/src/DeploymentActionsMenu.jsx
kubama: Yes, much better 👍.

widgets/common/src/ExecuteDeploymentModal.jsx
kubama: My experience when reading the code was that I had no hint from the function name that it is used as a prop type. Reading further I was actually surprised by the context in which it is used. Maybe a short function JSDoc would suffice?

widgets/common/src/DeploymentActionsMenu.jsx
qooban: Make sense. Changed.

widgets/common/src/DeploymentActionsMenu.jsx
qooban: Changed.

widgets/common/src/ExecuteDeploymentModal.jsx
qooban: I ended up with unnamed function directly assigned to the prop. Hope that is fine for you.
Also I added missing default value.

widgets/common/src/ExecuteDeploymentModal.jsx
qooban: I looked at it deeper and it turned out that actually we don't need to put general errors (not related to any particular form field) as a value of `error` key. `Form` component's `errors` prop accepts strings as well, so I modified it to just pass string to `setErrors` function and not create an object.

kubama: Cool, that's even better!

kubama (approved):

widgets/common/src/SetSiteModal.jsx
```diff
@@ -0,0 +1,112 @@
+function SetSiteModal({ deploymentId, onHide, open, toolbox }) {
```
Gelio, line 1: With such a large refactor you might have as well opted for TypeScript :smile: I don't expect any changes, thought :slightly_smiling_face:

widgets/common/src/SetSiteModal.jsx
```diff
@@ -0,0 +1,112 @@
...
+    useOpenProp(open, () => {
+        const actions = new DeploymentActions(toolbox);
+
+        setLoading();
+        clearDetachSite();
+        clearSiteName();
+        clearErrors();
+        resetSites();
+
+        actions
```
Gelio, line 28: Could we do `doGetSites` and `doGetSite` in parallel using `Promise.all`?

widgets/common/src/DeploymentActions.js
```diff
@@ -82,14 +82,18 @@ export default class DeploymentActions {
...
 
         return this.toolbox.getManager().doPost(`/deployments/${deploymentId}/set-site`, null, data);
     }
 
+    doGetSite(deploymentId) {
```
Gelio, line 90: A more suitable name IMO would be

```suggestion
    doGetSiteName(deploymentId) {
```

widgets/common/src/DeploymentActions.js
Gelio: Would it be worth it to document that here using a comment?

widgets/common/src/SetSiteModal.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+    function setSite() {
+        setLoading();
+
+        const actions = new DeploymentActions(toolbox);
+        actions
+            .doSetSite(deploymentId, siteName, detachSite)
+            .then(() => {
+                clearErrors();
+                toolbox.getEventBus().trigger('deployments:refresh');
```
Gelio, line 45: Who may be listening for such events and what will this line do? :slightly_smiling_face: (a general question)

widgets/common/src/SetSiteModal.jsx
```diff
@@ -0,0 +1,112 @@
...
+                onHide();
+            })
+            .catch(setMessageAsError)
+            .finally(unsetLoading);
+    }
+
+    return (
+        <div>
```
Gelio, line 53: Do we need a `Modal` to be wrapped in a div? :thinking:

widgets/common/src/ExecuteDeploymentModal.jsx
```diff
@@ -1,18 +1,25 @@
-/**
- * Created by kinneretzin on 19/10/2016.
- */
+function isWorkflowName(workflow) {
+    return _.isString(workflow);
```
Gelio, line 2: How about a native

```suggestion
    return typeof workflow === 'string';
```

widgets/common/src/ExecuteDeploymentModal.jsx
```diff
@@ -51,13 +94,18 @@ export default function ExecuteDeploymentModal({
         const { InputsUtils, DeploymentActions } = Stage.Common;
         const validationErrors = {};
 
-        if (!deployment || !workflow) {
-            setErrors({ error: 'Missing workflow or deployment' });
+        const name = getWorkflowName(workflow);
+        if (!getWorkflowName(workflow)) {
```
Gelio, line 98: Looks like a half-baked refactor :smile: 

```suggestion
        if (!name) {
```

widgets/common/src/ExecuteDeploymentModal.jsx
```diff
@@ -268,15 +315,28 @@ export default function ExecuteDeploymentModal({
 ExecuteDeploymentModal.propTypes = {
     toolbox: Stage.PropTypes.Toolbox.isRequired,
     open: PropTypes.bool.isRequired,
-    deployment: PropTypes.shape({ id: PropTypes.string }),
+    deploymentId: (props, propName, componentName) => {
+        const { [propName]: deploymentId, workflow } = props;
+
+        if (_.isString(workflow) && !(_.isString(deploymentId) && deploymentId)) {
```
Gelio, line 321: How about using native `typeof` instead of `isString`?

widgets/common/src/DeploymentActionsMenu.jsx
```diff
@@ -0,0 +1,91 @@
...
+    install: 'install',
+    manageLabels: 'manageLabels',
+    setSite: 'setSite',
+    uninstall: 'uninstall',
+    update: 'update'
+});
+
+const translate = (key, params) => Stage.i18n.t(`widgets.common.deployments.actionsMenu.${key}`, params);
+const menuItems = [
+    { key: actions.install, name: actions.install, icon: 'play', content: translate(actions.install) },
```
Gelio, line 14: Do we need both `key` and `name`? They seem to be the same at all times

widgets/common/src/DeploymentActionsMenu.jsx
```diff
@@ -0,0 +1,91 @@
...
+    { key: actions.update, name: actions.update, icon: 'edit', content: translate(actions.update) },
```
Gelio, line 15: Could we fill in `content` automatically, in order to avoid typos?

```ts
const menuItems = [
    { name: actions.install, icon: 'play' },
    // ... and other items
].map(item => ({ ...item, content: translate(item.name) }));
```

widgets/common/src/UpdateDeploymentModal.jsx
```diff
@@ -36,12 +33,16 @@ function UpdateDeploymentModal({ deployment, open, onHide, toolbox }) {
                 .then(fetchedBlueprint => {
                     const newDeploymentInputs = {};
                     const currentDeploymentInputs = deployment.inputs;
+                    const { data_types: dataTypes, inputs: plannedDeploymentInputs } = fetchedBlueprint.plan;
 
-                    _.forEach(fetchedBlueprint.plan.inputs, (inputObj, inputName) => {
-                        newDeploymentInputs[inputName] = Stage.Common.InputsUtils.getInputFieldInitialValue(
+                    _.forEach(plannedDeploymentInputs, (inputObj, inputName) => {
+                        const { type } = inputObj;
+                        const dataType = !_.isEmpty(dataTypes) && type ? dataTypes[type] : undefined;
```
Gelio, line 40: Isn't this equivalent to

```suggestion
                        const dataType = dataTypes?.[type];
```

Gelio (commented):
Looks cool overall :slightly_smiling_face: I don't have full confidence since I did not read the whole surrounding code (unmodified parts of the source). My review is more code-style/code-quality based :slightly_smiling_face:

widgets/common/src/SetSiteModal.jsx
qooban: Good catch. Didn't spot that when changing to functional component. Removed.

widgets/common/src/DeploymentActions.js
qooban: Changed.

widgets/common/src/SetSiteModal.jsx
qooban: Yep, I actually thought about that as well. Changed 👍🏻

widgets/common/src/SetSiteModal.jsx
qooban: Yeah, I'm slowly moving towards being ready for that 🙂 Next time.

widgets/common/src/DeploymentActionsMenu.jsx
qooban: Nice idea, implemented.

widgets/common/src/DeploymentActionsMenu.jsx
qooban: Yes, AFAIR React forces us to provide `key` prop when adding list of elements.

Ref.: https://reactjs.org/docs/lists-and-keys.html#keys

widgets/common/src/SetSiteModal.jsx
qooban: For example Deployments widget is listening for such events. It will make Deployments widget re-fetch the data.

widgets/common/src/ExecuteDeploymentModal.jsx
qooban: Hmm... In this case I think it will make the line longer and less clear, so I'd like to keep it as it is.

widgets/common/src/DeploymentActionsMenu.jsx
Gelio: Oh, yeah, if the array is used as a JSX expression, definitely :smile: 

In this case, could we add it via post-processing (inside the `map` that happens just below) :slightly_smiling_face: ?

widgets/common/src/ExecuteDeploymentModal.jsx
Gelio: Longer - yes, less clear - I'm not sure :smile: I prefer `typeof x !== 'string'` to `!_.isString(x)`, but it's fine the way it is :slightly_smiling_face:

widgets/common/src/SetSiteModal.jsx
Gelio: I see, alright :slightly_smiling_face:

widgets/common/src/SetSiteModal.jsx
Gelio: Sure sure :smile:

widgets/common/src/DeploymentActionsMenu.jsx
qooban: Sure, done.

kubama (approved):

widgets/common/src/SetSiteModal.jsx
```diff
@@ -25,11 +25,11 @@ function SetSiteModal({ deploymentId, onHide, open, toolbox }) {
...
-        actions
-            .doGetSites()
-            .then(setSites)
-            .then(() => actions.doGetSite(deploymentId))
-            .then(setSiteName)
+        Promise.all([actions.doGetSites(), actions.doGetSiteName(deploymentId)])
+            .then(([fetchedSites, fetchedSiteName]) => {
+                setSites(fetchedSites);
+                setSiteName(fetchedSiteName);
+            })
```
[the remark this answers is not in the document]
qooban: Thanks for sharing that 👍🏻

PR #1202, opened 2021-02-15
RD-1093 Fix endless loop in UI after migration tests

Gelio:
This PR fixes a couple of underlying problems that are related to this bug.

It is divided into individual commits for easier reviewing.

The problems are:

1. When there are no pages for a given tenant, the UI was redirected to the home screen (https://github.com/cloudify-cosmo/cloudify-stage/blob/994497a81882bd8512a1c04c7b5379a459417730/app/actions/userApp.js#L34). This is because `createPagesFromTemplate` did not throw any errors. There was a separate redirect when there were no pages (https://github.com/cloudify-cosmo/cloudify-stage/blob/209e56ed7bde25ef909bfd69ea45e927d461da3e/app/components/Home.jsx#L45-L52), but the redirect in `userApp` happened after the one in `Home`.
2. The migration that moves template resources into files was incorrectly reusing template's `pages`
3. `userData` directory from tests was reused during development. This is why the problem occurred in the first place, because the templates that stay after tests have no pages.
4. There were 2 templates after migration tests that conflicted (same `roles` and `tenants`)

Here are the reproduction steps invoked after the fix:

https://user-images.githubusercontent.com/889383/107953396-62046000-6f9b-11eb-8d98-f56e92be8cd4.mp4


kubama (commented):
If for any reason a user encountered this scenario (template without pages) I would say he should be still logged into application. This would give him two options - he could switch to another tenant, that could be associated with another, valid template, or he could use Reset Templates option that could possibly fix the issue. Looking at the changes (Layout.jsx) I understand this is not the case (`doLogout`), or is it?

Gelio:
@kubama Sounds like a good solution. This requires more changes to the code in multiple places, that I would rather not do at the moment, given the pressure with the Environments feature :slightly_smiling_face: The reported bug was fixed - the endless loop no longer exists, the tests are fixed, and there is a descriptive message saying what is wrong - I think it's good enough for this ticket.

FYI, why it's not a simple fix:
1. There are 2 redirects in `Home.jsx` https://github.com/cloudify-cosmo/cloudify-stage/blob/209e56ed7bde25ef909bfd69ea45e927d461da3e/app/components/Home.jsx#L45-L56 that would have to be changed
2. The error/404 page does not render the top bar, meaning that the user cannot do the actions you said on that page. The edit mode does not work too 

    image

3. Thus, we would have to either display the top bar on the error page, or display the error message in the regular layout when there are 0 pages

Let me know if you have a suggestion on how to easily change it :slightly_smiling_face: If you don't have a better idea, I will create another ticket for improving this scenario :slightly_smiling_face:

kubama:
The 'no pages' scenario is very unlikely, so sure we can leave it as it is if the proposed change is complicated.
However, this has further implications. The changes made to `migrations.test.js` are just a workaround (a bit dirty I'd say) that wouldn't be really required - I guess it would be just enough to be able to reset the template after logging in if the test corrupted the template.
These are just my thoughts I wanted to share - no expectations 🙂.

kubama (approved):

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
                     let pages = null;
 
                     try {
-                        pages = fs.readJsonSync(templateFilePath);
+                        pages = fs.readJsonSync(templateFilePath).pages;
```
qooban, line 33: Oh, wait, before 5.0 we had different file structure. Are you sure it should be like that, here?

JSON template file looked like:
```
[
  "app",
  "catalog",
  "blueprints",
  "deploy",
  "systemResources",
  "logs"
]
```

qooban (commented):
Changes generally looks good 👍🏻 
I'm not sure about the old migration change.

Gelio:
@kubama 

> The changes made to migrations.test.js are just a workaround (a bit dirty I'd say) that wouldn't be really required - I guess it would be just enough to be able to reset the template after logging in if the test corrupted the template.

The changes to `migrations.test.js` are not a workaround, but a developer-experience enhancement :slightly_smiling_face: Without them, when you logged into the UI and reset the template, you would encounter the error that there are 0 pages. This is because the test did not clean up after itself, which IMO it should :slightly_smiling_face: Nothing dirty here if you asked me :slightly_smiling_face: 

As an aside, I have created https://cloudifysource.atlassian.net/browse/RD-1423 to apply your suggestion :slightly_smiling_face:

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
Gelio: I did not know what :smile: As we agreed, I added logic that handles both new and old templates :slightly_smiling_face: I've also adjusted error handling to be more verbose and catch invalid cases instead of overwriting user data

Gelio:
With the recent changes, after putting `{}` as `userData/templates/test.json`, without the cleanup part in `migration.test.js`, here are the logs:

```
      [2021-02-16T14:41:34.928Z][DBMigration] INFO: 20190401130022-5_0-MoveTemplatesResourcesToFiles migrating
      [2021-02-16T14:41:34.932Z][DBMigration] INFO: Found 1 template rows to migrate.
      [2021-02-16T14:41:34.933Z][DBMigration] INFO: Updating /home/grzegorz/projects/cloudify/cloudify-stage/userData/templates/test.json
      [2021-02-16T14:41:34.933Z][DBMigration] INFO: File exists. Updating it...
      [2021-02-16T14:41:34.933Z][DBMigration] ERROR: Unexpected template file contents {}
      [2021-02-16T14:41:34.935Z][DBMigration] ERROR: UP ERROR
      [2021-02-16T14:41:34.935Z][DBMigration] ERROR: ========
      [2021-02-16T14:41:34.936Z][DBMigration] ERROR: Error: Unexpected template file at /home/grzegorz/projects/cloudify/cloudify-stage/userData/templates/test.json. Expected a whole template or just the pages array
          at /home/grzegorz/projects/cloudify/cloudify-stage/backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js:42:35

```

And the tests fail, as expected :slightly_smiling_face:

kubama:
> The changes to `migrations.test.js` are not a workaround, but a developer-experience enhancement 🙂 Without them, when you logged into the UI and reset the template, you would encounter the error that there are 0 pages. This is because the test did not clean up after itself, which IMO it should 🙂 Nothing dirty here if you asked me 🙂

If '0 pages' was handled by the UI in a way that would allow a user to actually use the application (I don't see a single reason why it shouldn't, of course except the amount of work needed for the implementation) then there would be no issue and thus no need for the setup/clean up logic in the test, right? So for me it's a workaround which could be removed as part of RD-1423. But I'm not going to argue on this 🙂.

kubama (approved):

Gelio:
@kubama Ok, let's agree to disagree here, as it's a minor matter :smile: Thanks for the review :slightly_smiling_face:

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
```diff
@@ -30,12 +30,27 @@ module.exports = {
                     let pages = null;
 
                     try {
-                        pages = fs.readJsonSync(templateFilePath);
                         logger.info('File exists. Updating it...');
```
qooban, line 33: This `logger` call was originally after `fs.readJsonSync` call. 

Right now it will always state that "File exists. (...)" no matter whether `fs.readJsonSync` throws an error or not. I think we can swap the lines.

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
```diff
@@ -30,12 +30,27 @@ module.exports = {
...
-                        pages = [];
-                        logger.info('File does not exist. Creating new one...');
-                        fs.mkdirSync(path.dirname(templateFilePath), { recursive: true });
+                        if (error.code === 'ENOENT') {
+                            pages = [];
+                            logger.info('File does not exist. Creating new one...');
+                            fs.mkdirSync(path.dirname(templateFilePath), { recursive: true });
+                        } else {
+                            throw error;
+                        }
```
qooban, line 53: I'm wondering how often we will have such situation in the production. Not sure it is the best solution here, as it will kill the whole migration process.

But on the other side I think this situation (invalid file structure) is very unlikely, so we should be good with it 🙂

qooban (commented):
Good change 👍🏻 Added 2 comments, 1st is a small change request, 2nd is just a note.

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
Gelio: IMO we should be surfacing such a problem :smile: If I was upgrading, I would rather be notified that the migration failed and did not overwrite my files :smile: I decided for such a change of logic since the previous behavior could potentially overwrite some user's file

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
Gelio: Right, good point :+1: I'll change it

Gelio: Done :slightly_smiling_face:

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
qooban: Yeah, I agree we should be surfacing that. 

Still I'm not sure we should stop the whole process. Snapshot restoration mechanism in Cloudify Manager is quite complex and it causes a lot of headaches, not only for UI team. It would probably be cleaner to handle that error.

Anyways, I'm good with the current solution 🙂

qooban (approved):

backend/migrations/20190401130022-5_0-MoveTemplatesResourcesToFiles.js
Gelio: By 

> It would probably be cleaner to handle that error.

you mean _swallow_ the error and overwrite user's files :smile: ? I'm not sure I follow, but we can leave it for now

qooban: No, not overwrite. We can e.g. do not process the the file, change it's name/extension and with some small update in getting templates/pages functions we could avoid using those files in the app. 

Let's leave it. It's edge case, very unlikely I think.

Gelio: Right, sure, that's an option :slightly_smiling_face: Alright

PR #1203, opened 2021-02-15
RD-1396 - Fix license_spec test

qooban:
After the change it looks like license test should be more reliable (no issues with redux state) and it is ~10 seconds faster (44s -> 35s).

Stage system tests for RD-1396 branch: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/167/


test/cypress/integration/license_spec.js
```diff
@@ -52,85 +52,80 @@ describe('License Management', () => {
...
-        verifyMessageHeader(validTrialLicense.header);
-    });
+        it('fetches license on page load', () => {
+            goToEditLicense();
+            uploadLicense(expiredTrialLicense.file).then(() => verifyMessageHeader(expiredTrialLicense.header));
 
-    it('shows active license', () => {
-        cy.visit('/console/license');
+            // Update license using REST call
+            cy.activate().reload();
```
kubama, line 100: I'm not sure what this comment means in the context of following line

test/cypress/support/commands.ts
```diff
@@ -26,6 +26,7 @@ import './widgets';
 import './secrets';
 import './snapshots';
 import { addCommands, GetCypressChainableFromCommands } from './command-utils';
+import emptyState from '../../../app/reducers/managerEmptyState';
```
kubama, line 29: Looks like this is the first time we are importing app code into cypress test. I'm not sure if such dependency is desired. Maybe we should move the imported file into cloudify-ui-common, so it's clear that it's reused from outside the app? Do you have any thoughts on this?

test/cypress/integration/license_spec.js
qooban: It was there before, what I actually changed in this line is just indentation.
I'll remove or rephrase it.

test/cypress/support/commands.ts
qooban: Yes, it's first time we are importing app code into cypress test. My first implementation used duplicated `emptyState` (I just copied it to cypress custom command code). Then I decided to extract common part to a separate file and import (directly importing `managerReducer` is not possible)  and validate it with you on the code review.

I think moving it to cloudify-ui-common is a bit too far (as that code is used only in cloudify-stage).

@kubama , @Gelio: For now I actually don't see any big problems with the current solution and one advantage: no code duplication. I would appreciate more comments, thoughts on that.

test/cypress/integration/license_spec.js
```diff
@@ -52,85 +52,80 @@ describe('License Management', () => {
         return cy
             .fixture(`license/${licenseFile}`)
             .then(license => {
-                cy.get('@licenseTextArea').type(license);
+                // Mimic copy-paste behavior to speed up the test
```
Gelio, line 55: Nice idea :+1: I reckon we could also use it in other places, e.g. pasting some URL instead of typing it :smile: 

Do you think it would be worth it to extract it to a custom command/function, either here, or in a separate ticket?

test/cypress/support/commands.ts
Gelio: To repeat what we talked about earlier: maybe we could get the current `manager` value from `localStorage` and just merge `auth` and `username` to the current value, instead of using the `emptyState` :smile: ? This could solve the problem of having to import it from `app`, and probably be more future-proof, as the tests are coupled loser :slightly_smiling_face:

Gelio (commented):
Well done :smile: 

Let's just see if importing `emptyState` can be avoided :slightly_smiling_face:

test/cypress/integration/license_spec.js
qooban: Agree, but the most benefit we get here, when we add long licenses strings. In case of URLs the gain is not as big.

Anyways, worth to consider in the future, so I created separate ticket for that: https://cloudifysource.atlassian.net/browse/RD-1436 .

test/cypress/support/commands.ts
qooban: OK, so the situation here is not as trivial as we could think of it. It seems to me it would be quite complex to update the local storage state after it is initialized. We would need to create kind of hook in the application code. It would have to be done before store is initialized. I didn't find quick way to do that.

When [RD-1106](https://cloudifysource.atlassian.net/browse/RD-1106) will be implemented, then we can improve `mockLogin` in much more convenient way. Basically after the fix we could always fetch user data (now we can't do that in case of no license or expired license) using `GET /user` endpoint. There will be then no need to pass user role from the backend (in `POST /auth/login` endpoint), because frontend side anyway is already calling `GET /user` endpoint a bit further in the starting procedure. If you are interested in the details, you can read comments in the issue. 

I reverted my change for now. And in case of problematic license tests, right after visiting `/console` (which makes it possible to initialize all the data) I'm switching to `/console/license`, that makes them safe from any problems related to user roles.

I suggest to keep it as it is and do not waste time as with [RD-1106](https://cloudifysource.atlassian.net/browse/RD-1106) it could be much easier in the future.

qooban:
System tests execution started after recent changes: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/174
Waiting for final results.

kubama (approved):

test/cypress/integration/license_spec.js
Gelio: Cool, thanks :slightly_smiling_face:

PR #1204, opened 2021-02-18
RD-223 - Make Stage system test results presented properly in Jenkins

qooban:
I removed `cypress-multi-reporters` from the project. I don't think we need it. Now we are aligned with Composer Cypress settings in the reporting area, so now we will use `junit` reporter only for Cypress tests. It should make our tests results visible the same way as in Composer.

Test results example available here:
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/185/testReport/(root)/


qooban:
@kubama There are some errors related to coverage reports in shared job: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/185/console . Could you elaborate what might the reason? Would it be hard to fix it? Maybe I could try to incorporate it in this PR?

kubama:
> UPDATE: Oh, I just recalled that you mentioned some time ago that coverage won't work when we are running only selected tests, right?

Yes, but actually I would expect coverage calculation to be done, and then a failure due to threshold levels not met. I don't think however what we have here is related to this PR. Since our build wasn't green for ages it might have been broken at same point and was not noticed. Maybe due to RD-220, or maybe due to TS introduction - I'll try to investigate.

kubama (approved):

kubama:
Looks like even when all tests are executed and passing, code coverage reporting is still failing: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/189/
So there is definitely an issue, but surely no related to this PR.

PR #1206, opened 2021-02-19
RD-1470 - Fix babel configuration to include istanbul plugin for coverage check builds

qooban:
This PR fixes coverage test checks which were broken by https://github.com/cloudify-cosmo/cloudify-stage/pull/1184/  

Current coverage status: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/193/Overall_20coverage/

Decreased level for functions coverage is probably due to Labels-related components (recently added, not yet covered with test). No tests for them is intentional as none of those new components are right now part of the production application. Once they will be included (in Deployments widget, in Labels Management widget), they will be covered by the tests.


package.json
```diff
@@ -21,7 +21,7 @@
     "build:analyse:main": "npm run build -- --env.analyse=main",
     "build:analyse:widgets": "npm run build -- --env.analyse=widgets",
     "build:analyse:widgets-common": "npm run build -- --env.analyse=widgets-common",
-    "build:coverage": "npm run build -- --env.coverage",
+    "build:coverage": "COVERAGE_CHECK=1 npm run build",
```
Gelio, line 24: I'm curious, should we not be using `cross-env` in case someone on Windows tries to run this command directly?

Gelio (approved):
Looks good, nice catch :+1:

package.json
qooban: I actually was thinking about that, already 🙂 

I originally didn't want to introduce new devDependency (`cross-env`) in front-end package (it is already in the stage back-end), but I just realized that actually Kuba M. is using Windows (or at least used AFAIR), so I'm going to add `cross-env`.

So... Good that you pointed that out and made me rethink it 🙂

PR #1207, opened 2021-02-22
RD-1495 - Update README.md

qooban:
This change is to avoid confusion in the future what is the compatible version of Node.js.


README.md
```diff
@@ -11,8 +11,8 @@ The Cloudify Console provides User Interface for managing and analyzing [Cloudif
 
 The following requirements should be met prior starting the application:
 
-- [Node.js](https://nodejs.org) (version >= 12.16.x) installed
```
kubama, line 14: If I understand correctly this is due to recently spotted issues with node version > 12.
That's OK, but was there no reason why at least 12.16 was required so far? Why are we relaxing it to include older versions now?

qooban: Actually my intention was to say v12 in general. We used 12.16.x, because that was the latest when we switched from v10 to v12.

Can be "(version = 12.16.x)", but I'm open to any suggestions how to describe it in clear and simple way :-)

kubama: > We used 12.16.x, because that was the latest when we switched from v10 to v12.

OK, I just thought that there was an actual requirement behind. But since there is not then all good 🙂.

kubama (approved):

PR #1209, opened 2021-02-23
RD-755 - Update Deployments widget

qooban:
### Changes:
* Splitted `DeploymentActionsMenu` into two components to reduce components (modals) duplication
* Added permissions support for DeploymentActionsMenu component
* Updated Deployments widget to:
  * reuse `DeploymentActionsMenu` and `DeploymentActionsModals` components
  * provide new menu for executing workflows

### Video:
https://user-images.githubusercontent.com/5202105/108836170-d0b37000-75d0-11eb-8581-a804bc07fca4.mp4

### TODOs:
- [x] Add system test


widgets/deployments/src/DeploymentsList.jsx
```diff
@@ -66,73 +58,24 @@ export default class DeploymentsList extends React.Component {
...
 
-    showModal = (value, deployment, workflow) => {
-        this.setState({
-            deployment,
-            workflow: workflow || {},
-            modalType: workflow ? MenuAction.WORKFLOW_ACTION : value,
-            showModal: true
-        });
+    openActionModal = (deploymentId, actionName) => {
+        this.setState({ deploymentId }, () => this.setState({ activeAction: actionName }));
```
kubama, line 74: Is there any reason why it couldn't be:
```suggestion
        this.setState({ deploymentId, activeAction: actionName });
```
?

widgets/deployments/src/DeploymentsTable.jsx
```diff
@@ -82,7 +78,16 @@ export default function DeploymentsTable({
...
+                            <WorkflowsMenu
+                                workflows={item.workflows}
+                                popupMenuProps={{ icon: 'cogs' }}
+                                onClick={workflow => onWorkflowAction(item.id, workflow.name)}
+                            />
+
+                            <DeploymentActionsMenu
+                                onActionClick={actionName => onMenuAction(item.id, actionName)}
+                                toolbox={toolbox}
+                            />
```
kubama, line 90: Looks like the exact same code fragment is used in DeploymentsSegment.jsx. Maybe worth extracting to a component?

widgets/deployments/src/DeploymentsList.jsx
qooban: Initially I thought it would be better to change it in two steps, but actually it works fine (no errors in browser console and in the UI), so I simplified it.

widgets/deployments/src/DeploymentsTable.jsx
qooban: Good point. Extracted it to `ActionsMenus`.

widgets/common/src/ManageLabelsModal.jsx
```diff
@@ -38,7 +38,16 @@ function ManageLabelsModal({ deploymentId, open, onHide, toolbox }) {
...
+        actions
+            .doSetLabels(deploymentId, labels)
+            .then(() => {
+                unsetLoading();
+                onHide();
+            })
+            .catch(error => {
+                unsetLoading();
+                setMessageAsError(error);
+            });
```
qooban, line 50: Not to call `unsetLoading` on removed component.

widgets/common/src/SetSiteModal.jsx
```diff
@@ -43,10 +43,13 @@ function SetSiteModal({ deploymentId, onHide, open, toolbox }) {
...
                 toolbox.getEventBus().trigger('deployments:refresh');
+                unsetLoading();
                 onHide();
             })
-            .catch(setMessageAsError)
-            .finally(unsetLoading);
+            .catch(error => {
+                setMessageAsError(error);
+                unsetLoading();
+            });
```
qooban, line 52: Not to call `unsetLoading` on removed component.

widgets/common/src/UpdateDeploymentModal.jsx
```diff
@@ -131,20 +131,23 @@ function UpdateDeploymentModal({ open, deploymentId, onHide, toolbox }) {
...
                     toolbox.getEventBus().trigger('executions:refresh');
+                    onHide();
                 }
             })
-            .catch(err => setErrors(InputsUtils.getErrorObject(err.message)))
-            .finally(unsetLoading);
+            .catch(err => {
+                setErrors(InputsUtils.getErrorObject(err.message));
+                unsetLoading();
+            });
```
qooban, line 150: Not to call `unsetLoading` on removed component.

widgets/deployments/src/DeploymentsList.jsx
```diff
@@ -1,24 +1,16 @@
...
+            activeAction: '',
+            deploymentId: '',
             error: null,
-            modalType: '',
-            showModal: false,
-            deployment: { id: '' },
-            deploymentUpdateId: null,
-            workflow: { name: '' }
+            showExecuteModal: false,
+            workflowName: ''
```
Gelio, line 13: How about using `null` / `undefined` for default values of state that is not yet set (`workflowName`, `activeAction`, `deploymentId`)?

Using a string is confusing to me, since we are using the "default value" of some type to denote a missing value, instead of using a more specialized `null` / `undefined`.

Gelio (commented):
I really like the changes :smile: Especially that this PR removes more lines than it adds :slightly_smiling_face: 

Some general comments about the UI:
1. Can we add tooltips for the `ActionsMenus`? At the moment, the icons are the only thing that guides the user into what they should expect after clicking the icon :slightly_smiling_face: 
2. It is very easy for me to misclick the icon and go into Deployments drill-down

    https://user-images.githubusercontent.com/889383/108976742-1ed77a80-7688-11eb-9073-9629a9fcce4e.mp4
    
    Could we add some padding around the icons to make them easier to click?

3. IMO it would be nice to somehow change the icon on hover, so the user knows that clicking will open a menu, and not go to drill-down

widgets/common/src/props/Workflows.js
```diff
@@ -0,0 +1,4 @@
+Stage.definePropType({
+    name: 'Workflows',
+    common: PropTypes.arrayOf(PropTypes.shape({ name: PropTypes.string, plugin: PropTypes.string }))
```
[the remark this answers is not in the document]
qooban: Added.

widgets/common/src/DeploymentActionsMenu.jsx
```diff
@@ -11,72 +10,43 @@ const actions = Object.freeze({
...
-    { name: actions.uninstall, icon: 'recycle' },
-    { name: actions.delete, icon: 'trash alternate' },
-    { name: actions.forceDelete, icon: 'trash' }
-].map(item => ({ ...item, key: item.name, content: translate(item.name) }));
-
-export default function DeploymentActionsMenu({ deploymentId, toolbox, trigger }) {
+    { name: actions.install, icon: 'play', permission: 'execution_start' },
+    { name: actions.update, icon: 'edit', permission: 'deployment_update_create' },
+    { name: actions.setSite, icon: 'building', permission: 'deployment_set_site' },
+    { name: actions.manageLabels, icon: 'tags', permission: 'deployment_create' },
```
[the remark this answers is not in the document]
qooban: That's the information from the backend team (Yoni) I received - use `deployment_create` as a manage labels permissions.
If you are interested in more details about the permissions please reach out someone from the core team.

widgets/common/src/DeploymentActionsModals.jsx
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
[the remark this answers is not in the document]
qooban: I don't feel ready, yet. Be patient 🙂

widgets/common/src/DeploymentActionsModals.jsx
```diff
@@ -0,0 +1,52 @@
...
+    } = Stage;
+
+    const commonProps = { deploymentId, open: true, onHide, toolbox };
+
+    let workflowName = '';
+    if (activeAction === actions.install) {
+        workflowName = 'install';
+    } else if (activeAction === actions.uninstall) {
+        workflowName = 'uninstall';
+    }
```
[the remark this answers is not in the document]
qooban: Implemented the second solution.

widgets/common/src/WorkflowsMenu.jsx
```diff
@@ -102,7 +102,7 @@ AccordionWorkflowsMenu.defaultProps = {
     onClick: _.noop
 };
 
-function WorkflowsMenu({ workflows, onClick, popupMenuProps, showInPopup, trigger }) {
+export default function WorkflowsMenu({ workflows, onClick, popupMenuProps, showInPopup, trigger }) {
```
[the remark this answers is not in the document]
qooban: Probably leftover of some debugging. Removed.

widgets/deployments/src/DeploymentsList.jsx
```diff
@@ -1,24 +1,16 @@
...
         this.state = {
+            activeAction: '',
+            deploymentId: '',
             error: null,
-            modalType: '',
-            showModal: false,
-            deployment: { id: '' },
-            deploymentUpdateId: null,
-            workflow: { name: '' }
+            showExecuteModal: false,
```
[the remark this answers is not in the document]
qooban: Implemented the first suggestion.

widgets/deployments/src/DeploymentsList.jsx
qooban: Used `null`, updated uses.

widgets/deployments/src/props/DeploymentsViewPropTypes.js
```diff
@@ -21,7 +21,8 @@ export default {
     fetchData: PropTypes.func,
     onSelectDeployment: PropTypes.func,
     onActOnExecution: PropTypes.func,
-    onMenuAction: PropTypes.func,
+    onDeploymentAction: PropTypes.func,
+    onWorkflowAction: PropTypes.func,
```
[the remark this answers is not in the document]
qooban: Yes, we can add `.isRequired` in most of the exisitng prop types in this file. Updated.

widgets/common/src/ManageLabelsModal.jsx
```diff
@@ -38,7 +38,16 @@ function ManageLabelsModal({ deploymentId, open, onHide, toolbox }) {
         clearErrors();
         setLoading();
 
-        actions.doSetLabels(deploymentId, labels).then(onHide).catch(setMessageAsError).finally(unsetLoading);
+        actions
+            .doSetLabels(deploymentId, labels)
+            .then(() => {
+                unsetLoading();
+                onHide();
+            })
```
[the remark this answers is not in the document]
qooban: `unsetLoading` on unmounted component throws an error in the browser console. That's why I changed in 3 modals.
About moving `finally` earlier - hmm... I am more confused reading such code, so I prefer to stay with what I did.

qooban:
@Gelio 
> Can we add tooltips for the ActionsMenus? At the moment, the icons are the only thing that guides the user into what they should expect after clicking the icon
Done by using PopupMenu based tooltip.

> It is very easy for me to misclick the icon and go into Deployments drill-down. Could we add some padding around the icons to make them easier to click?
Done by adding bordered icon (similar to what we have in different places).

> IMO it would be nice to somehow change the icon on hover, so the user knows that clicking will open a menu, and not go to drill-down
It is changed. Maybe it is not a big change, but this is how Semantic UI handles it, I don't want to change it for this specific case only.

NOTE: As we will be moving from current Deployments widget to new one, we don't need to be perfect in the old one.

widgets/common/src/ManageLabelsModal.jsx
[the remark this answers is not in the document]
Gelio: Sure, then :slightly_smiling_face:

widgets/deployments/src/props/DeploymentsViewPropTypes.js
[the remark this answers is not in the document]
Gelio: Great, looks clean now :slightly_smiling_face:

widgets/common/src/ManageLabelsModal.jsx
Gelio: Just to confirm - did you actually see the error, or is it your assumption?

I'm asking because I thought that Promise callbacks are microtasks that execute before the next event loop's task, and I thought that when React receives a state update, it queues a **task** to trigger rerenders, meaning that `then` would execute before the next rerender :slightly_smiling_face:

Gelio:
> It is changed. Maybe it is not a big change, but this is how Semantic UI handles it,

Yeah, right :smile: It's a slight change, thanks for mentioning that. I now see it

Thanks for addressing all the comments :slightly_smiling_face: I don't have more comments so far :heavy_check_mark:

Gelio:
Btw. 

> (optional) Convert class component DeploymentsList into functional component

As you said, we don't need to make the old component perfect, especially as it will be deprecated :smile: So, maybe let's limit the scope of changes?

widgets/common/src/ManageLabelsModal.jsx
qooban: Yes, I saw the error: `Warning: Can't perform a React state update on an unmounted component. (...)`.

Gelio: After digging in a bit into React, I have found the reason :smile: 

A quick summary for those interested:

React has 2 types of reacting to `setState` (or the setter in `useState`):
1. Asynchronous - in this "mode", updates are batched, and after the batching stops, updates are applied asynchronously, and then, the component rerenders. This is what happens in event listeners. For example, I changed `onHide` to execute synchronously in the event listener (the base level of `onApply`), and the order was as follows:

    image
    `setState` ran, scheduled an update, returned, `onHide` returned, and only then (in the next tick) did the component rerender.

2. Synchronous - in this mode the updates are applied synchronously and not batched. Every `setState` causes a rerender. You can offset it by wrapping multiple `setState`s in `ReactDOM.unstable_batchedUpdates`.

    This mode is the default outside of event handlers (and probably `useEffect` and other lifecycle callbacks). So, if we update state from a `.then` after a network request, the state is updated synchronously, the component rerenders, and this is why the 2nd update did not work :smile: 

    image
    Notice that `rerender` happens before `after setState`, meaning it was synchronous.

Fun fact: if you reverse the order of `onHide` and `unsetLoading`, the component will complain, but if you change the order **and** wrap in `batchedUpdates`, it works fine:

```ts
                // eslint-disable-next-line no-undef
                ReactDOM.unstable_batchedUpdates(() => {
                    onHide();
                    unsetLoading();
                });
```

qooban:
@Gelio @kubama 
I added basic system tests for Manage Labels functionality in Deployments widget as well as extended execute workflow tests to cover executing workflow from the new menu.

Removed **Convert class component DeploymentsList into functional component** from the TODO list as Deployments widget will probably not be as widely used as new Deployments View widget in the future.

PR is ready for final reviews.

widgets/common/src/DeploymentActionsMenu.jsx
```diff
@@ -32,13 +33,17 @@ export default function DeploymentActionsMenu({ onActionClick, toolbox, trigger
...
+        : {};
 
     function onItemClick(event, { name }) {
         onActionClick(name);
     }
 
     return (
-        <PopupMenu className="deploymentActionsMenu">
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        <PopupMenu className="deploymentActionsMenu" {...popupMenuProps}>
```
kubama, line 46: If we are justifying props spreading in such a case then what are the cases where it would be not justified?
We are disabling this rule constantly... how about disabling it globally?

widgets/common/src/SetSiteModal.jsx
kubama: How about adding a comment? I think someone in the future could be tempted to move it back to `finally`.

widgets/common/src/UpdateDeploymentModal.jsx
kubama: How about adding a comment? I think someone in the future could be tempted to move it back to `finally`.

widgets/common/src/DeploymentActionsMenu.jsx
qooban: 🙂 I think the rule in general makes sense - whenever it is possible we should be very clear in what we are passing. As in the rule description: _"This enhances readability of code by being more explicit (...) It is also good for maintainability by avoiding passing unintentional extra props"_.

In this particular case:
1. I found it a bit cleaner to do it like I did instead of:
```suggestion
        <PopupMenu
            className="deploymentActionsMenu"
            bordered={trigger ? undefined : true}
            help={trigger ? undefined : i18n.t('widgets.common.deployments.actionsMenu.tooltip')}
            offset={trigger ? undefined : [0, 5]}
        >
```
2. We exactly know what we are spreading, because `popupMenuProps` definition is 10 lines above.

What do you think? I'm open to any suggestions, but in general I'm for keeping the rule.

widgets/common/src/SetSiteModal.jsx
qooban: Make sense. Added.

widgets/common/src/DeploymentActionsMenu.jsx
kubama: I agree props spreading here is justified, I'm not against its usage here.
I just noticed we are disabling this rule frequently. At the same time I can see only one edge case where it would be unjustified: a scenario where props object is used only in one component and there is at most one conditional prop.
I think having this rule enabled globally actually discourages using this nice feature in cases where it would be justified and beneficial (actually this is my real experience). So I think it actually makes more harm then benefits. After all we have very picky code reviews, so if props spreading was used where it shouldn't it wouldn't be unnoticed.

widgets/common/src/DeploymentActionsMenu.jsx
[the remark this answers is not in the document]
Gelio: I discussed that with Yoni :slightly_smiling_face: He said he will consider creating a new permission set for updating deployments, as he agreed that having a `deployment_create` permission for an action that updates the deployment may be counter-intuitive :slightly_smiling_face: 

Problem clarified, I'm resolving the discussion

test/cypress/integration/widgets/deployments_spec.js
```diff
@@ -12,11 +12,17 @@ describe('Deployments widget', () => {
...
 
-    const actOnDeployment = (name, action) => {
-        searchForDeployment(name);
-        cy.contains('div.row', name).find('.deploymentActionsMenu').click();
+    const selectActionFromMenu = (deploymentId, menuClassName, action) => {
+        searchForDeployment(deploymentId);
+        cy.contains('div.row', deploymentId).find(menuClassName).click();
         cy.get('.popupMenu > .menu').contains(action).click();
     };
+    const actOnDeployment = (name, action) => {
```
Gelio, line 20: Maybe it's a good time to rename it to something more concrete, e.g.

```suggestion
    const executeDeploymentAction = (name, action) => {
```

What do you think?

test/cypress/integration/widgets/deployments_spec.js
```diff
@@ -168,6 +195,49 @@ describe('Deployments widget', () => {
         });
     });
 
+    it('should allow to manage deployment labels', () => {
```
Gelio, line 198: Very clean test :+1: I like it!

Gelio (commented):
Looks solid! :+1: Some more minor comments, and aside from that, looks good to me

test/cypress/support/deployments.ts
```diff
@@ -9,6 +9,11 @@ declare global {
     }
 }
 
+type Label = {
```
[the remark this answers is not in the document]
qooban: 😄

test/cypress/integration/widgets/deployments_spec.js
qooban: Thanks 🙂

test/cypress/integration/widgets/deployments_spec.js
qooban: Make sense. Changed.

test/cypress/integration/widgets/deployments_spec.js
```diff
@@ -108,15 +114,36 @@ describe('Deployments widget', () => {
...
+        };
+
+        it('install workflow from deployment actions menu', () => {
+            actOnDeployment(deploymentName, 'Install');
+            startAndVerifyWorkflowExecution();
+        });
 
-        cy.wait('@executeWorkflow');
-        cy.contains('div.row', deploymentName).find('.spinner.loading.icon').should('be.visible');
+        it('any workflow from workflows menu', () => {
```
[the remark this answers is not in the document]
qooban: Make sense. Changed to "a workflow ...". 

The main difference between those two tests are from where the workflow is taken - either from `WorkflowsMenu` or `DeploymentActionsMenu`.

test/cypress/integration/widgets/deployments_spec.js
```diff
@@ -12,11 +12,17 @@ describe('Deployments widget', () => {
...
+        searchForDeployment(deploymentId);
+        cy.contains('div.row', deploymentId).find(menuClassName).click();
         cy.get('.popupMenu > .menu').contains(action).click();
     };
+    const actOnDeployment = (name, action) => {
+        selectActionFromMenu(name, '.deploymentActionsMenu', action);
+    };
+    const executeWorkflow = (name, workflow) => {
+        selectActionFromMenu(name, '.workflowsMenu', workflow);
+    };
```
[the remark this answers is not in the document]
qooban: Yep, changed.

test/cypress/support/deployments.ts
```diff
@@ -9,6 +9,11 @@ declare global {
...
+    [key: string]: string | string[];
+};
```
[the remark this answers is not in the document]
qooban: Thanks for sharing, I'll keep that in mind 🙂

test/cypress/support/deployments.ts
```diff
@@ -9,6 +9,11 @@ declare global {
...
+type Labels = Array<Label>;
```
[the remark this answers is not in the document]
qooban: Thx, changed to inline type as you suggested.

widgets/common/src/DeploymentActionsMenu.jsx
qooban: I think I'm tending to agree with your approach. As long as we have agreement on what are good and what are the bad uses of the props spreading, I think we could disable it globally. Let's discuss on our slack channel, OK? I'll start a thread and resolve this comment here (I would like not to change the rule setting within this PR). Hope that's fine with you.

qooban:
We should probably merge https://github.com/cloudify-cosmo/cloudify-stage/pull/1210 first (due to cloudify-ui-components update). We'll see on Monday.

kubama (approved):

test/cypress/integration/widgets/deployments_spec.js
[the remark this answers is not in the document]
Gelio: Yeah, yeah :slightly_smiling_face: IMO it looks a bit clearer now, thanks

Gelio (approved):
Looks cool :smile: I have not run it in the UI after the final changes, but the code changes look solid :slightly_smiling_face:

PR #1211, opened 2021-02-26
RD-756 - Update Deployment Action Buttons widget

qooban:
The change is not complex. Basically I updated Deployment Action Buttons widget to reuse new component - `DeploymentActionsMenu`. At once I fixed some issues I spotted on the way. Hope some review comments + code comments are clear enough.

This change to work properly depends on https://github.com/cloudify-cosmo/cloudify-stage/pull/1209 .

### Video

https://user-images.githubusercontent.com/5202105/109295745-8aa81780-782f-11eb-9720-10291dba94f1.mp4


test/cypress/integration/widgets/blueprint_actions_spec.js
```diff
@@ -1,4 +1,4 @@
-describe('Blueprint Actions widget', () => {
+describe('Blueprint Action Buttons widget', () => {
```
qooban, line 1: This is just to align the naming.

widgets/common/src/RemoveDeploymentModal.jsx
```diff
@@ -23,6 +23,11 @@ function RemoveDeploymentModal({ open, deploymentId, force, onHide, toolbox }) {
         clearErrors();
         actions[deleteAction]({ id: deploymentId })
             .then(() => {
+                const contextDeploymentId = toolbox.getContext().getValue('deploymentId');
+                if (deploymentId === contextDeploymentId) {
+                    toolbox.getContext().setValue('deploymentId', null);
+                }
+                toolbox.goToParentPage();
```
qooban, line 30: As Deployment Action Buttons widget can be placed on the deployment drill-down page on a removal it make sense to go to the parent page.

Right now, I realized that it probably be necessary to revisit this implementation once Deployment View widget is implemented.

widgets/deploymentActionButtons/src/widget.jsx
```diff
@@ -19,31 +19,31 @@ Stage.defineWidget({
...
-                .then(deployment => {
-                    toolbox.loading(false);
-                    const workflows = Stage.Common.DeploymentUtils.filterWorkflows(
-                        _.sortBy(deployment.workflows, ['name'])
-                    );
 
-                    return Promise.resolve({ ...deployment, workflows });
-                });
+            return actions
+                .doGet({ id }) // FIXME: Once RD-1353 is implemented, pass { _include: ['worflows'] } to doGet
```
kubama, line 30:
```suggestion
                .doGet({ id }) // TODO: Once RD-1353 is implemented, pass { _include: ['workflows'] } to doGet
```

widgets/deploymentActionButtons/src/widget.jsx
```diff
@@ -19,31 +19,31 @@ Stage.defineWidget({
...
+                .then(deployment => {
+                    const workflows = DeploymentUtils.filterWorkflows(_.sortBy(deployment.workflows, ['name']));
```
kubama, line 32:
```suggestion
                    const workflows = DeploymentUtils.filterWorkflows(_.sortBy(deployment.workflows, 'name'));
```

widgets/deploymentActionButtons/src/widget.jsx
```diff
@@ -19,31 +19,31 @@ Stage.defineWidget({
...
         }
 
-        return Promise.resolve(DeploymentActionButtons.EMPTY_DEPLOYMENT);
+        return Promise.resolve({ id: '', workflows: [] });
     },
 
     fetchParams(widget, toolbox) {
         const deploymentId = toolbox.getContext().getValue('deploymentId');
+        // Deployment Actions Buttons widget does not support multiple actions, thus picking only one deploymentId
+        const firstDeploymentId = Array.isArray(deploymentId) ? deploymentId[0] : deploymentId;
```
kubama, line 44:
```suggestion
        const firstDeploymentId = _.castArray(deploymentId)[0];
```

widgets/deploymentActionButtons/src/widget.jsx
qooban: I went a bit further and moved `filterWorkflows` closer to the actual usage. Removed `DeploymentUtils` as it was not used anywhere else.

kubama (dismissed):

widgets/common/src/RemoveDeploymentModal.jsx
Gelio: Can we add a TODO comment for RD-1224? This way I will have a "marker" in the code when I'll be working on that other widget :slightly_smiling_face: and when searching for leftover TODOs, I'll spot this one and at least think about it

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -0,0 +1,56 @@
...
+            cy.interceptSp('POST', `/deployments/${deploymentName}/set-site`).as('setSite');
+
+            cy.get('button.deploymentActionsButton').should('not.have.attr', 'disabled');
+            cy.get('button.deploymentActionsButton').click();
+
+            cy.get('.popupMenu > .menu').contains('Set Site').click();
+            cy.get('.modal').within(() => {
+                cy.get('div[name="siteName"]').click();
+                cy.get(`div[option-value="${siteName}"]`).click();
+                cy.get('button.ok').click();
```
Gelio, line 49: This is more of a general comment - I have read and seen that relying on class names for selecting elements is fragile, since class names are usually applied for styling, and thus, the tests can break when using classes for selectors.

Also, if the tests are to closely resemble how the user would use the application, the user will not click elements using class names, but using text, captions, labels, or aria attributes :slightly_smiling_face: 

This is why [Testing Library recommends a prioritized list of query types by which to select elements](https://testing-library.com/docs/queries/about#priority), starting with text-based, continuing to adding artificial `data-test-id` attributes, and discourages using class names for selecting elements (selecting by class names is considered a hack :smile: )

I do not expect any changes here, I'm just putting this out there for you to consider. Maybe we can revisit the way wer query elements eventually :smile: 

If I were to make some changes here, I would probably do

```suggestion
            cy.contains('Set Site').click(); // Wow, I figured semantic UI would use some aria roles to show popups, but it does not :D
            cy.get('.modal').within(() => {
                cy.get('div[name="siteName"]').click();
                cy.get(`div[option-value="${siteName}"]`).click();
                cy.contains('Update').click(); // there's an icon there, so I'm not sure this will work
```

I thought Semantic UI is named that way because it is semantically correct :smile: I have been wrong - there are no aria attributes, roles, and `label` elements are not used correctly :scream:

widgets/common/src/RemoveDeploymentModal.jsx
Gelio: Btw. I reckon going to _previous page_, not to _parent page_ could be an ok solution

Gelio, line 30: Should we not do that inside the `if` above?

Otherwise, I don't see the point of clearing the value in the context that's done in the `if` :slightly_smiling_face:

widgets/common/src/WorkflowsMenu.jsx
```diff
@@ -4,6 +4,11 @@
 
 const WorkflowsPropType = Stage.PropTypes.Workflows;
 
+function filterWorkflows(workflows) {
```
Gelio, line 7: Minor:

How about 

```suggestion
function workflowsWithoutUpdate(workflows) {
```

for providing a bit more context :smile: ? I'm not 100% sure this makes sense

widgets/common/src/WorkflowsMenu.jsx
```diff
@@ -108,7 +113,8 @@ function WorkflowsMenu({ workflows, onClick, showInPopup, trigger }) {
         i18n
     } = Stage;
 
-    const workflowsGroups = _.chain(workflows)
+    const filteredAndSortedWorkflows = _.sortBy(filterWorkflows(workflows), 'name');
```
Gelio, line 116: Were they unsorted before :smile: ? Nice idea :+1:

widgets/deploymentActionButtons/src/DeploymentActionButtons.jsx
```diff
@@ -1,163 +1,70 @@
...
+                        icon="content"
+                        disabled={!id}
+                        content="Deployment actions"
+                    />
+                }
+            />
+
+            {id && workflow && (
+                <ExecuteDeploymentModal
+                    open={!_.isEmpty(workflow)}
```
Gelio, line 47: What's the point of checking `workflow` twice - once before rendering the element, and another time in the `open` prop? :thinking:

widgets/deploymentActionButtons/src/widget.jsx
```diff
@@ -19,31 +19,28 @@ Stage.defineWidget({
...
-                });
+
+            return actions
+                .doGet({ id }) // TODO: Once RD-1353 is implemented, pass { _include: ['workflows'] } to doGet
+                .then(deployment => ({ id, workflows: deployment.workflows }))
+                .finally(() => toolbox.loading(false));
         }
 
-        return Promise.resolve(DeploymentActionButtons.EMPTY_DEPLOYMENT);
+        return Promise.resolve({ id: '', workflows: [] });
```
Gelio, line 35: This looks like a good candidate to apply the early return pattern :slightly_smiling_face: Reversing the condition would make the most important code less indented

```ts
if (_.isEmpty(id)) {
  return Promise.resolve(...);
}

// The rest
```

widgets/deploymentActionButtons/src/widget.jsx
```diff
@@ -19,31 +19,28 @@ Stage.defineWidget({
...
     },
 
     fetchParams(widget, toolbox) {
         const deploymentId = toolbox.getContext().getValue('deploymentId');
+        // Deployment Actions Buttons widget does not support multiple actions, thus picking only one deploymentId
```
Gelio, line 40: In which situation can `deploymentId` be an array? :thinking:

widgets/common/src/WorkflowsMenu.jsx
Gelio: Nevermind, I see they were sorted in another place :slightly_smiling_face:

Gelio (commented):
Looks cool overall :slightly_smiling_face: I've posted a mix of general and specific comments, take a look :eyes:

widgets/deploymentActionButtons/src/widget.jsx
qooban: Resource Filter widget allows you to select multiple items.

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -0,0 +1,56 @@
+describe('Deployment Action Buttons widget', () => {
+    const blueprintName = 'deployment_action_buttons_test';
+    const deploymentName = 'deployment_action_buttons_test';
```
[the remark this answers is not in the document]
qooban: Hmm... I'm not sure. You will see blueprint only in the context of Blueprints (eg. in Blueprints widget) and deployment in the context of Deployments (eg. in Deployments widget). Is there any value in such change?

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -0,0 +1,56 @@
...
+        cy.get('button.deploymentActionsButton').should('have.attr', 'disabled');
+    });
+
+    describe('when deploymentId is set in the context', () => {
+        beforeEach(() => cy.setDeploymentContext(deploymentName));
+
+        it('should allow to execute a workflow', () => {
+            cy.interceptSp('POST', `/executions`).as('executeWorkflow');
+
+            cy.get('button.executeWorkflowButton').should('not.have.attr', 'disabled');
```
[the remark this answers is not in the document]
qooban: Yeah, I know that rules, but I think in this case it makes sense, I want to be explicit as in this place (as opposed to 'when deploymentId is not set in the context it should be disabled' test).

Regarding `be.visible` - Hm... I could consider removing it, but on the other hand with such check we have a bit clearer output in case of errors. Fail on the later line (`cy.get('.executeWorkflowModal button.ok').click();`) does not necessarily means that modal is visible.

widgets/common/src/WorkflowsMenu.jsx
qooban: Could be, but if we are going to be pure then shouldn't it be:
```suggestion
function getWorkflowsWithoutUpdate(workflows) {
```
?

widgets/deploymentActionButtons/src/widget.jsx
qooban: Applied.

widgets/deploymentActionButtons/src/DeploymentActionButtons.jsx
qooban: Yep, `open` prop is always true in this case, simplified.

widgets/common/src/RemoveDeploymentModal.jsx
qooban: Right, fixed by moving `toolbox.goToParentPage()` inside the condition block.

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -0,0 +1,56 @@
...
+            cy.get('.popupMenu > .menu').contains('Start').click();
+            cy.get('.executeWorkflowModal').should('be.visible');
+            cy.get('.executeWorkflowModal button.ok').click();
+
+            cy.wait('@executeWorkflow');
+            cy.get('.executeWorkflowModal').should('not.exist');
+        });
+
+        it('should allow to start an action on the deployment', () => {
+            const siteName = 'deployment_action_buttons_test';
```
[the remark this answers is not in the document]
qooban: Hmm... Not sure it is valuable. See my previous comment.

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Thanks for that comment. I can agree to some extent it is cleaner to use text-based selectors, for sure it is easier to read, but not in all cases it is unambiguous (as you can have multiple elements on the page with the same label). 

I checked [Cypress Best Practices for selecting elements](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements) and they recommend using either `data-*` attributes based selectors or text-based selectors.

So we can definitely consider switch in that direction.

kubama (approved):

qooban:
It's not a big PR, so I'm going to merge it with one approval.
I believe that I addressed/answered all (or most) of the comments.

test/cypress/integration/widgets/deployment_actions_spec.js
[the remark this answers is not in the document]
Gelio: Just naming entities differently :slightly_smiling_face: Right now we could use the wrong variable (e.g. use `deploymentName` instead of `blueprintName`) and the results will be the same :slightly_smiling_face: While it's nothing major here, in general, e.g. if a function accepted both `deploymentName` and `blueprintName`, I would definitely use different values to make sure that the inside the function proper arguments are used, and they are not switched by some mistake

test/cypress/integration/widgets/deployment_actions_spec.js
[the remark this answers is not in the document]
Gelio: Sure, explicitness here sounds ok :slightly_smiling_face: 

> Fail on the later line (cy.get('.executeWorkflowModal button.ok').click();) does not necessarily means that modal is visible.

I thought cypress displays the reason why an element was not clickable :thinking: Maybe I misread it

test/cypress/integration/widgets/deployment_actions_spec.js
Gelio: Ok, I'll raise it on Friday :slightly_smiling_face: I did not know Cypress had its own best practices page, so thanks for the resource

widgets/common/src/WorkflowsMenu.jsx
Gelio: Yeah, that would make sense :smile: But I see it's too late :slightly_smiling_face:

widgets/deploymentActionButtons/src/widget.jsx
Gelio: Oh... Ok :slightly_smiling_face: I didn't know, thanks

PR #1217, opened 2021-03-02
RD-1588 Migrate Widget component to TypeScript

Gelio:
This PR migrates the Widget component to TypeScript. It also merges the `Widget` container (the `connect` part) with the component.

The PR contains very few functional changes:
1. Extracting the `escapeKeyCode` variable in `Widget`
2. Changing the `_.isEqual` check for `readme` in `Widget`
2. Refactoring `mapDispatchToProps` to a shorthand version

Most of the changes are TypeScript related


app/components/Widget.tsx
```diff
@@ -1,37 +1,53 @@
...
+    hasError: boolean;
+    readmeContent: string;
+    showReadmeModal: boolean;
+}
+
+class Widget<Configuration> extends Component<WidgetProps<Configuration>, WidgetState> {
+    private widgetItemRef = createRef<any>();
 
-export default class Widget extends Component {
-    constructor(props, context) {
```
Gelio, line 18: No longer using the legacy `context` forwarding (it was probably always `undefined` since we do not have `contextTypes` defined on the component)

widgets/deploymentsView/src/columns.tsx
```diff
@@ -1,6 +1,7 @@
 import { mapValues } from 'lodash';
 import type { ReactNode } from 'react';
 import type { SemanticICONS } from 'semantic-ui-react';
+import type { WithOptionalProperties } from '../../../app/utils/types';
```
Gelio, line 4: I'm not sure about the path used here. I did not find a better place to store such shared types

Gelio:
Just FYI, I'm planning on changing https://github.com/cloudify-cosmo/cloudify-stage/blob/994497a81882bd8512a1c04c7b5379a459417730/app/utils/WidgetDefinition.js#L0-L1 in this PR to, to be more TypeScript-friendly :slightly_smiling_face: That should not impact the existing parts of the PR

widgets/deploymentsView/src/columns.tsx
qooban: I know we are preferring imports, but... Wouldn't it be better to expose TS types the same way as we exposed custom `PropTypes`? (through Stage global) What is the direction you see for `Stage` global?

In general - I think it's quite important topic if we are going to change the whole idea of that interface between application framework (`app` directory) and widgets (`widgets` directory). 

We also have to discuss custom widget development after TS introduction. How it affects custom widget developers wanting to use cloudify-stage as a starting base for widget development?

qooban (commented):
Added some comments. Focused mostly on general aspects.
Providing valuable feedback for TS code is still out of my possibilities.

app/components/Widget.tsx
Gelio: Oh, and even then it's the new context, not the legacy one :smile:

app/utils/StageAPI.ts
```diff
@@ -144,10 +150,26 @@ interface CommonOrPropTypeDefinition {
     common: any;
 }
 
+type WidgetDefinitionForDefining<Params, Data, Configuration> = WithOptionalProperties<
```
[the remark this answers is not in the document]
Gelio: Well... yeah... I agree :smile: I couldn't come up with a better one. If you have a suggestion, feel free to propose it

The reason for this type is that the `WidgetConfiguration` that can be passed to `defineWidget` has some properties marked as optional, whereas when you use `widget.configuration` somewhere later on, those properties are already defined, since the default values are supplied by [`WidgetDefinition`](https://github.com/cloudify-cosmo/cloudify-stage/blob/994497a81882bd8512a1c04c7b5379a459417730/app/utils/WidgetDefinition.js). Thus, we need 2 types - 1 with optional properties, 1 with those properties marked as required (meaning when you access `widget.configuration.hasStyle`, it's definitely a `boolean`, and not `boolean | undefined`)

The ideas I have for the name are:
1. `PartialWidgetDefinition`
2. `InitialWidgetDefinition`

:man_shrugging:

widgets/deploymentsView/src/columns.tsx
Gelio: That's a loaded question/comment :smile: Let me address each item individually:

1. > Wouldn't it be better to expose TS types the same way as we exposed custom PropTypes?
    
    I thought about it for a while but then thougth that `import type` is cheap, and is erased by Babel,
    so it does not influence the actual bundling.

    We could expose that type on the `Stage` global, but at the moment it is impossible AFAIK, since the
    [`Stage` global is an `interface`](https://github.com/cloudify-cosmo/cloudify-stage/blob/dd2f0b1b07563be0ed1655f4aaf3308d0efad7a4/app/utils/StageAPI.ts#L169-L189),
    and not a namespace. Types cannot be attached to interfaces, but they can be attached to namespaces.

    I have created RD-1409 some time ago to change Stage into a namespace, and the changes seem relatively small.
    I can try to do it now to avoid introducing this possible technical debt in the means of importing something from `app` inside of `widgets` 😃

    I'm leaning towards doing it now, but let me know what you think 😃

2. > What is the direction you see for Stage global?

    Personally, I would mark the `Stage` global as deprecated and tweak webpack to allow importing from `cloudify-stage`.

    Thus, the following code:

    ```ts
    Stage.defineWidget(...);
    ```

    would become

    ```ts
    import { defineWidget } from 'cloudify-stage';

    defineWidget(...);
    ```

    This way we avoid using global variables, use the standard ES modules. To me, that's a much nicer API 😃

3. > In general - I think it's quite important topic if we are going to change the whole idea of that interface between application framework (app directory) and widgets (widgets directory).

    I agree that this should be ironed out, but I don't see much overlap here really - the only interaction I see now is the `Stage` global and importing types in
    https://github.com/cloudify-cosmo/cloudify-stage/blob/d4e0188e181958aaba320a80d414166549999d44/widgets/tsconfig.json#L8-L9

4. > We also have to discuss custom widget development after TS introduction. How it affects custom widget developers wanting to use cloudify-stage as a starting base for widget development?

    There's no impact at the moment - the types are not exposed externally. There's RD-1292 to expose them 🙂

    If we combine that with the idea from point 2, then the consumers writing custom widgets would be able to use the Stage API through a regular ES import,
    which IMO is a much better solution than forcing them to use a global variable 😃

app/utils/normalizeWidgetDefinition.ts
```diff
@@ -0,0 +1,37 @@
+import log from 'loglevel';
+
+import GenericConfig from './GenericConfig';
+import { WidgetDefinition, WidgetDefinitionForDefining } from './StageAPI';
+
+export default function normalizeWidgetDefinition<Params, Data, Configuration>(
+    initialDefinition: WidgetDefinitionForDefining<Params, Data, Configuration>
+): WidgetDefinition<Params, Data, Configuration> {
+    const id = initialDefinition.id ?? document.currentScript?.id;
```
Gelio, line 9: Notice that this is a breaking change, but I'm not sure how to proceed :smile: Previously, the logic was https://github.com/cloudify-cosmo/cloudify-stage/commit/666e5d95fcf57e9c616067196a6feb24e48465a3#diff-74faea8610628fbca0503669f2b8ea6e104a402078f1560f56c89f06d924a99f which puzzles me.

1. `id` was always overridden, even if `document.currentScript` did not exist (then, `id` was set to `undefined`, see [the playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAJgUwGYEswqi8MC8MDeAUDCTCnAFwwCMANIQL4DchhokIANggHScgBzABRFSMHhMSp0mcPTHkqAVzBS0COIwCUQA))
2. This means that `id` defined in the widget definition **never mattered**, because the dynamically-loaded widget `script`'s ID was determined based on the widget folder name (https://github.com/cloudify-cosmo/cloudify-stage/blob/0aa73f914e29f3beacb89ba153dce977ef778a1d/backend/handler/WidgetHandler.js#L276-L277).

    Since it was always defined, it overrode the id from the definition object :smile: 

Thus being said, do we even need the `id` in the definition object? I reckon it can happen that a user will define a widget via some non-standard way. Then, if `document.currentScript.id` won't be passed, there's no way of knowing the id, and then, the `id` in the definition makes sense. For other cases, I would not encourage setting a property that does not matter at all, but looks like it matters :slightly_smiling_face: 

Therefore, my suggestion is to change this line to 

```suggestion
    const id = document.currentScript?.id ?? initialDefinition.id;
```

and make `id` in the definition optional :slightly_smiling_face:

Gelio:
As I have split RD-1223 into multiple subtasks, this PR is now for RD-1588

app/reducers/index.ts
```diff
@@ -4,6 +4,8 @@
 
 import { combineReducers } from 'redux';
 import { connectRouter } from 'connected-react-router';
+import { History } from 'history';
```
gregcfy, line 7: should we use import type { History } from 'history'; ?

Gelio: Yup, you're right :slightly_smiling_face: It's not required, but IMO it's good style to mark type-only imports as such to avoid unnecessary work for the bundler :smile: Done

gregcfy (dismissed):

app/utils/StageAPI.ts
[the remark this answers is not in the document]
qooban: I like `InitialWidgetDefinition` 👍🏻

app/utils/normalizeWidgetDefinition.ts
qooban: Looks like it is overcomplicated, but is that breaking change necessary for you to proceed? Or we can leave it for later?

If you insist on change, then:
1. Please check `edit_mode_spec.js` and run all tests, especially `'should open widget install modal and'` - this tests covers adding widgets. If everything will be fine, then probably you can introduce your change.
2. Verify it with official documentation - what we say to the user (custom widget developers), we have to be consistent.

qooban (commented):
Looks good overall 👍🏻 

I'm not sure about the breaking change.

app/utils/normalizeWidgetDefinition.ts
Gelio: Having thought about it, I'll revert the breaking change to the initial implementation, verify that it's the same by using unit tests, and then create a ticket to decide whether to fix it later :slightly_smiling_face:

app/utils/StageAPI.ts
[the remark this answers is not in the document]
Gelio: Done in 0b71f2e6491f95a95cad897b11b451e741ae0a1e :slightly_smiling_face: thanks for the suggestion

widgets/deploymentsView/src/columns.tsx
qooban: Thank you for the detailed answer. 

I think that if RD-1409 is relatively small and is going to allow us to use Stage.X global for that case, then I'd suggest to go this way.

Another point for deeper discussion about custom widget development is: What if a widget developer would like to import something? We explicitly list all the external dependencies which he can use (see [docs](https://docs.cloudify.co/developer/writing_widgets/widget-apis/#external-libraries)) without additional cost. But now he might be confused when seeing imports in our widgets' code. He will not be able to use our webpack changes when he is using [Cloudify-UI-Widget-boilerplate](https://github.com/cloudify-cosmo/Cloudify-UI-Widget-boilerplate) (probably not working, but still, it is open to use) or when he is developing the widget outside of cloudify-stage and Cloudify-UI-Widget-boilerplate.

This is of course separate discussion, but just to show you the wider context for any changes related to how we expose widgets API.

BTW:
```
import { defineWidget } from 'cloudify-stage';
```
is nice, but for that we'll need to have `cloudify-stage` in NPM or is there any other solution?

Gelio:
I have reverted the functional breaking change done in `normalizeWidgetDefinition` 

I have added tests to cover that behavior that I find buggy :smile: To make sure nothing changed in that area before and after my refactoring of creating `normalizeWidgetDefinition`, I have altered the branch history and added the tests before the commit for adding `normalizeWidgetDefinition`. This means that this PR requires an almost entire re-review :slightly_frowning_face: The benefit is that we can be sure that the behavior is the same :slightly_smiling_face: 

To see which commits have changed, you can run 

```sh
git range-diff origin/master 0b71f2e6491f95a95cad897b11b451e741ae0a1e origin/RD-1223-move-Widget-to-TS
```

which will show which commits have changed and require rereviewing, and which are the same as before (the ones with `=`) and do not require another look :slightly_smiling_face: 0b71f2e6491f95a95cad897b11b451e741ae0a1e is the previous version of this branch (before force pushing).

The tests for `widgetDefinitionsLoader` required upgrading `jsdom` (I could not modify `document.currentScript` on the previous version).

I will run UI System tests for this branch just to be 100% sure everything is working :slightly_smiling_face:

app/utils/normalizeWidgetDefinition.ts
Gelio: As promised, I have reverted the breaking change and added tests to make sure the behavior was not altered during the refactor. See https://github.com/cloudify-cosmo/cloudify-stage/pull/1217#issuecomment-790586456 for a summary :slightly_smiling_face:

app/utils/widgetDefinitionsLoader.ts
```diff
@@ -12,15 +13,10 @@ import StageUtils from './stageUtils';
...
 import GenericConfig from './GenericConfig';
-import WidgetDefinition from './WidgetDefinition';
 import * as PropTypes from './props';
 import * as Hooks from './hooks';
-import { StageAPI } from './StageAPI';
-
-// NOTE: why do we even use require here?
-// @ts-expect-error Allowed temporarily
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const ReactDOMServer = require('react-dom/server');
```
Gelio, line 23: This changed to a regular import to fix a TS error that started appearing after importing this file in the tests.

It still is working in the UI:

image

package.json
```diff
@@ -154,7 +157,7 @@
     "imagemin-webpack-plugin": "^2.4.2",
     "ip": "^1.1.3",
     "jest": "^26.4.2",
-    "jsdom": "^9.8.3",
+    "jsdom": "^16.4.0",
```
Gelio, line 160: Necessary upgrade for overriding `document.currentScript` to work in the tests

widgets/deploymentsView/src/columns.tsx
Gelio: > I think that if RD-1409 is relatively small and is going to allow us to use Stage.X global for that case, then I'd suggest to go this way.

Sure thing, I'll include RD-1409 in the current sprint and try to finish it :slightly_smiling_face: I'll do that in a separate PR, though. I will modify this place in that PR


> Another point for deeper discussion about custom widget development is

My idea is that the boilerplate will be configured in such a way that `import`s for those globally-exposed libraries will be changed to global imports, as opposed to importing the dependency itself :slightly_smiling_face: 

I didn't explore whether it is possible, but I would imagine it would be :slightly_smiling_face: The same way we can `import _ from 'lodash';` in our widgets (using webpack's `externals`).

> we'll need to have cloudify-stage in NPM or is there any other solution?

We could use the same solution as we do for `lodash` etc, meaning webpack's `externals` :slightly_smiling_face: Basically, translating imports to an import from the global variable

--- 

All in all, the user would be able to either use the global variables, or use imports which would be transformed to accessing global variables under the hood :smile: The best developer experience, IMO

Gelio:
Here are the system tests for this branch: https://jenkins.cloudify.co/job/Stage-UI-System-Test/226/

app/utils/widgetDefinitionsLoader.ts
```diff
@@ -82,7 +76,7 @@ export default class WidgetDefinitionsLoader {
         window.Stage = stageAPI;
     }
 
-    static loadWidgets(manager: any) {
+    private static loadWidgets(manager: any) {
```
qooban, line 79: `private static` - interesting :-) Good point is that it is clear right now that we have static function which is used only internally. Are there any real use-cases for that? 

I quickly checked the history and we've got such "internal" static functions since the beginning I think. Seems like this file is waiting for the major refactoring :-) 

Nothing wrong on your side, no need to change anything.

qooban (approved):
Thanks for applying the changes. All good 👍🏻

gregcfy (approved):

app/utils/widgetDefinitionsLoader.ts
Gelio: > Are there any real use-cases for that?

It's very similar to unexported module-scoped functions :smile: I don't see how `private static` can be more useful than a local function :slightly_smiling_face: Perhaps it helps with grouping if we have a class anyway

That probably comes from purely object-oriented languages, where you cannot have free-standing functions, but you need to group them in classes :slightly_smiling_face: 

Anyway, I wouldn't worry about it necessarily :slightly_smiling_face: Just keep in mind that even if it's marked as `private`, you can still call `WidgetDefinitionsLoader.loadWidgets` in JS - the `private` part is only enforced when using TypeScript

Gelio:
Thanks for reviews, guys :slightly_smiling_face:

PR #1219, opened 2021-03-03
RD-1593 Move Page's container to the component file

Gelio:
This PR moves the `Page` container to its component file.

This PR relies on #1217 being merged first

There should be no functional changes in this PR, just moving files around

PR #1220, opened 2021-03-03
RD-1589 Migrate Page component to TS

Gelio:
This PR continues the work of migrating some components / files to TS started in #1217 

Due to extracting the changes for merging `Page` container and component in #1219, git detects most migrations as renames. The only one not detected is for a small `ConfigLoader` (since now most of the file is an interface declaration :slightly_smiling_face: )

Most of the migrations excluding `Page.tsx` come from the fact that `Page` uses those types from other files in some way. They were relatively easy for me to write, so I decided to add type annotations for some of those dependencies :slightly_smiling_face: 

## Functional changes

The most severe functional change in this PR is removing the use of jQuery in `Page`'s `render` method. I replaced it with regular DOM operations.

I have marked all functional changes with PR comments.

---

This time there should be less TypeScript magic :smile: :mage: 

---

This PR relies on #1219 being merged first


app/components/Page.tsx
```diff
@@ -59,9 +64,8 @@ class Page extends Component {
             _(page.layout).flatMap('content').find({ maximized: true }) ||
             _(page.layout).flatMap('content').flatMap('widgets').find({ maximized: true });
 
-        $('body')
-            .css({ overflow: maximizeWidget ? 'hidden' : 'inherit' })
-            .scrollTop(0);
+        document.body.style.overflow = maximizeWidget ? 'hidden' : 'inherit';
+        window.scroll({ top: 0 });
```
Gelio, line 68: I've decided to use native DOM to avoid pulling in jQuery types

app/configureStore.ts
```diff
@@ -1,25 +1,26 @@
...
+        manager: emptyState
+    };
 
     // Clear login error if has any
-    initialState.manager.err = null;
-    initialState.manager.isLoggingIn = false;
+    reduxManagerState.manager.err = null;
+    reduxManagerState.manager.isLoggingIn = false;
+
+    const initialState: Partial<ReduxState> = { ...reduxManagerState, config };
```
Gelio, line 23: This is essentially the same as the previous implementation. I have just used 2 separate variables to guide TypeScript that `manager` is defined on `reduxManagerState`. I did not have any idea how to do that with the previous implementation that used a single variable :slightly_smiling_face:

app/reducers/widgetDefinitionsReducer.ts
```diff
@@ -1,25 +1,27 @@
...
+import type { Reducer } from 'redux';
+
 import * as types from '../actions/types';
+import type { WidgetDefinition } from '../utils/StageAPI';
 
-export default (state = {}, action) => {
-    let defs;
+export type WidgetDefinitionsState = WidgetDefinition[];
 
+const widgetDefinitionsReducer: Reducer<WidgetDefinitionsState> = (state = [], action) => {
```
Gelio, line 9: Notice the fixed default value for `state`

app/utils/widgetDefinitionsLoader.ts
```diff
@@ -228,7 +228,7 @@ export default class WidgetDefinitionsLoader {
                     i18n.t(
                         'widget.validationErrors.invalidPermission',
                         `Specified widget permission ('{{permission}}') not found in available permissions list.`,
-                        widgetDefinition
+                        _.pick(widgetDefinition, 'permission')
```
Gelio, line 231: The change here is to help with TypeScript. `widgetDefinition` somehow did not work well with `i18n.t` generic types, given that it also is generic

gregcfy (approved):

app/components/Page.tsx
qooban: Any change going towards leaving jQuery is appreciated. Just please verify maximizing works as before.

Just FYI - [RD-258](https://cloudifysource.atlassian.net/browse/RD-258) - there's plan to remove it completely.

app/configureStore.ts
qooban: Hmm... I'm not sure the name is the best as it suggest to me that there is only manager state inside, but actually there can be and in most case there would be `StatePersister.load(config.mode)` which is full state, just taken from the local storage.

Oh, but just checked that actually `StatePersister.load(config.mode)` gives you only `manager` part. That is strange. Hmm... In that case maybe we can change `StatePersister` name? WDYT? Maybe all of that presetting (emptyState + Clear login error if has any) could be moved to `StatePersister`? Just loudly thinking. Leaving it to you :-)

app/reducers/widgetDefinitionsReducer.ts
qooban: 👍🏻  Should be fine, I think.

app/reducers/widgetDefinitionsReducer.ts
```diff
@@ -1,25 +1,27 @@
...
     switch (action.type) {
         case types.STORE_WIDGETS:
             return [...action.widgetDefinitions];
         case types.INSTALL_WIDGET:
             return _.sortBy([...state, ...action.widgetDefinitions], ['name']);
-        case types.UPDATE_WIDGET_DEFINITION:
-            defs = _.reject(state, { id: action.widgetId });
+        case types.UPDATE_WIDGET_DEFINITION: {
+            const defs = _.reject(state, { id: action.widgetId });
```
qooban, line 16: Always good to kill such small things (let -> const + move inside case).

app/utils/ConfigLoader.ts
```diff
@@ -0,0 +1,28 @@
+import log from 'loglevel';
+import fetch from 'isomorphic-fetch';
+
+import StageUtils from './stageUtils';
+
+export interface ClientConfig {
```
qooban, line 6: Family friendly interface 🙂

app/utils/widgetDefinitionsLoader.ts
qooban: Make sense anyway. Passing whole `widgetDefinition` was not the best.

qooban (commented):
All good in general 👍🏻 

No blockers. If you find some time to think about `StatePersister` then it'd be great. If not let me know, I'll probably approve.

app/components/Page.tsx
Gelio: Sounds good :slightly_smiling_face: 

Maximizing looks to be working fine:


https://user-images.githubusercontent.com/889383/109968450-0227ec00-7cf3-11eb-92e0-a133a825b328.mp4

app/reducers/widgetDefinitionsReducer.ts
Gelio: Yeah, definitely :smile: This way it's also easier for TS to infer the type of the variable :slightly_smiling_face:

app/utils/ConfigLoader.ts
Gelio: Haha yup, looks fine :eyes:

app/configureStore.ts
Gelio: I would be glad to be able to refactor this part :smile: I wanted to avoid doing that in this PR, but since you're fine with it, I'll try to do some good here :slightly_smiling_face: 

I'll name it to `ManagerStatePersister` to be the most concrete :slightly_smiling_face:

Gelio: I have:

1. Renamed the class to `ManagerStatePersister`
2. Moved the logic of clearing login errors into the class
3. Changed the logic in the class so that the additional wrapping of the `manager` object is not needed.
4. Added a compatibility layer so that when the user logs in for the first time after this change, the `ManagerStateProvider` will load the state the old way
5. Added tests to make sure all of the above work

Looks like a nice cleanup :smile:

Gelio:
I've rebased this branch on top of the base branch, since the base branch was rebased on its base branch (`master`) :smile: It should be good to review :slightly_smiling_face: 

This time I have only appended commits

UI System tests for this branch will appear in https://jenkins.cloudify.co/job/Stage-UI-System-Test/230/

app/utils/ManagerStatePersister.ts
```diff
@@ -0,0 +1,63 @@
...
+
+        this.save(managerState, mode);
+        localStorage.removeItem(storageKey);
+
+        return managerState;
+    }
+}
+
+const getStorageKey = (mode: string) => `manager-state-${mode}`;
+const getDeprecatedStorageKey = (mode: string) => `state-${mode}`;
```
qooban, line 63: I understand the point of the name change, I hope it will not cause any issues.

Please check if there are no uses of `state-main` in our code. 
I found one in tests: `mockLogin` custom Cypress command. Let's move it to new name and structure.

app/configureStore.ts
qooban: Great 👍🏻

qooban (commented):
Looks good, just one comment. 

Let's align all parts of our code to the state persister change.

app/components/Page.tsx
gregcfy: what  do you think to use `window.scroll(0, 0)` instead of `window.scroll({ top: 0 });` ?
 or even scrollTo()

Gelio: I'm not sure :smile: What's the benefit?

I was mimicking the existing jquery code (`$.scrollTop(0)` https://github.com/cloudify-cosmo/cloudify-stage/pull/1220/files#diff-9a2ef18f11ca6f16e6751575f3bdbef8654d407e12380abb2fce4d806623f37eL46)

app/utils/ManagerStatePersister.ts
Gelio: Hmm... I did not think about checking for that string across the codebase :slightly_smiling_face: I'll do that, thanks for the suggestion

Gelio:
I've only rebased the branch on top of `master` to resolve the merge conflict. I'll address the review comments in a second

app/utils/ManagerStatePersister.ts
Gelio: I've only seen the `state-` prefix used in the Cypress command you mentioned :slightly_smiling_face:
image

I have adjusted it to the new format (although the old one would work fine, as shown by the UI system tests passing)

Gelio:
Stage UI system tests for the updated branch: https://jenkins.cloudify.co/job/Stage-UI-System-Test/235/

app/components/Page.tsx
gregcfy: For Edge `window.scroll(options)` was added at the beginning of 2020 - probably it was a time of migration into Chrome Engine. 
https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll

Gelio: Good catch :slightly_smiling_face: Done

qooban (approved):

PR #1221, opened 2021-03-03
RD-754 Labels widget implementation continued

kubama:
See commit log for list of changes.


widgets/common/src/labels/index.js
```diff
@@ -0,0 +1,14 @@
+import Modal from './LabelsModal';
+import ManageModal from './ManageLabelsModal';
+import ValidationErrorPopup from './ValidationErrorPopup';
+
+const Labels = {
```
qooban, line 5: I like that you implemented that idea.

Probably you will need to add `LabelsInput` here as well (for RD-757 - Update Deploy Blueprint modal). 
No need to change it now. Just noticed.

test/cypress/integration/widgets/labels_spec.ts
```diff
@@ -0,0 +1,63 @@
...
+            .activate()
+            .mockLogin()
+            .deleteDeployments(deploymentName, true)
+            .deleteBlueprints(blueprintName, true)
+            .uploadBlueprint('blueprints/simple.zip', blueprintName)
+            .deployBlueprint(blueprintName, deploymentName, { server_ip: 'localhost' })
+            .setDeploymentContext(deploymentName);
+    });
+
+    it('should allow to manage labels', () => {
```
qooban, line 16: Few questions about that test:
1. I'm wondering what was the reasoning behind testing whole functionality in a single test? Was it Cypress losing context (local storage) making it harder to make small and clean tests? Was that test-time or code overhead when splitting into smaller tests?
2. Export to JSON not tested probably due to lack of support from Cypress side for file downloads, right?

qooban (commented):
Very good changes, I like the direction 👍🏻 

Added some comments with questions, none of them are blockers for this PR.

widgets/common/src/labels/Input.jsx
```diff
@@ -1,16 +1,16 @@
...
+import { addSearchToUrl } from './common';
 
 const iconStyle = {
     position: 'absolute',
     top: '.7em',
     zIndex: 1
 };
 
-export default function LabelsInput({ hideInitialLabels, initialLabels, onChange, toolbox }) {
+export default function Input({ hideInitialLabels, initialLabels, onChange, toolbox }) {
```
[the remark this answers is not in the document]
Gelio: I agree :slightly_smiling_face: I would rather keep the components themselves named with the context prefix (`Labels...`), but the context prefix can be skipped in the global namespace :slightly_smiling_face:

test/cypress/integration/widgets/labels_spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        });
+
+        cy.log('Verify label value can be edited');
+        cy.get('.labelsTable').within(() => {
+            cy.get('.edit').click();
+            cy.get('.cancel').click();
+            cy.get('.edit').click();
+            cy.get('input').type('!');
+        });
+        cy.contains('Only letters, digits and the characters "-", "." and "_" are allowed');
```
Gelio, line 47: Is the intent here also to test the validation logic? If so, IMO this is not as clear as it could. I would encourage either doing unit tests to test the validation logic, or separate the validation checks into a separate test, and make them more detailed (test many cases, not only `!` :slightly_smiling_face: )

This way we will have a single place that tests the validation exclusively and extensively, rather than verifying it  _by the way_

test/cypress/integration/widgets/labels_spec.ts
Gelio: A side comment from me :slightly_smiling_face: 

For part 1 - I actually like the idea of having a single test here, as it is a single user journey of managing labels (a single label, actually :smile: ). If we split it, then there would be dependencies between tests (e.g. need to create a label before it can be edited), which seems like an antipattern, or each test would need additional setup overhead :slightly_smiling_face: 

The only thing I would encourage go break out is the validation

test/cypress/integration/widgets/labels_spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        cy.get('.labelsTable').within(() => {
+            cy.get('input').should('be.visible');
+            cy.get('.undo').should('not.exist');
+            cy.get('input').type('_changed');
+            cy.get('.undo').should('be.visible');
+            cy.get('.check').click();
+            cy.get('input').should('not.exist');
+            cy.contains('sample_value_changed');
```
Gelio, line 55: Due to optimistic UI, this will appear automatically even if the request to the backend will fail. Should we not check that the backend request actually happened, or perhaps refresh the page and then check if the value is changed?


This way we'll have confidence that the request works too, and it's not the optimistic UI part that makes the test pass

app/translations/en.json
```diff
@@ -507,6 +507,25 @@
...
+        "labels": {
+            "name": "Labels",
+            "description": "This widget shows a list of defined labels",
+            "columns": {
+                "key": "Key",
+                "value": "Value",
+                "actions": {
+                    "edit": "Edit label",
+                    "delete": "Delete label"
+              }
```
Gelio, line 520: This looks misaligned :slightly_smiling_face:

widgets/common/src/labels/CommonDropdown.jsx
```diff
@@ -1,4 +1,5 @@
 import { addSearchToUrl } from './common';
+import ValidationErrorPopup from './ValidationErrorPopup';
```
Gelio, line 2: I see we are going back and forth when it comes to using common components from `Stage.Common` or importing them :slightly_smiling_face:

widgets/common/src/labels/index.js
Gelio: Yup, very clean to only expose a few components :ok_hand:

widgets/common/src/labels/LabelsModal.jsx
```diff
@@ -79,34 +79,22 @@ function ManageLabelsModal({ deploymentId, existingLabels, header, applyButtonCo
...
 }
 
-ManageLabelsModal.propTypes = {
-    applyButtonContent: PropTypes.string,
+LabelsModal.propTypes = {
+    applyKey: PropTypes.string.isRequired,
     deploymentId: PropTypes.string.isRequired,
-    existingLabels: Stage.PropTypes.Labels,
-    header: PropTypes.string,
+    headerKey: PropTypes.string.isRequired,
```
Gelio, line 91: For giving a bit more context what those `key`s are, I encourage adding a `i18n` somewhere in the name, so it's clear that those are not React keys, but i18n ones :slightly_smiling_face: 

e.g. `applyI18nKey` or `i18nApplyKey`

Gelio: btw. good idea to extract it as prop :slightly_smiling_face: Makes things more readable, as opposed to the previous implementation that used default values :slightly_smiling_face:

widgets/common/src/labels/LabelsModal.jsx
```diff
@@ -1,34 +1,34 @@
...
     const [labels, setLabels, resetLabels] = useResettableState([]);
-    const [initialLabels, setInitialLabels, resetInitialLabels] = useResettableState(existingLabels || []);
+    const [initialLabels, setInitialLabels, resetInitialLabels] = useResettableState([]);
 
     useOpenProp(open, () => {
         clearErrors();
         resetLabels();
         resetInitialLabels();
 
-        if (!existingLabels) {
```
Gelio, line 18: Clever :ok_hand:

Gelio (commented):
I love the clear commit separation :ok_hand: It made the PR a brief to review :slightly_smiling_face:

widgets/common/src/labels/Input.jsx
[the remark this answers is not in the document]
kubama: Right, updated.

test/cypress/integration/widgets/labels_spec.ts
kubama: I split it into 3 test cases. No additional per-test setup overhead is needed.

> Export to JSON not tested probably due to lack of support from Cypress side for file downloads, right?

Yes, that's true.

test/cypress/integration/widgets/labels_spec.ts
kubama: The validation mechanism here is common with the modal. I agreed with @qooban that he's going to test the modal thoroughly, so here I'm only checking that the common validation logic gets fired.

qooban: Yep, I'm going to create better LabelsInput tests within [RD-781](https://cloudifysource.atlassian.net/browse/RD-781).

test/cypress/integration/widgets/labels_spec.ts
kubama: I've added backend call check.

app/translations/en.json
kubama: Fixed

widgets/common/src/labels/LabelsModal.jsx
kubama: I applied the second suggestion - I think having common prefix logically groups related props together.

kubama:
The only thing left is README.md content - since it's blocked by https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1730 I'd like to proceed with this one and eventually merge README.md later in separate PR.

widgets/common/src/hooks/useEventListener.js
```diff
@@ -0,0 +1,14 @@
...
+function useEventListener(toolbox, event, handler = toolbox.refresh) {
+    useEffect(() => {
+        if (event) {
+            toolbox.getEventBus().on(event, handler);
+            return () => toolbox.getEventBus().off(event, handler);
+        }
+
+        return undefined;
+    }, []);
+}
```
qooban, line 12: Very cool and useful addition 👍🏻

qooban (approved):
LGTM 👍🏻 

I assume that you verified after the refactoring (new hook `useEventListener` introduction) all changed widgets are properly working.

Gelio:
I'll take a look at this PR in 1h :slightly_smiling_face:

kubama:
> I assume that you verified after the refactoring (new hook `useEventListener` introduction) all changed widgets are properly working.

I ran system tests on the branch - all green.

test/cypress/integration/widgets/labels_spec.ts
Gelio: Makes sense, thanks for explanations :slightly_smiling_face:

test/cypress/integration/widgets/labels_spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        });
+        cy.wait('@labelsUpdate');
+        cy.contains('existing_changed');
+    });
+
+    it('should allow to remove labels', () => {
+        cy.get('.trash').each(deleteIcon => {
+            cy.wrap(deleteIcon).click();
+            cy.contains('Yes').click();
+            cy.wrap(deleteIcon).should('not.exist');
```
Gelio, line 59: Minor: Those could be extracted to outside the `each` (while refactoring the 2nd line to `cy.get('.trash').should('not.exist')`). Sounds more logical to me with less nesting :slightly_smiling_face: And the suggested assertion (the `cy.get` instead of `cy.wrap`) would make sure that there is actually no other trash icon (if I'm not mistaken, at the moment it only tests that this exact HTML element was removed)

widgets/common/src/labels/LabelsModal.jsx
```diff
@@ -40,13 +48,14 @@ function ManageLabelsModal({ deploymentId, existingLabels, header, applyButtonCo
...
 
-        const deploymentLabels = existingLabels ? [...existingLabels, ...labels] : labels;
+        const deploymentLabels = hideInitialLabels ? [...initialLabels, ...labels] : labels;
         actions
             .doSetLabels(deploymentId, deploymentLabels)
             .then(() => {
                 // State updates should be done before calling `onHide` to avoid React errors:
                 // "Warning: Can't perform a React state update on an unmounted component"
                 unsetLoading();
+                toolbox.getEventBus().trigger('labels:refresh');
```
Gelio, line 58: General comment/digression:

It would be awesome if we had a single repository with all known events that are triggered using the event bus. This would reduce the risk of introducing typos and allow for easier exploration of the codebase (one could use TS references instead of searching for strings) :slightly_smiling_face: 

Then, each event could also have its description to provide more context immediately, rather than having to read the source code to see what the intent of that event is

I'll introduce the idea during the weekly :slightly_smiling_face:

widgets/common/src/hooks/useEventListener.js
```diff
@@ -0,0 +1,14 @@
...
+
+function useEventListener(toolbox, event, handler = toolbox.refresh) {
+    useEffect(() => {
+        if (event) {
+            toolbox.getEventBus().on(event, handler);
+            return () => toolbox.getEventBus().off(event, handler);
+        }
+
+        return undefined;
+    }, []);
```
Gelio, line 11: Not using the variables that are closed over in the inner function inside of the dependency array makes this hook unsafe in my view :slightly_smiling_face: 

The crux of the matter is that this hook will not react to changes for `event` or `handler`. It will only add the listener on component's mount, and remove the listen on component's unmount. No reaction for the updates of the component. This is quite unexpected for a hook not to react to changes its parameters. This would be marked as an error if we were using https://www.npmjs.com/package/eslint-plugin-react-hooks

I suggest the following:

```suggestion
    }, [toolbox, event, handler]);
```

or 

```suggestion
    }, [event, handler]);
```

## Tests

To make sure the hook reacts to changes to the `event` or `handler`, it would be awesome to add some unit tests :slightly_smiling_face:

widgets/common/src/hooks/useEventListener.js
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from 'react';
+
+function useEventListener(toolbox, event, handler = toolbox.refresh) {
```
Gelio, line 3: How about using TypeScript for this new hook?

Gelio: I'm not sure `toolbox.refresh` is the best default value here. I don't think there is a good default value, actually :smile: I agree it's the most **common** one, but when I saw `useEventListener`, I would not immediately know that it would cause a refresh when the event occurred.

I encourage making the `handler` required

```suggestion
function useEventListener(toolbox, event, handler) {
```

Gelio (changes requested):
Nice cleanup with extracting `useEventListener` :+1: 

I'm putting _Request changes_ here to prevent preemptively merging the PR until we are sure there are no bugs/misleading APIs introduced here :slightly_smiling_face:

test/cypress/integration/widgets/labels_spec.ts
kubama: This would change test semantics and make it weaker. I want to make sure that after clicking trash icon not just any but the given row gets removed from the table. This is a noticeable difference IMO.

widgets/common/src/labels/LabelsModal.jsx
kubama: I agree, however I'm not sure this should be part of this PR.

widgets/common/src/hooks/useEventListener.js
kubama: Change applied.

> To make sure the hook reacts to changes to the event or handler, it would be awesome to add some unit tests 🙂

I'm not sure what you mean here, reacting on changes is an internal feature of `useEffect` hook, I don't think this is something we need to test.

widgets/common/src/hooks/useEventListener.js
kubama: > How about using TypeScript for this new hook?

I'm not sure how should we handle type for `toolbox`, I can see in `renderDeploymentRow.tsx` you are importing it directly from `StageApi`, but is this the way to go? I guess `StageApi` should be part of `Stage` to adhere to how we share things across `app` and `widgets`. Anyway, for me it seems this area is not settled yet, so I'd rather not mess with it at the moment.

kubama: > I'm not sure `toolbox.refresh` is the best default value here.

Actually, after revisiting how this hook is used it makes perfect sense to name it `useRefreshEvent`, what I just did. With such name it's no longer surprising that `toolbox.refresh` is the default handler.

test/cypress/integration/widgets/labels_spec.ts
Gelio: Ok, I see your point now

widgets/common/src/labels/LabelsModal.jsx
Gelio: It was not my intent :slightly_smiling_face: I wanted to leave a comment here to have something to refer to during the weekly discussion

widgets/common/src/hooks/useEventListener.js
Gelio: Thanks for applying the change :slightly_smiling_face: 

> I'm not sure what you mean here, reacting on changes is an internal feature of useEffect hook, I don't think this is something we need to test.

I was thinking of adding tests in general. One of the tests could make sure that the correct `handler` gets invoked after it is changed, but I guess we can get the PR merged without the test

widgets/common/src/hooks/useEventListener.js
Gelio: > Anyway, for me it seems this area is not settled yet, so I'd rather not mess with it at the moment.

Sure, that's fine :slightly_smiling_face: 

---

About `useRefreshEvent` - good idea :smile: the default value makes sense now :slightly_smiling_face:

widgets/common/src/hooks/useRefreshEvent.js
```diff
@@ -0,0 +1,14 @@
+import { useEffect } from 'react';
+
+function useRefreshEvent(toolbox, event, handler = toolbox.refresh) {
```
Gelio, line 3: I'm in two minds about this name. From one side, yeah, the default value now makes sense, due to the name of the function.

From the other side, there is nothing in this implementation that prevents consumers from using it with any event, and doing any action. Having a specific name here (related to refreshing) in my mind is more confusing if the hook is used in other contexts (in which it functionally works well).

Having thought about it, I would suggest keeping the old name, so as not to limit the usage contexts, and removing the default value, as mentioned before :slightly_smiling_face: 

```suggestion
function useEventListener(toolbox, event, handler) {
```

Gelio: Alternatively, we could do something to make sure this hook is not _hackably_ used in some other context, to listen to a non-refresh event

kubama: I'd say it's hard to agree. Here are my points:

1. Function name should reflect its implementation as close as possible, that's obvious. However, it should be the implementation that should drive the name, not the other way around.

2. After reviewing events usage in widgets I found out that almost all events handled are `[resourceName]:refresh`. There is only one exception which is `topology:selectNode` event. It would still make sense to use this event with hook named `useRefreshEvent` as we need to refresh widget state to reflect application state change. On the other hand, if we used this event with hook named `useEventListener` the `toolbox.refresh` default handler would still make sense for the same reason - we need to refresh widget state to reflect application state change.

Summing up I'm OK with both proposed names, but I can see no good reason for removing the default handler.

Gelio: Hmm... I can see how removing the default handler can be inconvenient. I still would not expect `toolbox.refresh` to be the default handler when seeing:

```ts
useEventListener(toolbox, 'abc:refresh');
```

How about exposing 2 hooks?

1. `useEventListener` without the default value, for general purposes
2. `useRefreshEvent` that would call `toolbox.refresh` automatically inside and not expose any `handler`

That could solve our problems with an unexpected default value, and with repeating `toolbox.refresh` multiple times

kubama: Makes perfect sense now, implemented.

Gelio: Great :tada: I like it!

Gelio (approved):
LGTM :+1: 

Thanks for updating all the readmes :smile: I'm hoping this PR will get in soon so we don't have to do the same in other PRs :slightly_smiling_face:

qooban (approved):

PR #1225, opened 2021-03-05
RD-1217 - Enhance LabelsInput component

qooban:
This PR contains multiple improvements/fixes to `LabelsInput` component.

* Add label duplication popup to Labels modal and Labels widget (RD-1231)
* Limit height of `LabelsList` component and add scroll to it, removed truncation (RD-779)
* Add label sorting by key first, then by value in `LabelsList` component
image

* Improve auto-focus for keyboard-only input scenario (focus on key dropdown once label is added) (RD-1274)

https://user-images.githubusercontent.com/5202105/110439117-5ba76680-80b7-11eb-8604-f9b6e5cdb4e4.mp4

* Improved fetching data on URL changes in `DynamicDropdown` component
  * Added `useUpdateEffect` hook to handle component's updates excluding a update triggered on component mount
  * Provide better user experience when working with DynamicDropdown (delete options only after they are fetched)
  * Stopped fetching all keys/values on load

https://user-images.githubusercontent.com/5202105/110439468-cd7fb000-80b7-11eb-8e9a-1250ddde4875.mp4

* Made Add button disabled when the label existence check request is in flight (RD-1267)
* Added cypress tests for `LabelsInput` component (RD-781)
* Fixed `resetInput` function in `useLabelsInput` hook

Stage system tests: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/241/ (all passed)


widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -11,19 +11,29 @@ const iconStyle = {
...
     const [labels, setLabels, resetLabels] = useResettableState(hideInitialLabels ? [] : initialLabels);
     const [open, toggleOpen] = useToggle();
     const [newLabelKey, setNewLabelKey, resetNewLabelKey] = useResettableState('');
     const [newLabelValue, setNewLabelValue, resetNewLabelValue] = useResettableState('');
+    const keyDropdownRef = useRef();
+
+    const newLabelIsProvided = !!newLabelKey && !!newLabelValue;
+    const newLabelIsAlreadyPresent = (function isNewLabelAlreadyPresent() {
+        const label = { key: newLabelKey, value: newLabelValue };
+        return !!(_.find(labels, label) || (hideInitialLabels && _.find(initialLabels, label)));
```
Gelio, line 33: Every time I read this line I need to think for a bit longer than is probably necessary :smile: This is because of how the condition is structured.

I have 2 suggestions here:
1. How about using `_.includes` instead of `_.find` here? We could then skip the `!!`
2. Optional: IMO it would be more readable to do 

    ```suggestion
            return _.includes([...labels, ...(hideInitialLabels ? initialLabels : [])], label);
    ```

widgets/labels/src/LabelsTable.jsx
```diff
@@ -51,6 +47,12 @@ export default function LabelsTable({ data, toolbox }) {
 
     const tdStyle = { textOverflow: 'ellipsis', overflow: 'hidden' };
 
+    const currentLabelAlreadyUsed = !!_.find(
```
Gelio, line 50: A bit shorter :slightly_smiling_face: 

```suggestion
    const currentLabelAlreadyUsed = _.includes(
```

widgets/common/src/labels/LabelsList.jsx
```diff
@@ -1,32 +1,28 @@
 export default function LabelsList({ labels, onChange }) {
-    const { Label, Icon, Popup } = Stage.Basic;
-    const maxLength = 20;
+    const { Label, Icon } = Stage.Basic;
     const newLabelColor = 'blue';
+    const sortedLabels = _.sortBy(labels, ['key', 'value']);
```
Gelio, line 4: Does it make sense to always sort the values? I'm not sure :smile: The reason is that as a user, after adding a new label, I would expect it to appear at the end, and not somewhere in the middle. Only after refreshing the data from the backend do I expect the labels to be sorted. When working with the labels, I would expect them to be in the order I added :slightly_smiling_face: 

Thus, I would suggest to only sort the labels when interacting with the API, and otherwise, keep them in the order they are :slightly_smiling_face: 

Other than that, I see sorting done in multiple places, and each time the exact sorting rules are defined explicitly. Having such duplication can yield unexpected differences in the behavior, where in one place we sort by key and value, and in other ones, only by key. I would encourage extracting that to a function in the future, to make sure the rules are the same :slightly_smiling_face:

widgets/common/src/hooks/useUpdateEffect.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Defines a function that is triggered each time component is updated.
+ * It is not triggered when the component mounts.
+ *
+ * @param onUpdate function to be triggered after each component's update
+ * @param dependencies dependencies list passed to React.useEffect
+ */
+function useUpdateEffect(onUpdate: () => void, dependencies: React.DependencyList) {
```
Gelio, line 8: Yay, TypeScript :smile: Nice work reusing the `React.DependencyList` type :+1:

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -53,4 +53,109 @@ describe('Deployment Action Buttons widget', () => {
...
+        function typeLabelKey(key) {
+            cy.get('div[name=labelKey] > input').clear().type(key);
+        }
+        function typeLabelValue(value) {
+            cy.get('div[name=labelValue] > input').clear().type(value);
+        }
+        function addLabel(key, value) {
+            typeLabelKey(key);
+            typeLabelValue(value);
+            cy.get('button .add').click();
```
Gelio, line 67: Optional:

Some aria labels to select the button via name would be awesome :slightly_smiling_face:

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -53,4 +53,109 @@ describe('Deployment Action Buttons widget', () => {
...
+        before(() => {
+            cy.setLabels(deploymentName, [{ existing_key: 'existing_value' }]);
+            cy.setDeploymentContext(deploymentName);
+            cy.interceptSp('GET', `/deployments/${deploymentName}?_include=labels`).as('fetchLabels');
+            cy.get('button.deploymentActionsButton').click();
+            cy.get('.popupMenu > .menu').contains('Manage Labels').click();
+            cy.get('.modal').within(() => {
+                cy.wait('@fetchLabels');
+                cy.get('form.loading').should('not.exist');
+                cy.get('.selection').click();
```
Gelio, line 79: Optional:

Is there a better way to select this element, e.g. by its label somehow :smile: ?

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -53,4 +53,109 @@ describe('Deployment Action Buttons widget', () => {
...
+            });
+        });
+
+        beforeEach(() => {
+            // NOTE: Clicking at the header to close opened dropdowns / popups
+            cy.get('.modal .tags').click();
```
Gelio, line 85: If we use classes anyway, let's at least make the classname a bit less confusing IMO :smile: 

```suggestion
            cy.get('.modal .header').click();
```

I did not verify that it works :slightly_smiling_face: but it's a bit more readable to me

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -33,15 +43,9 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
...
-        const label = { key: newLabelKey, value: newLabelValue };
-        return (
-            newLabelKey &&
-            newLabelValue &&
-            !_.find(labels, label) &&
-            (!hideInitialLabels || !_.find(initialLabels, label))
-        );
-    }
+    useEffect(() => {
+        resetNewLabelValue();
```
Gelio, line 47: This makes it so that in the following scenario:

1. User provides label key
2. User provides label value
3. User notices a typo in the key
4. User changes the label key

the label value is reset :smile: I would say this is not ideal :slightly_smiling_face: What's the use for this check?

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -33,12 +35,17 @@ function DynamicDropdown({
     const [hasMore, setHasMore] = useState(true);
     const [currentPage, setCurrentPage] = useState(-1);
     const [searchQuery, setSearchQuery] = useState('');
-    const [loaderVisible, setLoaderVisible] = useState(false);
+    const [shouldLoadMore, setShouldLoadMore] = useState(false);
     const [isLoading, setLoading] = useState(false);
+    const [overrideOptions, setOverrideOptions, resetOverrideOptions] = useBoolean();
```
Gelio, line 40: I don't exactly understand the necessity of this flag, so I would appreciate a bit of explanation :slightly_smiling_face:

widgets/common/src/labels/CommonDropdown.jsx
```diff
@@ -11,36 +11,25 @@ function useDebouncedSetValue(value, setValue, deps) {
...
-export default function CommonDropdown({ baseFetchUrl, onChange, toolbox, value, ...rest }) {
+export default function CommonDropdown({ innerRef, baseFetchUrl, onChange, toolbox, value, ...rest }) {
     const { useEffect, useState } = React;
     const {
         Common: { DynamicDropdown },
-        Hooks: { useLabelInput, useResettableState }
+        Hooks: { useLabelInput }
     } = Stage;
 
-    const [selectedValue, setSelectedValue, resetSelectedValue] = useResettableState('');
```
Gelio, line 21: Does this make it so that the selected value is not duplicated in multiple components?

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -88,10 +100,15 @@ function DynamicDropdown({
...
+        if (shouldLoadMore) {
             loadMore();
         }
-    }, [loaderVisible]);
+    }, [shouldLoadMore]);
+
+    useUpdateEffect(() => {
+        refreshData();
+        setShouldLoadMore(true);
+    }, [fetchUrl]);
```
Gelio, line 111: I'm not entirely sure we need this hook. What's the reason for using this hook, as opposed to using a regular `useEffect`?

Gelio (commented):
I was hoping for a bit of TS action, seeing TS changes in your previous PR, but it's fine too :cat: 

---

I see you've been cooking this PR for quite some time, and it looks to be a bit on the lengthier side :slightly_smiling_face: This can be confirmed by looking at the description, which lists multiple changes, and the fact that multiple tickets are addressed by this PR :smile:

Do you think it would be possible to split it into a few PRs? Or at least remember to try that in the future, since I've already reviewed it and breaking it into multiple PRs would break comments history :slightly_smiling_face: 

---

I like that we allow adding values via Enter :smile: We could do 1 thing that IMO is more UX-friendly, which is support adding values by pressing enter when either label key or label value input is focused. This is similar to filling out regular forms - you can press enter anytime to submit.

Actually, could we a regular form (HTML or Semantic UI's one) to capture this behavior? Then, we could listen for `submit` events, and not have to listen for enter via `onKeyPress`

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -53,4 +53,109 @@ describe('Deployment Action Buttons widget', () => {
...
+            typeLabelKey(key);
+            typeLabelValue(value);
+            cy.get('button .add').click();
+        }
+
+        before(() => {
+            cy.setLabels(deploymentName, [{ existing_key: 'existing_value' }]);
+            cy.setDeploymentContext(deploymentName);
+            cy.interceptSp('GET', `/deployments/${deploymentName}?_include=labels`).as('fetchLabels');
+            cy.get('button.deploymentActionsButton').click();
```
kubama, line 74: Is there a way for referencing this button by using Semantic's built in class name? Our custom class names are much less stable, and it's much more likely that it will somehow change in the future.

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -53,4 +53,109 @@ describe('Deployment Action Buttons widget', () => {
...
+            validateIfNotPermitted(' ');
+            validateIfNotPermitted('$');
+            validateIfNotPermitted('&');
+        });
+
+        it('allows to remove label from the list', () => {
+            cy.get('.modal').within(() => {
+                const existingLabel = 'existing_key existing_value';
+                cy.contains('a.label', existingLabel).should('be.visible');
+                cy.contains('a.label', existingLabel).within(() => cy.get('.delete').click());
```
kubama, line 146:
```suggestion
                cy.contains('a.label', existingLabel).find('.delete').click();
```

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -53,4 +53,109 @@ describe('Deployment Action Buttons widget', () => {
...
+
+        before(() => {
+            cy.setLabels(deploymentName, [{ existing_key: 'existing_value' }]);
+            cy.setDeploymentContext(deploymentName);
+            cy.interceptSp('GET', `/deployments/${deploymentName}?_include=labels`).as('fetchLabels');
+            cy.get('button.deploymentActionsButton').click();
+            cy.get('.popupMenu > .menu').contains('Manage Labels').click();
+            cy.get('.modal').within(() => {
+                cy.wait('@fetchLabels');
+                cy.get('form.loading').should('not.exist');
```
kubama, line 78: When checking if something doesn't exist, we should at some point (before or after) check if the same element exists by using the same selector. If we don't then such assertion is a bit flaky - it will be passing when the element doesn't exist, but also when the element exists but the selector used is not valid. But if the selector is invalid and we also check for element's existence then the issue will be caught. In other words the pattern should be:
```
cy.get(selector).should('not.exist')
...
cy.get(selector);
```
or
```
cy.get(selector);
...
cy.get(selector).should('not.exist')
```

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -11,19 +11,29 @@ const iconStyle = {
...
 
+    const [addingLabel, setAddingLabel, unsetAddingLabel] = useBoolean();
     const [labels, setLabels, resetLabels] = useResettableState(hideInitialLabels ? [] : initialLabels);
     const [open, toggleOpen] = useToggle();
     const [newLabelKey, setNewLabelKey, resetNewLabelKey] = useResettableState('');
     const [newLabelValue, setNewLabelValue, resetNewLabelValue] = useResettableState('');
+    const keyDropdownRef = useRef();
+
+    const newLabelIsProvided = !!newLabelKey && !!newLabelValue;
+    const newLabelIsAlreadyPresent = (function isNewLabelAlreadyPresent() {
```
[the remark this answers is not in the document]
qooban: Yep, changed it to an arrow function.

widgets/common/src/labels/LabelsInput.jsx
qooban: `_.includes` cannot handle objects properly (comparing references), that's why I didn't use it here.
But I implemented part of your suggestion.

widgets/labels/src/LabelsTable.jsx
qooban: See my previous comment about `_.includes`.

widgets/common/src/labels/LabelsList.jsx
qooban: Good point, extracted common function for sorting labels and reused it in our code.

I'm not sure about this one:
> The reason is that as a user, after adding a new label, I would expect it to appear at the end, and not somewhere in the middle.

Will need to think it over it make sense in all cases.

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Done.

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Didn't find a better way, unfortunately.

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Done.

widgets/common/src/DynamicDropdown.jsx
qooban: It make sense to execute the content of this effect only in case of real updates, and not on component mount (because on mount the `fetchUrl` is not changing, it is only being initialized). That is basically to maintain backward compatibility with DynamicDropdown's not using dynamic fetchUrl's as well as optimize number of fetches.

widgets/common/src/labels/LabelsList.jsx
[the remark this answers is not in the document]
qooban: Extracted common function for sorting labels and reused it.

widgets/common/src/labels/LabelsList.jsx
```diff
@@ -1,32 +1,28 @@
...
-                                <Icon
-                                    name="delete"
-                                    onClick={() => onChange(_.differenceBy(labels, [{ key, value }], { key, value }))}
-                                />
-                            </Label>
-                        </Popup.Trigger>
-                        <strong>{key}</strong> {value}
-                    </Popup>
+                    <Label
+                        key={`${key}_${value}`}
```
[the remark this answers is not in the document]
qooban: Good point, changed.

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -100,7 +116,17 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
                             />
                         </Form.Field>
                         <Form.Field width={2}>
-                            <AddButton onClick={onAddLabel} disabled={!isAddAllowed()} />
+                            <Popup
```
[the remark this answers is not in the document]
qooban: Make sense. Did some refactoring around popups:
* Extracted common `LabelErrorPopup` component and use it in both - existing `ValidationErrorPopup` and new `DuplicationErrorPopup` components
* Used `DuplicationErrorPopup` component in `LabelsInput` and `LabelsTable` components, attached to labels value dropdown

(common popup can be useful for RD-1606)

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Yep, changed it to use button labels.

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Changed.

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Added pre-check for that selector.

widgets/common/src/labels/LabelsInput.jsx
qooban: Removed it.

qooban:
@Gelio 

> I was hoping for a bit of TS action, seeing TS changes in your previous PR, but it's fine too cat

I did some small in new hook ;-) 

> I see you've been cooking this PR for quite some time, and it looks to be a bit on the lengthier side :slightly_smiling_face This can be confirmed by looking at the description, which lists multiple changes, and the fact that multiple tickets are addressed by this PR smile

> Do you think it would be possible to split it into a few PRs? Or at least remember to try that in the future, since I've already reviewed it and breaking it into multiple PRs would break comments history slightly_smiling_face

Yes, unfortunately. But multiple tickets are actually small sub-tasks, these are not individual issues (user stories, tasks).

I remember that rule we discussed recently, so I'm definitely going to follow it. 
This time - as you both reviewed it already and as I started working on that before the Friday's agreement - I think we can have it like that.

> I like that we allow adding values via Enter smile We could do 1 thing that IMO is more UX-friendly, which is support adding values by pressing enter when either label key or label value input is focused. This is similar to filling out regular forms - you can press enter anytime to submit.

> Actually, could we a regular form (HTML or Semantic UI's one) to capture this behavior? Then, we could listen for submit events, and not have to listen for enter via onKeyPress

Got your point, not sure it will work in all cases here, but AFAIR we have even an issue in JIRA for that, need to find it.

widgets/common/src/labels/CommonDropdown.jsx
qooban: `selectedValue` was actually not necessary in `CommonDropdown` because it was handled in the parent component.

widgets/common/src/labels/LabelsInput.jsx
Gelio: Good to know :slightly_smiling_face: I assumed `_.includes` and `_.find` work similarly

Thanks :+1:

widgets/common/src/labels/LabelsList.jsx
Gelio: > Will need to think it over it make sense in all cases.

What's your suggestion here :slightly_smiling_face: ? Will you talk about it with Ofer?

test/cypress/integration/widgets/deployment_actions_spec.js
Gelio: Welp, ok :slightly_smiling_face:

widgets/common/src/labels/LabelsList.jsx
qooban: Yes, I'm planning to do Labels leftovers review tomorrow with Ofer.

widgets/common/src/labels/CommonDropdown.jsx
Gelio: Thanks for the info :slightly_smiling_face:

Gelio:
@qooban 
> I remember that rule we discussed recently, so I'm definitely going to follow it.
Sure :slightly_smiling_face: 

> Got your point, not sure it will work in all cases here, but AFAIR we have even an issue in JIRA for that, need to find it.

By this you mean deferring the change to that other ticket, right? I'm fine with that :slightly_smiling_face: as long as there is something that tracks it :smile: 

I'll rereview the PR in 1h

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -56,14 +58,21 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
...
             setLabels(newLabels);
             resetNewLabelKey();
             resetNewLabelValue();
+            unsetAddingLabel();
         });
     }
 
+    function onEnterPressOnAddButton() {
+        onAddLabel();
+        keyDropdownRef.current.click();
```
Gelio, line 73: Optional:

If I'm not mistaken, this makes it so that the dropdown is focused before the input is cleared https://github.com/cloudify-cosmo/cloudify-stage/blob/d4b016bc3c46ca547733c3f8fdf53a5d49d3661e/widgets/common/src/labels/LabelsInput.jsx#L65-L66

I suggest waiting making `onAddLabel` return a `Promise`, and waiting for it to resolve before executing `click` :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -33,12 +35,17 @@ function DynamicDropdown({
     const [hasMore, setHasMore] = useState(true);
     const [currentPage, setCurrentPage] = useState(-1);
     const [searchQuery, setSearchQuery] = useState('');
-    const [loaderVisible, setLoaderVisible] = useState(false);
+    const [shouldLoadMore, setShouldLoadMore] = useState(false);
```
Gelio, line 38:
```suggestion
    const [shouldLoadMore, setShouldLoadMore] = useState(!!prefetch);
```

would probably allow us to remove https://github.com/cloudify-cosmo/cloudify-stage/blob/6d2ed65d065b058601fcd6c1f680a9a46f99b0dc/widgets/common/src/DynamicDropdown.jsx#L84-L86 and slightly reduce the complexity of this component :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
Gelio: Thanks for an explanation :slightly_smiling_face: 

1. I think `appendToOptions` or even `appendToOptionsDuringFetch` would provide more context and be easier to comprehend
2. I encourage leaving some comments behind to explain the purpose of this flag :smile: See https://github.com/google/eng-practices/blob/master/review/reviewer/comments.md#accepting-explanations-explanations

widgets/common/src/DynamicDropdown.jsx
qooban: Good idea. Changed. 

WIth a small change, as I think in case of boolean value we don't need `!!`. Or am I missing something?

widgets/common/src/DynamicDropdown.jsx
qooban: > I think `appendToOptions` or even `appendToOptionsDuringFetch` would provide more context and be easier to comprehend

Appending is default behavior in most use cases for this component, so I think `overrideOptions` is better. I can consider changing it to be `overrideOptionsAfterFetch`. WDYT? I can add comment as well.

> I still cannot grasp how the data fetching happens in this component. It seems quite complex slightly_smiling_face Maybe we could rewrite it somehow

It's quite complex, that is for sure. But definitely this PR is not aiming to handle it 🙂

widgets/common/src/DynamicDropdown.jsx
Gelio: Yup, you're right :slightly_smiling_face: I thought `prefetch` can be undefined, but it has a default value

Looks good :+1:

widgets/common/src/DynamicDropdown.jsx
Gelio: I agree with both points :slightly_smiling_face: `overrideOptionsAfterFetch` would be a tad better, so let's do it and resolve the thread :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
Gelio: Welp, welp, I got some parts, but did not see the overall picture :slightly_smiling_face: I'll not pour more time over this component, as it looks to be working :smile:

widgets/common/src/DynamicDropdown.jsx
qooban: > IMO the logic related to this flag is scattered across the file. For me, this is hard to read, as I have to jump across the file and think how each piece influences other ones slightly_smiling_face

Hmm... There is single line for flag definition (`useState`), there is actual use in `loadMore` function and there is setting it in `refreshData` function. I don't think it is scattered across the file more than usual state variable.

qooban: Done.

qooban:
@Gelio 
> By this you mean deferring the change to that other ticket, right? I'm fine with that slightly_smiling_face as long as there is something that tracks it smile

There is a ticket to handle Enter key press properly. I think it is enough for now: [CYBL-133](https://cloudifysource.atlassian.net/browse/CYBL-133).

In case of labels input:
* I would not expect that Enter key press on `KeyDropdown` input will add the label (as Enter can be used to select the key from the list in the dropdown)
* I could expect that Enter key press on `ValueDropdown` input will add the label (right now user would need to press Tab to get to Add button)
I will take that with Ofer tomorrow.

qooban:
After code review updates triggered Stage system tests again. Results will be here: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/246/

Gelio (dismissed):
Looks good, if you asked me :slightly_smiling_face:

widgets/common/src/DeploymentActions.js
```diff
@@ -7,6 +7,10 @@ export default class DeploymentActions {
         this.toolbox = toolbox;
     }
 
+    static sortLabels(labels) {
+        return _.sortBy(labels, ['key', 'value']);
```
kubama, line 11:
```suggestion
        return _.sortBy(labels, 'key', 'value');
```

widgets/common/src/labels/DuplicationErrorPopup.jsx
```diff
@@ -0,0 +1,11 @@
+import LabelErrorPopup from './LabelErrorPopup';
+
+export default function DuplicationErrorPopup({ open }) {
+    const { i18n } = Stage;
+
+    return <LabelErrorPopup open={open} content={i18n.t('widgets.common.labels.labelDuplicationError')} />;
+}
```
kubama, line 7: Just wondering, I know we are using this pattern consistently, but why use the `open` prop when we can use conditional component rendering? The benefit I can think of is that toggling `open` prop keeps component state, which can be useful in some cases and troublesome in others, but in case of stateless component like this it's not an argument. What do you think?

widgets/common/src/DeploymentActions.js
```diff
@@ -7,6 +7,10 @@ export default class DeploymentActions {
...
+    }
+
```
kubama, line 13: I'm not sure if this is the best place to put this function. The `*Actions` classes contain methods related to manager requests, while this function is purely related to data presentation. How about placing it in `Stage.Common.Labels`?

widgets/common/src/DeploymentActions.js
qooban: Done.

widgets/common/src/DeploymentActions.js
qooban: Make sense, good point, moved to `Stage.Common.Labels`.

widgets/common/src/labels/DuplicationErrorPopup.jsx
qooban: Nice idea, removed `open` prop from `*ErrorPopup` components and used conditional rendering. Looks like it works equally good.

qooban:
Triggered Stage system tests again. Results will be here: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/248/

BTW Previous job failed due to environment issues, DevOps already notified.

widgets/common/src/labels/common.js
```diff
@@ -3,3 +3,7 @@ export function addSearchToUrl(url, search) {
     const { appendQueryParam } = Stage.Utils.Url;
     return search ? appendQueryParam(url, '_search', search) : url;
 }
+
+export function sortLabels(labels) {
+    return _.sortBy(labels, 'key', 'value');
+}
```
kubama, line 9: I think comment in L1 can be removed after adding this function.

qooban: Removed.

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -74,7 +74,6 @@ describe('Deployment Action Buttons widget', () => {
             cy.contains('button', 'Deployment actions').click();
             cy.get('.popupMenu > .menu').contains('Manage Labels').click();
             cy.get('.modal').within(() => {
-                cy.get('form.loading').should('be.visible');
```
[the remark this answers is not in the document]
qooban: Yes, I found that the change I provided for https://github.com/cloudify-cosmo/cloudify-stage/pull/1225#discussion_r590184957 did not reduce flakiness, but it introduce more flakiness :-) See last test results. So I decided to remove the added line (see: https://github.com/cloudify-cosmo/cloudify-stage/pull/1225/commits/fde2780d6e8a57fe47ff884e09359e2a7c387721) 

I think that my original solution is perfectly OK as it:
* first wait for the labels fetching to finish (`cy.wait('@fetchLabels');`) 
* then expect that loader is not visible anymore (`cy.get('form.loading').should('not.exist');`)
I don't think in that case we can expect any flakiness. During the fetch (if it is long enough) loader can be visible, but don't have to. After the fetch loader should not be visible. Do you see still any flakiness here?

qooban:
@Gelio @kubama Final test results (only for one failing test) will be here: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/250/console . 

Once they are fine, I'd like to finalize this PR as soon as possilbe, meaning if you have no more comments, merge it.

test/cypress/integration/widgets/deployment_actions_spec.js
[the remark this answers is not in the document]
kubama: > During the fetch (if it is long enough) loader can be visible, but don't have to.

If the fetch is done asynchronously then the loader should always become visible, right? I suspect it's just the test runner that starts the `visible` assertion after the loader is gone.

> Do you see still any flakiness here?

Yes, I do. My initial comment is still valid. However I understand it is not straightforward to address due to timing issue - if so lets leave it.

kubama (dismissed):

qooban:
I had to push one more commit (https://github.com/cloudify-cosmo/cloudify-stage/pull/1225/commits/b2e089b15b087ede941afd6bbc75dd97e8a97141) due to changes introduced by https://github.com/cloudify-cosmo/cloudify-stage/pull/1227

Gelio:
I'm beginning to review the latest changes

widgets/common/src/labels/DuplicationErrorPopup.jsx
Gelio: Just FYI, AFAIK, the `open` prop is useful not only for preserving state, but also for exit animations. When the component is unmounted suddenly, it cannot animate its exit in any way :slightly_smiling_face: If it is mounted, but the `open` prop is toggled, then the component can stay in the DOM and determine how to disappear

This is a general comment. I don't think it matters here, but just something to consider :slightly_smiling_face:

widgets/common/src/hooks/useUpdateEffect.ts
```diff
@@ -0,0 +1,27 @@
...
+    useEffect(() => {
+        if (isInitialMount.current) {
+            isInitialMount.current = false;
+        } else {
+            onUpdate();
+        }
+    }, dependencies);
+}
+
+declare namespace Stage {
```
Gelio, line 21: YES! :tada:

kubama (approved):

Gelio (approved):
Reads well from my perspective :slightly_smiling_face:

widgets/common/src/labels/DuplicationErrorPopup.jsx
qooban: Good point for the future. Thx.

PR #1226, opened 2021-03-05
RD-1409 Transform Stage into a namespace

Gelio:
This PR transforms `Stage` (exposed on `window`) into a namespace. This allows using declaration merging to dynamically enhance the types of `Stage` based on what's defined (e.g. using `Stage.defineCommon`).

In this PR I have also changed the way reusable TS types (only `WithOptionalProperties` at the moment) are exposed to the application. Before, they were accessible only by importing them, which was not ideal, as we had to import it from `app` when in `widgets` (risking importing actual code into `widgets`). Now, those types are accessible via `Stage.Types`

There will be a separate PR for RD-1408 that builds upon this one


app/utils/widgetDefinitionsLoader.ts
```diff
@@ -68,7 +68,7 @@ export default class WidgetDefinitionsLoader {
 
             Hooks,
             defineHook: def => {
-                window.Stage.Hooks = { ...window.Stage.Hooks, ...def };
+                Object.assign(window.Stage.Hooks, def);
```
Gelio, line 71: You can consider this a breaking change, since the `window.Stage.Hooks` object will not be created again, but the expected result should be the same

app/utils/types.ts
```diff
@@ -1,2 +1,9 @@
...
-    Partial<Pick<T, OptionalProperties>>;
+declare global {
+    // eslint-disable-next-line @typescript-eslint/no-namespace
+    namespace Stage.Types {
+        type WithOptionalProperties<T, OptionalProperties extends keyof T> = Omit<T, OptionalProperties> &
+            Partial<Pick<T, OptionalProperties>>;
+    }
+}
+
+export {};
```
Gelio, line 9: The export is necessary since we are using `isolatedModules` and each file has to be a module (with ES imports/exports)

app/utils/widgetDefinitionsLoader.ts
kubama: So why it was needed to change it?

test/jest/widgets/common/DeploymentActions.test.ts
```diff
@@ -5,6 +5,7 @@ describe('(Widgets common) DeploymentActions', () => {
     const doGetExecutions = jest.fn();
 
     beforeEach(() => {
+        // @ts-expect-error Necessary when overriding
```
kubama, line 8: I think it can be quite common to disable this rule in jest tests when creating mocks as it is done here - how about disabling this rule globally for `test/jest` dir?

kubama (commented):
Since this PR now ads `Stage.Types` maybe we could add `Toolbox` type there and remove import that is present in `renderDeploymentRow.tsx`?

app/utils/StageAPI.ts
```diff
@@ -174,24 +174,33 @@ export type InitialWidgetDefinition<Params, Data, Configuration> = WithOptionalP
...
+        const Common: Record<string, unknown>;
+        const defineCommon: (definition: CommonOrPropTypeDefinition) => void;
+
+        const PropTypes: typeof StagePropTypes & Record<string, any>;
+        const definePropType: (definition: CommonOrPropTypeDefinition) => void;
+
+        const Hooks: typeof StageHooks & Record<string, any>;
+        const defineHook: (definition: any) => void;
+
+        const i18n: typeof import('i18next').default;
```
qooban, line 198: Just wondering - why moving the import here? How it works with type imports? 
Is it dynamic right now (type is fetched during runtime)?

qooban (commented):
Added one comment.

app/utils/widgetDefinitionsLoader.ts
Gelio: Due to this

image

test/jest/widgets/common/DeploymentActions.test.ts
Gelio: Hmm... You would still need `ts-expect-error` comments. This is because those are TSC errors, not ESLint ones.

The only thing we could do is disable the rule that says that the explanation is necessary, but IMO that's useful :slightly_smiling_face:

kubama: So we can't configure tsc to ignore this one specific error type?

app/utils/StageAPI.ts
Gelio: If `i18n` was imported in the module scope, then there was an error with `const i18n` reusing the same name for a variable. This is similar to if we did:

```ts
import i18n from 'i18next';

const i18n = 1;
```

Hence I moved it to this syntax of the import. 

Every type in TypeScript is resolved at compile time, and is erased. This only looks like a dynamic import :slightly_smiling_face: This is used when you don't want the import to have a variable name (like with `import i18n from 'i18next';`, it declares the `i18n` symbol) :slightly_smiling_face:

Gelio:
> Since this PR now ads Stage.Types maybe we could add Toolbox type there and remove import that is present in renderDeploymentRow.tsx?

Awesome idea! :tada: 

I will add a couple of types into the `Stage.Types` (mainly those exposed in `StageAPI.ts`)

test/jest/widgets/common/DeploymentActions.test.ts
Gelio: Not that I am aware of. Alternatively, we could do `as any`, but that's more obscure IMO

Gelio:
After @kubama's suggestion, I have exposed some types in `Stage.Types` to avoid importing `app` from `widgets` :tada:

qooban (approved):

app/utils/StageAPI.ts
```diff
@@ -18,9 +18,9 @@ import type External from './External';
 import './types';
```
kubama, line 18: My general comment to this file is that it introduces an awkward types naming scheme. For example, `toolbox` type is `StageToolbox` - I believe type name should be a more general term than an instance name, so in this case it would be more intuitive if it was the other way around - `stageToolbox` was an instance of type `Toolbox`. Do you think it would be possible to refactor it?

Gelio: I'm not sure I understand the part about `stageToolbox` being an instance of `Toolbox`. Do you suggest here renaming the parameter names to `stageToolbox` (and changing the type name)? If yes, then this part (changing the parameter names is doable without any problem).

When it comes to naming types, I changed the type names in the module scope to `Stage*` to be able to reference those in the `Stage.Types` namespace. If I did

```ts
type Toolbox = ...;

namespace Stage.Types {
  type Toolbox = Toolbox;
}
```

then `Stage.Types.Toolbox` would be referring to itself, not to the type from the parent scope. I did not find a way to export a type from a parent scope with the same name. Hence the name change.

Those `Stage*` names are only local to this file. When using the type (either by importing from `StageAPI.ts` or by using `Stage.Types.X`), you use the name without the `Stage` prefix, e.g. `Stage.Types.Toolbox`.

What I could do is register those `Stage.Types` types in a separate file (e.g. in `stage-api.d.ts`). Then, we would stay with the `Toolbox` name everywhere. I'm not sure it's that much of a benefit though :slightly_smiling_face: 

Let me know if you'd rather keep the name as `Toolbox` still. @qooban you can also weigh in :smile:

kubama: > I'm not sure I understand the part about `stageToolbox` being an instance of `Toolbox`. Do you suggest here renaming the parameter names to `stageToolbox` (and changing the type name)?

Actually it was just an example to show what I mean by saying that type names should use more general terms than instance names.

However, if this is local to this file then I think I'm fine with it.

kubama (approved):

app/utils/StageAPI.ts
Gelio: Oh ok, sure :slightly_smiling_face:

PR #1227, opened 2021-03-08
RD-1409 Allow strict typing for Stage.Common, Hooks, PropTypes

Gelio:
This PR builds upon #1226 and allows dynamically registering items into `Stage.Common`, `Stage.Hooks`, or `Stage.PropTypes`, all in a way that allows strict type checking in TS files, and finding references, going to declaration etc in IDEs (in both TS and JS files).

Example IDE experience:

https://user-images.githubusercontent.com/889383/110321937-a9b65e80-8012-11eb-9df3-e00bd92c0060.mp4

Unfortunately, not all file migrations were caught as renames by git :slightly_frowning_face:


Gelio:
@gregcfy I've added you here so you can take a look at what will be possible :slightly_smiling_face: Your review is optional

app/utils/widgetDefinitionsLoader.ts
```diff
@@ -56,6 +56,8 @@ export default class WidgetDefinitionsLoader {
             GenericConfig,
             Utils: StageUtils,
 
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore Surpress error related to missing items when compiling the widgets project
```
Gelio, line 60: The problem here was that when compiling the `widgets` project, `Stage.Common` has items defined, and then tsc is explaining that they are not specified here.

It is fine when compiling the `app` project, though, since there are no items in `Stage.Common`

Gelio:
I've rebased the branch on `master` to resolve merge conflicts

Gelio:
UI system tests results for this branch will be here: https://jenkins.cloudify.co/job/Stage-UI-System-Test/240/

app/utils/hooks/useErrors.ts
```diff
@@ -3,15 +3,13 @@ import { useState } from 'react';
...
- * @returns {{clearErrors: (function(): void), setErrors: React.Dispatch<React.SetStateAction<{}>>, errors: {}, setMessageAsError: (function(*): void)}}
  */
 export default function useErrors() {
-    const [errors, setErrors] = useState({});
+    const [errors, setErrors] = useState({} as Record<string, any>);
 
     return {
         errors,
-        setMessageAsError: err => setErrors({ errors: err.message }),
+        setMessageAsError: (err: Error) => setErrors({ errors: err.message }),
```
qooban, line 12: I'm not sure this is always the case (have argument with type `Error`) when we get error message from the Cloudify API.

widgets/common/src/DeploymentStates.ts
```diff
@@ -61,6 +61,14 @@ export default class DeploymentStates {
     }
 }
 
+declare global {
+    namespace Stage {
+        interface Common {
+            DeploymentStates: typeof DeploymentStates;
+        }
+    }
+}
```
qooban, line 70: I assume it's not possible to have a function for such kind of declarations as they must be provided in a global scope and it is not possible to declare anything dynamically, right?

The context of my question is that we will be adding such 7 lines declarations to all files. And if it would be possible to avoid it, then great. I suspect it is not possible, so just asking to be sure.

Gelio: I'm not aware of any way to make this shorter :frowning_face: This is the cost we pay for not using ES modules, but attaching stuff in the global object. Adding such properties has to be done statically, and quite explicitly, as you can see.

We could make it shorter if we grouped all the `Common`s into a single object (as we do with Cypress commands in each file), and then have a single line that extends the `Common` interface with `typeof theVariableThatHoldsItemsToAttach`.

This will be possible after RD-1609. For now, since we have individual files that define commons, I don't see a better way :slightly_frowning_face:

qooban: Got it, clear.

app/utils/hooks/useErrors.ts
qooban: Thanks.

qooban (approved):

app/utils/widgetDefinitionsLoader.ts
```diff
@@ -56,6 +56,8 @@ export default class WidgetDefinitionsLoader {
             GenericConfig,
             Utils: StageUtils,
 
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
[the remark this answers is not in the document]
Gelio: Good point, I like the direction :+1: 

I have taken a look at [the docs for the rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-ts-comment.md) and I'm opting for a similar, but different solution.

Instead of disabling this rule, I would encourage changing the options so that both `ts-ignore` and `ts-expect-error` have `allow-with-description` set. This means that we would always have to provide an explanation (which I advocate for and have been doing :slightly_smiling_face: ) rather than allowing just disabling TS without any notes.

Other options would stay as is (`ts-nocheck` would be disallowed).

Since it's a wider topic, I suggest taking this to the UI weekly and not blocking this PR :slightly_smiling_face: I have added the topic for Friday. Let me know if that suits you

widgets/common/src/props/Labels.ts
```diff
@@ -7,5 +15,5 @@ Stage.definePropType({
 
 Stage.definePropType({
     name: 'Labels',
-    common: PropTypes.arrayOf(LabelPropType)
+    common: LabelsPropType
 });
```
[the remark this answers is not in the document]
Gelio: I'm all for it :slightly_smiling_face: that would reduce some boilerplate since usually the `name` matches the name of the value passed to `common`.

I will add it to the UI weekly meeting's agenda :slightly_smiling_face: Since I have 4 topics for the next weekly, feel free to take over this topic (change the initials in the agenda :slightly_smiling_face: )

app/utils/widgetDefinitionsLoader.ts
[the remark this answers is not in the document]
kubama: Yes it suits me, thanks.

kubama (approved):

widgets/common/src/props/Labels.ts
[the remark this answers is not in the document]
kubama: Actually on last meeting we discussed the idea of introducing a central `index.ts` for defining Stage.[Common|PropTypes|Hooks] and we agreed to go this direction. I think my proposal here is in fact in scope of this bigger change, so I guess there is no need to discuss it separately.

PR #1228, opened 2021-03-08
RD-757 Added labels input to deployment modal

kubama:
image


Gelio:
Yay for TS :smile: 

btw. git did not pick up those migrations to TS as renames. When I was doing similar things, it helped to do the following as separate commits
1. Change the file extension
2. Make the changes
then, when viewing by commits, we could see the changes to the file as a separate commit, which could reduce the diff :slightly_smiling_face: You could try that next time

However, even if you do that, the full diff (e.g. the one displayed in GitHub for the whole PR) will not catch those renames :smile: git's weird this way

widgets/common/src/BlueprintActions.js
```diff
@@ -58,6 +58,7 @@ export default class BlueprintActions {
         deploymentId,
         inputs,
         visibility,
+        labels,
         skipPluginsValidation = false,
         siteName = null,
         runtimeOnlyEvaluation = false
```
Gelio, line 64: How about a small cleanup :smile: ? Instead of accepting 8 formal parameters, this function could accept an options object. This would provide a bit more context seeing how values map to parameters when calling this function :slightly_smiling_face: 

```ts
doDeploy({
  blueprint: a,
  deploymentId: b,
  inputs: c,
  visibility: d
})
```

instead of

```ts
doDeploy(a, b, c, d)
```

widgets/common/src/BlueprintActions.js
```diff
@@ -66,6 +67,7 @@ export default class BlueprintActions {
             blueprint_id: blueprint.id,
             inputs,
             visibility,
+            labels: _.map(labels, label => ({ [label.key]: label.value })),
```
Gelio, line 70: Is there a benefit of using lodash here?

widgets/common/src/DeployBlueprintModal.jsx
```diff
@@ -262,6 +262,7 @@ class DeployBlueprintModal extends React.Component {
             blueprint,
             deploymentName,
             runtimeOnlyEvaluation,
+            labels,
```
Gelio, line 265: At the moment, `labels` can be `undefined`. In my opinion, it would be easier to reason about if `labels` were always an array (initialized to `[]`). Then, you would not have to think whether `labels` have already been changed to an empty array, or have not been set and are `undefined`. You would always know that they are a valid array, that could be empty if it hasn't been touched

Thus, I suggest declaring it as `[]` in the `initialState` static :slightly_smiling_face:

widgets/common/src/BlueprintActions.js
```diff
@@ -58,6 +58,7 @@ export default class BlueprintActions {
         deploymentId,
         inputs,
         visibility,
+        labels,
```
Gelio, line 61: `labels` are not passed in in `ConfirmationStep`. I would be able to understand it easier if there was a default value, rather than relying on this function handling both `undefined` and a valid array :slightly_smiling_face: 

```suggestion
        labels = [],
```

widgets/common/src/BlueprintActions.js
kubama: Done

widgets/common/src/BlueprintActions.js
kubama: After introducing default value for `labels` - no, removed

widgets/common/src/BlueprintActions.js
kubama: Done

widgets/common/src/DeployBlueprintModal.jsx
kubama: > Then, you would not have to think whether `labels` have already been changed to an empty array, or have not been set and are `undefined`. You would always know that they are a valid array, that could be empty if it hasn't been touched

Why anybody would want to think about it 🙂? `doDeploy` handles nil values for `labels` correctly as it can be called without them, so it really doesn't matter.

test/cypress/support/editMode.ts
```diff
@@ -0,0 +1,52 @@
+import { addCommands, GetCypressChainableFromCommands } from 'cloudify-ui-common/cypress/support';
```
qooban, line 1: Is that JS->TS transformation related to the RD-757? (the same question for `plugins.ts`)

Have you tried the solution suggested by @Gelio to make changes (extension, content) in separate PRs? That way GitHub is more likely to provide real diff instead of seeing the change as file removal and new file addition.

widgets/common/src/BlueprintActions.js
```diff
@@ -53,19 +53,21 @@ export default class BlueprintActions {
...
         siteName = null,
         runtimeOnlyEvaluation = false
-    ) {
+    }) {
         const data = {
-            blueprint_id: blueprint.id,
+            blueprint_id: blueprintId,
             inputs,
             visibility,
+            labels: labels.map(label => ({ [label.key]: label.value })),
```
qooban, line 70: We have similar mapping in `DeploymentActions.doSetLabels`. 
What do you think about extracting common function for this?

qooban (commented):
Looks good. Added two questions.

Gelio (approved):
Looks good :+1: Thanks for applying the changes

I did not test this in the UI

widgets/common/src/BlueprintActions.js
kubama: Good point - extracted.

test/cypress/support/editMode.ts
kubama: I admit I didn't at this time. I'll try to remember about it in the future.

qooban (approved):
LGTM 👍🏻

PR #1229, opened 2021-03-08
RD-1629 Migrate pages reducer to TS

Gelio:
Another TS PR, this time migrating the pages reducer to TS.

I've decided to do that right away since parts of the code that I will be modifying later in the Environments feature will require `PageDefinition` and other similar interfaces, so I thought it would make sense to extract them from `Page` component to the reducer, since that's closer to where it is defined.

Very few functional changes here, I will mark them with comments :slightly_smiling_face:


app/actions/page.ts
```diff
@@ -171,18 +224,18 @@ export function addPage(name) {
...
 }
 
-export function removePage(page) {
+export function removePage(page: PageDefinition): ThunkAction<void, ReduxState, never, AnyAction> {
     return (dispatch, getState) => {
         const pagesMap = createPagesMap(getState().pages);
-        const removePageWithChildren = p => {
-            if (!_.isEmpty(p.children)) {
+        const removePageWithChildren = (p: PageDefinition) => {
+            if (p.children && !_.isEmpty(p.children)) {
```
Gelio, line 238: `isEmpty` does not narrow the type good enough :slightly_smiling_face:

app/reducers/pageReducer.ts
```diff
@@ -1,21 +1,36 @@
...
+    | ReturnType<typeof updateTab>
+    | ReturnType<typeof moveTab>;
+
+const tabs: Reducer<TabContent[], TabsAction> = (state = [], action) => {
     switch (action.type) {
         case types.ADD_TAB:
             return [...state, { name: i18n.t('editMode.tabs.newTab'), widgets: [] }];
         case types.REMOVE_TAB:
-            return _.without(state, _.nth(state, action.tabIndex));
+            return _.without(state, state[action.tabIndex]);
```
Gelio, line 33: `nth` has  `undefined` in its return type, and array indexing does not :slightly_smiling_face: `state` is always defined (has a default value), so this should be safe

app/reducers/pageReducer.ts
```diff
@@ -124,11 +132,11 @@ const page = (state = {}, action) => {
     }
 };
 
-const pages = (state = [], action) => {
+const pages: Reducer<PageDefinition[]> = (state = [], action) => {
     switch (action.type) {
         case types.ADD_PAGE:
         case types.CREATE_DRILLDOWN_PAGE:
-            return [...state, page(undefined, action)];
+            return [...state, createPage(action as any)];
```
Gelio, line 139: Not calling the reducer with `undefined` allows for fewer `as any` assertions or other workarounds in the `page` reducer, since it can assume that it will always get a valid `PageDefinition` as a parameter

Gelio:
UI system tests results will be in https://jenkins.cloudify.co/job/Stage-UI-System-Test/242/

kubama (dismissed):

app/actions/page.ts
```diff
@@ -248,31 +301,34 @@ export function createPagesFromTemplate() {
...
 
+        // @ts-expect-error Missing type definitions for app reducer
         const pageId = state.app.currentPageId || state.pages[0].id;
 
         const page = _.find(state.pages, { id: pageId });
         if (page && page.parent) {
-            const parentPage = _.find(state.pages, { id: page.parent });
+            // NOTE: assume page is always found
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const parentPage: PageDefinition = _.find(state.pages, { id: page.parent })!;
```
qooban, line 331: When `page.parent` is set, I believe it is very unlikely that the page will not be found.

TS-beginner questions: post-fix exclamation mark is to provide non-null (not `null` and not `undefined`) value, right? 
So what kind of `PageDefinition` will be assigned to `parentPage` in case `_.find` result is `undefined`?

app/components/Page.tsx
```diff
@@ -115,19 +118,6 @@ class Page extends Component<PageProps, never> {
 }
 
 // NOTE: these should be extracted to appropriate reducers when those are migrated to TS
```
qooban, line 120: Probably comment could be removed now?

qooban (commented):
Looks good in general. Added some questions.

app/actions/page.ts
```diff
@@ -102,25 +150,30 @@ function createPageId(name, pages) {
...
         pageId,
         description: newDescription
     };
 }
 
-export function selectPage(pageId, isDrilldown, drilldownContext, drilldownPageName) {
+export function selectPage(
+    pageId: string,
+    isDrilldown?: boolean,
+    drilldownContext?: any,
```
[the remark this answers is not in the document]
Gelio: To be frank, I did not spend much time for types that are not obvious :slightly_smiling_face: Here, I was not sure if it's the `drilldownContext` from `toolbox`, or from some other place :smile:

I was already fooled by variable `layoutSection` that I thought was an object, but after further investigation, turned out to be the index of the section (meaning of type `number`). I would rather add very general type annotations (such as `any`) rather than risk lying and saying that the type is different then the values that are passed in reality :smile: We can start with `any` and make it stricter later

So, summing up both points, I did not spend much time investigating each parameter, as I have spent a significant (IMO) time on TS during RD-1223, and I should be progressing with the actual functionality, and then, to avoid risking lies in the type system, I used `any` :slightly_smiling_face: 

The use of `any` is not related to any fallback values :slightly_smiling_face:

app/actions/page.ts
Gelio: > When page.parent is set, I believe it is very unlikely that the page will not be found.

Makes sense :slightly_smiling_face: 

>  post-fix exclamation mark is to provide non-null (not null and not undefined) value, right?

Yes, that's correct :slightly_smiling_face: However, this is only in the type system - the exclamation mark will be erased by the TS parser

> So what kind of PageDefinition will be assigned to parentPage in case _.find result is undefined?

In short, `parentPage` will be `undefined`

The exclamation mark does not influence the JavaScript code in any way. If the result of `_.find` is `undefined`, then `parentPage === undefined` :slightly_smiling_face: And since the code later assumes that `parentPage` is a valid `PageDefinition` (and not `PageDefinition | undefined`), there is nothing that stops this code from resulting in a runtime TypeError.

This is why the `no-non-null-assertion` rule exists, as using `!` willy-nilly results in less confidence about the possible values in the program

app/components/Page.tsx
Gelio: I considered that, but I believe `DrilldownContext` should belong to `app/reducers/drilldownContextReducer.js`, which I did not want to convert to TS in this PR :smile: So I decided to let it be for now

However, since `app/reducers/drilldownContextReducer.js` looks small, I can convert it to TS and remove this comment then. Let me know which option you'd prefer (I don't have a preference here)

app/actions/page.ts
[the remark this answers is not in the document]
qooban: Got it, clear 👍🏻

app/components/Page.tsx
qooban: Yep, looking at how small `drilldownContextReducer.js` is, I think you can do it within this PR.

app/actions/page.ts
qooban: I see, OK, fine.

app/components/Page.tsx
Gelio: Done :slightly_smiling_face:

Gelio:
I have added 1 more commit to resolve https://github.com/cloudify-cosmo/cloudify-stage/pull/1229#discussion_r590356146 :slightly_smiling_face: Hence, rereviews will be welcome

qooban (approved):

Gelio:
Thanks for the reviews :slightly_smiling_face: The UI tests passed recently, and the changes from the latest commit can be verified by `npm run check-types` alone (only TS changes), so I'll merge the PR

PR #1230, opened 2021-03-08
RD-1590 Move resolving widget definiton to Widget component

Gelio:
This PR moves resolving `widget.definition` (from `string` to `WidgetDefinition` object) from `Page` to the `Widget`'s `mapStateToProps`.

This will allow rendering `Widget` components inside of other widgets, allowing recursive rendering of widgets (needed for the details pane of the Deployments View widget).

The aim of this PR is to be transparent in terms of UI changes - everything should keep working as previously. Hence, I did not write any tests, since if all existing tests pass, then it should mean there are no problems and the UI keeps working even with this change :smile: 

As far as I have clicked around the UI, it looks fine.

There will be a separate PR for RD-1591 that adds a unit test that will verify that rendering widgets inside one another works correctly

## UI system tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/263/ (the failed one comes from https://cloudifysource.atlassian.net/browse/RD-1648)


Gelio:
UI system tests results will be in https://jenkins.cloudify.co/job/Stage-UI-System-Test/243/

app/components/Widget.tsx
```diff
@@ -14,15 +14,16 @@ import WidgetDynamicContent from './WidgetDynamicContent';
...
 import type { ReduxState } from '../reducers';
 import type { Widget as WidgetObj } from '../utils/StageAPI';
+import { getWidgetDefinitionById, SimpleWidgetObj } from '../actions/page';
 
 export interface WidgetOwnProps<Configuration> {
     isEditMode: boolean;
     onWidgetUpdated: (widgetId: string, params: Partial<WidgetObj<Configuration>>) => void;
     onWidgetRemoved: (widgetId: string) => void;
-    widget: WidgetObj<Configuration>;
+    widget: SimpleWidgetObj;
```
kubama, line 23: I know this `SimpleWidgetObj` type is not introduced in this PR, but looking at how it is used maybe we could come up with a better naming? Here we have `WidgetObj` and `SimpleWidgetObj` and no clue what is the actual difference.

app/components/useWidgetsFilter.ts
```diff
@@ -0,0 +1,26 @@
...
+    const { manager, widgetDefinitions } = useSelector((state: ReduxState) =>
+        _.pick(state, 'manager', 'widgetDefinitions')
+    );
+
+    return (widgets: SimpleWidgetObj[]) =>
+        _.filter(widgets, widget => {
+            if (!widget.definition) {
+                return false;
+            }
+            const resolvedDefinition = getWidgetDefinitionById(widget.definition, widgetDefinitions);
```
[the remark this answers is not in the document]
Gelio: Hmm. I'm not sure I get the question :smile: `getWidgetDefinitionById` will return `WidgetDefinition | undefined` (https://github.com/cloudify-cosmo/cloudify-stage/blob/b0f75c74bcf0dd7fd14997577027da61d4ab94aa/app/actions/page.ts#L135-L140)

It is the function that takes the ID (which is a string) and resolves that to an object :slightly_smiling_face:

app/components/WidgetsList.tsx
```diff
@@ -8,16 +8,24 @@ import React from 'react';
...
+export default function WidgetsList({ onWidgetUpdated, onWidgetRemoved, isEditMode, widgets }: WidgetsListProps) {
     return (
         <Grid
             isEditMode={isEditMode}
             onGridDataChange={onWidgetUpdated}
             style={{ zIndex: _(widgets).filter({ maximized: true }).size() }}
         >
             {widgets.map(widget => {
-                const widgetDefId = (widget.definition || {}).id;
+                const widgetDefId = widget.definition;
```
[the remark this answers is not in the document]
Gelio: Keep in mind that `widget.definition` here is a `string`

image

since `widget` is `SimpleWidgetObj` (the one for which `definition` has not been resolved yet)

image

Thus, as far as I know, this is backward compatible :slightly_smiling_face:

app/components/Widget.tsx
```diff
@@ -14,15 +14,16 @@ import WidgetDynamicContent from './WidgetDynamicContent';
...
 }
 
-type WidgetProps<Configuration> = WidgetOwnProps<Configuration> & PropsFromRedux;
+type WidgetProps<Configuration> = Omit<WidgetOwnProps<Configuration>, 'widget'> & PropsFromRedux;
```
kubama, line 26: Maybe it's just because I'm a newbie, but my general impression so far is that using `Omit` makes type hierarchy hard to maintain and understand. Personally I would find it cleaner if we tried to use inheritance based hierarchy as much as possible.
For example, the `SimpleWidgetObj` omits one field from `Widget` but adds another. The cleaner way around would be to have a common base type without the `definition` field. What do you think?

app/components/useWidgetsFilter.ts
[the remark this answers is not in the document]
kubama: I think it's confusing because depending on the type `definition` is a string or an object. If possible this should be refactored to have a clean, inheritance based types separation (see my other comments). Variables/properties rename should be done in line, so they reflected their types.

Gelio:
@qooban 

> Loudly thinking: Maybe that's the right time to clean it and fix? Looking at the Redux state it seems like in pages we use definition as a string (widget definition ID) and in widgetDefinitions the widget definition is an object containing the same ID, but inside a field named id.

Well, yeah, the fact that `widget.definition` is `string` in some places and `WidgetDefinition` in others is definitely confusing. TS can help with that, since you can see the type of `widget.definition` and know whether it's an object or a string. This is why I have done all those PRs that convert various files to TS, since it helped me think about the problem without having to remember the type of `widget.definition` in each place :smile: 

What we could do is instead of using the same property (`definition`), have 2 properties:

1. `definitionId` of type `string`
2. `resolvedDefinition` (or just `definition`, for simplicity) of type `WidgetDefinition` (which will be added by the `Widget` component, and is absent otherwise)

This could reduce the confusion. It looks like a more advanced change, so I would rather postpone it to a separate ticket :slightly_smiling_face:

app/actions/page.ts
```diff
@@ -132,6 +132,13 @@ export function forEachWidget(
     );
 }
 
+export function getWidgetDefinitionById(
+    definitionId: string,
+    definitions: ReduxState['widgetDefinitions']
+): WidgetDefinition | undefined {
+    return _.find(definitions, { id: definitionId });
+}
+
```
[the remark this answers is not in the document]
Gelio: To answer your question: there is no benefit when it comes to typing

The benefits I see are:
1. There is a single place in the system that defines how widget definitions are resolved
2. There is a clear name for this operation. Seeing `_.find(state.widgetDefinitions, { id: widget.definition })` is harder to read for me
3. By looking for the function references you can clearly see where the resolution takes place (in 2 places). If the logic was inlined, I don't see how you could easily do that

app/components/Widget.tsx
Gelio: Let me know if you have a better idea for a name :slightly_smiling_face: I'm not proud of it, and when creating it, I thought about alternatives, yet none seemed suitable.

* `DefinitionlessWidget` - no, it has a definition
* `UnresolvedWidget` - sounds very vague. It's not the widget that is resolved, but the definition

thus, I ended up choosing `Simple`. I figured going to the definition and seeing how the type is composed will be sufficient for now

app/components/Widget.tsx
Gelio: Generally, `Omit` is quite natural in TS, and widely used. It's a different mental model than inheritance, so I can see how that's difficult initially

The solution you proposed (having an "abstract base class" (but an interface, in reality)) makes sense, but creates yet another interface (for which we already don't have a good naming - see https://github.com/cloudify-cosmo/cloudify-stage/pull/1230#discussion_r590225333).

Besides, I don't see where such a base interface would be defined where it would make sense. If we put it in `StageAPI.ts`, then you have 2 interfaces, both exported, one extends another, and you don't really see why there are 2 interfaces there.

The proposed interface hierarchy could be `DefinitionlessWidget`, `WidgetWithResolvedDefinition`, and `WidgetWithUnresolvedDefinition` (the names could be better, I did not spend much time on them), and IMO, that's worse than the current 2 interface solution in terms of readability :slightly_smiling_face:

Gelio: Also, a side note - in TS there are "flat type hierarchies" - you do not use many `extends`. Composition is preferred. Types are treated like values, meaning you can easily modify them. AFAIK this is a different paradigm from Java, where you only append properties/methods, and have chains in the hierarchy :slightly_smiling_face:

app/actions/page.ts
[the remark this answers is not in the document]
Gelio: > What do you mean by 'resolved'? It's a simple array lookup.

You are correct, a resolution in terms of an array lookup. I still prefer to have this logic not duplicated across the codebase, especially that we know we want to do the same operation multiple times :slightly_smiling_face: 

> It's easy to read and understand

You are probably better at reading lodash than I am, then :slightly_smiling_face: 

> Maybe I'm missing something, but what's so special about this single line that anybody would want to trace references to it?

In my view, it's the other way around. When I was making this change I was wondering "where in the system are widget definitions resolved?". I did not find a clear answer. 

With this being extracted, a reference lookup yields the exact places that leverage resolving widget definitions :slightly_smiling_face: 

---

Thanks for agreeing to disagree :slightly_smiling_face: I'll leave it as is

kubama:
> What we could do is instead of using the same property (`definition`), have 2 properties:
> 
> 1. `definitionId` of type `string`
> 2. `resolvedDefinition` (or just `definition`, for simplicity) of type `WidgetDefinition` (which will be added by the `Widget` component, and is absent otherwise)
> 
> This could reduce the confusion. It looks like a more advanced change, so I would rather postpone it to a separate ticket 🙂

It would be a welcome change. I can remember being confused by the same thing working in the same area some time ago. So I think it's worth addressing.

app/components/Widget.tsx
kubama: How about:
- widget with string `definition` (preferably renamed to `definitionId`) - just `Widget`
- widget with object `definition` - ResolvedWidget

Gelio: Hmm... makes sense. I could do the change here, but that would expand the diff considerably :slightly_smiling_face: I would rather do that in a separate PR, especially that `SimpleWidgetObj` was not introduced here.

Let me know if creating a separate task for that is fine :slightly_smiling_face: I will not do it immediately, since I feel pressured with delivering features for Environments

app/components/Widget.tsx
kubama: I can see I'm not yet "thinking the TS way". But still, if we take `SimpleWidgetObj` as an example, which removes one prop from the 'base' type but ads another - is this really a good practice?

Gelio: Can't really say :slightly_smiling_face: It looks fine to me

I reckon both of these can be avoided if the underlying JS code was not reusing the same property for 2 types of data (resolved and unresolved) :slightly_smiling_face: 

For instance, in this line that you commented on, both `WidgetOwnProps` and `PropsFromRedux` have a `widget` property which are of different types. If `mapStateToProps` used a different property (e.g. `resolvedWidget`), then we would not technically need the `Omit`. I would still encourage using it to show the fact that `Widget.props.widget` (the unresolved one) is not needed in the component.

All in all, I don't see much wrong here.

btw. I agree with the "extension" approach when it comes to more complex/larger types. For example, in `StageAPI.ts` there is `CommonWidgetDefinition` and `ReactWidgetDefinitionPart` and `HTMLWidgetDefinitionPart` that are combined together, rather than using `Omit`s.

app/components/Widget.tsx
kubama: > Let me know if creating a separate task for that is fine 🙂 I will not do it immediately, since I feel pressured with delivering features for Environments

Sure, separate task is just fine.

app/components/Widget.tsx
kubama: OK, thanks for explaining.

kubama (approved):

app/components/useWidgetsFilter.ts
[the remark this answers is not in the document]
qooban: So I assume we'll have separate issue for that subject (refactoring around `widget.definition` to make it clear when it is a string and when it is an object), right?

Gelio:
I've only rebased the branch on `master` having merged #1229. I'll work on creating tickets and investigating UI tests now

app/components/Widget.tsx
Gelio: Great :slightly_smiling_face: I have captured that in https://cloudifysource.atlassian.net/browse/RD-1645

I will add TODO comments in the code

Gelio:
@qooban @kubama I have added a new commit that adds TODO comments for the tickets I have created to address the points you brought up :slightly_smiling_face: 
1. RD-1645
2. RD-1649

If you have any comments on the JIRA tickets, feel free to post them :slightly_smiling_face: 

I'm going to take care of the system tests now

qooban:
> If you have any comments on the JIRA tickets, feel free to post them slightly_smiling_face

Thanks for reporting it. Just one comment: We should probably add an epic for them. **UI Technical Debt**? Or you had another plan for them? With tons of the issues reported not only in the UI, I believe there's lower probability that we'll forget about them when we assign it to an epic (especially that eg. there's a plan for UI Technical Debt review somewhere in the future).

Gelio:
> We should probably add an epic for them

Yes, yes, my bad! I have forgotten to add them to the epic. I was thinking about it

btw. wouldn't it be nice if there was an "Epic" field when creating a ticket :smile: ? I would probably forget about it less often. I'll bring it up during the retro :slightly_smiling_face:

qooban:
> btw. wouldn't it be nice if there was an "Epic" field when creating a ticket  ? I would probably forget about it less often. I'll bring it up during the retro

Yeah, I think it was present in the previous projects: CY and CYBL. Worth to point it on retro.

kubama (approved):

Gelio:
Looks like I have missed this part of code that also deals with widget definition resolution

https://github.com/cloudify-cosmo/cloudify-stage/blob/209e56ed7bde25ef909bfd69ea45e927d461da3e/app/components/templates/PageManagement.jsx#L49-L56

and that causes the system tests for page edit to fail, because after adding widgets to the page in `PageManagement` component, the widgets are not added.

I will investigate it further.

I plan on migrating `PageManagement` component to TypeScript to clearly mark when `definition` is a `string`, and when it's an object. If there will be many code changes, I will do that in a separate PR

app/components/WidgetsList.tsx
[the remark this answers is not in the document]
Gelio: Yup, that's in RD-1649

app/components/useWidgetsFilter.ts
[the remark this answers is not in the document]
Gelio: Yup :slightly_smiling_face: RD-1649

Gelio:
I have managed to fix the problem in template management. To avoid adding more and more changes to this PR, I have created a new one with the fixes, that builds on top of this one: https://github.com/cloudify-cosmo/cloudify-stage/pull/1232

https://github.com/cloudify-cosmo/cloudify-stage/pull/1232 will have to be merged first into this PR before this PR is merged into `master` to avoid breaking master

Gelio:
The last commit comes from merging https://github.com/cloudify-cosmo/cloudify-stage/pull/1232 (reviewed separately into this branch)

Gelio:
UI system tests will be in https://jenkins.cloudify.co/job/Stage-UI-System-Test/263/

Gelio:
I believe I have addressed every comment in this PR so far and with #1232  merged, the UI system tests should be passing (I'll be waiting for the results :crossed_fingers: )

I'm looking for re-reviews (but the code has mostly been reviewed already) and possible approvals :slightly_smiling_face:

kubama (approved):

qooban (approved):

Gelio:
The only test that failed in https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/263/pipeline/ is the one for `widgets/deployment_button_spec.js`, which is related to https://cloudifysource.atlassian.net/browse/RD-1648

With all other tests having passed, I think it's safe to merge the PR.

@qooban @kubama, any objections/comments :slightly_smiling_face: ?

kubama:
Just merge it man 🙂.

PR #1232, opened 2021-03-10
RD-1590 Fix widget definition resolution in PageManagement component

Gelio:
This PR fixes the previously missed widget definition resolution in `PageManagement` component. This component can now operate on `SimpleWidgetObj`s (whose `definition` is a string), because the widget definition resolution happens in the nested `Widget` component (which was changed in https://github.com/cloudify-cosmo/cloudify-stage/pull/1230)

I am doing these changes as a separate PR to make the review process easier by having smaller PR, and since https://github.com/cloudify-cosmo/cloudify-stage/pull/1230 is already reviewed in the most part.

## Merge plan

Since the system tests only pass with the changes from this PR, the suggested merge plan is:

1. Merge this PR into https://github.com/cloudify-cosmo/cloudify-stage/pull/1230
2. Merge https://github.com/cloudify-cosmo/cloudify-stage/pull/1230 into `master` (without having to re-review it again, since the changes have been reviewed in this PR)

## Changes in this PR

Most of the changes are related to migrating files to TS. I will mark changes to the logic with comments.

## UI system tests

1. https://jenkins.cloudify.co/job/Stage-UI-System-Test/253/
    They failed on tests for "Create Deployment Button widget handles errors during deploy & install process", which is a known error, probably unrelated to this PR, and has been reported: https://cloudifysource.atlassian.net/browse/RD-1648

    All other tests passed, meaning that the problems with template management should be fixed :slightly_smiling_face: 
2. A rerun: https://jenkins.cloudify.co/job/Stage-UI-System-Test/256
    Failed on some Jenkins error.
3. A rerun: https://jenkins.cloudify.co/job/Stage-UI-System-Test/257
    Failed on some Jenkins error. I have contacted Yuval to resolve the Jenkins issue.

When running locally only the 2 test suites that failed for https://github.com/cloudify-cosmo/cloudify-stage/pull/1230, they passed.


app/components/templates/PageManagement.tsx
```diff
@@ -9,47 +6,66 @@ import { useDispatch, useSelector } from 'react-redux';
...
-export default function PageManagement({ pageId, isEditMode }) {
-    const dispatch = useDispatch();
+export default function PageManagement({ pageId, isEditMode = false }: PageManagementProps) {
+    const dispatch = useDispatch<ThunkDispatch<ReduxState, never, AnyAction>>();
 
     useEffect(() => {
         dispatch(setActive(true));
-        return () => dispatch(setActive(false));
+        // NOTE: use void to return `undefined` from the cleanup handler (fix a type error)
+        // eslint-disable-next-line no-void
```
Gelio, line 34: I will suggest disabling the `no-void` rule during the weekly meeting. I don't see how it can be harmful, and it's tedious in cases such as these :slightly_smiling_face:

app/components/templates/PageManagement.tsx
```diff
@@ -77,70 +93,84 @@ export default function PageManagement({ pageId, isEditMode }) {
...
+            name: i18n.t('editMode.tabs.newTab'),
+            widgets: []
+        });
         updatePage();
     };
-    const onTabRemoved = (layoutSection, tabIndex) => {
+    const onTabRemoved = (layoutSection: number, tabIndex: number) => {
         page.layout[layoutSection].content = _.without(
             page.layout[layoutSection].content,
-            _.nth(page.layout[layoutSection].content, tabIndex)
```
Gelio, line 138: `nth` returns `value | undefined` and TypeScript does not like that

app/components/templates/PageManagement.tsx
```diff
@@ -149,24 +179,27 @@ export default function PageManagement({ pageId, isEditMode }) {
...
         page.layout = [..._.slice(page.layout, 0, position), layoutSection, ..._.slice(page.layout, position)];
         updatePage();
     };
 
     const isWidgetMaximized = findWidget({ maximized: true });
 
-    $('body')
-        .css({ overflow: isWidgetMaximized ? 'hidden' : 'inherit' })
-        .scrollTop(0);
+    document.body.style.overflow = isWidgetMaximized ? 'hidden' : 'inherit';
```
Gelio, line 197: The jQuery logic replacement here is the same as in https://github.com/cloudify-cosmo/cloudify-stage/blob/65e961f4b11cec9a8e2a5ad1ad80da3c75c8dc84/app/components/Page.tsx#L71-L72

app/components/templates/PageManagement.tsx
```diff
@@ -192,7 +225,7 @@ export default function PageManagement({ pageId, isEditMode }) {
                             <Breadcrumb.Section active>
                                 <EditableLabel
                                     value={page.name}
-                                    placeHolder={i18n.t(
+                                    placeholder={i18n.t(
```
Gelio, line 224: Looks like there was a typo. The `EditableLabel` uses `placeholder`, not `placeHolder` :slightly_smiling_face: bug fixed

app/utils/templatesLoader.ts
```diff
@@ -3,30 +3,32 @@
...
+                const promises: Promise<void>[] = [];
                 templateList.forEach(item => {
                     promises.push(
                         LoaderUtils.fetchResource(`templates/${item.id}.json`, item.custom)
                             .then(templateData => {
                                 templatesDef[item.id] = templateData;
                             })
                             .catch(e => {
-                                throw new Error(`Error loading template ${item.id}`, e);
+                                throw new Error(`Error loading template ${item.id}: ${e}`);
```
Gelio, line 31: `Error` constructor only accepts a single argument, AFAIK. This was caught by TS first

qooban:
Will review tomorrow morning.

app/components/templates/PageManagement.tsx
```diff
@@ -149,24 +175,27 @@ export default function PageManagement({ pageId, isEditMode }) {
...
+            page.layout[layoutSection].content as TabContent[],
+            oldTabIndex,
+            newTabIndex
+        );
         updatePage();
     };
-    const onLayoutSectionRemoved = layoutSection => {
-        page.layout = _.without(page.layout, _.nth(page.layout, layoutSection));
+    const onLayoutSectionRemoved = (layoutSection: number) => {
+        page.layout = _.without(page.layout, page.layout[layoutSection]);
```
qooban, line 187: I assume we are safe here with removing `_.nth`?

app/utils/templatesLoader.ts
```diff
@@ -3,30 +3,32 @@
...
+import type { ManagerData } from '../reducers/managerReducer';
+import type { TemplatesState } from '../reducers/templatesReducer';
 import Internal from './Internal';
 import LoaderUtils from './LoaderUtils';
 
 export default class TemplateLoader {
-    static load(manager) {
-        log.log('Load templates');
+    static load(manager: ManagerData) {
+        log.debug('Load templates');
```
qooban, line 13: General:

I saw in the previous PRs you also changed log levels for some of the logs we had? What is the reason? I understand that in some cases we can have `debug` instead of `log/info`. But on the other hand, as we have `warn` set by default, I see no big value in changing it. 

Also, this kind of logs are for us, not for the users.

app/utils/templatesLoader.ts
qooban: Can you share an example of error thrown by `LoaderUtils.fetchResource`?

app/actions/templateManagement.js
```diff
@@ -33,10 +33,7 @@ export function createPageId(name, pages) {
 export function persistPage(page) {
     return (dispatch, getState) => {
         function prepareWidgetData(widget) {
-            return {
-                ..._.pick(widget, 'name', 'width', 'height', 'x', 'y', 'configuration'),
-                definition: widget.definition.id
```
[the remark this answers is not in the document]
qooban: Looking at the history, this code have been here since 2017. 
So just to be sure - you are fully confident that in this case `widget.definition` is string, right?

app/actions/page.ts
```diff
@@ -29,6 +29,7 @@ export interface WidgetsSection {
 export interface TabContent {
     name: string;
     widgets: SimpleWidgetObj[];
+    isDefault?: boolean;
```
qooban, line 32: So it means that `isDefault` field sometimes is inside the variable with a type `TabContent` and sometimes it is not there?

app/actions/page.ts
```diff
@@ -108,12 +109,12 @@ export function createPagesMap(pages: PageDefinition[]) {
 }
 
 export function forAllWidgets(
-    page: PageDefinition,
+    page: Pick<PageDefinition, 'layout'>,
```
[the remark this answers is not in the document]
qooban: General:

Looking at TS-related changes in general, it seems to me like we are going in a direction where we have tons of types definitions which are similar. They're having slightly different "flavours" - coming from the same types, sometimes with additional field(s), sometimes without some field(s). 

I'm not sure it is the right direction. But maybe that is how TS works. Maybe the reason is that cloudify-stage code needs general refactoring before going fully into TS. Just loudly thinking.

qooban (commented):
Added some mix of specific/general comments.

app/actions/templateManagement.js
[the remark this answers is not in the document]
Gelio: TL;DR: yes, I'm confident :slightly_smiling_face: 

Explanation:

Let's dive into the call hierarchy for this code
1. https://github.com/cloudify-cosmo/cloudify-stage/blob/d63b4702a9ba70a2e58c4ad6c90125732c4ff5f7/app/actions/templateManagement.js#L43
2. `persistePage` is only used in `savePage`: https://github.com/cloudify-cosmo/cloudify-stage/blob/d63b4702a9ba70a2e58c4ad6c90125732c4ff5f7/app/actions/templateManagement.js#L58-L60
3. References (uses) of `savePage`
    image
4. https://github.com/cloudify-cosmo/cloudify-stage/blob/209e56ed7bde25ef909bfd69ea45e927d461da3e/app/components/templates/PageManagement.jsx#L117-L11
    and with the rest of the code changed in `PageManagement`, `definition`s are no longer resolved to be objects, and they remain as strings, due to:

    https://github.com/cloudify-cosmo/cloudify-stage/blob/209e56ed7bde25ef909bfd69ea45e927d461da3e/app/components/templates/PageManagement.jsx#L98-L105

    https://github.com/cloudify-cosmo/cloudify-stage/blob/209e56ed7bde25ef909bfd69ea45e927d461da3e/app/components/templates/PageManagement.jsx#L49-L56

app/utils/templatesLoader.ts
Gelio: It can be any `fetch` error, or `response.json()` error, e.g.:

image

app/actions/page.ts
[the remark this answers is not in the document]
Gelio: > They're having slightly different "flavours" - coming from the same types, sometimes with additional field(s), sometimes without some field(s).

Correct :slightly_smiling_face: Good observation, I've had the same ideas too

> I'm not sure it is the right direction. But maybe that is how TS works. Maybe the reason is that cloudify-stage code needs general refactoring before going fully into TS. Just loudly thinking.

I see 2 causes:
1. The JS values have different flavors :slightly_smiling_face: If the objects always followed the same structure, we would always be using the same types.
2. I am not familiar with the whole codebase and when adding types, I don't know whether somewhere in the system there is a value that is very similar to a type I am adding for another value, but e.g. with some additional fields/missing fields. 

    For example in this PR, that was the case with `TemplatePageDefinition` and `PageDefinition`. I thought we are always dealing with full `PageDefinition`s, but then it turns out that there is a different set of values that are similar, but have some properties that are missing.

If there was static typing from the beginning, types and the underlying JS values would probably be used differently and thought about differently. When designing new features, I usually start with designing the types first. If possible, I never work with JS objects first, and then sprinkle types.

I reckon the problem is in JS values. If we make the JS values consistent, then the types would be consistent too. Adding some non-standard fields (like the `id` field added to `page` (https://github.com/cloudify-cosmo/cloudify-stage/blob/bf67e15f38c28a6c710fe673f05687ad27b59960/app/components/templates/PageManagement.tsx#L55, https://github.com/cloudify-cosmo/cloudify-stage/blob/bf67e15f38c28a6c710fe673f05687ad27b59960/app/components/templates/PageManagement.tsx#L144-L148) confused me since the property did not appear pages in other parts of the system. It may not even be used. I did not investigate, but properties such as these are what I would call "type smells" :smile: 

Overall, type reuse in TypeScript is encouraged (using `Pick`, `Omit`, `extends`). Typing less code is always nice, and reusing the same properties lets you get richer references and explorability in the IDE. So, I would not worry about the general type reuse (e.g. this line), but try to align the JS values first, and then the types will follow :slightly_smiling_face:

app/components/templates/PageManagement.tsx
Gelio: I think so. The reason for removing `nth` is the same as in other places I do that - to avoid dealing with `undefined` being possibly returned.

The only place I have seen where `page` is created is:
https://github.com/cloudify-cosmo/cloudify-stage/blob/209e56ed7bde25ef909bfd69ea45e927d461da3e/app/components/templates/TemplateManagement.jsx#L207-L211

and there, `layout` is always defined. Thus, the chance of a possible regression is pretty small to me - one would have to remove the `layout` property. I have not noticed any errors while testing myself, and the system tests also passed after the changes

app/utils/templatesLoader.ts
Gelio: Oh, so it was not clear why I did that :smile: Let me make up for it by explaining it now

The reason is that there is no `log` method on a `loglevel` instance in TypeScript typings for this module:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b4683d7/types/loglevel/index.d.ts

A `log` method is not even described in the documentation of `loglevel` :slightly_smiling_face: https://www.npmjs.com/package/loglevel/v/1.4.1#documentation

I reckon it's exposed only for compatibility with the `console` API. Even then, `log` is the same as `debug`:
image

Which is done here:

https://github.com/pimterry/loglevel/blob/v1.6.2/dist/loglevel.js#L72-L73

app/actions/page.ts
Gelio: Yup, very good observation :slightly_smiling_face:

`isDefault` is not set in `catalog` template:
image

It is also not present in my newly created page:
image

But :exclamation: it gets set after I edit the tab name/set a tab to be the default one:
image

:exploding_head:

Gelio:
Thanks for all the comments @qooban and @kubama :slightly_smiling_face: 

I have responded to all of them :slightly_smiling_face: I'm looking for another round of comments with possible approvals :smile:

kubama (approved):

app/utils/templatesLoader.ts
qooban: Got it, thanks for the explanation.

app/utils/templatesLoader.ts
qooban: I was more interested how the error string is now displayed, but I assume it's fine?

app/actions/page.ts
qooban: Hmm... I'm wondering will it not cause issues in the future. Do you think it would be worth to create a ticket for that to fix it at some point?

app/actions/page.ts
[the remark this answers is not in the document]
qooban: Yep, I think we have similar observations.

>If there was static typing from the beginning, types and the underlying JS values would probably be used differently and thought about differently.

That is true.

app/actions/templateManagement.js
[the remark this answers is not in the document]
qooban: If you are confident, then I'm fine 👍🏻

app/components/templates/PageManagement.tsx
qooban: OK 👍🏻

qooban (approved):
LGTM 👍🏻

app/actions/templateManagement.js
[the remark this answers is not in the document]
Gelio: My confidence stems from 2 factors:
1. I followed TypeScript references to those functions and did not notice it being called from elsewhere than the code I modifed
2. In the code I modified I am always (AFAIK) using string references

this is not to say I'm 100% confident, as there may always be unexpected things, but overall we should be good here :slightly_smiling_face:

app/utils/templatesLoader.ts
Gelio: Oh, this part :slightly_smiling_face: Then I believe what may be the answer you are looking for is: how does `Error` behave when used in a template literal?

image

Answer: the error's `message` property is being used. This is due to the `toString` defined on errors: https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Global_Objects/Error/toString

app/actions/page.ts
Gelio: Hmm... good idea :smile: Here it is https://cloudifysource.atlassian.net/browse/RD-1675

Gelio:
Thanks for quick reviews :slightly_smiling_face: I'll be merging this PR into https://github.com/cloudify-cosmo/cloudify-stage/pull/1230 and run system tests there

app/utils/templatesLoader.ts
qooban: Got it, thx.

app/actions/page.ts
qooban: Thanks 👍🏻

PR #1235, opened 2021-03-11
RD-1606 - Add validation for keys with Cloudify owned prefix

qooban:
This PR adds new validation popup to prevent adding labels with not permitted keys.

https://user-images.githubusercontent.com/5202105/110759420-e28e4780-824d-11eb-99e9-3f864b5d883d.mp4


widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -34,23 +36,43 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
...
     useEffect(() => {
         onChange(labels);
     }, [labels]);
 
+    useEffect(() => {
+        if (open) {
+            resetNewLabelKey();
+            resetNewLabelValue();
+        }
+    }, [open]);
```
qooban, line 53: Addition: Resetting label key and values on opening labels input. I think that is worth to add it as it clears both inputs and hides popups.

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -34,23 +36,43 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
...
+
     function onAddLabel() {
         function isLabelInSystem() {
-            const fetchUrl = addSearchToUrl(`/labels/deployments/${newLabelKey}`, newLabelValue);
-            return toolbox
-                .getManager()
-                .doGet(fetchUrl)
+            const actions = new DeploymentActions(toolbox);
+            return actions
+                .doGetLabel(newLabelKey, newLabelValue)
```
qooban, line 75: Addition: I took a chance to move fetching action to more accurate place - `DeploymentActions`.

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -89,14 +117,21 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
                         style={{ ...iconStyle, right: '2em' }}
                     />
                 )}
-                <Icon name="dropdown" link onClick={toggleOpen} style={{ ...iconStyle, right: '0.5em' }} />
+                <Icon
+                    data-cy="labels-input-switch"
```
qooban, line 121: Following https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements I introduced `data-cy` attribute for this specific component (labels input toggle).

I'm open for your suggestions for the better name here.

Gelio: If we named it `data-testid` or `data-test`, the name would be general enough that we could use it in jest tests too :slightly_smiling_face: Aside from that, I am all for introducing such an attribute :smile:

Gelio:
Some general comments :slightly_smiling_face: 

1. I noticed that the popup with the list of possible labels goes away when the user types more than `csys-`. I reckon it would be a nice addition if the known and allowed `csys-` labels were present in the autocomplete dropdown. This way the user could type less. This would mean that they would appear even if there are no labels with that key yet (they would be dynamically injected in the dropdown).
    What do you think? This could be a separate ticket, ofc :slightly_smiling_face: 

2. In the video I noticed you typed an incorrect `csys-` label, and there was no error message. Is this expected?

widgets/common/src/labels/InvalidKeyErrorPopup.jsx
```diff
@@ -0,0 +1,15 @@
+import LabelErrorPopup from './LabelErrorPopup';
+
+export default function InvalidKeyErrorPopup({ reservedKeys }) {
```
Gelio, line 3: Optional:

This is a good candidate for a TypeScript component :slightly_smiling_face: It seems quite small with simple props

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -12,16 +12,18 @@ const iconStyle = {
...
-        Common: { RevertToDefaultIcon },
+        Common: { DeploymentActions, RevertToDefaultIcon },
         Hooks: { useBoolean, useResettableState, useToggle },
         Utils: { combineClassNames },
         i18n
     } = Stage;
 
     const [addingLabel, setAddingLabel, unsetAddingLabel] = useBoolean();
+    const [reservedKeys, setReservedKeys] = useState([]);
+    const [fetchingReservedKeys, setFetchingReservedKeys, unsetFetchingReservedKeys] = useBoolean();
```
Gelio, line 26: How about extracting all logic related to fetching and exposing `reservedKeys` into a custom hook? This would make the component itself less complex :slightly_smiling_face:

widgets/common/src/labels/LabelsInput.jsx
Gelio: Yeah, good addition :+1:

Gelio, line 53: I reckon we can use the `useOpenProp` hook here :slightly_smiling_face:

widgets/common/src/labels/LabelsInput.jsx
Gelio: Good idea :+1:

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -34,23 +36,43 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
...
     useEffect(() => {
         if (!hideInitialLabels) setLabels(initialLabels);
     }, [initialLabels]);
 
+    useEffect(() => {
+        const actions = new DeploymentActions(toolbox);
+        setFetchingReservedKeys();
+        actions
+            .doGetReservedLabelKeys()
+            .then(setReservedKeys)
```
Gelio, line 64: General:

This will trigger a warning if the component is unmounted before the request finishes. Instead of solving it case by case by adding some kind of `isMounted` variable, I believe we should solve that on a system level by either:
1. Adding some common logic that would encapsulate the pattern of only running some side-effects after the request is finished **and** the component is not mounted
2. Cancelling requests on component unmount
with the 2nd option being the preferred one :smile: 

I'm pretty sure there is already a ticket for cancelling requests, I just wanted to bring it up that not having it causes more and more technical debt to pile up :slightly_smiling_face:

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -89,14 +117,21 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
...
+                    style={{ ...iconStyle, right: '0.5em' }}
+                />
                 <LabelsList labels={labels} onChange={setLabels} />
             </div>
             {open && (
                 <div style={{ padding: '0 0.5em' }}>
                     <Divider hidden={_.isEmpty(labels)} />
                     <Form.Group>
                         <Form.Field width={7}>
+                            {newLabelKeyIsNotPermitted && <InvalidKeyErrorPopup reservedKeys={reservedKeys} />}
```
Gelio, line 134: Optional:

IMO we should have 2 popups:

1. User started typing a `csys` key -> explain the concept of reserved labels to the user and show the reserved labels (the current popup)
2. User typed an unexpected reserved label -> explicitly tell the user that the value is wrong

This way the user will notice that the label is invalid when they are typing. At the moment, the popup remains the same no matter if the user typed a prefix for a valid reserved label (e.g. `csys-env-...`), or an invalid reserved label

test/cypress/support/deployments.ts
```diff
@@ -30,6 +30,8 @@ const commands = {
             siteName !== '' ? { site_name: siteName } : { detach_site: true }
         );
     },
+    getReservedLabelKeys: () =>
+        cy.cfyRequest('/labels/deployments?_reserved=true', 'GET', null).then(response => response.body.items),
```
Gelio, line 34: If we added a type annotation for `items`, the whole function would be safer :slightly_smiling_face: 

```suggestion
        cy.cfyRequest('/labels/deployments?_reserved=true', 'GET', null).then(response => response.body.items as string[]),
```

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -66,6 +66,17 @@ describe('Deployment Action Buttons widget', () => {
...
+        function checkIfPopupIsDisplayed(key, popupContent) {
+            typeLabelKey(key);
+            cy.contains('.popup', popupContent).should('be.visible');
+        }
+        function checkIfPopupIsNotDisplayed(key) {
+            typeLabelKey(key);
+            cy.get('.popup').should('not.exist');
+        }
+        function toggleLabelsInput() {
+            cy.get('[data-cy="labels-input-switch"]').click();
```
Gelio, line 78: How about adding a custom command for selecting elements by the `data-cy` attribute, as recommended in https://docs.cypress.io/guides/references/best-practices.html#Real-World-Example? :smile: 

This would reduce the boilerplate related to such selections

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -118,25 +130,37 @@ describe('Deployment Action Buttons widget', () => {
                 typeLabelKey('existing_key');
                 typeLabelValue('existing_value');
 
-                cy.get('button.disabled').should('have.attr', 'disabled');
+                cy.get('button[aria-label=Add]').should('have.attr', 'disabled');
```
Gelio, line 133: :+1:

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -118,25 +130,37 @@ describe('Deployment Action Buttons widget', () => {
...
+                typeLabelValue('a');
+                cy.get('button[aria-label=Add]').should('have.attr', 'disabled');
+            }
+            function checkIfInternalKeyIsPermitted(key) {
+                checkIfPopupIsNotDisplayed(key);
+                typeLabelValue('a');
+                cy.get('button[aria-label=Add]').should('not.have.attr', 'disabled');
+            }
+            cy.getReservedLabelKeys().then(reservedKeys => {
+                reservedKeys.forEach(key => checkIfInternalKeyIsPermitted(key));
```
Gelio, line 160: Optional:

Some code golf :slightly_smiling_face: 

```suggestion
                reservedKeys.forEach(checkIfInternalKeyIsPermitted);
```

Gelio:
Looks well overall :slightly_smiling_face:

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Yeah, make sense, but I don't like the name they suggested `getBySel`. The argument for me is not a `selector`, but attribute value. I'm not sure we also needs additional values for `cy.get`, so we can go with simple one. 

I can add it, but to avoid ping-pong in comments: Do you have suggestion what would be the best name for it?

widgets/common/src/labels/InvalidKeyErrorPopup.jsx
qooban: Yes, but as other `*ErrorPopup` components are already written in JS, I don't want to do an exception for this specific one.

widgets/common/src/labels/LabelsInput.jsx
qooban: Yeah, could be, but I think in that case we won't be using Jest tests for it.

widgets/common/src/labels/LabelsInput.jsx
qooban: Nice idea, but as it was already discussed and accepted with Ofer, I'd like not to change it. I think it's good enough.

qooban (commented):
>  I noticed that the popup with the list of possible labels goes away when the user types more than csys-. 

In which case exactly? It goes away when the key is valid.

> I reckon it would be a nice addition if the known and allowed csys- labels were present in the autocomplete dropdown. This way the user could type less. This would mean that they would appear even if there are no labels with that key yet (they would be dynamically injected in the dropdown). What do you think? This could be a separate ticket, ofc slightly_smiling_face

Yes, that I was thinking about, too, but I think it is something for the future. I'll create a ticket for that.

> In the video I noticed you typed an incorrect csys- label, and there was no error message. Is this expected?

What do you mean by incorrect csys- label? `csys-env-type` is permitted, so the popup disappeared. `csys-my-key` is not permitted, so the popup is still visible and the add button is disabled.

test/cypress/support/deployments.ts
qooban: Done.

widgets/common/src/labels/LabelsInput.jsx
qooban: Yep, good point, done.

widgets/common/src/labels/LabelsInput.jsx
qooban: Good idea, done.

qooban:
@Gelio

> I reckon it would be a nice addition if the known and allowed csys- labels were present in the autocomplete dropdown. This way the user could type less. This would mean that they would appear even if there are no labels with that key yet (they would be dynamically injected in the dropdown). What do you think? This could be a separate ticket, ofc slightly_smiling_face

Created [RD-1676](https://cloudifysource.atlassian.net/browse/RD-1676) for that.

widgets/common/src/labels/LabelsInput.jsx
Gelio: We have Enzyme installed, meaning we can test React components using jest. I don't see other disadvantages to using `data-test` or `data-testid`, aside from having to type 2 more letters :slightly_smiling_face: and selecting in jest tests will get less clumsy

Do you see risks in naming it something that's not related to Cypress?

widgets/common/src/labels/InvalidKeyErrorPopup.jsx
Gelio: Hmm... why not? There does not seem to be a rule that TS components cannot use JS components, or that similarly named components cannot be written in JS/TS :smile: How is this making an exception?

widgets/common/src/labels/LabelsInput.jsx
Gelio: What are your thoughts :slightly_smiling_face: ?

widgets/common/src/labels/LabelsInput.jsx
Gelio: Do you see my suggestion as an improvement/enhancement to the UX? If so, creating a ticket will mean this gets reported and considered :slightly_smiling_face: 

I see this as an improvement, but not required at the moment :slightly_smiling_face:

test/cypress/integration/widgets/deployment_actions_spec.js
Gelio: Depending on what we decide in https://github.com/cloudify-cosmo/cloudify-stage/pull/1235#discussion_r592166734, I suggest:

1. `getByTestId`
2. `getByTestAttribute`

And finally, I suggest the following combo: 
* attribute name: `data-testid`
* command name: `getByTestId`

Gelio:
Aside from the currently open discussions, I don't have any more comments :slightly_smiling_face:

widgets/common/src/labels/InvalidKeyErrorPopup.jsx
qooban: As it was Optional comment, I don't want to spend more time discussing it here. Also, there's already a topic about JS -> TS files migration you want to bring up on a weekly.

widgets/common/src/labels/LabelsInput.jsx
qooban: Added few lines about it to https://cloudifysource.atlassian.net/browse/RD-1676 . 
If sb decide to do that in the future, I think it should be done at once and should not be separated.

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: Implemented your suggestion.

widgets/common/src/labels/LabelsInput.jsx
qooban: Does not matter much for me. If you are insisting, I'm fine with changing it.

Gelio (dismissed):
Looks good :slightly_smiling_face:

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -1,27 +1,54 @@
...
+    const [fetchingReservedKeys, setFetchingReservedKeys, unsetFetchingReservedKeys] = useBoolean();
+
+    useEffect(() => {
+        const actions = new DeploymentActions(toolbox);
+        setFetchingReservedKeys();
+        actions
+            .doGetReservedLabelKeys()
+            .then(setReservedKeys)
+            .catch(error => {
+                log.error(i18n.t('widgets.common.labels.fetchingReservedKeysError'), error);
```
kubama, line 32: Do we need i18n for log messages?

qooban: I think it is definitely not mandatory. But on the other hand, I see no issues with having it. 
I originally planned to have that message visible to the user, but as it is very, very unlikely `log.error` seems to me like perfectly enough.

widgets/common/src/labels/LabelsInput.jsx
qooban: Oh, missed that. Yeah, I'm all for the second option. I don't know if we already have a ticket for cancelling requests.

app/translations/en.json
```diff
@@ -410,9 +410,11 @@
                 "newValue": "New value",
                 "valuePlaceholder": "Select or type in a value",
                 "validationError": "Only letters, digits and the characters \"-\", \".\" and \"_\" are allowed and it must begin with a letter.",
+                "invalidKeyError": "All labels with a `csys-` prefix are reserved for internal use. Allowed `csys-` prefixed labels are: {{ reservedKeys }}.",
```
kubama, line 413: I'm really not sure what this message says. It first says "all are reserved" (my understanding - not available for use) but then "allowed are" (my understanding - some are actually available for use). For me it's really confusing and kind of contradicting.

widgets/common/src/labels/LabelsInput.jsx
```diff
@@ -34,23 +61,29 @@ export default function LabelsInput({ hideInitialLabels, initialLabels, onChange
         const allLabels = [...labels, ...(hideInitialLabels ? initialLabels : [])];
         return !!_.find(allLabels, newLabel);
     })();
-    const addLabelNotAllowed = !newLabelIsProvided || newLabelIsAlreadyPresent || addingLabel;
+    const newLabelKeyIsNotPermitted = newLabelKey.startsWith('csys-') && !reservedKeys.includes(newLabelKey);
```
kubama, line 64: Optional: this could be extracted to a constant and passed as a parameter to the popup message, so the message was always in line with the logic behind.

widgets/common/src/labels/LabelsInput.jsx
kubama: I can see several issues:
- it's inconvenient 
- if there is no value in it then it's better to stay aligned with existing approaches
- it bloats translation file - imagine we started adding log messages to the translation file at a greater scale - we could even end up with the number of log messages greater than the number of actual UI strings defined there
- translation file is often handed directly to the translation team - whoever did the translation he could be not interested in investing time in translating log messages, but there is no clear indication which messages are used in UI directly, and which are just log messages
- logs are intended for technical staff (support, developers), as such they should stick to English - I would be rather annoyed if I received logs from production to analyse and they were using any other language then EN

widgets/common/src/labels/LabelsInput.jsx
qooban: By this you mean `csys-`? 
Good idea. Can do.

kubama: Yes

app/translations/en.json
qooban: I actually was not thinking about the message as it was copied directly from the REST API response and Ofer accepted it.
In my understanding "are reserved for internal use" does not mean "not available for use". But I understand it is ambiguous, so please provide your suggestion. Maybe we can just drop "All" word at the beginning?

widgets/common/src/labels/LabelsInput.jsx
qooban: Done.

widgets/common/src/labels/LabelsInput.jsx
qooban: Removed.

app/translations/en.json
kubama: My proposal:
Allowed `csys-` prefixed labels are: {{ reservedKeys }}. All other `csys-` prefixed labels are reserved for internal use and cannot be used directly (or maybe "cannot be added/assigned").
Also, the list of the allowed labels in this popup is a piece of continuous text which is really hard to read. I'd rather used bullet points for labels list.

qooban: I'll go with: 
image

Any other improvements (like changing the list to be bulleted) to the dropdown/popup can be done within: [RD-1676](https://cloudifysource.atlassian.net/browse/RD-1676). Added your suggestion about bullet points there.

kubama (approved):

qooban:
PR ready for final reviews.

kubama (approved):

qooban:
Assuming the change is small and there's one approval already, I'm merging the PR. 

In case of more comments, we can continue in [RD-1676](https://cloudifysource.atlassian.net/browse/RD-1676) or have another issue.

widgets/common/src/labels/LabelsInput.jsx
Gelio: I did not find a ticket about aborting/canceling requests (here's [the jql I used to search for it](https://cloudifysource.atlassian.net/issues/?jql=labels%20%3D%20UI%20AND%20(summary%20~%20%22cancel%22%20OR%20description%20~%20%22cancel%22%20OR%20summary%20~%20%22abort%22%20OR%20description%20~%20%22abort%22)%20%20AND%20status%20in%20(Open)%20order%20by%20created%20DESC))), so I created one: RD-1695

PR #1236, opened 2021-03-11
RD-1661 - Increase max-height of LabelsList component

qooban:
Before:
image

Now:
image

### DIFF
--- a/widgets/common/src/labels/LabelsList.jsx
+++ b/widgets/common/src/labels/LabelsList.jsx
@@ -1,15 +1,23 @@
 import { sortLabels } from './common';
 
+const newLabelColor = 'blue';
+const labelLinesVisibleWithoutScroll = 6;
+const maxListHeight = `${labelLinesVisibleWithoutScroll * 2 + 0.2}em`;
+
 export default function LabelsList({ labels, onChange }) {
     const { Label, Icon } = Stage.Basic;
-    const newLabelColor = 'blue';
     const sortedLabels = sortLabels(labels);
 
     return (
         <div
             className="ui multiple dropdown"
-            /* maxHeight set to 12.2em guarantees 6 lines of labels to be visible without scroller */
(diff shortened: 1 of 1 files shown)


widgets/common/src/labels/LabelsList.jsx
```diff
@@ -8,8 +8,8 @@ export default function LabelsList({ labels, onChange }) {
     return (
         <div
             className="ui multiple dropdown"
-            /* maxHeight set to 8.2em guarantees 4 lines of labels to be visible without scroller */
-            style={{ paddingRight: '4.1em', minHeight: '2em', maxHeight: '8.2em', overflow: 'auto', maxWidth: '100%' }}
+            /* maxHeight set to 12.2em guarantees 6 lines of labels to be visible without scroller */
+            style={{ paddingRight: '4.1em', minHeight: '2em', maxHeight: '12.2em', overflow: 'auto', maxWidth: '100%' }}
```
Gelio, line 12: Optional:

Maybe instead of having a comment explaining, we could compute the `12.2em` :smile: ? That could show how to change the height in the future in a safe way

```ts
const labelLinesVisibleWithoutScroll = 6;
const maxListHeight = `${labelLinesVisibleWithoutScroll * 2 + 0.2}em`;
```

What do you think?

Gelio (dismissed):
A pleasure to review :+1: I appreciate the screenshot comparison

widgets/common/src/labels/LabelsList.jsx
qooban: I like the idea. Changed 👍🏻 

BTW I moved outside of the functional component `const`'s not needed for re-rendering.

Gelio: Yup, sounds good :slightly_smiling_face:

Gelio (approved):
:+1:

PR #1237, opened 2021-03-11
RD-1663 - Change fields order in Deploy Blueprint modal

qooban:
This PR introduce small re-ordering of the fields in Deploy Blueprint modal. The change was previously consulted with Product Manager and R&D Manager. See screenshots below (they present both flavors of the modal - with/without blueprint dropdown).

The change is cosmetic, so I expect one approval before merging it.

I decided to include small refactorings which are strictly connected with the area affected by the change:
* In `DeployBlueprintModal` updated toolbox prop type
* In `RevertToDefaultIcon` component added `data-testid` attribute to get the element easier in the tests
* In `deployment_button_spec`
  * Extracted `revertToDefaultValue` custom Cypress command and reused it across different test cases
  * Improved getting `Cancel` button
* In `deployment_actions_spec` reused `revertToDefaultValue` custom Cypress command

#### Before
image
image

#### After
image
image


widgets/common/src/DeployBlueprintModal.jsx
```diff
@@ -430,6 +415,25 @@ class DeployBlueprintModal extends React.Component {
...
+                        >
+                            <DynamicDropdown
+                                value={siteName}
+                                onChange={value => this.setState({ siteName: value })}
+                                name="siteName"
+                                fetchUrl="/sites?_include=name"
+                                valueProp="name"
+                                toolbox={toolbox}
+                            />
+                        </Form.Field>
```
qooban, line 435: This part was just moved from L373. 

Only one change: removed `(Optional)` from the help description as with red asterisks for mandatory fields it is not necessary IMHO. Also in the Labels field we do not have such prefix in help description, so that's another argument for me (consistency) not to have it here.

qooban:
Stage system tests results: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/267

widgets/common/src/RevertToDefaultIcon.jsx
```diff
@@ -8,7 +8,18 @@ function RevertToDefaultIcon({ value, defaultValue, onClick, onMouseDown, popupC
     }
 
     return (
-        <Popup trigger={<Icon name="undo" link onClick={onClick} onMouseDown={onMouseDown} style={style} />}>
+        <Popup
+            trigger={
+                <Icon
+                    data-testid="revert-to-default"
```
kubama, line 14: In this case, what's the benefit of using `data-testid` instead of `aria-label`?

widgets/common/src/DeployBlueprintModal.jsx
```diff
@@ -430,6 +415,25 @@ class DeployBlueprintModal extends React.Component {
...
                         )}
 
+                        <Form.Divider>
+                            {i18n.t('widgets.common.deployments.deployModal.deploymentMetadata')}
+                        </Form.Divider>
+
+                        <Form.Field
+                            error={errors.siteName}
+                            label="Site name"
+                            help="Specify a site to which this deployment will be assigned."
```
kubama, line 425: I can see `help` was modified, maybe it's a good moment for placing i18n here?

widgets/common/src/RevertToDefaultIcon.jsx
qooban: Good point, of course changed.

widgets/common/src/DeployBlueprintModal.jsx
qooban: IMHO I can either change all in this file or just newly added. In this case (help description for site name) it just removed `(Optional)` prefix, so I don't consider it is a newly added. I see that when you were adding Labels here you didn't move to 18n.

I considered moving to 18n, but I didn't do that at once because: there's quite big amount of strings in this file and it will significantly increase the size of this PR if I do that.

Anyways, if you both (@Gelio , @kubama ) are OK, I will move all strings to `i18n` within this PR.

kubama (approved):

widgets/common/src/RevertToDefaultIcon.jsx
```diff
@@ -8,7 +8,18 @@ function RevertToDefaultIcon({ value, defaultValue, onClick, onMouseDown, popupC
     }
 
     return (
-        <Popup trigger={<Icon name="undo" link onClick={onClick} onMouseDown={onMouseDown} style={style} />}>
+        <Popup
+            trigger={
+                <Icon
+                    aria-label="revert-to-default"
```
Gelio, line 14: `aria-label` should be a user-readable and a user-friendly string :slightly_smiling_face: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

It's like a tooltip

Thus, I suggest:

```suggestion
                    aria-label="Revert value to default"
```

test/cypress/support/deployments.ts
```diff
@@ -50,6 +50,11 @@ const commands = {
             cy.get('.input input').clear().type(deploymentId);
             cy.get('.input.loading').should('not.exist');
         });
+    },
+    revertToDefaultValue: () => {
+        const revertToDefaultAriaLabel = '[aria-label="revert-to-default"]';
+        cy.get(revertToDefaultAriaLabel).click();
+        cy.get(revertToDefaultAriaLabel).should('not.exist');
```
Gelio, line 57: Optional:

I have recently learned that this can be a single like, like done in https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Example-2-Reversing-the-Default-Assertion

```suggestion
        cy.get(revertToDefaultAriaLabel).click().should('not.exist');
```

Not sure if it's better, but I'm leaving it out here for your consideration :slightly_smiling_face:

Gelio (commented):
I like the new order better, it makes the form cleaner, so :+1: and thanks for all the local refactorings you've done :slightly_smiling_face: 

There is 1 comment that I think is worth addressing :slightly_smiling_face:

Gelio (commented):
I like the new order better, it makes the form cleaner, so :+1: and thanks for all the local refactorings you've done :slightly_smiling_face: 

There is 1 comment that I think is worth addressing :slightly_smiling_face:

widgets/common/src/DeployBlueprintModal.jsx
Gelio: I'm fine with that :slightly_smiling_face: Although doing that in a separate PR is probably not too much overhead for you, but would make it easier to review, as it would be a separate change :slightly_smiling_face:

qooban: 👍🏻 I'll do that that in a separate PR.

test/cypress/support/deployments.ts
qooban: Checked, works equally well, so changed.

widgets/common/src/RevertToDefaultIcon.jsx
Gelio: I had to double-check it :smile: so I guess it's new to both of us

Gelio (approved):
Looks good from my side :slightly_smiling_face:

kubama (approved):

PR #1238, opened 2021-03-12
RD-1591 Preparation for recursive widget rendering tests

Gelio:
This PR does the preparation work necessary for recursive widget rendering. Its aim is to have no breaking changes and very little impact overall. To gain context about how the changes done in this PR are exercised and why&how they are used, see #1239.

Short description of the point of these changes:
1. `WidgetsGrid` is exposed to allow rendering the `components/layout/Grid` (a grid of widgets).

    Note that there is a similarly named `Stage.Basic.Grid` component, but it's not the same one.
2. `renderWidgetGridItem` is exposed to allow easy rendering of `Widget`s alongside other components when using `WidgetsGrid`.

    Without it, the consumer would have to render `GridItem` and put `Widget` inside of it themselves, essentially doing the same work that `renderWidgetGridItem` is already doing.
3. Cypress' `widgets` commands will be modified to add the `interceptWidgetScript`, so I migrated it to TS in this PR to have it be a simple rename
4. `Widget`, `WidgetsList`, `WidgetsGrid` components are exposed to allow easy nested widget rendering. 
    They are exposed in `Stage.Common`, because exposing them in `Stage.Basic` caused circular references.
5. `@babel/cli` is installed and some other changed in the cypress ESLint confg and tsconfig are done to allow writing custom widgets as Cypress' `fixtures` and compiling them on the fly before the tests.

    This lets us write custom widgets in TypeScript and have them as plain TSX files (instead of ZIP files) in the codebase, which IMO is easier to maintain and explore.

## System tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/268/

The only failure is the known `widgets/deployment_button_spec.js`

### DIFF
--- a/app/components/shared/widgets/Widget.tsx
+++ b/app/components/shared/widgets/Widget.tsx
@@ -5,16 +5,16 @@ import i18n from 'i18next';
 import React, { Component, createRef, ErrorInfo, ReactElement } from 'react';
 import { connect, ConnectedProps, MapStateToProps } from 'react-redux';
 
-import { setValue } from '../actions/context';
-import { fetchWidgetData as fetchWidgetDataThunk } from '../actions/WidgetData';
-import EditWidget from './EditWidget';
-import stageUtils from '../utils/stageUtils';
-import { EditableLabel, ErrorMessage, Header, Icon, Loading, Message, ReadmeModal, Segment } from './basic';
-import WidgetDynamicContent from './WidgetDynamicContent';
-import type { ManagerData } from '../reducers/managerReducer';
-import type { ReduxState } from '../reducers';
-import type { Widget as WidgetObj } from '../utils/StageAPI';
-import { getWidgetDefinitionById, SimpleWidgetObj } from '../actions/page';
+import { setValue } from '../../../actions/context';
+import { fetchWidgetData as fetchWidgetDataThunk } from '../../../actions/WidgetData';
+import EditWidget from '../../EditWidget';
+import stageUtils from '../../../utils/stageUtils';
+import { EditableLabel, ErrorMessage, Header, Icon, Loading, Message, ReadmeModal, Segment } from '../../basic';
+import WidgetDynamicContent from '../../WidgetDynamicContent';
+import type { ManagerData } from '../../../reducers/managerReducer';
+import type { ReduxState } from '../../../reducers';
+import type { Widget as WidgetObj } from '../../../utils/StageAPI';
+import { getWidgetDefinitionById, SimpleWidgetObj } from '../../../actions/page';
 
 export interface WidgetOwnProps<Configuration> {
     isEditMode: boolean;
--- a/app/utils/StageAPI.ts
+++ b/app/utils/StageAPI.ts
@@ -203,7 +203,7 @@ declare global {
         const GenericConfig: typeof GenericConfigType;
         const Utils: typeof StageUtils;
 
-        // NOTE: Common items are mostly defined in widgets
+        // NOTE: Common items are defined in widgets
         // eslint-disable-next-line @typescript-eslint/no-empty-interface
         interface Common {}
         /** Common widget utilities */
--- a/app/utils/StageAPI.ts
+++ b/app/utils/StageAPI.ts
@@ -207,6 +207,7 @@ declare global {
         interface Common {
             Widget: typeof import('../components/Widget').default;
             WidgetsList: typeof import('../components/WidgetsList').default;
+            WidgetsGrid: typeof import('../components/WidgetsGrid').default;
         }
         /** Common widget utilities */
         const Common: Common;
--- a/app/utils/StageAPI.ts
+++ b/app/utils/StageAPI.ts
@@ -204,11 +204,8 @@ declare global {
         const Utils: typeof StageUtils;
 
         // NOTE: Common items are mostly defined in widgets
-        interface Common {
-            Widget: typeof import('../components/Widget').default;
-            WidgetsList: typeof import('../components/WidgetsList').default;
-            WidgetsGrid: typeof import('../components/WidgetsGrid').default;
-        }
+        // eslint-disable-next-line @typescript-eslint/no-empty-interface
+        interface Common {}
         /** Common widget utilities */
         const Common: Common;
         const defineCommon: <Name extends keyof Common>(definition: CommonOrPropTypeDefinition<Common, Name>) => void;
--- a/app/utils/StageAPI.ts
+++ b/app/utils/StageAPI.ts
@@ -203,9 +203,11 @@ declare global {
         const GenericConfig: typeof GenericConfigType;
         const Utils: typeof StageUtils;
 
-        // NOTE: Common items are defined in widgets
-        // eslint-disable-next-line @typescript-eslint/no-empty-interface
-        interface Common {}
+        // NOTE: Common items are mostly defined in widgets
+        interface Common {
+            Widget: typeof import('../components/Widget').default;
+            WidgetsList: typeof import('../components/WidgetsList').default;
+        }
         /** Common widget utilities */
         const Common: Common;
         const defineCommon: <Name extends keyof Common>(definition: CommonOrPropTypeDefinition<Common, Name>) => void;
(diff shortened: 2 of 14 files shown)


test/cypress/fixtures/widgets/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+    "extends": "../../../../widgets/.eslintrc.json"
+}
```
kubama, line 3: What's the reason for adding thid in this directory?

Gelio: See point number 5 in the PR description

qooban:
Planning to review this one after our today's weekly.

test/cypress/support/widgets.ts
```diff
@@ -1,8 +1,8 @@
 Cypress.Commands.add('getWidgets', () => cy.stageRequest('/console/widgets/list'));
 
 Cypress.Commands.add('removeCustomWidgets', () => {
-    cy.getWidgets().then(response => {
-        response.body.forEach(widget => {
+    (cy as any).getWidgets().then((response: any) => {
```
qooban, line 4: Just wondering - `cy` has `any` type as not all custom commands in `cypress/support` are migrated to TS? Or there's different reason?

package.json
```diff
@@ -94,6 +94,7 @@
     "uuid": "^3.3.2"
   },
   "devDependencies": {
+    "@babel/cli": "^7.13.10",
```
qooban, line 97: Why do we need `@babel/cli`?

I'm guessing it's probably for https://github.com/cloudify-cosmo/cloudify-stage/pull/1239? Optional: Maybe it should be moved there?

qooban (commented):
In general the code looks good 👍🏻 Left some comments.

I was not sure about some parts which seems to stick more to https://github.com/cloudify-cosmo/cloudify-stage/pull/1239 . But that's probably minor thing.

app/utils/StageAPI.ts
```diff
@@ -203,9 +203,12 @@ declare global {
...
         const Utils: typeof StageUtils;
 
-        // NOTE: Common items are defined in widgets
-        // eslint-disable-next-line @typescript-eslint/no-empty-interface
-        interface Common {}
+        // NOTE: Common items are mostly defined in widgets
+        interface Common {
+            Widget: typeof import('../components/Widget').default;
+            WidgetsList: typeof import('../components/WidgetsList').default;
+            WidgetsGrid: typeof import('../components/WidgetsGrid').default;
```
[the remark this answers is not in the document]
Gelio: Right, I forgot about `Stage.Shared`. Makes sense to move the components there :slightly_smiling_face: I'll do that

test/cypress/support/widgets.ts
Gelio: The actual reason is that since I made 2 PRs anyway, I decided to split the migration of this file to TS into 2 parts. This PR only renames the file and gets it to compile (in a hacky way :smile: ), and the proper migration is done in #1239 :slightly_smiling_face: If I made the full migration here, GitHub would not recognize the rename :slightly_smiling_face: 

In #1239, there are no `any` in this file :slightly_smiling_face:

package.json
Gelio: Yeah, right :smile: Good guess, it is used in #1239. I wanted to balance out the number of changes in both PRs, so I put the installation of it here, but it makes more sense to have babel-related parts in the other PR, when it is actually used.

Since I am going to be making changes to this PR (and probably #1239) anyway, I'll move the babel-related parts there

Thanks for the suggestion

kubama:
My general comment is that I think this PR shouldn't be separated from #1239. Looks like most changes here do not make much sense per se, and have to be reviewed in the context of #1239. I think this makes the review more difficult than if there was a single PR.

Gelio:
@kubama Thanks for the opinion :slightly_smiling_face: When splitting the single branch into 2 I had the idea of smaller PRs in mind, as well as the approach of:
1. Make the change easy
2. Make the easy change

I agree that the changes in this PR are themselves not useful and not visible from the app standpoint, but they are the breaking changes in the application, and thus, can be isolated. You can say that #1239 is merely optional - that's just tests. Without this PR, using the feature in production would be harder

As the author, I do not have a preference over whether the work is split or not here. @qooban let me know if you would also prefer to have a single PR rather than 2, and I will close this one

qooban:
@Gelio 
> As the author, I do not have a preference over whether the work is split or not here. @qooban let me know if you would also prefer to have a single PR rather than 2, and I will close this one

I see advantages and disadvantages of the splitting in this case. Maybe it was not perfect, but for me it can stay as it is. 
Anywasy, I'm OK with both approaches (having 1 or 2 PRs) as long as the PRs are not big.

kubama (approved):

app/utils/StageAPI.ts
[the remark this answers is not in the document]
Gelio: I've moved the 3 components there. Unfortunately, git now does not moving `WidgetsList` as a rename :worried:

package.json
Gelio: I've moved the commit that installs `@babel/cli` to #1239

Gelio:
I've applied both suggestions to this PR:
1. Move `Stage.Common.Widget*` components to `Stage.Shared.Widgets.*`
2. Move the commit that installs `@babel/cli` to #1239

app/components/shared/widgets/Widget.tsx
```diff
@@ -5,16 +5,16 @@ import i18n from 'i18next';
...
+import { setValue } from '../../../actions/context';
+import { fetchWidgetData as fetchWidgetDataThunk } from '../../../actions/WidgetData';
+import EditWidget from '../../EditWidget';
+import stageUtils from '../../../utils/stageUtils';
+import { EditableLabel, ErrorMessage, Header, Icon, Loading, Message, ReadmeModal, Segment } from '../../basic';
+import WidgetDynamicContent from '../../WidgetDynamicContent';
+import type { ManagerData } from '../../../reducers/managerReducer';
+import type { ReduxState } from '../../../reducers';
+import type { Widget as WidgetObj } from '../../../utils/StageAPI';
+import { getWidgetDefinitionById, SimpleWidgetObj } from '../../../actions/page';
```
qooban, line 17: General:

I'm wondering would it be possible to introduce similar functionality as we have in some other places to resolve modules from the root directory (that would allow us to avoid such long `../../../` in the import paths)? 🙂

qooban (commented):
Looks good in general 👍🏻

app/utils/StageAPI.ts
```diff
@@ -203,7 +203,7 @@ declare global {
         const GenericConfig: typeof GenericConfigType;
         const Utils: typeof StageUtils;
 
-        // NOTE: Common items are defined in widgets
+        // NOTE: Common items are mostly defined in widgets
```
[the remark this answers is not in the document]
Gelio: Oh, yeah, good catch, let me quickly fix that

app/components/shared/widgets/Widget.tsx
Gelio: Makes sense :slightly_smiling_face: I can add a ticket to support resolving paths with `app` as root

qooban: Sounds good 👍🏻

app/utils/StageAPI.ts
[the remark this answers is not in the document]
Gelio: Done :slightly_smiling_face:

qooban (approved):

app/components/shared/widgets/Widget.tsx
Gelio: Here it goes: https://cloudifysource.atlassian.net/browse/RD-1724

Gelio:
I'm waiting for the system tests for #1239 to pass and I'll be merging both PRs to master then

Gelio:
The system tests failed to reasons related to the backend. See https://github.com/cloudify-cosmo/cloudify-stage/pull/1239#issuecomment-800876704 for more information. As agreed, I'll be merging this PR anyway since the test suite ran locally passed

PR #1239, opened 2021-03-12
RD-1591 Add tests for rendering widgets recursively

Gelio:
This PR builds upon https://github.com/cloudify-cosmo/cloudify-stage/pull/1238 and adds the tests that verify rendering nested widgets.

The tests verify that:
1. Rendering nested widgets works
2. Nested widgets can make API requests, have their own data, etc (_Plugins Catalog_ fetches and displays the list of plugins)
3. Widgets can pass configuration to nested widgets

Additional checks I have done by hand:
1. Showing nested widget's readme works
2. Interacting with the nested widget works (I wanted to install some plugin using _Plugins Catalog_)

https://user-images.githubusercontent.com/889383/110922882-8ba57280-8320-11eb-8f44-768c22a229b2.mp4

### New patterns used in this PR
1. Widgets can be rendered inside other widgets, thanks to components exposed in https://github.com/cloudify-cosmo/cloudify-stage/pull/1238
2. Custom test-only widgets are defined in `fixtures/widgets` in TypeScript. They are compiled by babel before the test starts and the compiled script is injected in the intercepted request.
    This seems to work fine, have little overhead, and improve code explorability and maintainability (can view and modify source directly, instead of zipping it)

### Known issues
1. Maximizing the nested widgets does not work.

    Even when I simulated the `onWidgetUpdate` function and changed `maximized = true` when the maximization button is pressed, the icon changed, but the component was not maximized. I didn't see a way to mitigate it and actually allow nested widgets to be maximized (this is probably related to how maximization is handled in parent components, e.g. `Page`, `PageContent`).

    Thus, I left it intact. In RD-1224 I will most likely add some `Widget` prop to hide the maximization button when it does not make sense (when rendering nested widgets).
2. The nested grid does not seem to be aligned well.
    Upon initial render, the nested grid rendered inside `SplitViewWidget` is not aligned, even though all grid-related properties (`x`, `y`, `width`, `height`) are set.

    image

    I assume this is an issue with Cypress + react-grid-layout in combination.

    The problem goes away when I click the "Toggle device toolbar" in DevTools, or switch to another page and come back (which forces the grid to remount).

    Uploading recording-2021-03-12_10.43.19.mp4…

    Thus, this is probably an issue with the tests, and will not appear in the actual app.

## Cypress test run

https://user-images.githubusercontent.com/889383/110922216-b8a55580-831f-11eb-9aa6-1d257b018bd1.mp4

image
the blue overlay comes from Cypress' pinning the UI state

image

## System tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/283/

Run 2021-03-16 14:24 CET

(failed due to errors in the Manager - the nightly build failed too. See https://github.com/cloudify-cosmo/cloudify-stage/pull/1239#issuecomment-800876704 for more information)

### DIFF
--- a/test/cypress/fixtures/widgets/splitViewWidget.tsx
+++ b/test/cypress/fixtures/widgets/splitViewWidget.tsx
@@ -3,7 +3,7 @@ Stage.defineWidget({
     name: 'Split view widget',
     isReact: true,
     render: () => {
-        const { WidgetsGrid } = Stage.Common;
+        const { WidgetsGrid } = Stage.Shared.Widgets;
 
         return (
             <div>
--- a/test/cypress/fixtures/widgets/splitViewWidget.tsx
+++ b/test/cypress/fixtures/widgets/splitViewWidget.tsx
@@ -1,0 +1,41 @@
+Stage.defineWidget({
+    id: 'split-view-widget',
+    name: 'Split view widget',
+    isReact: true,
+    render: () => {
+        const { WidgetsGrid } = Stage.Common;
+
+        return (
+            <div>
+                This is a split view widget
+                <WidgetsGrid isEditMode={false} onGridDataChange={_.noop}>
+                    <WidgetsGrid.Item height={4} width={4} key="content-left" id="content-left">
+                        Content on the left
+                    </WidgetsGrid.Item>
+                    <WidgetsGrid.Item height={4} width={4} x={10} key="content-right" id="content-right">
+                        Content on the right, above plugins catalog
+                    </WidgetsGrid.Item>
+                    {WidgetsGrid.renderWidgetGridItem({
+                        isEditMode: false,
+                        onWidgetRemoved: _.noop,
+                        onWidgetUpdated: _.noop,
+                        widget: {
+                            id: 'pluginsCatalog',
+                            configuration: {
+                                jsonPath: 'http://repository.cloudifysource.org/cloudify/wagons/plugins.json'
+                            },
+                            definition: 'pluginsCatalog',
+                            drillDownPages: {},
+                            height: 20,
+                            width: 10,
+                            maximized: false,
+                            name: 'Plugins Catalog',
+                            x: 0,
+                            y: 4
+                        }
+                    })}
+                </WidgetsGrid>
+            </div>
+        );
+    }
+});
--- a/test/cypress/integration/widgets/user_defined_widgets_spec.ts
+++ b/test/cypress/integration/widgets/user_defined_widgets_spec.ts
@@ -1,0 +1,28 @@
+describe('User-defined widgets', () => {
+    describe('Fibonacci sequence widget', () => {
+        before(cy.activate);
+
+        beforeEach(() => {
+            cy.compileScriptFixture('widgets/fibonacciSequenceWidget.tsx').then(compiledScriptSource =>
+                cy.interceptWidgetScript('fibonacci-sequence-widget', compiledScriptSource)
+            );
+        });
+        const initialSequence = [1, 1];
+
+        it('render itself once with the initial sequence', () => {
+            cy.usePageMock('fibonacci-sequence-widget', { targetSequenceLength: initialSequence.length }).mockLogin();
+
+            cy.contains(`Current sequence: ${JSON.stringify(initialSequence)}`);
+        });
+
+        it('render itself recursively with intermediate sequences', () => {
+            const sequence = [1, 1, 2, 3, 5, 8, 13, 21, 34];
+            cy.usePageMock('fibonacci-sequence-widget', { targetSequenceLength: sequence.length }).mockLogin();
+            const initialSequenceLength = initialSequence.length;
+
+            Array.from({ length: sequence.length - initialSequenceLength }).forEach((_, index) => {
+                cy.contains(`Current sequence: ${JSON.stringify(sequence.slice(0, initialSequenceLength + index))}`);
+            });
+        });
+    });
+});
--- a/test/cypress/integration/widgets/user_defined_widgets_spec.ts
+++ b/test/cypress/integration/widgets/user_defined_widgets_spec.ts
@@ -25,4 +25,24 @@ describe('User-defined widgets', () => {
             });
         });
     });
+
+    describe('Split view widget', () => {
+        before(cy.activate);
+
+        beforeEach(() => {
+            cy.compileScriptFixture('widgets/splitViewWidget.tsx').then(compiledScriptSource =>
+                cy.interceptWidgetScript('split-view-widget', compiledScriptSource)
+            );
+        });
+
+        it('should render itself and the nested Plugins Catalog widget', () => {
+            cy.usePageMock('split-view-widget').mockLogin();
+
+            cy.contains('Content on the left');
+            cy.contains('Content on the right');
+            cy.contains('Plugins Catalog');
+            // NOTE: make sure the content of the widget is rendered
+            cy.get('th').contains('Name');
+        });
+    });
 });
--- a/test/cypress/integration/widgets/user_defined_widgets_spec.ts
+++ b/test/cypress/integration/widgets/user_defined_widgets_spec.ts
@@ -2,22 +2,21 @@ describe('User-defined widgets', () => {
     before(cy.activate);
 
     describe('Fibonacci sequence widget', () => {
-        beforeEach(() => {
-            cy.compileScriptFixture('widgets/fibonacciSequenceWidget.tsx').then(compiledScriptSource =>
-                cy.interceptWidgetScript('fibonacci-sequence-widget', compiledScriptSource)
-            );
-        });
         const initialSequence = [1, 1];
+        const widgetFixturePath = 'widgets/fibonacciSequenceWidget.tsx';
+        const widgetId = 'fibonacci-sequence-widget';
 
         it('render itself once with the initial sequence', () => {
-            cy.usePageMock('fibonacci-sequence-widget', { targetSequenceLength: initialSequence.length }).mockLogin();
+            const widgetConfiguration = { targetSequenceLength: initialSequence.length };
+            useMockWidgetFixture(widgetFixturePath, widgetId, widgetConfiguration);
 
             cy.contains(`Current sequence: ${JSON.stringify(initialSequence)}`);
         });
 
         it('render itself recursively with intermediate sequences', () => {
             const sequence = [1, 1, 2, 3, 5, 8, 13, 21, 34];
-            cy.usePageMock('fibonacci-sequence-widget', { targetSequenceLength: sequence.length }).mockLogin();
+            const widgetConfiguration = { targetSequenceLength: sequence.length };
+            useMockWidgetFixture(widgetFixturePath, widgetId, widgetConfiguration);
             const initialSequenceLength = initialSequence.length;
 
             Array.from({ length: sequence.length - initialSequenceLength }).forEach((_, index) => {
@@ -27,14 +26,11 @@ describe('User-defined widgets', () => {
     });
 
     describe('Split view widget', () => {
-        beforeEach(() => {
-            cy.compileScriptFixture('widgets/splitViewWidget.tsx').then(compiledScriptSource =>
-                cy.interceptWidgetScript('split-view-widget', compiledScriptSource)
-            );
-        });
+        const widgetFixturePath = 'widgets/splitViewWidget.tsx';
+        const widgetId = 'split-view-widget';
 
         it('should render itself and the nested Plugins Catalog widget', () => {
-            cy.usePageMock('split-view-widget').mockLogin();
+            useMockWidgetFixture(widgetFixturePath, widgetId);
 
             cy.contains('Content on the left');
             cy.contains('Content on the right');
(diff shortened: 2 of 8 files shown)


kubama (commented):
I'm quite surprised by what's happening here.
So far we were using cypress to test production level features.
I think that what you are testing here is a candidate for a unit test, not an integration test. There is quite some overhead to make it testable with cypress.
I understand that at some point in the near future the feature you implemented will be used in production solution, so it will be much easier to test it as it will be readily available. No additional overhead will be needed to access the feature. And what then, are you going to remove the tests you added here and replace them with tests working on production solution?

qooban (commented):
I went quickly through the PR description and the changes. Will read the code deeper on Monday, but generally I must say that I like the idea of compiling the widget's on the fly for the testing purposes 👍🏻 

We maybe would like to divide production-like testing from development-like (or artificial) testing we have here, but still I see benefits:
* we can test custom widgets easier (widgets archives are hard to maintain as the code itself is hidden)
* we can provide kind of Cypress testing suite for custom widget developers

kubama:
> We maybe would like to divide production-like testing from development-like (or artificial) testing we have here

If this is going to stay then I think we should divide it.

But I still don't understand why we can't just test this widget-in-widget functionality once it's reachable from production code.

I think we recently had a similar situation in labels. @qooban developed label management modal but it was not yet part of any production feature, so temporarily there were no tests covering it. Tests were added once the modal was used in widgets. I can see no issue with it. Is there any reason why we couldn't do the same with nested widgets feature?

> * we can test custom widgets easier (widgets archives are hard to maintain as the code itself is hidden)

Not sure what you mean, we are not testing custom widgets anywhere. What we're testing is using custom widgets in the application (installation, upgrading, removing and adding to the page). Widgets that we use for this purpose are just black-boxes and we don't care about their actual implementation. And we still need zip packages to test this functionality. So the approach proposed in this PR brings no help to this area.

test/cypress/integration/widgets/user_defined_widgets_spec.ts
```diff
@@ -0,0 +1,48 @@
+describe('User-defined widgets', () => {
```
qooban, line 1: Seems to me that `usePageMock` and `interceptWidgetScript` should be executed "closer" to each other as the setup of each test depends on both. So I would consider adding a function which defines interceptor and page mock.

test/cypress/integration/widgets/user_defined_widgets_spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    describe('Fibonacci sequence widget', () => {
+        before(cy.activate);
```
qooban, line 3: You can move `cy.activate` to a root `describe`, one level above, I think.

test/cypress/integration/widgets/user_defined_widgets_spec.ts
qooban, line 1: As we already mentioned that in general comments, I think it would make sense to kind of separate this type of tests from the rest (as this is not production-like). I don't know what are the best solutions for that in Cypress. Maybe separate folder? Maybe environment variable? Maybe custom configuration for that test?

qooban (commented):
Added a bit more comments after second review.

Tested the code internally, works quite good. I think we can disable maximize icon for widget-in-widget, I don't think anyone will need it.

@kubama 
> But I still don't understand why we can't just test this widget-in-widget functionality once it's reachable from production code.

I assume that there will be production code test anyway in the future. @Gelio Am I correct?

> And we still need zip packages to test this functionality. So the approach proposed in this PR brings no help to this area.

I'm see it differently. With that changes from this PR it's much closer. On-the-fly babel transpiling is one thing. If we add in-test zipping of a specific folder we have full solution.

And the second bullet I mentioned here - Cypress testing suite for custom widget developers - is even more interesting for me and it is aligned in the long term direction of improving custom widget development ([RD-237](https://cloudifysource.atlassian.net/browse/RD-237)).

kubama:
> I assume that there will be production code test anyway in the future. @Gelio Am I correct?

If so, then why do we need this complex solution to test it? I can see no point.
I can understand this solution may be valuable in the context of future developments, as you pointed out. I just don't think it's applicable in this particular case (testing implementations that are not yet used in production code).

> I'm see it differently. With that changes from this PR it's much closer. On-the-fly babel transpiling is one thing. If we add in-test zipping of a specific folder we have full solution.

OK, but it would be a very complex solution to address a really minor issue. I can't remember any changes that ever required updating these zip files. Not sure if the effort would be worth it.

> And the second bullet I mentioned here - Cypress testing suite for custom widget developers - is even more interesting for me and it is aligned in the long term direction of improving custom widget development ([RD-237](https://cloudifysource.atlassian.net/browse/RD-237)).

Very cool, I'm all for it. But since I can see no justification for using this solution in the scope of this PR I think it should be developed within RD-237.

Gelio:
@qooban @kubama Looks like we are not on the same page when it comes to these tests :slightly_smiling_face: I reckon it might be faster and more productive to have a meeting today/tomorrow (depending on the bug scrub progress) to quickly get up to speed and decide how to proceed :slightly_smiling_face: I'll schedule one once I'm done with the bug scrub

test/cypress/fixtures/widgets/splitViewWidget.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <div>
+                This is a split view widget
+                <WidgetsGrid isEditMode={false} onGridDataChange={_.noop}>
+                    <WidgetsGrid.Item height={4} width={4} key="content-left" id="content-left">
+                        Content on the left
+                    </WidgetsGrid.Item>
+                    <WidgetsGrid.Item height={4} width={4} x={10} key="content-right" id="content-right">
+                        Content on the right, above plugins catalog
+                    </WidgetsGrid.Item>
+                    {WidgetsGrid.renderWidgetGridItem({
```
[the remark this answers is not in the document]
Gelio: Hmm... I'm not sure what exactly you have in mind, so let me provide a couple of options :smile: Widgets can:

1. Render `WidgetsList` with multiple widgets (but can be a single one)
2. A single `Widget` component
3. If they want to have a split view, like here (and in the Deployments view, most likely), we can define the `Grid` and then use this existing function

What usage do you have in mind?

test/cypress/integration/widgets/user_defined_widgets_spec.ts
Gelio: I thought about it for some time and I fail to see the benefits of extracting making this test suite something separate :thinking: Yes, it does not interact with the production application, so that's definitely the differentiating factor, but do we care :smile: ? Personally, I don't think I do, but I also don't see the problem with moving it to some separate folder :slightly_smiling_face: 

The thing I'm not sure about is what the benefits will be of moving it to a separate directory. It should be run on the same conditions the other tests are run, and when it fails, the same thing should happen when other tests fail. So I see it that separation will only allow the developers to easily see that this test is different :slightly_smiling_face: 

I did not see any concrete guidelines about it for Cypress. The only similar thing I found was to run a different set of tests by their path https://stackoverflow.com/questions/50993299/easy-way-of-running-different-groups-of-cypress-tests which means that this test suite would have to be put into a separate directory (which is fine)

Still, let me know if you see other benefits that I do not see, or you think that such tests should be handled differently from other tests :slightly_smiling_face:

test/cypress/integration/widgets/user_defined_widgets_spec.ts
Gelio: Good point, some deduplication is always welcome :slightly_smiling_face:

test/cypress/integration/widgets/user_defined_widgets_spec.ts
Gelio: Right, that does make sense. I will make it a local function in this file. 

To provide more info, I don't want to make it a global custom command, I'm worried that we will produce a very specialized `compileFixtueWidgetAndShowMockPage` (name blown out of proportion on purpose :smile: ) command that will be used just in this very specialized case :smile: Whereas the `compileScripFixture` and `interceptWidgetScript` are a bit more general

test/cypress/fixtures/widgets/splitViewWidget.tsx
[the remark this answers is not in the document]
qooban: I'll describe what I had in mind 🙂 

We've got list of pages, e.g. https://github.com/cloudify-cosmo/cloudify-stage/blob/master/templates/pages/deployment.json - which file content's are saved in the redux store.

I was thinking about a function which could process a predefined page and render all the content (widgets/tabs from it) in a widget.

test/cypress/integration/widgets/user_defined_widgets_spec.ts
qooban: Writing "Maybe separate folder? Maybe environment variable? Maybe custom configuration for that test?" I was just loudly thinking about potential solutions for logical separation 🙂 I'm not saying any of listed should be implemented. I'm more interested in having logically separated test suites with "Cypress blessing" meaning implemented according to Cypress best practices. 

Let's discuss on the meeting.

BTW One of the reasons I'm interested in that separation is that in the long-term I see that as an entrance for having multiple test groups. This is something we will be introducing anyway sooner or later. I'll probably talk about it on testing roadmap review somewhere in March/April.

kubama (approved):

test/cypress/fixtures/widgets/splitViewWidget.tsx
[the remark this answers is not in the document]
Gelio: Hmm... I see :smile: That could be possible, if there are use cases for it :slightly_smiling_face:  What would be the benefits in this case?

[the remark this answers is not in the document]
qooban: 🙂 There is use case. The reason why we are doing all that stuff (widget in widget) is to include current deployment drill down page (https://github.com/cloudify-cosmo/cloudify-stage/blob/master/templates/pages/deployment.json) inside the widget, as the right pane of Deployment View widget 🙂 Isn't it?

The benefits I see, at least right now, are the following:
1. We keep the definition of the list of widgets/tabs in the already known structure (page JSON)
2. We can modify page in Template Management feature, which makes it easier to make any changes in the widgets/tabs settings
3. We can allow users to include their custom pages inside the their widgets to create better UIs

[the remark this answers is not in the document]
Gelio: Oh, ok. I thought I will just hardcode the widgets in the details pane of Environments :slightly_smiling_face: I'll keep that in mind as I will be working on it, thanks!

test/cypress/integration/widgets/user_defined_widgets_spec.ts
Gelio: Extracted `useMockWidgetFixture` :slightly_smiling_face:

test/cypress/integration/widgets/user_defined_widgets_spec.ts
Gelio: Done

Gelio:
I've rebased the branch to apply the suggestions from the comments

test/cypress/integration/widgets/user_defined_widgets_spec.ts
qooban: Looks good 👍🏻

qooban (dismissed):
I don't have any good proposal for another place for `user_defined_widgets_spec.ts`, so I think we can keep it as it is for now. 

BTW how long it normally takes to execute that test?

Gelio:
@qooban 

> BTW how long it normally takes to execute that test?

I would have shown you the system test logs, but they are gone :smile: As I checked last time - 19 seconds in CI

---

Thanks for the approvals. I queued a new set of system tests to run for this branch just now. I'll be waiting for them to pass, and then I plan on merging both this and #1238 to master :slightly_smiling_face: (starting with #1238 )

Gelio:
The system tests failed since the CI is broken at the moment (remember yesterday's nightly started failing). The error is in the before-all hook for the tests that ran on Jenkins:
image

Meaning that `cy.activate()` failed. This does not seem to be the issue with this specific test suite

I ran the `user_defined_widgets_spec.ts` test suite locally, and it passed fine:

https://user-images.githubusercontent.com/889383/111433345-76548d80-86fe-11eb-9094-3ea213a4153d.mp4

Thus, I suggest merging both this and #1238 anyway to get the task closed.

@qooban do you have anything against it :slightly_smiling_face: ?

qooban:
@Gelio Yes, I'm OK with merging. 
I'm going to report an issue or few issues on system tests, soon. Looks like some fixes in the backend didn't solve all the problems we had.

qooban:
@Gelio As I updated `ui_components_link` in the docs, I think we should merge https://github.com/cloudify-cosmo/cloudify-stage/pull/1246 first.

Gelio:
@qooban I'm not sure if there will be any conflicts or issues related to that fact, but sure, go ahead :smile: 

---

I have rebased the branch on `master` to get rid of the duplicated commits on this branch, since the previous base branch (#1238 ) was merged

qooban:
@Gelio I suspect so as I changed the `ui_components_link` to 2.4.4 and you have here cloudify-ui-components v2.4.3.

Gelio:
Oh yeah, right. I thought the CI ensured that the version in the docs is >= than the one in the repo, but looks like they have to match. Thus, I have to wait anyway :smile:

Gelio:
I'm looking for a re-approve now that #1238 and #1246 are merged :slightly_smiling_face:

qooban (approved):

PR #1240, opened 2021-03-12
RD-1648 - Update deployment's test case after changes in the REST API

qooban:
https://github.com/cloudify-cosmo/cloudify-manager/pull/2766 changed the following:
1. Reverted constraint validation in deployment API
2. Introduced a small change in the error message for the constraint validation. 

This PR adjusts our test code for the second change.

Error message is now prefixed with `dsl_parser.exceptions.ConstraintException:`:
image

I think we can keep that prefix. I don't think that cutting it on the UI side is good idea.
Our Stage System Tests should all pass after this PR is merged.

### DIFF
--- a/test/cypress/integration/widgets/deployment_button_spec.js
+++ b/test/cypress/integration/widgets/deployment_button_spec.js
@@ -218,7 +218,7 @@ describe('Create Deployment Button widget', () => {
             cy.wait('@deployBlueprint');
 
             cy.get('div.error.message > ul > li').should(
-                'have.text',
+                'contain.text',
                 'Value CentOS 7.6 of input string_constraint_pattern violates ' +
                     'constraint pattern(Ubuntu \\d{2}\\.\\d{2}) operator.'
             );
--- a/test/cypress/integration/widgets/deployment_button_spec.js
+++ b/test/cypress/integration/widgets/deployment_button_spec.js
@@ -196,8 +196,7 @@ describe('Create Deployment Button widget', () => {
         it('parses constraint error message from /deployments REST API', () => {
             selectBlueprintInModal('string');
 
-            const deploymentName = 'test';
-
+            const deploymentName = `${resourcePrefix}constraintError`;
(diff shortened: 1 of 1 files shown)


test/cypress/integration/widgets/deployment_button_spec.js
```diff
@@ -196,8 +196,7 @@ describe('Create Deployment Button widget', () => {
         it('parses constraint error message from /deployments REST API', () => {
             selectBlueprintInModal('string');
 
-            const deploymentName = 'test';
-
+            const deploymentName = `${resourcePrefix}constraintError`;
```
Gelio, line 199: General question:

What is the benefit to changing the name here?

Gelio (approved):
Looks good :slightly_smiling_face: I'm looking forward to having this fixed :smile:

Gelio:
Also, will you be running the system tests for this PR to make sure all tests pass now?

test/cypress/integration/widgets/deployment_button_spec.js
qooban: In `before` we remove all the deployments which name starts with a `resourcePrefix`:
https://github.com/cloudify-cosmo/cloudify-stage/blob/ca77549188317b042369739c96bbcae48ff5b37a/test/cypress/integration/widgets/deployment_button_spec.js#L11-L13
It helps to prepare for the test and not have issues like (deployment duplication).

This test case had different deployment name which caused problems when re-running the test locally. In the CI it does not happen (as we don't re-run the whole test).

Gelio: Cool, thanks for the explanation :slightly_smiling_face:

qooban:
> Also, will you be running the system tests for this PR to make sure all tests pass now?

I could do so, but I feel quite confident it works. Locally it works (and the problem we encountered was reproducible locally). In case of the problems not reproducible locally I would definitely give it a try, but in this case I'd prefer to merge and see :-)

Gelio:
Sounds good :slightly_smiling_face:

kubama (approved):

PR #1241, opened 2021-03-12
RD-1663 - Add i18n support to Deploy Blueprint modal

qooban:
As a continuation of https://github.com/cloudify-cosmo/cloudify-stage/pull/1237 I added i18n support to Deploy Blueprint modal.

### DIFF
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -400,9 +400,62 @@
                     "updateButton": "Update"
                 },
                 "deployModal": {
-                    "deploymentMetadata": "Deployment metadata",
-                    "executionParameters": "Execution parameters",
-                    "labelsLabel": "Labels"
+                    "inputs": {
+                        "deploymentName": {
+                            "label": "Deployment name",
+                            "help": "Specify a name for this deployment instance."
+                        },
+                        "blueprintName": {
+                            "label": "Blueprint",
+                            "help": "Select the blueprint based on which this deployment will be created."
+                        },
+                        "siteName": {
+                            "label": "Site name",
+                            "help": "Specify a site to which this deployment will be assigned."
+                        },
+                        "labels": {
+                            "label": "Labels"
+                        },
+                        "skipPluginsValidation": {
+                            "label": "Skip plugins validation",
+                            "message": "The recommended path is uploading plugins as wagons. This option is designed for plugin development and advanced users only."
+                        },
+                        "runtimeOnlyEvaluation": {
+                            "label": "Runtime only evaluation",
+                            "help": "If set, then get_property and get_input intrinsic functions will be evaluated on demand at runtime. If not set, then evaluation will be done at deployment creation time."
+                        },
+                        "deploymentInputs": {
+                            "loading": "Loading inputs...",
+                            "noInputs": "No inputs available for the blueprint"
+                        }
+                    },
+                    "sections": {
+                        "deploymentMetadata": "Deployment metadata",
+                        "executionParameters": "Execution parameters"
+                    },
+                    "buttons": {
+                        "deploy": "Deploy",
+                        "deployAndInstall": "Deploy & Install"
+                    },
+                    "steps": {
+                        "deploy": {
+                            "validatingData": "1/2: Validating data...",
+                            "deployingBlueprint": "2/2: Deploying blueprint..."
+                        },
+                        "deployAndInstall": {
+                            "validatingData": "1/4: Validating data...",
+                            "deployingBlueprint": "2/4: Deploying blueprint...",
+                            "waitingForDeployment": "3/4: Creating deployment environment...",
+                            "installingDeployment": "4/4: Installing deployment..."
+                        }
+                    },
+                    "errors": {
+                        "loadingYamlFileFailed": "Loading values from YAML file failed: {{ reason }}",
+                        "noBlueprintName": "Please select blueprint from the list",
+                        "noDeploymentName": "Please provide deployment name",
+                        "deploymentCreationFailed": "Deployment {{ deploymentId }} environment creation failed. {{ error }}",
+                        "deploymentInstallationFailed": "Deployment {{ deploymentId }} installation failed: {{ error }}"
+                    }
                 }
             },
             "labels": {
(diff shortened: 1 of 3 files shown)


app/translations/en.json
```diff
@@ -413,8 +466,10 @@
                 "valuePlaceholder": "Select or type in a value",
                 "validationError": "Only letters, digits and the characters \"-\", \".\" and \"_\" are allowed and it must begin with a letter.",
                 "invalidKeyError": "Allowed `{{ keyPrefix }}` prefixed labels are: {{ reservedKeys }}. All other `{{ keyPrefix }}` prefixed labels are reserved for internal use and cannot be added.",
-                "inputName": "Labels list",
-                "inputHelp": "A list of labels - key-value pairs - that can be assigned with a deployment",
+                "input": {
+                    "label": "Labels",
```
qooban, line 470: Intentionally changed `Labels list` to `Labels`. I think it is good enough and I'm using it in both places - Manage/Add Labels modals and Deploy Blueprint modal.

app/translations/en.json
```diff
@@ -400,9 +400,62 @@
                     "updateButton": "Update"
                 },
                 "deployModal": {
-                    "deploymentMetadata": "Deployment metadata",
-                    "executionParameters": "Execution parameters",
-                    "labelsLabel": "Labels"
+                    "inputs": {
```
qooban, line 403: For the `inputs` I used similar structure as in `blueprintUpload` section.

widgets/common/src/DeployBlueprintModal.jsx
```diff
@@ -1,6 +1,5 @@
-/**
- * Created by kinneretzin on 05/10/2016.
- */
+const { i18n } = Stage;
+const t = (key, options) => i18n.t(`widgets.common.deployments.deployModal.${key}`, options);
```
kubama, line 2: I'm wondering if using such shortcut function is a good pattern. In terms of code reuse this seems to be fine, but in practice it's a common scenario when working with translation file to check if/where given i18n key is used in the code. This won't work with such functions where i18n key is concatenated. I would say i18n key is a single identifier and as such should be used as a whole.

qooban: I had those concerns when I introduced it here, but I realized that I never look for the whole key string when searching for key usage. We have quite deep structure in `en.json` (in `deployModal` we get to 6 nesting levels) and I normally use the last part or last 2-3 parts of the key in most of the cases.

Anyways, from my perspective it's really minor issue, doesn't really matter much for me what solution will we use, so if you both see that such code reuse is bad, then I'm perfectly OK with changing it to use full key.

kubama: > I normally use the last part or last 2-3 parts of the key in most of the cases.

It's of course dead easy to use only the last part (just copy a single word) but this is sometimes insufficient. But in WebStorm (I guess in any modern IDE in fact) it's equally easy to copy the full key (Ctrl+Shift+Alt+C), so the nesting level is not relevant at all. Using last 2-3 parts requires some additional effort, as you need to copy the full path and then truncate it manually - not very handy.
One of the common headaches when maintaining translation files are unused keys that are bloating the file. So I think it's quite important to be able to locate key usages easily so the translation file maintenance could be done efficiently.

qooban: @Gelio Please add your 5 cents to the discussion. I see that you are using variables for prefixing i18n keys in Deployments View widget.

Anyways, I prefer to not spend more time on the long discussions about this topic. I'm OK with both approaches (variable prefix or full keys), it doesn't matter much for me, so please let me know which to choose.

Gelio: I'm leaning towards @qooban's take on this matter. Less duplication is always welcome, especially considering the fact that the strings passed to `i18n.t` can be arbitrary since they are not validated at compile/development time in any way. By duplicating the full string every time we have more ways to make typos (rare, but still). It's also more tedious for me to read the same string every time just to make sure it's from the same place :smile: 

If removal of unused entries from the translation file is a concern, one idea to mitigate it would be to set up a practice similar to _code coverage_ - translation coverage. We could monkey-patch `i18n.t` and check the strings passed, and then cross-check it against the translations file to make sure:
1. There are no unused strings
2. No invalid strings are used
but that's probably a minor concern :slightly_smiling_face: 

btw. we could add TypeScript support to make sure the i18n only accepts strings that are valid paths in the translation JSON (and we could also configure it in such a way that such concatenations will also be valid), giving us more safety :smile: 

To sum up, I would leave it as is, as I appreciate the deduplication and being able to read less :slightly_smiling_face:

Gelio, line 2: Optional:

Not sure about the naming of this function :smile: Seems a bit short to me. How about

```suggestion
const translate = (key, options) => i18n.t(`widgets.common.deployments.deployModal.${key}`, options);
```

Gelio (approved):
Solid work :ok_hand:

widgets/common/src/DeployBlueprintModal.jsx
kubama: Imagine class with methods `doUpdateSomethingVeryLongCommonPrefixA`, `doUpdateSomethingVeryLongCommonPrefixB`, `doUpdateSomethingVeryLongCommonPrefixC`. Would you create a wrapper function `update` and call it with arguments `A`, `B` or `C` for sake of deduplication?

Gelio: I'm not sure I imagined the same thing :slightly_smiling_face: I don't know if there are hard rules that apply everywhere, so the answer is probably _it depends_ :smile: 

I see the benefit in reducing the deduplication in repeating the same string prefixes over and over :slightly_smiling_face: The case is different if we would be talking about logic and actual code, so I cannot say

kubama (approved):

widgets/common/src/DeployBlueprintModal.jsx
qooban: I was also thinking about that when I introduce this arrow function, but I think in this case (wrapper for `t` function from `i18n`) it is OK 🙂

qooban: @Gelio 

> If removal of unused entries from the translation file is a concern, one idea to mitigate it would be to set up a practice similar to code coverage - translation coverage. We could monkey-patch i18n.t and check the strings passed, and then cross-check it against the translations file to make sure:

> 1. There are no unused strings
> 2. No invalid strings are used
> but that's probably a minor concern slightly_smiling_face
> btw. we could add TypeScript support to make sure the i18n only accepts strings that are valid paths in the translation JSON (and we could also configure it in such a way that such concatenations will also be valid), giving us more safety smile

I like this ideas, maybe at some point in the future we would be able to do that. I believe that it could help in cleaning up translation file in the Composer (ref. [RD-610](https://cloudifysource.atlassian.net/browse/RD-610)).

I know the rules, but in that case - can I ask you to create JIRA ticket for that? I'm not sure I would be able to provide good description there 🙂

Gelio: Sure deal :+1: let's skip it

Gelio: You got it :slightly_smiling_face: 

Here's one for adding TS strict typing: https://cloudifysource.atlassian.net/browse/RD-1716

Here's one for collecting localization file coverage: https://cloudifysource.atlassian.net/browse/RD-1717

qooban: Thanks a lot 👍🏻

PR #1242, opened 2021-03-15
RD-1353 - Optimize fetching of deployment workflows

qooban:
As RD-1353 was fixed on the backend side, I took a chance to create small PR and update two places in the code which were waiting for it.

### DIFF
--- a/widgets/common/src/DeploymentActions.js
+++ b/widgets/common/src/DeploymentActions.js
@@ -125,6 +125,10 @@ export default class DeploymentActions {
             .then(({ items }) => items);
     }
 
+    doGetWorkflows(deploymentId) {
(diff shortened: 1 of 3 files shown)


kubama (approved):

Gelio (approved):
The change looks sensible, but I'm wondering how it worked before. Was it always returning `workflows`? Why is this change needed?

qooban:
@Gelio Before we had to fetch all the deployment resource fields (even if we needed just `workflows`) as `_include` selector was not working properly with `workflows`.

Gelio:
Oh, I see :slightly_smiling_face: So omitting the `_include` fetches all nested resources? I assumed the `_include` behavior is opt-in, meaning you only get what you ask for, and if you don't ask for it, then you get only simple fields :smile: Anyway, thanks for explaining

PR #1243, opened 2021-03-15
CY-3642 - Filter agents list by agent state

qooban:
According to the description provided in the https://cloudifysource.atlassian.net/browse/CY-3642 comment I just added a query parameter to `GET /agents` request. The solution is aligned with CLI's `cfy agents list` command right now.

Unfortunately `state` field is not present in the response, so I don't see any reasonable way of testing it. 
I'm not familiar with setting up environment which provides different agents, so I asked issue reported. He promised to set up something today, so I'm going to do manual tests there.

This video shows that `state=started` is added to the URL:

https://user-images.githubusercontent.com/5202105/111127599-4c6f5f80-8574-11eb-86d3-f0c40aad222d.mp4

### DIFF
--- a/widgets/agents/src/widget.jsx
+++ b/widgets/agents/src/widget.jsx
@@ -12,7 +12,7 @@ Stage.defineWidget({
     initialWidth: 12,
     initialHeight: 24,
     color: 'olive',
-    fetchUrl: '[manager]/agents?[params:gridParams,deployment_id,node_ids,node_instance_ids,install_methods]',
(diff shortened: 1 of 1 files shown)


kubama (approved):

gregcfy (approved):

qooban:
Consulted with issue reporter. Managed to reproduce the issue on the original code. With the fix provided in this PR the issue is not present anymore. 

Cypress test for Agents passed. Once CI is finished I'm going to merge this PR.

PR #1244, opened 2021-03-15
RD-1252 Align Plugin Name vertically in System Resources

Gelio:
This PR aligns the Plugin Name column cell vertically so the plugin name and the icon are roughly on the same line.

I have noticed the same bug in the Sites widget, so I fixed it too in this PR

I assume the change is small enough for 1 approval to be sufficient

## Plugins widget
Before:
image

After:
image

## Sites widget

Before:
image

After:
image

## Snapshots widget
Before:
image

After:
image

## Secret Store Managment widget
Before:
image

After:
image

## Approaches tried

I have tried the following approaches, but they did not work well:
1. Adding the class to the `td` itself - the padding was then too small
2. Using `vertical-align: middle` (it's already set)
3. Using `line-height` - this one works quite well, but requires the height to be set, and it would have to be set on a case-by-case basis, and not once


app/styles/style.scss
```diff
@@ -443,6 +443,12 @@ table.servicesData button.refreshButton {
   width: auto !important;
 }
 
+td .verticallyAlignedCell {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
```
kubama, line 451: How about adding a simple wrapper component instead of adding a global css rule?

Gelio: Good idea :slightly_smiling_face: Where would you recommend to put it? `Stage.Common`?

kubama: I think so.

qooban: Agree, good idea.

qooban (commented):
I think that Snapshots widget can get the same treatment 🙂 

image

app/styles/style.scss
Gelio: Done :slightly_smiling_face:

Gelio:
> I think that Snapshots widget can get the same treatment :slightly_smiling_face: 

Good point :+1: I've fixed the alignment in the Snapshots table and the Secrets table too. I have updated the PR description with the screenshots

widgets/common/src/VerticallyAlignedCell.tsx
```diff
@@ -0,0 +1,22 @@
...
+        display: flex;
+        justify-content: space-between;
+        align-items: center;
+    }
+`;
+
+Stage.defineCommon({
+    name: 'VerticallyAlignedCell',
+    common: VerticallyAlignedCell
+});
```
kubama, line 22: Revisiting how and where this component is now used I think that an even cleaner solution would be to add a prop to `DataTable.Data` and have it handled internally. Yes, I know it's not in line with my previous comment...

Gelio: That would definitely compose better, but I would not add it to the public API prematurely. The reason is that we have not tested this component with other content (a single component, many components), thus, I would consider this experimental.

I would rather move on to the next bug rather than spend time making this perfect. If there will be cases that the component is widely used, I would recommend extracting it at that point.

What do you think?

kubama: > That would definitely compose better, but I would not add it to the public API prematurely. The reason is that we have not tested this component with other content (a single component, many components), thus, I would consider this experimental.

My idea was to add a prop that would enable this only on demand, by default everything would work the old way. So I can see no risk here.

> I would rather move on to the next bug rather than spend time making this perfect. If there will be cases that the component is widely used, I would recommend extracting it at that point.

There is a ticket for adding `className` support for all cloudify-ui-common components, maybe it could be extended to cover this new additional prop for `DataTable.Data` as well?

widgets/common/src/VerticallyAlignedCell.tsx
```diff
@@ -0,0 +1,22 @@
+import styled from 'styled-components';
```
qooban, line 1: Is the `size` stage error because of this one? Probably need to add to `externals` in webpack config?

Gelio: Oh, yeah, right... I forgot about it :slightly_smiling_face: Thanks!

Gelio: `styled-components` is not exposed globally, so I reckon moving the component to `Stage.Shared` would have a smaller impact

widgets/common/src/VerticallyAlignedCell.tsx
Gelio: I have found the issue you mentioned: https://cloudifysource.atlassian.net/browse/RD-1151

While it is somewhat related (both mention the components repo), I find moving `VerticallyAlignedCell` to be somewhat separate. I have created https://cloudifysource.atlassian.net/browse/RD-1701 to move the component to `DataTable.Data` prop and linked it to RD-1151 :slightly_smiling_face:

kubama: Cool, thanks.

kubama (approved):

Gelio:
Looks like the bundle sizes are now OK after moving the component to `Stage.Shared`. 

I'm merging the PR :slightly_smiling_face: Thanks for the reviews

widgets/common/src/VerticallyAlignedCell.tsx
qooban: `Stage.Shared` was meant to keep components shared between `app` and `widgets`, so it's not the best places (as the new component is not used in `app`), but as we have plan to remove it within https://cloudifysource.atlassian.net/browse/RD-1701, I'm OK.

PR #1245, opened 2021-03-15
RD-795 - Fix for stage-backend crash upon downloading huge files

qooban:
This PR provide fix for the most important issue reported within RD-795 - fixes stage-backend crashes by preventing sending headers from `ServerProxy` route when they already have been sent. This happens when e.g. timeout occurs after download has started. Timeout can occur when there is no response from the server (no more bytes sent) during 10 seconds period of time (that is default setting for GET request we have right now).

### DIFF
--- a/backend/routes/ServerProxy.js
+++ b/backend/routes/ServerProxy.js
@@ -26,7 +26,9 @@ function errorHandler(url, res, err) {
     }
 
     logger.error(`${urlMsg} ${exMsg}`, err);
-    res.status(500).send({ message: `${urlMsg} ${exMsg}` });
+    if (!res.headersSent) {
+        res.status(500).send({ message: `${urlMsg} ${exMsg}` });
+    }
 }
 
 function buildManagerUrl(req, res, next) {


kubama:
Just a question to better understand the context - why in timeout scenario response headers are getting sent, whereas in other error scenarios they don't?

qooban:
> Just a question to better understand the context - why in timeout scenario response headers are getting sent, whereas in other error scenarios they don't?

I would call it data download with timeout scenario to be more precise.

In `ServerProxy.js` we have the following piping:
```
    req.pipe(
        request(req.su, options).on('error', err => {
            errorHandler(req.su, res, err);
        })
    ).pipe(res);
```
and `errorHandler` has the following part:
```
function errorHandler(url, res, err) {
    /* (...) */
    res.status(500).send({ message: `${urlMsg} ${exMsg}` });
}
```

In case of file downloads (`Content-Disposition` header present in the response), headers are sent first and then the data is being sent. It might happen that error occurs during data transfer (e.g. 10s timeout). That was the direct reason why Express.js crashed with `ERR_HTTP_HEADERS_SENT` error.

kubama (approved):
I think headers are always sent first, regardless of request/response type, so it's purely a matter of the stage at which the error occurs - before or after headers are sent. File download, due to response payload size, just makes it more probable for an error to occur after headers are sent. But it can still happen for any request/response.
Anyway, the fix seems to be correct 👍.

PR #1247, opened 2021-03-17
RD-1723 - Adjust ExecutionProgress component to use new fields from executions API

qooban:
Recently Executions API was updated and it now provides two new fields:
* `total_operations`
* `finished_operations`
which can be used to calculate execution progress.

PR changelog:
* used new fields in `ExecutionProgress`
* removed workflow limitation for `ExecutionProgress` (now it works on all workflows)
* converted `ExecutionProgress` component to support TypeScript

Video:

https://user-images.githubusercontent.com/5202105/111434578-fd563580-86ff-11eb-9a95-feb33943401b.mp4


kubama (dismissed):

widgets/deployments/src/ExecutionProgress.tsx
```diff
@@ -0,0 +1,32 @@
...
+        total_operations: number;
+        /* eslint-enable camelcase */
+    };
+}
+
+const ExecutionProgress: React.FunctionComponent<ExecutionProgressProps> = ({ execution }) => {
+    const { Progress } = Stage.Basic;
+    const { Execution } = Stage.Utils;
+
+    const color = Execution.isActiveExecution(execution) ? 'yellow' : 'green';
```
Gelio, line 15: Optional:

To show that `yellow` and `green` are not arbitrary strings, we could do:

```suggestion
    const color: SemanticCOLORS = Execution.isActiveExecution(execution) ? 'yellow' : 'green';
```

which would require an `import type` from `semantic-ui-react'` :slightly_smiling_face:

widgets/deployments/src/ExecutionProgress.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const error = Execution.isFailedExecution(execution);
+
+    const { finished_operations: finishedOperations, total_operations: totalOperations } = execution;
+    const ratio = finishedOperations / totalOperations;
+    const percent = Number.isFinite(ratio) ? Math.round(ratio * 100) : 0;
```
Gelio, line 20: Looks like I will need a very similar operation in Deployments View.

Optional:

Would you mind extracting that to `DeploymentActions` or some other file?

I may also do it in my PR :slightly_smiling_face:

Gelio (approved):
Sad no file rename detection :worried: I suggest separating the functional changes from the changes related to TS into separate commits, so it's easier to review the functional changes :slightly_smiling_face: 

I appreciate the TS file :smile: Good work!

widgets/deployments/src/ExecutionProgress.tsx
qooban: Sure, good point. Added `getProgress` to `ExecutionUtils` as I think that's the right place for it.

widgets/deployments/src/ExecutionProgress.tsx
qooban: Sure, make sense, added 👍🏻

qooban:
> Sad no file rename detection worried I suggest separating the functional changes from the changes related to TS into separate commits, so it's easier to review the functional changes slightly_smiling_face

In this case, I thought that it is such small that it'll not be a big problem in review.
But yeah, I'll definitely do that next time.

widgets/deployments/src/ExecutionProgress.tsx
Gelio: Sounds good :+1:

Gelio (approved):
:ok_hand:

PR #1249, opened 2021-03-17
RD-1740 Show real deployments status in Deployments View widget

Gelio:
This PR adds displaying the real deployment status icon in the leftmost column of the table, based on the data returned from the API.

There is also a popup that displays the information about the status on hover. The information is localized.

In this PR I have also done some miscellaneous things:
1. Added types for `Deployment` (I am aware they should be extracted to some top-level directory after we decide on the correct place for such types)
2. Renamed the file with sites commands to TypeScript
    This will make it so that GItHub will recognize the change as a rename now, and will produce a nicer diff when I'll be migrating that file to TypeScript in a follow-up PR

image
image
image


package.json
```diff
@@ -48,7 +48,7 @@
     "better-console": "^0.2.4",
     "blueimp-md5": "^2.7.0",
     "cloudify-blueprint-topology": "^2.4.2",
-    "cloudify-ui-common": "^3.1.2",
+    "cloudify-ui-common": "^3.1.3",
```
Gelio, line 51: Fixes the issue with `no-shadow` rule (see https://github.com/cloudify-cosmo/cloudify-ui-common/pull/51)

widgets/deploymentsView/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface Deployment {
+    id: string;
+    // NOTE: the property names come from the backend
+    /* eslint-disable camelcase */
+    site_name: string;
+    blueprint_id: string;
+    latest_execution_status: LatestExecutionStatus;
+    deployment_status: DeploymentStatus;
+    installation_status: InstallationStatus;
+    environment_type?: string;
```
Gelio, line 35: I've marked some properties as optional since they are not present on `master` yet, but are listed in https://cloudifysource.atlassian.net/wiki/spaces/WR/pages/1572864001/Environment#Data-Model

test/cypress/support/sites.ts
```diff
@@ -1,5 +1,5 @@
 Cypress.Commands.add('createSite', site => {
```
Gelio, line 1: I'm changing the file name for this file in this PR so the diff will be nicer in the next one :slightly_smiling_face:

app/translations/en.json
```diff
@@ -557,7 +557,11 @@
             },
             "columns": {
                 "status": {
-                    "name": ""
+                    "name": "",
+                    "iconLabels": {
+                        "in_progress": "In progress",
+                        "requires_attention": "Requires attention"
```
kubama, line 563: We are using camel case as keys consistently throughout this file - how about using `_.camelCase` before using `deployment.deployment_status` as a key?

qooban:
@Gelio I'll review it tomorrow morning.

widgets/deploymentsView/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+    InProgress = 'in_progress',
+    RequiresAttention = 'requires_attention'
+}
+
+export enum SubdeploymentStatus {
+    Good = 'good',
+    InProgress = 'in_progress',
+    Failed = 'failed',
+    Pending = 'pending'
+}
```
qooban, line 24: General:

Do we have issues for other widget alignment to the new statuses (Sites Map widget, Deployments widget)?
It would be good to reuse some code from here there probably. I know that we discussed we may drop some old widgets completely, but I suspect it would be better to have at least an issue(s) to consider.

widgets/deploymentsView/src/types.ts
```diff
@@ -0,0 +1,43 @@
...
+    // NOTE: the property names come from the backend
+    /* eslint-disable camelcase */
+    site_name: string;
+    blueprint_id: string;
+    latest_execution_status: LatestExecutionStatus;
+    deployment_status: DeploymentStatus;
+    installation_status: InstallationStatus;
+    environment_type?: string;
+    total_operations?: number;
+    finished_operations?: number;
```
qooban, line 37: Will they be part of the deployment API?

If so, then good, if not, we should get them from executions API.

qooban (commented):
LGTM 👍🏻 Added just two comments with questions.

app/translations/en.json
Gelio: Right, makes sense, I'll update it

widgets/deploymentsView/src/types.ts
Gelio: > Do we have issues for other widget alignment to the new statuses (Sites Map widget, Deployments widget)?

I'm not aware of any. At least I did not create them :smile: I have been focusing only on the Deployments View widget itself

> It would be good to reuse some code from here there probably

Yeah yeah, I agree :slightly_smiling_face:  as I have said in the PR description :smile: 

> Added types for Deployment (I am aware they should be extracted to some top-level directory after we decide on the correct place for such types)

widgets/deploymentsView/src/types.ts
Gelio: Yup, I have asked Mohammed to include it, so they will eventually be included in the responses, as said in the design doc: https://cloudifysource.atlassian.net/wiki/spaces/WR/pages/1572864001/Environment#Get-deployment-API

Meaning that we don't need to make another request for each entry in the table :smile:

app/translations/en.json
Gelio: Done in 6ff039485ff92df989d9bed0ca8402d655faf316

kubama (approved):

widgets/deploymentsView/src/types.ts
qooban: OK, so could you create one per each widget? (I'm not sure these are all, maybe there are more to take care, but not coming to my mind now)

(I know that we maybe will deprecate Deployments widget, but until then I think it would be wise to have "alignement issues" for both, to have at least a trace)

widgets/deploymentsView/src/types.ts
qooban: Got it, clear.

qooban (approved):

widgets/deploymentsView/src/types.ts
Gelio: Yeah, it's worth having those issues. Here's one for Deployments view: https://cloudifysource.atlassian.net/browse/RD-1779 (it's outside of any epic, since I did not see it as strictly related with Environments)

Here's one for Sites map (in the Environments Map epic): https://cloudifysource.atlassian.net/browse/RD-1780

Gelio:
Thanks for the reviews, guys :slightly_smiling_face: I'll merge the PR

widgets/deploymentsView/src/types.ts
qooban: Thanks 👍🏻

PR #1252, opened 2021-03-18
RD-1753 Fix Sites widget tests

Gelio:
The selectors were too strict and required the cells to have a specific structure. Introducing an additional `div` in the cell broke the `td > span` selector. `td span` works just fine :slightly_smiling_face: 

image

You might notice quite a large commit when reviewing by commits. That's because I tried to refactor the verification logic to avoid repetition, but it looks like there were not enough waiting/verification I could do to make it work. I decided to revert it and do a simple, dirty, but working fix instead

## System tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/293/

### DIFF
--- a/test/cypress/integration/widgets/sites_spec.js
+++ b/test/cypress/integration/widgets/sites_spec.js
@@ -66,29 +66,42 @@ describe('Sites Management', () => {
     };
 
     const verifySiteRow = (index, site) => {
-        const siteRow = `tbody > :nth-child(${index})`;
-        cy.get(`${siteRow} > :nth-child(1)`).should('have.text', site.name);
+        const siteRow = `tbody > tr:nth-child(${index})`;
 
-        let visibilityColor = 'green';
-        if (site.visibility === 'private') {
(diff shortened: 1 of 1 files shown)


test/cypress/integration/widgets/sites_spec.js
```diff
@@ -66,29 +66,32 @@ describe('Sites Management', () => {
     };
 
     const verifySiteRow = (index, site) => {
-        const siteRow = `tbody > :nth-child(${index})`;
-        cy.get(`${siteRow} > :nth-child(1)`).should('have.text', site.name);
+        const siteRow = `tbody > tr:nth-child(${index})`;
```
Gelio, line 69: Added `tr` and `td` element selectors to improve readability

qooban (approved):
Looks good to me 👍🏻

kubama (approved):

Gelio:
All system tests except for _topology_ passed, thus, looks like the issue with sites is indeed fixed :slightly_smiling_face:

PR #1253, opened 2021-03-18
RD-1751 Add basic tests for Deployments View widget

Gelio:
The PR adds a few basic tests for the Deployments View widget.

Don't be afraid by the number of lines added :smile: Most of them come from the test fixture for deployments

All passed locally:
image

## Changes in this PR

1. Migrate the sites commands to TypeScript

    Sites commands are used in the new test suite.

2. Basic tests:

    1. The created deployment is displayed in the table
    2. Hiding and showing columns (`fieldsToShow` in the widget configuration) is working correctly
    3. The deployment status icon is displayed correctly and has a popup

    I have tried to make the tests rely on accessible properties (`aria-label`, `role`, text), rather than classes. I think I managed to do that well, except for a few cases :slightly_smiling_face: 

    The added fixture will make it easier to test functionalities such as displaying the number and statuses of subdeployments, and the latest execution progress. Those will be implemented and tested in future tickets.

3. Add handling `process.env.TEST` (set to `'1'` during e2e tests)

    Necessary to get `deploymentsViewWidget` to render during system tests.

    I was hoping to set `process.env.NODE_ENV` to `test` during e2e tests, but webpack overrides `process.env.NODE_ENV` based on the `mode` it runs in. Thus, in e2e tests it always has `process.env.NODE_ENV === 'production'` :worried: 

    To verify that it does not break production, I ran the production build locally and here's the value of `process.env.TEST`:
    image

## System tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/315/

Ran 2021-03-23 10:56 CET

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "docWidgets": "node scripts/updateReadmes.js",
     "e2e": "cypress run -C node_modules/cloudify-ui-common/cypress/cypress.json",
     "e2e:open": "cypress open -C node_modules/cloudify-ui-common/cypress/cypress.json",
-    "e2e:ci": "npm run beforebuild && run-system-tests ${TEST_SPEC}",
+    "e2e:ci": "npm run beforebuild && cross-env TEST=1 run-system-tests ${TEST_SPEC}",
     "lint": "eslint --cache --ignore-path .gitignore --ext js,jsx,ts,tsx --max-warnings 10 .",
     "size": "[ -d \"dist\" ] && size-limit",
     "test": "npm run test:frontend && npm run test:backend",
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -1,4 +1,5 @@
 import type { RouteHandler } from 'cypress/types/net-stubbing';
+import { without } from 'lodash';
 
 describe('Deployments View widget', () => {
     const widgetId = 'deploymentsView';
@@ -27,9 +28,12 @@ describe('Deployments View widget', () => {
             .setSite(deploymentName, siteName);
     });
 
-    const useDeploymentsViewWidget = (routeHandler?: RouteHandler) => {
+    const useDeploymentsViewWidget = ({
+        routeHandler,
+        configurationOverrides = {}
+    }: { routeHandler?: RouteHandler; configurationOverrides?: Record<string, any> } = {}) => {
         cy.interceptSp('GET', 'deployments', routeHandler).as('deployments');
-        cy.usePageMock([widgetId], widgetConfiguration).mockLogin();
+        cy.usePageMock([widgetId], { ...widgetConfiguration, ...configurationOverrides }).mockLogin();
         cy.wait('@deployments');
     };
 
@@ -41,14 +45,10 @@ describe('Deployments View widget', () => {
         const toggleFieldsDropdown = () => getFieldsDropdown().find('.dropdown.icon').click();
 
         it('should allow changing displayed columns', () => {
-            useDeploymentsViewWidget();
-
-            cy.log('Hide some columns');
-            cy.editWidgetConfiguration(widgetId, () => {
-                getFieldsDropdown().within(() => {
-                    cy.contains('Blueprint Name').find('.delete.icon').click();
-                    cy.contains('Location').find('.delete.icon').click();
-                });
+            useDeploymentsViewWidget({
+                configurationOverrides: {
+                    fieldsToShow: without(widgetConfiguration.fieldsToShow, 'blueprintName', 'location')
+                }
             });
 
             getDeploymentsViewTable().within(() => {
@@ -86,7 +86,9 @@ describe('Deployments View widget', () => {
 
     it('should display various deployment statuses', () => {
         useDeploymentsViewWidget({
-            fixture: 'deployments/various-statuses.json'
+            routeHandler: {
+                fixture: 'deployments/various-statuses.json'
+            }
         });
 
         getDeploymentsViewTable().within(() => {
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -1,0 +1,36 @@
+describe('Deployments View widget', () => {
+    const blueprintName = 'deployments_view_test';
+    const deploymentName = 'deployments_view_test_dep';
+    const siteName = 'Olsztyn';
+    const blueprintUrl =
+        'https://github.com/cloudify-community/blueprint-examples/releases/download/latest/simple-hello-world-example.zip';
+
+    before(() => {
+        cy.activate()
+            .deleteSites(siteName)
+            .deleteDeployments(deploymentName, true)
+            .deleteBlueprints(blueprintName, true)
+            .uploadBlueprint(blueprintUrl, blueprintName)
+            .deployBlueprint(blueprintName, deploymentName, { webserver_port: 9123 })
+            .createSite({ name: siteName })
+            .setSite(deploymentName, siteName);
+    });
+
+    beforeEach(() => {
+        cy.interceptSp('GET', 'deployments').as('deployments');
+        // NOTE: add widget through Edit Mode to apply default configuration automatically
+        cy.usePageMock([], {}, ['deploymentsView']).mockLogin().addWidget('deploymentsView');
+    });
+
+    // TODO: test changing the visible columns in the configuration
+
+    it('should display the deployments', () => {
+        cy.wait('@deployments');
+
+        cy.get('tr').within(() => {
+            cy.contains(deploymentName);
+            cy.contains(blueprintName);
+            cy.contains(siteName);
+        });
+    });
+});
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -10,8 +10,8 @@ describe('Deployments View widget', () => {
 
     before(() => {
         cy.activate()
-            .deleteSites(siteName)
             .deleteDeployments(deploymentName, true)
+            .deleteSites(siteName)
             .deleteBlueprints(blueprintName, true)
             .uploadBlueprint(blueprintUrl, blueprintName)
             .deployBlueprint(blueprintName, deploymentName, { webserver_port: 9123 })
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -77,5 +77,29 @@ describe('Deployments View widget', () => {
         });
     });
 
-    it('should display various deployment statuses', () => {});
+    it('should display various deployment statuses', () => {
+        useDeploymentsViewWidget({
+            fixture: 'deployments/various-statuses.json'
+        });
+
+        getDeploymentsViewTable().within(() => {
+            cy.contains('deployments_view_test_deployment')
+                .parents('tr')
+                .find('i.exclamation[aria-label="Requires attention"]')
+                .as('requiresAttentionIcon')
+                .trigger('mouseover');
+            cy.root().parents('body').find('.popup').contains('Requires attention');
+            cy.get('@requiresAttentionIcon').trigger('mouseout');
+
+            cy.contains('hello-world-one')
+                .parents('tr')
+                .find('i.spinner[aria-label="In progress"]')
+                .as('inProgressIcon')
+                .trigger('mouseover');
+            cy.root().parents('body').find('.popup').contains('In progress');
+            cy.get('@inProgressIcon').trigger('mouseout');
+
+            cy.contains('one-in-warsaw').parents('tr').find('td:nth-child(1) i').should('not.exist');
+        });
+    });
 });
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -7,6 +7,14 @@ describe('Deployments View widget', () => {
     const siteName = 'Olsztyn';
     const blueprintUrl =
         'https://github.com/cloudify-community/blueprint-examples/releases/download/latest/simple-hello-world-example.zip';
+    const widgetConfiguration = {
+        filterByParentDeployment: false,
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "audit": "npm audit --production && cd backend && npm run audit",
     "beforebuild": "npm ci --prefer-offline && (cd backend && npm ci --prefer-offline)",
-    "build": "cross-env NODE_ENV=production webpack --config webpack.config.js --mode=production --bail",
+    "build": "webpack --config webpack.config.js --mode=production --bail",
```
Gelio, line 20: See message of 509afa4aebe2ebba781ec4c8a7b308abe55fc283 for more details on this change

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+            cy.contains('hello-world-one')
+                .parents('tr')
+                .find('i.spinner[aria-label="In progress"]')
+                .as('inProgressIcon')
+                .trigger('mouseover');
+            cy.root().parents('body').find('.popup').contains('In progress');
+            cy.get('@inProgressIcon').trigger('mouseout');
+
+            cy.contains('one-in-warsaw').parents('tr').find('td:nth-child(1) i').should('not.exist');
```
Gelio, line 102: I could not find a better way to check that the icon does not exist

test/cypress/support/commands.ts
```diff
@@ -176,13 +176,14 @@ const commands = {
         }
         cy.waitUntilPageLoaded();
     },
-    usePageMock: (widgetIds?: string | string[], widgetConfiguration: any = {}) => {
+    usePageMock: (
+        widgetIds?: string | string[],
+        widgetConfiguration: any = {},
+        additionalWidgetIdsToLoad: string[] = []
```
Gelio, line 182: Allows using widgets that are not displayed initially in the tests

Gelio:
It seems that the changes around `NODE_ENV` do not work, since webpack overrides `NODE_ENV` to `production` or `development` when the `mode` property is set. Thus, `NODE_ENV` will always be set to `production` in e2e tests.

I'm working on a different solution, will push it out in 1h

widgets/deploymentsView/src/widget.tsx
```diff
@@ -24,8 +24,8 @@ interface DeploymentsViewWidgetConfiguration {
 
 const i18nPrefix = 'widgets.deploymentsView';
 
-// TODO(RD-1224): remove environment check
-if (process.env.NODE_ENV === 'development') {
+// TODO(RD-1226): remove environment check
+if (process.env.NODE_ENV === 'development' || process.env.TEST) {
```
Gelio, line 28: `process.env.TEST` is undefined in production:

image

And `process.env.NODE_ENV === 'development'` is substituted to `true` in development:
image

Gelio:
Turns out `process.env.TEST` is still `undefined` during test. `EnvironmentPlugin` does not forward it. I need to investigate

qooban:
@Gelio Try using `supportedEditions` field on widget definition. You can just pass something like `['my-super-edition']` and the widget should not be presented to the users.

Gelio:
> @Gelio Try using `supportedEditions` field on widget definition. You can just pass something like `['my-super-edition']` and the widget should not be presented to the users.

Unfortunately, that's not what I was looking for :sweat_smile: I am already disabling the widget by not calling `Stage.defineWidget` in production. What I was looking for is a way to know whether we are in actual production, or in e2e tests :slightly_smiling_face: I ended up adding another `process.env.TEST` variable to determine it

---

Turns out it was not working because `EnvironmentPlugin` was not being applied at all :slightly_smiling_face: It was inserted (by me :smile: ) in the wrong place some time ago.

webpack.config.js
```diff
@@ -216,9 +217,9 @@ module.exports = (env, argv) => {
                             }
                         }
                     }),
+                    environmentPlugin,
                     isProduction && getProductionPlugins(env && env.analyse === 'main')
-                ]),
-                environmentPlugin
```
Gelio, line 221: It was placed outside of the array :sweat_smile:

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    const blueprintName = 'deployments_view_test_blueprint';
+    const deploymentName = 'deployments_view_test_deployment';
+    const siteName = 'Olsztyn';
+    const blueprintUrl =
+        'https://github.com/cloudify-community/blueprint-examples/releases/download/latest/simple-hello-world-example.zip';
+
+    before(() => {
+        cy.activate()
+            .deleteSites(siteName)
+            .deleteDeployments(deploymentName, true)
```
kubama, line 14: Not sure if it matters in this particular case, but since it's a deployment that is assigned a site, not the other way around, I think deployments should be deleted first, then sites.

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -0,0 +1,105 @@
...
+            .uploadBlueprint(blueprintUrl, blueprintName)
+            .deployBlueprint(blueprintName, deploymentName, { webserver_port: 9123 })
+            .createSite({ name: siteName })
+            .setSite(deploymentName, siteName);
+    });
+
+    const useDeploymentsViewWidget = (routeHandler?: RouteHandler) => {
+        cy.interceptSp('GET', 'deployments', routeHandler).as('deployments');
+        // NOTE: add widget through Edit Mode to apply default configuration automatically
+        cy.usePageMock([], {}, [widgetId]).mockLogin().addWidget(widgetId);
```
kubama, line 25: Is there any reason why we want the default configuration? Using specified configuration would make the test faster, and speed was the actual reason behind introducing `usePageMock` allowing any widget to be added to the test page out of the box, without the need for Edit Mode.

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Good point :slightly_smiling_face: Copy-paste didn't work well :smile:

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: I understand the concern about the speed of the test. In this case, I decided that the pros outweigh the cons

The default configuration is quite lengthy (it has the list of columns visible in the table). I did not want to duplicate it and hardcode it here, because:
1. The test gives less confidence that the functionality works
2. It makes the test not pick up changes to the default widget configuration
3. That would tie the test to the implementation, which I did not want to do, as I prefer black-box testing (changing the field name in the table would require changing the tests)

---

As I see it, the problem with the widget configuration lies in **when** the default configuration is applied. As I understand, applying default configuration happens when the page is created or the widget is added to the page, and not when it's rendered. Thus, when using `usePageMock` the configuration is empty, the widget does not merge the default configuration into the passed `configuration`, but just uses the empty config, which makes the widget fail.

What I would suggest as a global fix would be to use the widget's default configuration when the widget is being rendered and its configuration is missing. Then, I would happily render the widget with the empty config here, and it would use the default one.

What do you think about such a solution? If you're ok, I will create a tech debt item to introduce such changes in the future

kubama: I'm all for the proposed solution.

However, since it is unlikely that the tech debt item will be implemented any soon I would still consider using specified configuration. The points that you made are general and apply to any widget test, not only to this particular one, so if we agreed on them then we should start moving all widget tests to this approach. As it would increase tests execution time considerably it would be an opposite direction to what we were trying to achieve so far (system tests execution time was an issue and `usePageMock` was introduced to address it).

package.json
```diff
@@ -34,7 +34,7 @@
     "docWidgets": "node scripts/updateReadmes.js",
     "e2e": "cypress run -C node_modules/cloudify-ui-common/cypress/cypress.json",
     "e2e:open": "cypress open -C node_modules/cloudify-ui-common/cypress/cypress.json",
-    "e2e:ci": "npm run beforebuild && run-system-tests ${TEST_SPEC}",
+    "e2e:ci": "npm run beforebuild && cross-env TEST=1 run-system-tests ${TEST_SPEC}",
```
qooban, line 37: Hmm... We are adding new environmental variable to be able to include specific widget in the test-build. Not sure I like this solution as I see it is temporary or we plan to use it more extensively?

Maybe in `widgets/deploymentsView/src/widget.tsx` instead of `TEST` env-var we could check some env-var available only in the Jenkins CI? Loudly thinking, I'm not sure it would be the best.

qooban (commented):
In general looks good 👍🏻 

However, I'm not sure about introducing new env-var for that specific test.

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Solid points, I will refactor the test to use the hardcoded version of the configuration and create tickets to introduce the solution I described earlier :slightly_smiling_face:

package.json
Gelio: > Not sure I like this solution [...]

Yeah, I was not too sure about it either, but I did not come up with a better way. I was trying to use `NODE_ENV = 'test'`, but not only that did not work, that would also alter the bundle (since other packages, e.g. `React`, strip away some code in production, meaning that the behavior in tests could be a bit different than in production). Thus, I decided that a new variable would be a good enough solution

> [...] or we plan to use it more extensively?

My intent is to be able to show and test widgets, as well as enable some functionality only in test and development environments, but not in production. This applies to Environments, and probably future features I will be working on :slightly_smiling_face: I see benefits to this approach (I call it "poor man's feature flag") compared to feature branches.

That said, I didn't come up with a better solution for now. Whether this solution stays, or we implement something else, is an open question. My only aim is what I described in the paragraph above :smile: 

> Maybe in widgets/deploymentsView/src/widget.tsx instead of TEST env-var we could check some env-var available only in the Jenkins CI?

Keep in mind that the list of environment variables that are available in the code has to be listed in the `EnvironmentPlugin` in webpack's config. This means that if we want to use some Jenkins environment variable, we would have to specify it there, along with a default value. It's not a large overhead, but if we use an environment variable just to detect the environment, I would prefer a service-agnostic `TEST` to some obscure `JENKINS_...` :smile: 

Also, if we run `npm run e2e:ci` locally, then the Jenkins variable will not be set, meaning that the code being tested will be different than the one on Jenkins :smile: To fix it, we would have to define that variable ourselves, probably in the `npm run e2e:ci` script. Then, it would not solve the problem of having to define an env variable anyway :slightly_smiling_face: 

To sum up, I don't see a better way, and, in my view, a service-agnostic `TEST` variable is superior to something Jenkins-related :slightly_smiling_face:

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: The execution time when running tests locally went down 20% (from 55 seconds to 44 seconds) :slightly_smiling_face: 

Here are the tickets:
1. https://cloudifysource.atlassian.net/browse/RD-1820
2. https://cloudifysource.atlassian.net/browse/RD-1821

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Done :slightly_smiling_face:

kubama (approved):

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        it('should allow changing displayed columns', () => {
+            useDeploymentsViewWidget();
+
+            cy.log('Hide some columns');
+            cy.editWidgetConfiguration(widgetId, () => {
+                getFieldsDropdown().within(() => {
+                    cy.contains('Blueprint Name').find('.delete.icon').click();
+                    cy.contains('Location').find('.delete.icon').click();
+                });
+            });
```
kubama, line 52: Actually, this could be part of the initial widget configuration.

test/cypress/support/commands.ts
kubama: Is this still needed?

package.json
qooban: OK, got it, thanks for detailed explanation. 

I think that maybe we can go further and have that variable turned on in `run-system-tests` script. But for now we can leave it as it is.

qooban (dismissed):

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Yeah, very much so. Would it be ok if I leave it be as-is for now? :smile: I'd like to continue working on other tasks that depend on this PR and reduce the "branch-chaining"

Gelio:
I've just queued up the system tests for this branch: https://jenkins.cloudify.co/job/Stage-UI-System-Test/315/

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: Wouldn't it be a 5-10 minutes change? Adding an optional parameter to `useDeploymentsViewWidget` allowing to override `fieldsToShow`.

package.json
Gelio: > I think that maybe we can go further and have that variable turned on in run-system-tests script. But for now we can leave it as it is.

Yup, that would make sense :+1:

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Yes, it would. I did not want to context-switch to introduce a change I thought was minor and could be optional :slightly_smiling_face:

Gelio:
I've added more more commit to address @kubama's comment https://github.com/cloudify-cosmo/cloudify-stage/pull/1253#discussion_r599398075

qooban (approved):

Gelio:
The 2 system tests that failed are:
1. Known `topology_spec`
2. A not-updated test for `deployment_actions_spec.js` after #1256  (which I've reported on slack)

Both are unrelated to this PR, so I'm merging it

test/cypress/support/commands.ts
Gelio: Sorry, I did not notice your comment earlier. It's not leveraged in this PR, but will be used in a future PR once the details pane is implemented for this widget. We need to load widgets that are rendered in that details pane, which are not rendered in the top level

PR #1259, opened 2021-03-22
RD-1782 Blueprint sources widget cluster environment fix

kubama:
System tests execution: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/311/ (all green except those failing on master)


backend/routes/BlueprintUserData.js
```diff
@@ -15,17 +15,16 @@ router.get('/layout/:blueprintId', (req, res) => {
...
-                    if (path) {
-                        SourceHandler.browseArchiveFile(path)
-                            .then(yaml.safeLoad)
-                            .then(layout => res.send(layout));
-                    } else {
-                        res.status(404).send({});
-                    }
-                });
+            SourceHandler.browseArchiveTree(req).then(data => {
+                const path = _.chain(data).get('children[0].children').find({ title: 'info.yaml' }).get('key').value();
```
qooban, line 19: For a person not familiar with blueprint topology layout feature it might be not clear. Having that line extracted to a simple named function (eg. `getBlueprintLayoutFilePath` or sth explicitly saying what path we are getting here) would explain what is it for. WDYT?

backend/routes/SourceBrowser.js
```diff
@@ -10,13 +10,14 @@ const router = express.Router();
 
 router.use(passport.authenticate('token', { session: false }));
 
-router.get('/browse/file', (req, res, next) => {
-    const { path } = req.query;
+router.get('/browse/:blueprintId/file/:timestamp/*', (req, res, next) => {
```
qooban, line 13: Just wondering - what was the reasoning behind changing it from query param? Only that it is simpler and shorter?

I assume encoding is done the same way on the frontend side, so it should still be fine.

widgets/blueprintSources/src/BlueprintSources.jsx
```diff
@@ -90,7 +89,7 @@ export default function BlueprintSources({ data, toolbox, widget }) {
                 );
             return (
                 <NodesTree.Node
-                    key={item.key}
+                    key={`${blueprintId}/file/${timestamp}/${item.key}`}
```
qooban, line 92: Is `item.key` encoded?

qooban (commented):
Added some comments.

backend/routes/BlueprintUserData.js
kubama: I renamed the `path` variable to be more descriptive - I guess this makes the job

qooban: 👍🏻

widgets/blueprintSources/src/BlueprintSources.jsx
kubama: This is actually a good catch - I think it would only be an issue if it contained `#` or `?` characters, but it's still worth addressing so I added proper encoding and a unit test for this.

backend/routes/SourceBrowser.js
kubama: > what was the reasoning behind changing it from query param? Only that it is simpler and shorter?

Basically yes. Since virtually every path contains slashes it is always encoded when passed as query parameter, which makes it hard to read when it appears in any kind of logs.

qooban: Got it, thx for the explanation.

backend/handler/SourceHandler.js
```diff
@@ -25,16 +26,18 @@ module.exports = (() => {
...
 
         if (stats.isSymbolicLink() || isUnixHiddenPath(name)) {
             return null;
         }
 
         const item = {
-            key: archivePath.replace(pathlib.join(root, pathlib.sep), ''),
+            key: url
+                .pathToFileURL(pathlib.relative(rootDir, scannedFileOrDirPath))
+                .pathname.substring(url.pathToFileURL('').pathname.length + 1),
```
qooban, line 40: Optional:

It took me a while to understand what is happening here. 
Maybe worth to extract it to a const/function or add a comment?

qooban (approved):
Thanks for source files handling improvements. Looks good 👍🏻

backend/handler/SourceHandler.js
kubama: Done

qooban (approved):

backend/handler/SourceHandler.js
qooban: Thanks 👍🏻

widgets/blueprintSources/src/BlueprintSources.jsx
qooban: 👍🏻

PR #1260, opened 2021-03-23
RD-1738 - Update passport-saml version

qooban:
As https://github.com/node-saml/passport-saml/issues/556 was fixed, I created small PR to make Stage audit happy.


Gelio (approved):
:tada:

kubama (approved):

PR #1265, opened 2021-03-23
RD-1743 Use existing table features in Deployments View table

Gelio:
This PR is a base for #1267  and includes standalone changes that I have separated out to make the code review a bit easier.

Notable changes:
1. Add `searchable`, `noDataMessage`, and `sortable` to the table
2. Refactor the way class names are combined in `WidgetDynamicContext` (non-breaking change, and will be useful in future PRs to combine 1 more class)

## Video

https://user-images.githubusercontent.com/889383/112164892-8e407b80-8bee-11eb-85be-a084f35b65e7.mp4

## System tests

I will run system tests for #1267, which builds upon this one, to avoid having to run system tests for a linear path of branches

PR #1254, opened 2021-03-19
RD-1786 Show Environment Type column

Gelio:
This PR adds support for showing data in the Environment Type column in the Deployments View.

I hope the commit log gives enough guidance about the changes :slightly_smiling_face: 

One unrelated change was fixing the way the "Status" column was presented in the widget's configuration modal (it was empty).

The functionality is covered by a Cypress test. So far, the backend does not return the `environment_type` property, so it's tested only using the mocked fixture.

## System tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/301/

Ran 2021-03-19 16:50 CET

PR #1267, opened 2021-03-23
RD-1743 Add deployment details pane in Deployments View

Gelio:
This PR builds upon https://github.com/cloudify-cosmo/cloudify-stage/pull/1265 and adds the details pane in the Deployments View widget.

There are many changes in this PR. I did not find a better way to split it without requiring large changes. Around 300 lines of changes is the hardcoded details pane structure.

## Features implemented

1. Deployment name is displayed in the header
2. Deployment actions buttons in the details pane
3. Tabs in the details pane, the same as on the deployment drill-down page.

    The details' pane structure (the tabs and the widgets inside) is hardcoded for now. I did not have enough bandwidth to investigate the option to render a page there, which would be created from a template, to allow for pane content customization. I have created RD-1826 to address it as a future enhancement.

## Video

https://user-images.githubusercontent.com/889383/112169060-24c26c00-8bf2-11eb-841a-b4b22062269e.mp4

## Known issues/future work

The items below are known issues or future work that will enhance this widget:

1. RD-1826
2. RD-1827
3. RD-1828
4. RD-1829
5. RD-1819 (addressed in #1268)
6. RD-1822 (addressed in #1269)
7. RD-1830

## System tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/327/

Ran 2021-03-24 15:05 CET


app/components/WidgetDynamicContent.jsx
```diff
@@ -292,11 +292,12 @@ export default class WidgetDynamicContent extends Component {
     }
 
     render() {
-        const { widget } = this.props;
+        const { widget, standalone } = this.props;
         const { loading } = this.state;
         const baseWidgetContentClassName = combineClassNames([
             'widgetContent',
-            widget.definition.showHeader ? '' : 'noHeader'
+            widget.definition.showHeader ? '' : 'noHeader',
```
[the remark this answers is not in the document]
Gelio: Done :slightly_smiling_face: It appeared as a merge conflict when I was rebasing this PR on https://github.com/cloudify-cosmo/cloudify-stage/pull/1265, since this change was already fixed in https://github.com/cloudify-cosmo/cloudify-stage/pull/1265

widgets/deploymentsView/src/detailsPane/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <DetailsPaneWidgets />
+        </div>
+    );
+};
+export default DetailsPane;
+
+DetailsPane.propTypes = {
+    // NOTE: `as any` to comply with TypeScript
+    deployment: PropTypes.shape({}) as any
+};
```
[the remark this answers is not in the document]
Gelio: Yup, the rule that enforces specifying `propTypes` (https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/prop-types.md) is still on for tsx files. This is because not all code is in TypeScript, so we still get benefits from propTypes and can detect errors easier this way

app/components/Tabs.jsx
```diff
@@ -7,7 +7,7 @@ import EditTabModal from './EditTabModal';
 import EditModeButton from './EditModeButton';
 import { Confirm, Menu } from './basic';
 import AddWidget from '../containers/AddWidget';
-import { WidgetsList } from './shared/widgets';
+import WidgetsList from './shared/widgets/WidgetsList';
```
qooban, line 10: Just wondering - why did you change it here? 

In `app/components/Page.tsx` you used:
```
import { PageContent } from './shared/widgets';
```

IMHO we can keep importing single export from `./shared/widgets`.

app/components/shared/widgets/Widget.tsx
```diff
@@ -176,7 +182,8 @@ class Widget<Configuration> extends Component<WidgetProps<Configuration>, Widget
                     className={stageUtils.combineClassNames([
                         'widgetItem',
                         isEditMode && widget.definition && !widget.definition.showBorder && 'borderOnHover',
-                        !widget.definition.showHeader && 'headerless'
+                        !widget.definition.showHeader && 'headerless',
+                        standalone && 'standalone'
```
qooban, line 186: I see that in `app/components/WidgetDynamicContent.jsx` we've got similar classes. But when `!widget.definition.showHeader === true` we use `noHeader` in `WidgetDynamicContent` and `headerless` here. Is that intentional? If not, maybe worth to have it simplified and defined in one place only?

widgets/deploymentsView/src/detailsPane/widgets.tsx
```diff
@@ -0,0 +1,293 @@
...
+                onWidgetAdded={noop}
+                onLayoutSectionAdded={noop}
+                onLayoutSectionRemoved={noop}
+            />
+        </div>
+    );
+};
+export default DetailsPaneWidgets;
+
+const page = {
```
qooban, line 28: Do we need `id` fields here per each widget? 
Did you consider moving it to custom page JSON file? IMHO that would be much cleaner solution.

Ref.: https://github.com/cloudify-cosmo/cloudify-stage/pull/1239#discussion_r595132782

widgets/deploymentsView/src/styles.scss
```diff
@@ -38,4 +38,15 @@ $deploymentStateColors: (
             background-color: var(--deployment-color-#{$stateName});
         }
     }
+
+    .grid {
+        display: grid;
+        grid-template-columns: minmax(50%, 1fr) 1fr;
```
qooban, line 44: Oh, nice, something new to me (fraction unit). 
I think, I must invest some time to get more familiar with all aspects of CSS Grid Layout 🙂

widgets/deploymentsView/src/detailsPane/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        return (
+            <div className="unknownDeploymentMessage">
+                <Message warning>
+                    <Message.Header>Unknown deployment selected</Message.Header>
+                    <p>
+                        The selected deployment is either not in the table, or you are using a Resource Filter with
+                        multiple selection.
+                    </p>
+                    <p>Please select a deployment shown in the table on the left.</p>
+                </Message>
```
qooban, line 33: Are you planning to move it to the translation file?

widgets/deploymentsView/src/detailsPane/header.tsx
```diff
@@ -0,0 +1,48 @@
+import { noop } from 'lodash';
+import { ComponentProps, FunctionComponent, useMemo, useRef } from 'react';
+
+import './header.scss';
+
+export interface DetailsPaneHeaderProps {
+    deploymentName: string;
+}
+
+const DetailsPaneHeader: FunctionComponent<DetailsPaneHeaderProps> = ({ deploymentName }) => {
```
qooban, line 10: General:

Having `DetailsPaneHeader` is a nice idea. I assume this is to get part of the details always visible at the top. That's good, but I'm wondering - from maintenance perspective - wouldn't it be better to have an option in widget containers to make them kind of fixed to the place where they are added? That will probably allow you to use single JSON page for the whole details pane. I know it might be quite big change and not easy to implement. I'm interested in your opinion on that.

widgets/deploymentsView/src/detailsPane/index.scss
```diff
@@ -0,0 +1,4 @@
+.unknownDeploymentMessage {
+    // NOTE: 48px to align with the table
+    margin: 48px 10px 0;
```
qooban, line 3: Did you consider using less pixel-related unit here? (`rem`, `em`)

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -25,19 +39,29 @@ describe('Deployments View widget', () => {
...
 
     const useDeploymentsViewWidget = ({
         routeHandler,
         configurationOverrides = {}
     }: { routeHandler?: RouteHandler; configurationOverrides?: Record<string, any> } = {}) => {
-        cy.interceptSp('GET', 'deployments', routeHandler).as('deployments');
-        cy.usePageMock([widgetId], { ...widgetConfiguration, ...configurationOverrides }).mockLogin();
+        cy.interceptSp('GET', /^\/deployments/, routeHandler).as('deployments');
+        // NOTE: larger viewport since the widget requires more width to be comfortable to use
+        cy.viewport(1600, 900)
```
qooban, line 52: Optional:

Maybe we can change default viewport.

widgets/deploymentsView/src/detailsPane/header.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const { Header } = Stage.Basic;
+    const { Widget } = Stage.Shared.Widgets;
+    const uuidRef = useRef(Stage.Utils.uuid);
```
qooban, line 13: Nice, didn't know that we can use `useRef` this way 👍🏻

qooban (approved):
Nicely done 👍🏻 Added some comments.

qooban:
@Gelio Sorry, wanted to pass review as "Comment". By mistake clicked "Approve" 🙂

kubama (dismissed):

app/components/Tabs.jsx
Gelio: My bad, I did not comment on it :slightly_smiling_face:  That's due to a circular dependency. If `Tabs.js` imported `shared/widgets/index.ts`, the chain would be

1. `Tabs.js`
2. `shared/widgets/index.ts`
3. `shared/widgets/PageContent.js` (exported from the `index.ts`)
4. back to `Tabs.js`, since `PageContent.js` imports it

Had it not been for that issue, I would have kept the import intact :slightly_smiling_face:

app/components/shared/widgets/Widget.tsx
Gelio: I'm not particularly interested in the header-related classes :slightly_smiling_face: I have left them as they are to avoid breaking existing behaviors. I have only added the `standalone` classes :slightly_smiling_face: 

If that's a concern to have the `standalone` class in 2 places, let me know, I will try to modify the SCSS to add it only to `Widget` :slightly_smiling_face:

widgets/deploymentsView/src/detailsPane/widgets.tsx
Gelio: > Do we need id fields here per each widget?

I'm pretty sure they would not be able to display the fetched data otherwise :slightly_smiling_face: 

https://github.com/cloudify-cosmo/cloudify-stage/blob/51abc96a4d0713361499e2f6c3ebdd60bf5c83ae/app/components/shared/widgets/Widget.tsx#L280

> Did you consider moving it to custom page JSON file? IMHO that would be much cleaner solution.

That's what RD-1826 is for (mentioned in the description) :slightly_smiling_face: I agree that's a miles better solution, but also a more complex one, and one that requires a larger time investment. Given the circumstances, I think we are pretty short on time here, thus, I've decided to postpone it :slightly_smiling_face:

widgets/deploymentsView/src/styles.scss
Gelio: Haha yeah :smile: I'm not an expert by any means, but I very much like CSS Grid :smile: And as you can probably see, it's quite easy to use :smile: I can only recommend :slightly_smiling_face:

widgets/deploymentsView/src/detailsPane/index.tsx
Gelio: Whoops, good point :smile: I've forgotten about it. Since the message will be changed to a more precise one anyway (RD-1829), so to expedite the work here, I'll add a note to that ticket to also extract the strings to `en.json` :slightly_smiling_face: 

Let me know if you'd rather do it now

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Do you mean the viewport for all tests :smile: ? Maybe, I don't see a problem with that, but I would rather not experiment with it in this PR and risk breaking something and then having to fix it with quite little time :slightly_smiling_face:

I can create a ticket for it, let me know if that's what you'd suggest here :slightly_smiling_face:

widgets/deploymentsView/src/detailsPane/header.tsx
Gelio: > Having DetailsPaneHeader is a nice idea. I assume this is to get part of the details always visible at the top. 

Thanks, and yes :slightly_smiling_face: 

> wouldn't it be better to have an option in widget containers to make them kind of fixed to the place where they are added?

I'm not sure how that would work inside a grid, e.g. if you marked one of the widgets that appear on the 2nd row as "fixed" :smile: 

Overall, I do believe that having some notion of fixed-ness is welcome. I would approach it by allowing having a fixed height for widget containers. Then, you could add 2 widget containers, one under another. If the 2nd one had a fixed height, if it had more widgets, a scrollbar would appear, and the first widget container would always appear on the top :slightly_smiling_face: 

Styling that would be definitely a challenge, though, if e.g. we wanted to align some widgets to match, as I have aligned the header to match the tabs below

Anyway, we could investigate that direction some time in the future :slightly_smiling_face:

I don't think that it would be usable here, though. It looks like at the current step that would work (we would have to add a widget that renders the `deploymentId` from the context, though :smile: not a useful one), but if we add the _drill-down buttons_, then either:
1. We pass them the information about the number of subdeployments and their status, from the table data
2. They make a request and fetch the data again

and I would lean towards the first solution to avoid sending multiple requests for the same data. 

To conclude, IMO the pattern of having some widgets always appear visible sounds useful and can be investigated in the future (e.g. a separate epic), but it would not help here :slightly_smiling_face:

widgets/deploymentsView/src/detailsPane/index.scss
Gelio: Actually, no, I didn't :slightly_smiling_face: Thanks for the reminder

Although, using `3.428571rem` and `0.714286rem` do not sound too appealing, since we are using `font-size: 14px` :slightly_smiling_face: 

Personally, I would leave it as is for brevity, and make a systemic change if we decide to embrace such units :slightly_smiling_face: Setting the base font size to 10px would make the calculations much nicer and the units easier to work with

Let me know if that sounds ok for now :slightly_smiling_face:

widgets/deploymentsView/src/detailsPane/header.tsx
Gelio: I'm glad I could show some new approaches with this PR :smile:

Gelio:
I've rebased my tree of branches after #1254 was merged. No need to reapprove it now, since I will need to rebase it anyway after the parent PR gets merged to master :slightly_smiling_face:

app/components/Tabs.jsx
qooban: OK, clear 👍🏻

widgets/deploymentsView/src/detailsPane/widgets.tsx
qooban: > I'm pretty sure they would not be able to display the fetched data otherwise 🙂 

If that's the case, then couldn't we have them generated somehow and not hardcoded? What if we end up with having the same UUIDs?

> That's what RD-1826 is for (mentioned in the description) 🙂 

Hmm... I'm not sure https://cloudifysource.atlassian.net/browse/RD-1826 is covering what I meant. Having a way to modify the widgets easily is one thing, but another thing is an API option to include the whole page JSON file instead of specifying the whole list of widgets inside the code. Will that issue cover both aspects? I think the second one (at least from my perspective) seems quite important, we can discuss that on planning or other 6.0 content related meeting.

Anyways, for the time being I suggest at least moving all that page object structure to a separate JSON file. That would bring more clarity and make this file smaller.

widgets/deploymentsView/src/detailsPane/index.tsx
qooban: OK, it could be done in https://cloudifysource.atlassian.net/browse/RD-1829 as long as it is in our plan.

widgets/deploymentsView/src/detailsPane/header.tsx
qooban: > I'm not sure how that would work inside a grid, e.g. if you marked one of the widgets that appear on the 2nd row as "fixed" smile

For me it seems like it would be only useful in the top widget containers, we shouldn't be afraid of not handling all the cases. I think that (fixed top header) would be the most common (and maybe the only one) case.

No need to change anything here, just a question: Can we have it reported in our JIRA?

widgets/deploymentsView/src/detailsPane/index.scss
qooban: Oh, if the alternative is to use `3.428571rem` and `0.714286rem` then let's leave it as it is 🙂

test/cypress/integration/widgets/deployments_view_spec.ts
qooban: OK, sure, having a ticket for it is perfectly fine.

qooban (commented):
Replied to replies 🙂

widgets/deploymentsView/src/detailsPane/widgets.tsx
Gelio: > If that's the case, then couldn't we have them generated somehow and not hardcoded? What if we end up with having the same UUIDs?

Hmm... For a base case, we need the uuids on the same page to be unique. `widgetData` is cleared on page change (https://github.com/cloudify-cosmo/cloudify-stage/blob/1581d65cae8999cdbec8d7db732cc93ab6d48be3/app/actions/page.ts#L193-L194) anyway :smile: So there's not much risk

> an API option to include the whole page JSON file instead of specifying the whole list of widgets inside the code

Do you mean a page in the `templates/pages` directory? I will assume yes in the rest of the comment below

> Will that issue cover both aspects? 

Yes :slightly_smiling_face: I have described that in the ticket, and added the path to the directory to hopefully make it clearer:
> One way this could be changed is to have a template (in templates/pages) based on which a new page will be created. That page will be then rendered inside the details pane.

---

> Anyways, for the time being I suggest at least moving all that page object structure to a separate JSON file. That would bring more clarity and make this file smaller.

Sure, I'll extract that to a JSON file

widgets/deploymentsView/src/detailsPane/index.tsx
Gelio: Great, I'll resolve the comment then :slightly_smiling_face:

widgets/deploymentsView/src/detailsPane/header.tsx
Gelio: Yup, here it is: https://cloudifysource.atlassian.net/browse/RD-1843

I have described both approaches, and very vaguely :slightly_smiling_face: IMO the matter needs further refinement

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Great, here it is: https://cloudifysource.atlassian.net/browse/RD-1844

I was not sure about the epic, as it's not tech debt per se, so I left it epicless :slightly_smiling_face:

Gelio:
I've rebased this branch on `master` to resolve merge conflicts since #1265 is merged. I'll be making 1 more commit to address https://github.com/cloudify-cosmo/cloudify-stage/pull/1267#discussion_r600228702

After that, I'll be looking for re-reviews :slightly_smiling_face:

widgets/deploymentsView/src/detailsPane/widgets.tsx
Gelio: Done, I've separated out the JSON file with page content :slightly_smiling_face:

Gelio:
I've queued up system tests:

https://jenkins.cloudify.co/job/Stage-UI-System-Test/327/

widgets/deploymentsView/src/detailsPane/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+     */
+    deployment: Deployment | undefined;
+}
+
+const DetailsPane: FunctionComponent<DetailsPaneProps> = ({ deployment }) => {
+    if (!deployment) {
+        const { Message } = Stage.Basic;
+
+        return (
+            <div className="unknownDeploymentMessage">
```
kubama, line 25: I can see this class is used only to set the `margin`. How about using inline style here?
Also, is there a need for the wrapping `div`? Can't the margin be set on `Message` directly?

widgets/deploymentsView/src/detailsPane/header.scss
```diff
@@ -0,0 +1,20 @@
...
+        flex-wrap: wrap;
+
+        // NOTE: adds a gap between the tabs headers when the items in the header wrap
+        margin-bottom: 10px;
+    }
+
+    .detailsPaneWidgets {
+        overflow: auto;
+    }
+}
```
kubama, line 20: Why not inline styles?

widgets/deploymentsView/src/detailsPane/widgets.json
```diff
@@ -0,0 +1,266 @@
+{
```
kubama, line 1: Rather then just widget definitions this file contains complete layout definition - I'd try to name this file in a more accurate way.

test/cypress/integration/widgets/deployments_view_spec.ts
qooban: Thanks. Added to RD-228 - UI System Tests epic.

qooban (dismissed):
Looks good 👍🏻 

From my side, I see only one thing worth to change - I agree with @kubama that `widgets.json` is not the most accurate name for the layout definition.

widgets/deploymentsView/src/detailsPane/widgets.json
Gelio: Makes sense :slightly_smiling_face: I would appreciate a suggestion next time, if you have a concrete name in mind. This way we avoid another back and forth

`page.json`? `pageLayout.json`?

kubama: I'd say `pageLayout.json` is perfect.

qooban: `pageLayout.json` sounds good 👍🏻

widgets/deploymentsView/src/detailsPane/index.tsx
Gelio: Using inline styles here makes sense since that would get rid of the `index.scss` file.

I didn't see a `style` attribute in the types for the `Message` component. Besides, here's how it looks without the div:
image

So I guess keeping the div is advised. I'll only move the styles to inline ones

widgets/deploymentsView/src/detailsPane/header.scss
Gelio: Why inline styles :slightly_smiling_face: ?

I prefer using CSS for styling, not inline styles using objects. We do use CSS for other components, so I don't see why it would be discouraged here.

widgets/deploymentsView/src/detailsPane/index.tsx
Gelio: Done, inline styles applied

widgets/deploymentsView/src/detailsPane/widgets.json
Gelio: Done, used `pageLayout.json`

Gelio:
I'll not be rerunning the system tests as the last 2 commits are without functional changes. The widget still works in the UI.

widgets/deploymentsView/src/detailsPane/header.scss
qooban: This is maybe something to discuss in the weekly meeting.

Some points to the discussion. My current way of thinking about styling in React is that:
1. It's better to use inline styles wherever the style is applied to a single component and no need to reuse it.
2. It's better to use styled-components or pure CSS styles whenever inline styles cannot handle desired functionality (e.g. hovering)

qooban (approved):

widgets/deploymentsView/src/detailsPane/header.scss
kubama: I'm not expecting change here, but I think our approach so far was to use inline styles when possible and other options when for some reason they suited better. External (s)css files provide some overhead in terms of code navigation as well as maintenance - class names are just strings and they have to match between component and css file. This is inconvenient and may be error prone.
Nevertheless, this is a side comment.

kubama (approved):

widgets/deploymentsView/src/detailsPane/header.scss
Gelio: Oh, so you both have different opinions here :smile: Let's take it during the weekly :slightly_smiling_face:

qooban: I don't think our opinions are different. I think they are quite similar 🙂 
Anyways, worth to discuss it in the weekly.

Gelio: I forgot to say "from me" :smile: It should have been

> Oh, so you both have different opinions from me here :smile:

PR #1270, opened 2021-03-23
RD-1741 Show real latest execution progress

Gelio:
This PR adds showing the real latest execution progress as it comes from the API. It bases on #1267 

The rules describing the behavior of the progress bar are in https://cloudifysource.atlassian.net/wiki/spaces/WR/pages/1674215425/Deployment+states ("Progress bar" point)

The API documentation for those fields is in https://cloudifysource.atlassian.net/wiki/spaces/WR/pages/1572864001/Environment#Data-Model

## Screenshot

image

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/335/pipeline

Ran 2021-03-25 15:15 CET

### DIFF
--- a/app/utils/Context.ts
+++ b/app/utils/Context.ts
@@ -1,6 +1,16 @@
 import { MiddlewareAPI } from 'redux';
 import { setValue as setContextValue } from '../actions/context';
 
+export interface WellKnownContextEntries {
+    deploymentId: string | string[] | null;
+    blueprintId: any;
+    nodeId: any;
+    executionId: any;
+    nodeInstanceId: any;
+    executionStatus: any;
+    siteName: any;
+}
+
 /** @see https://docs.cloudify.co/developer/writing_widgets/widget-apis/#getcontext */
 export default class Context {
     private readonly store: MiddlewareAPI;
@@ -21,11 +31,22 @@ export default class Context {
      * This means that calling `getValue` after calling `setState` will keep yielding the value
      * from before `setState` until the widget rerenders.
      */
-    public setValue(key: any, value: any) {
+    public setValue<K extends keyof WellKnownContextEntries>(
+        key: K,
+        value: WellKnownContextEntries[K] | undefined
+    ): void;
+
+    public setValue(key: string, value: any): void;
+
+    public setValue(key: string, value: any) {
         this.store.dispatch(setContextValue(key, value));
     }
 
-    public getValue(key: any) {
+    public getValue<K extends keyof WellKnownContextEntries>(key: K): WellKnownContextEntries[K] | undefined;
+
+    public getValue(key: string): any;
+
+    public getValue(key: string) {
         return this.context[key];
     }
 }
--- a/app/utils/Context.ts
+++ b/app/utils/Context.ts
@@ -1,16 +1,6 @@
 import { MiddlewareAPI } from 'redux';
 import { setValue as setContextValue } from '../actions/context';
 
-export interface WellKnownContextEntries {
-    deploymentId: string | string[] | null;
-    blueprintId: any;
-    nodeId: any;
-    executionId: any;
-    nodeInstanceId: any;
-    executionStatus: any;
-    siteName: any;
-}
-
 /** @see https://docs.cloudify.co/developer/writing_widgets/widget-apis/#getcontext */
 export default class Context {
     private readonly store: MiddlewareAPI;
@@ -31,10 +21,7 @@ export default class Context {
      * This means that calling `getValue` after calling `setState` will keep yielding the value
      * from before `setState` until the widget rerenders.
      */
-    public setValue<K extends keyof WellKnownContextEntries>(
-        key: K,
-        value: WellKnownContextEntries[K] | undefined
-    ): void;
+    public setValue<K extends keyof Stage.ContextEntries>(key: K, value: Stage.ContextEntries[K] | undefined): void;
 
     public setValue(key: string, value: any): void;
 
@@ -42,7 +29,7 @@ export default class Context {
         this.store.dispatch(setContextValue(key, value));
     }
 
-    public getValue<K extends keyof WellKnownContextEntries>(key: K): WellKnownContextEntries[K] | undefined;
+    public getValue<K extends keyof Stage.ContextEntries>(key: K): Stage.ContextEntries[K] | undefined;
 
     public getValue(key: string): any;
 
--- a/widgets/deploymentsView/src/renderDeploymentRow.tsx
+++ b/widgets/deploymentsView/src/renderDeploymentRow.tsx
@@ -8,7 +8,7 @@ const renderDeploymentRow = (toolbox: Stage.Types.Toolbox, fieldsToShow: Deploym
     deployment: Deployment
 ) => {
     const { DataTable } = Stage.Basic;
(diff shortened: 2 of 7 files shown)


widgets/deploymentsView/src/renderDeploymentRow.tsx
```diff
@@ -1,11 +1,14 @@
...
 import { deploymentsViewColumnDefinitions, DeploymentsViewColumnId } from './columns';
+import { Deployment, LatestExecutionStatus } from './types';
 
 const renderDeploymentRow = (toolbox: Stage.Types.Toolbox, fieldsToShow: DeploymentsViewColumnId[]) => (
-    deployment: any
+    deployment: Deployment
 ) => {
     const { DataTable } = Stage.Basic;
-    const selectedDeploymentId = toolbox.getContext().getValue('deploymentId');
+    const selectedDeploymentId: string | string[] | null | undefined = toolbox.getContext().getValue('deploymentId');
```
qooban, line 11: Can it (`string | string[] | null | undefined`) be extracted to a single type?

widgets/deploymentsView/src/renderDeploymentRow.tsx
```diff
@@ -1,11 +1,14 @@
+import { pick } from 'lodash';
 import { ReactNode } from 'react';
+
 import { deploymentsViewColumnDefinitions, DeploymentsViewColumnId } from './columns';
+import { Deployment, LatestExecutionStatus } from './types';
```
qooban, line 5: I see we've got small difference in naming here. You used `LatestExecutionStatus`, I used `LastExecutionStatus` (eg. `LastExecutionStatusIcon`). I think right now that your name is better, so could you create a ticket to align it?

widgets/deploymentsView/src/styles.scss
```diff
@@ -1,16 +1,15 @@
...
-$deploymentStateColors: (
-    'good': green,
-    'in-progress': orange,
-    'failure': red,
-    'pending': blue
+$deploymentProgressBarColors: (
+    // NOTE: use Semantic UI colors
+    // See https://semantic-ui.com/usage/theming.html
+    'in-progress': #f2711c,
+    'failed': #db2828
```
qooban, line 5: Hmm... I'm wondering if Semantic UI expose variables for the colors?
(probably you checked that, so I assume not, but wanted to ask anyway to be sure)

qooban (commented):
Looks good, added some minor comments.

kubama (dismissed):

widgets/deploymentsView/src/renderDeploymentRow.tsx
Gelio: Yup, definitely a worthy addition :+1: In fact, I can add some overloads for `getValue` in `Context` so that it will be well-typed whenever we call `getValue('deploymentId')` :smile:  This way we will not have to repeat it or have it only in this feature, but across the system

I'll add a commit for it soon :slightly_smiling_face:

widgets/deploymentsView/src/renderDeploymentRow.tsx
Gelio: Here you go: https://cloudifysource.atlassian.net/browse/RD-1836 :slightly_smiling_face:

widgets/deploymentsView/src/styles.scss
Gelio: As far as I have checked, Semantic UI does expose such variables, but it's written in Less, so I'm not sure how to use them from inside SCSS :slightly_smiling_face: 

https://semantic-ui.com/usage/theming.html#sitewide-defaults

The colors for existing Semantic UI components (e.g. the color of the icon) are hardcoded very hard :smile: 
image

Notice the `!important`. That's what I noticed when I was looking to see if there are Semantic UI's classes that I can reuse (I did not find any)

---

To conclude, I did not find a way to reuse the colors

Gelio:
I've rebased my tree of branches after #1254 was merged. I'll add 1 more commit to address https://github.com/cloudify-cosmo/cloudify-stage/pull/1270#discussion_r600240914

widgets/deploymentsView/src/renderDeploymentRow.tsx
Gelio: It turns out that was a great idea :smile: 

It now allows adding safe typing for context entries that may be added by customers in custom widgets

image

app/utils/Context.ts
```diff
@@ -21,11 +21,19 @@ export default class Context {
      * This means that calling `getValue` after calling `setState` will keep yielding the value
      * from before `setState` until the widget rerenders.
      */
-    public setValue(key: any, value: any) {
+    public setValue<K extends keyof Stage.ContextEntries>(key: K, value: Stage.ContextEntries[K] | undefined): void;
+
+    public setValue(key: string, value: any): void;
+
+    public setValue(key: string, value: any) {
```
qooban, line 28: Probably TS-dummy question: Why we have it separated?

widgets/deploymentsView/src/renderDeploymentRow.tsx
qooban: Thanks 👍🏻

widgets/deploymentsView/src/styles.scss
qooban: Thanks for the explanation.

I'm now just wondering if there is anything to do here to have less work on https://cloudifysource.atlassian.net/browse/RD-611 in the future.

If you don't see anything, I'm not blocking it.

qooban (approved):

app/utils/Context.ts
Gelio: That separation is necessary since that's the way of describing method overloads in TypeScript :smile: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

FYI, if the method on line 26 was removed, then TS would assume that there's only 1 overload (the one above), and not allow passing any `string` as `key`:
image

qooban: Clear 👍🏻

widgets/deploymentsView/src/styles.scss
Gelio: Here, since the styling is implemented using CSS custom properties, the customer can just override those custom properties from within their external CSS file :slightly_smiling_face: 

```scss
.deploymentsViewWidget deploymeent-progress-bar {
  --deployment-progress-bar-color-in-progress: blue; // maybe !important
}
```

So IMO that's already good enough :slightly_smiling_face: And having custom properties is IMO the easiest way to allow for customer customization :smile: but that's slightly off topic

qooban: Make sense. Updated RD-611 to not forget about CSS custom properties.

qooban (dismissed):

Gelio:
I've rebased my tree of branches after #1265 was merged

Still, #1267 needs to be merged first before this one

Gelio:
As with the last 3 PRs, I've rebased this one on `master`, ready to be merged. Queued up system tests:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/335/pipeline

I'm looking for reapprovals slightly_smiling_face

qooban (approved):

kubama (approved):

PR #1271, opened 2021-03-24
RD-1754 Prevent `creating` snapshot from being deleted, updated cypress test accordingly

kubama:
See latest Michael's comment to RD-1754.
System test job run: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/323/

### DIFF
--- a/widgets/snapshots/src/SnapshotsTable.jsx
+++ b/widgets/snapshots/src/SnapshotsTable.jsx
@@ -7,10 +7,6 @@ import RestoreModal from './RestoreSnapshotModal';
 import UploadModal from './UploadSnapshotModal';
 import SnapshotPropType from './props/SnapshotPropType';
 
-function isSnapshotUseful(snapshot) {
-    return snapshot.status === 'created' || snapshot.status === 'uploaded';
-}
-
 export default class SnapshotsTable extends React.Component {
     constructor(props, context) {
         super(props, context);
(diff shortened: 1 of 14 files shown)


widgets/snapshots/src/SnapshotsTable.jsx
```diff
@@ -140,6 +140,7 @@ export default class SnapshotsTable extends React.Component {
 
                     {data.items.map(item => {
                         const isUseful = isSnapshotUseful(item);
```
qooban, line 142: What do you think about changing:
```suggestion
                        const isUsable = isSnapshotUsable(item);
```
?

I know it was not introduced in this PR, but when I look at it now it is misleading.

widgets/snapshots/src/SnapshotsTable.jsx
```diff
@@ -140,6 +140,7 @@ export default class SnapshotsTable extends React.Component {
...
+                        const isRemovable = isUseful || item.status === 'failed';
```
qooban, line 143: Maybe to have `isRemovable` and `isUseful` consts definition more consistent you can either remove `isSnapshotUsable` function or extract this part to a function?

qooban (commented):
Added two minor comments.

kubama:
I addressed the comments.
Snapshots and topology widget tests passed on CI.

qooban (approved):

PR #1272, opened 2021-03-24
RD-1742 Show subdeployments count and status

Gelio:
This PR bases on #1270 to show counts and statuses for subdeployments (subenvironments and subservices).

The icon rendering logic is shared with the main deployment status column (the first column). The rules are described in https://cloudifysource.atlassian.net/wiki/spaces/WR/pages/1674215425/Deployment+states.

## Screenshot

image

The point of focus in the screenshot are the last 2 columns of the table :slightly_smiling_face: 

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/340/pipeline

Ran 2021-03-26 12:46 CET


Gelio:
I've rebased my tree of branches after #1265 was merged

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -157,27 +184,44 @@ describe('Deployments View widget', () => {
...
+                        iconLabel: 'Requires attention'
                     });
                     verifyProgressBar('failed', '60%');
                 });
 
             cy.contains('hello-world-one')
                 .parents('tr')
                 .within(() => {
-                    cy.get('i.spinner[aria-label="In progress"]').as('inProgressIcon').trigger('mouseover');
+                    cy.get(`td:nth-of-type(${columnNumbers.status}) i.spinner[aria-label="In progress"]`)
```
Gelio, line 210: 2 icons matched inside the row, so I had to add selecting inside the specific column

widgets/deploymentsView/src/columns.tsx
```diff
@@ -28,7 +28,42 @@ export interface DeploymentsViewColumnDefinition {
...
+    inProgress: {
+        labelI18nKey: 'inProgress',
+        props: { name: 'spinner', color: 'orange' }
+    },
+    requiresAttention: {
+        labelI18nKey: 'requiresAttention',
+        props: { name: 'exclamation', color: 'red' }
+    }
+});
+type StatusIconName = keyof typeof statusIconDescriptions;
```
qooban, line 54: Hmm... Intuition keeps saying to me that it is over-engineered 🙂 

1. Do we really need `labelI18nKey` as it is the same as the object key? 
2. Maybe we could have enum or consts defined for `inProgress` and `requiresAttention` keys. That way we could avoid providing strings later on in this code, e.g. here:
```
[DeploymentStatus.InProgress]: 'inProgress',
[DeploymentStatus.RequiresAttention]: 'requiresAttention'
```
3. `StatusIconName` can have predefined set of keys based on what you define in p. 2 
(it has now just two keys, so maybe not worth to create it dynamically using `keyof typeof statusIconDescriptions`).

widgets/deploymentsView/src/columns.tsx
```diff
@@ -28,7 +28,42 @@ export interface DeploymentsViewColumnDefinition {
...
+const renderStatusIcon = (iconName: StatusIconName) => {
+    const { Icon, Popup } = Stage.Basic;
+    const iconDescription = statusIconDescriptions[iconName];
+    const label = Stage.i18n.t(`${i18nPrefix}.iconLabels.${iconDescription.labelI18nKey}`);
```
qooban, line 58: If you simplify the code (see my first comment), you should be able to do just:
```suggestion
    const label = Stage.i18n.t(`${i18nPrefix}.iconLabels.${iconName}`);
```

widgets/deploymentsView/src/columns.tsx
```diff
@@ -28,7 +28,42 @@ export interface DeploymentsViewColumnDefinition {
...
+
+    return (
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        <Popup trigger={<Icon aria-label={label} {...iconDescription.props} />} position="top center">
```
qooban, line 62: If you simplify the code (see my first comment), you should be able to change it to:
```suggestion
        <Popup trigger={<Icon aria-label={label} {...iconDescription} />} position="top center">
```

I would consider changing `iconDescription` to have better name, probably `iconProps` is more accurate.

widgets/deploymentsView/src/columns.tsx
```diff
@@ -37,24 +72,17 @@ const partialDeploymentsViewColumnDefinitions: Record<
...
         width: '20px',
         render(deployment) {
-            const { Icon, Popup } = Stage.Basic;
-            const deploymentStatusIconProps: Record<DeploymentStatus, Pick<IconProps, 'name' | 'color'> | undefined> = {
+            const deploymentStatusIconProps: Record<DeploymentStatus, StatusIconName | undefined> = {
                 [DeploymentStatus.Good]: undefined,
-                [DeploymentStatus.InProgress]: { name: 'spinner', color: 'orange' },
-                [DeploymentStatus.RequiresAttention]: { name: 'exclamation', color: 'red' }
+                [DeploymentStatus.InProgress]: 'inProgress',
+                [DeploymentStatus.RequiresAttention]: 'requiresAttention'
```
qooban, line 78: If you simplify the code (see my first comment), you should be able to avoid using strings and have something like:
```suggestion
                [DeploymentStatus.InProgress]: statusIcon.inProgress,
                [DeploymentStatus.RequiresAttention]: statusIcon.requiresAttention
```

widgets/deploymentsView/src/columns.tsx
```diff
@@ -86,30 +114,54 @@ const partialDeploymentsViewColumnDefinitions: Record<
...
         }
+        /* eslint-enable camelcase, react/prop-types */
     }
 };
 
+const subdeploymentStatusToIconMapping: Record<SubdeploymentStatus, StatusIconName | undefined> = {
+    [SubdeploymentStatus.InProgress]: 'inProgress',
+    [SubdeploymentStatus.Good]: undefined,
+    [SubdeploymentStatus.Failed]: 'requiresAttention',
+    [SubdeploymentStatus.Pending]: undefined
```
qooban, line 153: If you simplify the code (see my first comment), you should be able to avoid using strings and have something like:
```suggestion
    [SubdeploymentStatus.InProgress]: statusIcon.inProgress,
    [SubdeploymentStatus.Good]: undefined,
    [SubdeploymentStatus.Failed]: statusIcon.requiresAttention,
    [SubdeploymentStatus.Pending]: undefined
```

qooban (commented):
I believe that the code could be simplified, so I added one comment with some ideas and 5 other comments are just referring to the first one 🙂

widgets/deploymentsView/src/columns.tsx
Gelio: > Hmm... Intuition keeps saying to me that it is over-engineered

Hmm... :thinking: maybe



> 1. Do we really need `labelI18nKey` as it is the same as the object key?

That part can be deduplicated I guess, good point, so the object will contains the props directly



> 2\. Maybe we could have enum or consts defined for `inProgress` and `requiresAttention` keys.

That's what I wanted to avoid by having the Constrained Identity Function approach :slightly_smiling_face: That's the point to use JavaScript and infer the types in TypeScript :slightly_smiling_face: 

The benefit? Less duplication :smile:  You don't need to first define the enum, then define the object with props for each enum value. It feels more JavaScripty to me, and TS is just a cherry on top that infers the types :slightly_smiling_face: 

I can see how it can be more complex, though. I encourage reevaluating it :slightly_smiling_face: 

---

Keep in mind that the keys in the snippet you shared are type-safe - they have to match the keys in the object

widgets/deploymentsView/src/columns.tsx
Gelio: I'm not addressing this one and the one below yet, to see your reply to my arguments above :slightly_smiling_face:

Gelio:
I've only rebased the PR since one of the parent branches have been merged. I have not applied any changes yet

Gelio:
Again, I have just rebased the branch on `master` now, no functional changes

widgets/deploymentsView/src/columns.tsx
qooban: OK, I think I don't have work-load to investigate it deeper.

If you could just use props directly in the object, that'd probably be good enough.

widgets/deploymentsView/src/columns.tsx
qooban: Not relevant anymore. Resolving.

widgets/deploymentsView/src/columns.tsx
qooban: Not relevant anymore. Resolving.

qooban (commented):
Answered and resolved some comments.

widgets/deploymentsView/src/columns.tsx
Gelio: Ok, sounds good :slightly_smiling_face: If you would have some free time, I recommend the blogpost that I linked in the comment:

https://kentcdodds.com/blog/how-to-write-a-constrained-identity-function-in-typescript

That will probably explain the use-case better :slightly_smiling_face: 

---

I'll change the structure in a few minutes

widgets/deploymentsView/src/columns.tsx
Gelio: I've applied both suggestions :slightly_smiling_face:

widgets/deploymentsView/src/columns.tsx
Gelio: Done

widgets/deploymentsView/src/columns.tsx
Gelio: I've removed the unnecessary wrapping for `props` - the object contains `IconProps` as values now

Gelio:
System tests:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/340/pipeline

qooban (approved):

Gelio:
Looks like both CI pipelines passed :smile: I'm finally merging this PR, no more rebases

PR #1273, opened 2021-03-25
RD-1835 - Improve deployment_actions_spec to be more robust

qooban:
This is a simple change to make `deployment_actions_spec` more robust by:
* waiting for "labels existence check" request to be send
* waiting for the label to be added to the labels list

before starting the next actions.

In my local environment test is passing. CI (only deployment_actions_spec) results will be here: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/330

See RD-1835 for more details.

### DIFF
--- a/test/cypress/integration/widgets/deployment_actions_spec.js
+++ b/test/cypress/integration/widgets/deployment_actions_spec.js
@@ -62,9 +62,14 @@ describe('Deployment Action Buttons widget', () => {
             cy.get('div[name=labelValue] > input').clear().type(value);
         }
         function addLabel(key, value) {
+            cy.interceptSp('GET', `/labels/deployments/${key}?_search=${value}`).as('fetchLabel');
+
             typeLabelKey(key);
             typeLabelValue(value);
             cy.get('button[aria-label=Add]').click();
+
+            cy.wait('@fetchLabel');
+            cy.contains('a.label', `${key} ${value}`).should('exist');
         }
         function checkIfPopupIsDisplayed(key, popupContent) {
             typeLabelKey(key);
@@ -100,8 +105,6 @@ describe('Deployment Action Buttons widget', () => {
         it('adds new label by typing', () => {
             cy.get('.modal').within(() => {
                 addLabel('sample_key', 'sample_value');
-
-                cy.contains('a.label', 'sample_key sample_value').should('be.visible');
             });
         });
 


test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -62,9 +62,14 @@ describe('Deployment Action Buttons widget', () => {
...
         }
         function addLabel(key, value) {
+            cy.interceptSp('GET', `/labels/deployments/${key}?_search=${value}`).as('fetchLabel');
+
             typeLabelKey(key);
             typeLabelValue(value);
             cy.get('button[aria-label=Add]').click();
+
+            cy.wait('@fetchLabel');
+            cy.contains('a.label', `${key} ${value}`).should('exist');
```
Gelio, line 72: Do we care about `exist` or `be.visible`? The latter may give additional confidence

test/cypress/integration/widgets/deployment_actions_spec.js
```diff
@@ -62,9 +62,14 @@ describe('Deployment Action Buttons widget', () => {
             cy.get('div[name=labelValue] > input').clear().type(value);
         }
         function addLabel(key, value) {
+            cy.interceptSp('GET', `/labels/deployments/${key}?_search=${value}`).as('fetchLabel');
+
             typeLabelKey(key);
             typeLabelValue(value);
             cy.get('button[aria-label=Add]').click();
+
+            cy.wait('@fetchLabel');
```
Gelio, line 71: Shouldn't this wait be before clicking _Add_?

Gelio (commented):
Some general comments :slightly_smiling_face: And I see that something weird happened in CI WRT test coverage :smile:

kubama (approved):

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: That fetch happens after clicking the *Add* button. It checks if the label exists already in the system. Depending on the result we have either grey or blue label.

test/cypress/integration/widgets/deployment_actions_spec.js
qooban: In that case I think we don't need to care, I think `exist` is enough. `be.visible` checks if it is available in the current view, which can lead to problems in case of scrolls, IIRC.

Anyways, I'm OK with both - `exist` or `be.visible` - in this case. If you prefer changing, I can do the change.

qooban:
> Some general comments  And I see that something weird happened in CI WRT test coverage

All fine, I think. Test coverage will fail when specifying single test execution.

test/cypress/integration/widgets/deployment_actions_spec.js
Gelio: No need :smile: I forgot to mark the comment as optional, or maybe a general question :slightly_smiling_face:

test/cypress/integration/widgets/deployment_actions_spec.js
Gelio: Alright :smile: I didn't check, thanks for clarifying

Gelio:
> All fine, I think. Test coverage will fail when specifying single test execution.

Right, I forgot :slightly_smiling_face:

Gelio (approved):

PR #1274, opened 2021-03-25
RD-1123 Added basic filters widget implementation (read-only)

kubama:
Just the very basic, read-only widget implementation.

New and noteworthy: 100% TS 🙂.

image


widgets/filters/src/FilterActions.ts
```diff
@@ -0,0 +1,11 @@
+export default class FilterActions {
+    toolbox: Stage.Types.Toolbox;
+
+    constructor(toolbox: Stage.Types.Toolbox) {
+        this.toolbox = toolbox;
+    }
```
Gelio, line 6: Optional:

Here's a cool trick to consider :slightly_smiling_face: 

```suggestion
    constructor(private toolbox: Stage.Types.Toolbox) {}
```

Although, `no-useless-constructor` is not too happy about it. We would need to disable it and enable https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-useless-constructor.md to fix the false-positive. I guess this should be done in `cloudify-ui-common` at some point

If you decide to apply the suggestion, I'm ok with having a `eslint-disable-next-line` or changing the ESLint config file and adding the `@typescript-eslint` version there :slightly_smiling_face:

widgets/filters/src/FilterActions.ts
```diff
@@ -0,0 +1,11 @@
...
+
+    doList(params: unknown) {
```
Gelio, line 8: Do we know the type of params and the response type?

If yes, supplying it here would be great for readability and discoverability :slightly_smiling_face: Not having to jump to the devtools/docs/explore uses of the function is miles worse than having the type right here :slightly_smiling_face: 

```suggestion
    doList(params: ???): Promise<Stage.Types.PaginatedResonse<Filter> {
```

widgets/filters/src/FiltersTable.tsx
```diff
@@ -0,0 +1,44 @@
...
+    total: number;
+}
+
+interface FiltersTableProps {
+    data: FiltersTableData;
+    toolbox: Stage.Types.Toolbox;
+    widget: Widget;
+}
+
+export default function FiltersTable({ data, toolbox, widget }: FiltersTableProps) {
```
Gelio, line 16: Optional:

I suggest using the `FunctionComponent` or `FC` type from `react` as a type annotation here:

```suggestion
const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, widget }) => {
```

and then manually
```ts
export default FiltersTable;
```

The benefits are:
1. The return type is strictly validated (here you could have returned anything, no type safety for the return type)
1. The function is automatically detected as a component, thus allowing you to specify `displayName`, `defaultProps`, `propTypes` etc (does not apply here, just FYI)
1. `children` are automatically included in the props (does not apply here, though, just FYI)

Given that only one benefit applies here, treat this comment more as a general note for the future :slightly_smiling_face:

widgets/filters/src/FiltersTable.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const { i18n } = Stage;
+    const { DataTable } = Stage.Basic;
+    const { Time } = Stage.Utils;
+
+    return (
+        <>
```
Gelio, line 22: What's the fragment for? Looks like we are returning a single element (`DataTable`) anyway :thinking:

widgets/filters/src/FiltersTable.tsx
```diff
@@ -0,0 +1,44 @@
...
+                noDataMessage={i18n.t('widgets.filters.noFilters')}
+                fetchData={toolbox.refresh}
+                totalSize={data.total}
+                pageSize={widget.configuration.pageSize}
+                searchable
+                sortable
+            >
+                <DataTable.Column width="1%" label={i18n.t('widgets.filters.columns.name')} name="id" />
+                <DataTable.Column width="1%" label={i18n.t('widgets.filters.columns.creator')} name="created_by" />
+                <DataTable.Column width="1%" label={i18n.t('widgets.filters.columns.created')} name="created_at" />
```
Gelio, line 33: What's the effect of `width="1%"`? That sounds peculiar to me :smile:

widgets/filters/src/types.ts
```diff
@@ -0,0 +1,13 @@
...
+    created_by: string;
+    // eslint-disable-next-line camelcase
+    created_at: string;
+}
+
+export interface WidgetConfiguration {
+    pageSize: number;
+}
+
+export type Widget = Stage.Types.Widget<WidgetConfiguration>;
```
Gelio, line 13: I'm worried that such generally named types (without any `Filters` prefix) will start appearing globally when using intellisense, e.g.


```
function acceptsWidget(widget: Widget
```

and if you ran some kind of "auto import", it could import the file from here.

I would encourage naming them with a prefix

```suggestion
export interface FiltersWidgetConfiguration {
    pageSize: number;
}

export type FiltersWidget = Stage.Types.Widget<WidgetConfiguration>;
```

widgets/filters/src/types.ts
```diff
@@ -0,0 +1,13 @@
+export interface Filter {
```
Gelio, line 1: I believe that''s the type we can use in `FilterActions` :slightly_smiling_face:

widgets/filters/src/widget.tsx
```diff
@@ -0,0 +1,37 @@
...
+    initialConfiguration: [Stage.GenericConfig.POLLING_TIME_CONFIG(30), Stage.GenericConfig.PAGE_SIZE_CONFIG(10)],
+
+    fetchData(_widget, toolbox, params) {
+        return new FilterActions(toolbox).doList(params);
+    },
+
+    render(widget, data, _error, toolbox) {
+        if (Stage.Utils.isEmptyWidgetData(data)) {
+            const { Loading } = Stage.Basic;
+            return <Loading />;
```
Gelio, line 27: Optional:

We can do

```suggestion
            return <Stage.Basic.Loading />;
```

To save one line :slightly_smiling_face:

widgets/filters/src/widget.tsx
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        const formattedData = {
```
Gelio, line 30: Specifying the type here would improve the error message should there be a type mismatch:

```suggestion
        const formattedData: FiltersTableData = {
```

For that we would have to either export `FiltersTableData` type or use:

```suggestion
        const formattedData: ComponentProps<typeof FiltersTable>['data'] = {
```

Gelio (commented):
Awesome work using TypeScript :tada: 

Looks quite good overall :slightly_smiling_face: I did not test it in the UI, though

btw. how do you create filters? Do you use the CLI?

kubama:
> btw. how do you create filters? Do you use the CLI?

Yes, it's extremely easy.

widgets/filters/src/FilterActions.ts
kubama: I applied the suggestion and used `eslint-disable-next-line` for now.

widgets/filters/src/FilterActions.ts
kubama: My understanding is that here `params` is a black box that we are just passing further, without caring what it actually is. So, its type should be determined only by what `doGet` accepts - but it is not yet in TS so it accepts `any`.

Question: if `doGet` specified its parameters types then would it be necessary to specify type for `params`? Would type inference work here?

Regarding return type - I just added it.

Gelio: > My understanding is that here `params` is a black box that we are just passing further, without caring what it actually is. So, its type should be determined only by what `doGet` accepts - but it is not yet in TS so it accepts `any`.

The `any` type (could have been `unknown` though) in `doGet` is because `doGet` is very general. It should accept any URL, and thus, any parameters.

Here we have a specialized function - one that fetches the filters from a very specific URL. That API endpoint also has a known set of parameters that it accepts. Knowing them, we can do better than `any` or `unknown` :slightly_smiling_face: 

> Question: if `doGet` specified its parameters types then would it be necessary to specify type for `params`? Would type inference work here?

Unfortunately, not out of the box. You would have to import the type that `doGet` uses for parameter, and then use it here, or, do some magic with the `Parameters` type:

```ts
type Manager = ReturnType<Stage.Types.Toolbox['getManager']>;
type Params = Parameters<Manager['doGet']>[1];
```

I would recommend the first approach, though, if `doGet`'s parameters are ever strongly typed :smile:

widgets/filters/src/FiltersTable.tsx
qooban: Yep, that's interesting for me, too 🙂

widgets/filters/src/FiltersTable.tsx
kubama: If I do this, eslint forces me to add `propTypes`. My understanding is that I don't need it if this component is used only in TS code base, right?

widgets/filters/src/FiltersTable.tsx
kubama: Yes, you are right. I'll leave it though as I would need to add it in the next PR anyway (modals).

widgets/filters/src/FiltersTable.tsx
Gelio: Hmm... right. I guess we should do something about this rule, since it may get more and more annoying. Maybe we could disable it today :smile: A topic for the weekly?

widgets/filters/src/FiltersTable.tsx
kubama: In this case it means 1/3 of total width.

widgets/filters/src/FiltersTable.tsx
Gelio: Ok, no problem :slightly_smiling_face:

widgets/filters/src/widget.tsx
```diff
@@ -0,0 +1,37 @@
+import FilterActions from './FilterActions';
+import FiltersTable from './FiltersTable';
+import type { Filter, WidgetConfiguration } from './types';
+
+Stage.defineWidget<unknown, Stage.Types.PaginatedResponse<Filter>, WidgetConfiguration>({
+    id: 'filters',
+    name: Stage.i18n.t('widgets.filters.name'),
+    description: Stage.i18n.t('widgets.filters.description'),
+    initialWidth: 12,
+    initialHeight: 28,
```
qooban, line 10: Nothing bad with it, I'm just wondering why `28`? 
Most of the widgets are smaller in terms of height.

widgets/filters/src/widget.tsx
```diff
@@ -0,0 +1,37 @@
...
+    color: 'olive',
```
qooban, line 11: General:

I'm thinking that at some point we'll have to review all the colors we applied to the widgets and maybe group them by the colors or so. Now we are assigning it randomly, or maybe there is a rule which I'm not aware of?

qooban (commented):
Good TypeScript job 👏🏻 

Added some minor/general comments.

widgets/filters/src/FiltersTable.tsx
kubama: > A topic for the weekly?

Sure

widgets/filters/src/widget.tsx
kubama: I set default page size to 10, so I wanted 10 rows to be displayed without the scrollbar.

widgets/filters/src/widget.tsx
kubama: > or maybe there is a rule which I'm not aware of?

I'm not aware of any rule either. In my case it depends on which existing widget I choose as a base to copy from to start work on the new one.

widgets/filters/src/widget.tsx
kubama: Some time ago @qooban suggested to always use destructuring when using things from `Stage`, so I'm trying to align to this rule.

widgets/filters/src/FiltersTable.tsx
Gelio: Wow, some table wizardry :open_mouth: 

Can we have a comment explaining that? Or do we need it to be 1/3 of the width? I'm concerned that this will be perplexing to other devs (me included) in the future too

widgets/filters/src/widget.tsx
Gelio: Hmm... I thought that it was impossible before due to some ESLint rule that prevented it. I was not informed about such spoken rules :smile: 

@qooban do you think the rule still makes sense? Do you see benefits destructuring here?

Personally I prefer the one-line version :slightly_smiling_face:

widgets/filters/src/FiltersTable.tsx
kubama: It's temporary - I will have to change it once fixed-width actions column is introduced.

Gelio: Ok then, let's leave this magic spell here :slightly_smiling_face:

widgets/filters/src/FilterActions.ts
```diff
@@ -0,0 +1,10 @@
+import { Filter } from './types';
```
Gelio, line 1: Optional:

`import type` would save webpack some work :slightly_smiling_face: 

```suggestion
import type { Filter } from './types';
```

widgets/filters/src/FilterActions.ts
kubama: > The any type (could have been unknown though) in doGet is because doGet is very general. It should accept any URL, and thus, any parameters.

Not really. `params` are just query parameters, so the expected type is an object keyed by string. It's just not typed yet. If it was, the type for `params` would be naturally enforced.

> Here we have a specialized function - one that fetches the filters from a very specific URL. That API endpoint also has a known set of parameters that it accepts. Knowing them, we can do better than any or unknown 🙂

What for if nobody would use parameters defined by this type, at least not at this point?

But generally, looking at how the REST API works, I suspect it should be possible to create a general type describing query params - there are some common parts (pagination, sorting, searching, including specific properties) while the endpoint specific part is based on the response item type. So I imagine a parametrized type, like this: `QueryParams<Filter>`. However, since I don't need such type at the moment I'd rather not spend time on it now.

Gelio: > Not really. `params` are just query parameters, so the expected type is an object keyed by string. It's just not typed yet. If it was, the type for `params` would be naturally enforced.

Yeah, I guess `Record<string, any>` sounds reasonable there

> So I imagine a parametrized type, like this: QueryParams<Filter>

Makes sense

> However, since I don't need such type at the moment I'd rather not spend time on it now.

Sounds good. If the params accepted by the endpoint are the same as any other endpoint (essentially the `gridParams`), that's fine :+1:

widgets/filters/src/FiltersTable.tsx
kubama: I applied your suggestion and disabled `react/prop-types` rule for tsx files.

widgets/filters/src/types.ts
kubama: I agree, changed.

app/.eslintrc.json
```diff
@@ -10,5 +10,14 @@
...
+    },
+    "overrides": [
+        {
+            // NOTE: disable globally once TS migration is done
+            "files": ["**/*.tsx"],
+            "rules": {
+                "react/prop-types": "off"
+            }
+        }
+    ]
```
Gelio, line 22: Isn't `./.eslintrc.temp-overrides.json` a more suitable place for this override :slightly_smiling_face: ? It would not have to be repeated in 2 places, and then name suggests that it's temporary, so the `NOTE` comment (which I appreciate nonetheless) may be redundant :slightly_smiling_face:

widgets/filters/src/FiltersTable.tsx
```diff
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 
-import type { Filter, Widget } from './types';
+import { FunctionComponent } from 'react';
```
Gelio, line 3: Optional:

Could we do

```suggestion
import type { FunctionComponent } from 'react';
```

?

To prevent situations like these, where we do not add `import type` manually, I have found out recently there's a setting in `tsconfig.json` we can enable that will mark these as errors. I will add a ticket in the backlog to enable that in `cloudify-ui-common`. That will let us avoid some review comments :slightly_smiling_face: 

The ticket: https://cloudifysource.atlassian.net/browse/RD-1862

Gelio (commented):
Some more comments. Looks good, again :slightly_smiling_face:

app/.eslintrc.json
kubama: Done

Gelio (approved):
Looks good from my side :slightly_smiling_face: 

I only have an optional comment left, so I may as well approve :slightly_smiling_face:

widgets/filters/src/FilterActions.ts
```diff
@@ -0,0 +1,10 @@
+import type { Filter } from './types';
+
+export default class FilterActions {
+    // eslint-disable-next-line no-useless-constructor
+    constructor(private toolbox: Stage.Types.Toolbox) {}
```
qooban, line 5: If this is the way to create and assign a class instance property from a constructor parameter, then I think we could do sth with `no-useless-constructor` rule for `.ts*` files. Maybe reconfigure it or turn off, I didn't check details. I'm fine with having a ticket for that.

widgets/filters/src/widget.tsx
qooban: I'm OK with both approaches in this case, but to be consistent with other parts I would stay with destructuring. I don't see real benefits with having one line instead of two.

widgets/filters/src/widget.tsx
qooban: Was there a requirement to show 10 filters at once?

Anyways, probably let it be.

widgets/filters/src/widget.tsx
qooban: I think that sooner or later we'll have to make it more consistent. As it is general issue, I'll create ticket for that.

qooban: Issue created: RD-1898 .

qooban (dismissed):

widgets/filters/src/widget.tsx
kubama: No requirement. Probably I'll tune it further once action column is added and row height is increased.

widgets/filters/src/FilterActions.ts
kubama: There is nothing that can be configured for this rule - I turned it off for ts(x) files.

widgets/filters/src/widget.tsx
kubama: I'm OK with both approaches as well.
Since this comment was marked as optional and there is no consistent opinion I'm leaving it as it is for now.

qooban (approved):

.eslintrc.temp-overrides.json
```diff
@@ -4,5 +4,19 @@
...
+            "rules": {
+                "react/prop-types": "off"
+            }
+        },
+        {
+            "files": ["**/*.ts", "**/*.tsx"],
+            "rules": {
+                "no-useless-constructor": "off"
+            }
+        }
```
Gelio, line 20: Optional:

Instead of having 2 sections, I think we can do away with just one, and have both rules disabled in both `ts` and `tsx` files. That would be a bit shorter :slightly_smiling_face:

Gelio (approved):
Looks good, although I see potential for a small improvement :slightly_smiling_face:

.eslintrc.temp-overrides.json
kubama: Done

qooban (approved):
LGTM 👍🏻 

Gelio (approved):
Looks solid, thanks for addressing all the comments :+1:

PR #1276, opened 2021-03-29
RD-1142 Respect filters in Deployments View

Gelio:
This PR adds handling filters in Deployments View in a very basic way:
* filter rules are fetched every time before fetching the deployments
* there is no caching of filter rules. That will come in RD-1852 or RD-1854 (I'm thinking about using [react-query](https://react-query.tanstack.com/overview) instead of writing the logic for polling and aborting requests myself)

The tests need to create filters to be used for the test, so I added necessary commands that facilitate that too.

## Video

https://user-images.githubusercontent.com/889383/112830684-2927d280-9093-11eb-9f19-f1bdd3a68cc4.mp4

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/357/pipeline

Ran 2021-03-30 15:23 CEST

### DIFF
--- a/app/components/shared/widgets/WidgetsList.tsx
+++ b/app/components/shared/widgets/WidgetsList.tsx
@@ -37,7 +37,7 @@ export default function WidgetsList({ onWidgetUpdated, onWidgetRemoved, isEditMo
 WidgetsList.propTypes = {
     widgets: PropTypes.arrayOf(PropTypes.shape({})),
     onWidgetRemoved: PropTypes.func.isRequired,
-    onWidgetUpdated: PropTypes.func.isRequired,
+    onWidgetUpdated: PropTypes.func,
     isEditMode: PropTypes.bool.isRequired
 };
 
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -3,8 +3,10 @@ import { without } from 'lodash';
 
 describe('Deployments View widget', () => {
     const widgetId = 'deploymentsView';
-    const blueprintName = 'deployments_view_test_blueprint';
-    const deploymentName = 'deployments_view_test_deployment';
+    const specPrefix = 'deployments_view_test_';
+    const blueprintName = `${specPrefix}blueprint`;
+    const deploymentName = `${specPrefix}deployment`;
+    const deploymentNameThatMatchesFilter = `${specPrefix}precious_deployment`;
     const siteName = 'Olsztyn';
     const blueprintUrl =
         'https://github.com/cloudify-community/blueprint-examples/releases/download/latest/simple-hello-world-example.zip';
@@ -40,7 +42,8 @@ describe('Deployments View widget', () => {
 
     before(() => {
         cy.activate()
-            .deleteDeployments(deploymentName, true)
+            // NOTE: will remove all deployments used in this test spec
+            .deleteDeployments(specPrefix, true)
             .deleteSites(siteName)
             .deleteBlueprints(blueprintName, true)
             .uploadBlueprint(blueprintUrl, blueprintName)
@@ -139,6 +142,47 @@ describe('Deployments View widget', () => {
             cy.root().parents('body').find('.modal').contains('button', 'Execute').click();
             cy.wait('@restartDeployment');
             cy.contains('Last Execution');
+        });
+    });
+
+    describe('with filters', () => {
+        const filterId = 'only-precious';
+        before(() => {
+            cy.deleteFilter('deployments', filterId, { ignoreFailure: true })
+                .createFilter({
+                    id: filterId,
+                    rules: [{ type: 'label', key: 'precious', values: ['yes'], operator: 'any_of' }],
+                    resource: 'deployments'
+                })
+                .deleteDeployments(deploymentNameThatMatchesFilter, true)
+                .deployBlueprint(blueprintName, deploymentNameThatMatchesFilter, { webserver_port: 9124 })
+                .setLabels(deploymentNameThatMatchesFilter, [{ precious: 'yes' }]);
+        });
+
+        const getFilterIdInput = () =>
+            cy.contains('Name of the saved filter to apply').parent().get('input[type="text"]');
+
+        it('should take the fiter into account when displaying deployments', () => {
+            useDeploymentsViewWidget({ configurationOverrides: { filterId } });
+
+            cy.log('Show only precious deployments');
+            cy.contains(deploymentNameThatMatchesFilter);
+            cy.contains(deploymentName).should('not.exist');
+
+            cy.log('Show all deployments');
+            cy.editWidgetConfiguration(widgetId, () => {
+                getFilterIdInput().clear();
+            });
+
+            cy.contains(deploymentNameThatMatchesFilter);
+            cy.contains(deploymentName);
+
+            cy.log('Invalid filter id');
+            cy.editWidgetConfiguration(widgetId, () => {
+                getFilterIdInput().type('some-very-gibberish-filter-id');
+            });
+
+            cy.contains(/with ID .* was not found/);
         });
     });
 
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -1,6 +1,5 @@
 import type { RouteHandler } from 'cypress/types/net-stubbing';
 import { without } from 'lodash';
-import { FilterResource } from '../../support/filters';
 
 describe('Deployments View widget', () => {
     const widgetId = 'deploymentsView';
@@ -148,8 +147,8 @@ describe('Deployments View widget', () => {
     describe('with filters', () => {
         const filterId = 'only-precious';
         before(() => {
-            cy.deleteFilter(FilterResource.Deployments, filterId, { ignoreFailure: true })
-                .createFilter(FilterResource.Deployments, filterId, [
+            cy.deleteDeploymentsFilter(filterId, { ignoreFailure: true })
+                .createDeploymentsFilter(filterId, [
                     { type: 'label', key: 'precious', values: ['yes'], operator: 'any_of' }
                 ])
                 .deployBlueprint(blueprintName, deploymentNameThatMatchesFilter, { webserver_port: 9124 })
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -54,7 +54,7 @@ describe('Deployments View widget', () => {
         routeHandler,
         configurationOverrides = {}
     }: { routeHandler?: RouteHandler; configurationOverrides?: Record<string, any> } = {}) => {
-        cy.interceptSp('GET', /^\/deployments/, routeHandler).as('deployments');
+        cy.interceptSp('POST', /^\/searches\/deployments/, routeHandler).as('deployments');
         // NOTE: larger viewport since the widget requires more width to be comfortable to use
         cy.viewport(1600, 900)
             .usePageMock(
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -1,5 +1,6 @@
 import type { RouteHandler } from 'cypress/types/net-stubbing';
 import { without } from 'lodash';
+import { FilterResource } from '../../support/filters';
 
 describe('Deployments View widget', () => {
     const widgetId = 'deploymentsView';
@@ -42,7 +43,6 @@ describe('Deployments View widget', () => {
 
     before(() => {
         cy.activate()
-            // NOTE: will remove all deployments used in this test spec
             .deleteDeployments(specPrefix, true)
             .deleteSites(siteName)
             .deleteBlueprints(blueprintName, true)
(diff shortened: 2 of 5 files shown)


app/components/shared/widgets/WidgetsList.tsx
```diff
@@ -37,7 +37,7 @@ export default function WidgetsList({ onWidgetUpdated, onWidgetRemoved, isEditMo
 WidgetsList.propTypes = {
     widgets: PropTypes.arrayOf(PropTypes.shape({})),
     onWidgetRemoved: PropTypes.func.isRequired,
-    onWidgetUpdated: PropTypes.func.isRequired,
+    onWidgetUpdated: PropTypes.func,
```
Gelio, line 40: This caused console warnings. `onWidgetUpdated` is optional in TypeScript

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -3,8 +3,10 @@ import { without } from 'lodash';
 
 describe('Deployments View widget', () => {
     const widgetId = 'deploymentsView';
-    const blueprintName = 'deployments_view_test_blueprint';
-    const deploymentName = 'deployments_view_test_deployment';
+    const specPrefix = 'deployments_view_test_';
+    const blueprintName = `${specPrefix}blueprint`;
+    const deploymentName = `${specPrefix}deployment`;
+    const deploymentNameThatMatchesFilter = `${specPrefix}precious_deployment`;
```
Gelio, line 9: Initially I wanted to take `deploymentName` and append something to it

```ts
const deploymentNameThatMatchesFilter = `${deploymentName}_something`;
```

but then:

```ts
cy.contains(deploymentName).should('not.exist');
```

will fail when `deploymentNameThatMatchesFilter` is visible, because it contains `deploymentName`. Thus, the names cannot be prefixes of each other

.size-limit
```diff
@@ -13,7 +13,7 @@
     },
     {
         "name": "Widgets",
-        "limit": "2.5 MB",
+        "limit": "2.51 MB",
```
Gelio, line 16: There is a problem with transpiling `async/await` which includes a bunch of code into widgets on use :smile: I've reported it to fix later: https://cloudifysource.atlassian.net/browse/RD-1894

IMO that should not be a blocker for the PR, though, since we are already using async/await in other widgets

test/cypress/support/filters.ts
```diff
@@ -0,0 +1,48 @@
...
+    operator: string;
+    type: 'label' | 'attribute';
+}
+
+export interface Filter {
+    id: string;
+    resource: 'deployments' | 'blueprints';
+    rules: FilterRule[];
+    visibility?: string;
+}
```
kubama, line 26: I find this type to be quite artificial. The only place it is used is `createFilter` argument, but inside this command it's broken down and not used as a whole. Why not just use 4 (or even 3 if you considered my other comment) arguments instead of creating a type?

test/cypress/support/filters.ts
```diff
@@ -0,0 +1,48 @@
...
+            visibility: filter.visibility
+        };
+
+        return cy.cfyRequest(`/filters/${filter.resource}/${filter.id}`, 'PUT', null, payload);
+    },
+
+    deleteFilter: (
+        resource: Filter['resource'],
+        filterId: string,
+        { ignoreFailure }: { ignoreFailure?: boolean } = {}
```
kubama, line 41: Why object if only one property is used?

test/cypress/support/filters.ts
```diff
@@ -0,0 +1,48 @@
...
+    createFilter: (filter: Filter) => {
+        const payload = {
+            filter_rules: filter.rules,
+            visibility: filter.visibility
+        };
+
+        return cy.cfyRequest(`/filters/${filter.resource}/${filter.id}`, 'PUT', null, payload);
+    },
+
+    deleteFilter: (
```
kubama, line 38: How about `createDeploymentsFilter`, `createBlueprintsFilter`, `deleteDeployemntsFilter`, ... ?

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -142,6 +145,47 @@ describe('Deployments View widget', () => {
...
+                })
+                .deleteDeployments(deploymentNameThatMatchesFilter, true)
+                .deployBlueprint(blueprintName, deploymentNameThatMatchesFilter, { webserver_port: 9124 })
+                .setLabels(deploymentNameThatMatchesFilter, [{ precious: 'yes' }]);
+        });
+
+        const getFilterIdInput = () =>
+            cy.contains('Name of the saved filter to apply').parent().get('input[type="text"]');
+
+        it('should take the fiter into account when displaying deployments', () => {
```
kubama, line 165:
```suggestion
        it('should take the filter into account when displaying deployments', () => {
```

test/cypress/support/filters.ts
Gelio: I prefer an options object for readability purposes. Consider seeing a call:

```ts
cy.createFilter({
  id: 'a',
  resource: 'deployments',
  rules: [/* ... */]
});
```

vs

```ts
cy.createFilter('a', 'deployments', [/* ... */]);
```

I find the former to be more readable

test/cypress/support/filters.ts
Gelio: When viewing a function being called, seeing `true` being passed in tells you very little as to what it changes. Having an options object aids that

test/cypress/support/filters.ts
Gelio: I don't see much of a benefit and having more methods introduces more code :slightly_smiling_face:

kubama: Let me answer the comment above and https://github.com/cloudify-cosmo/cloudify-stage/pull/1276/files#r603274205 at the same time.

Looking into it further I can see that `visibility` property is not used. So, if it was removed you could have API as simple as `createDeploymentFilter(id, rules)` - simple and intuitive without the options object pattern (BTW blueprint filters are not used in this PR, so actually no reason for adding support for them at this point).
But what's more important, `resource` is by no means a filter attribute - it is an actual type of the filter. In other words, deployments filter is a different entity type than blueprints filter. The sets of rule types they may use are not same. At some point it may be reflected in the TS type for the rules. Creating an enum with two possible values and using it as function parameter does not address this distinction.
Lastly, `id` is a 'primary' attribute - having it as the first create function parameter is very intuitive and almost self descriptive.

test/cypress/support/filters.ts
kubama: > When viewing a function being called, seeing true being passed in tells you very little as to what it changes. Having an options object aids that

Using an IDE aids that as well 🙂.

I may be missing something, but I don't see a clear rule by which you are deciding to use or not use options object pattern. Here, it is used just for the third argument, but not for first or second, while the justification that you gave could be applied to them as well. It's also not in line with `createFilter` signature, where all arguments are passed as an object, and they include `resource` and `filterId` properties as well.
Additionally, when extending any API (in this case a set of custom cypress commands) it's good to stay in line with existing patterns. There are existing commands that use `boolean` parameters passed directly, including those setting `failOnStatusCode` option, so using a different pattern here is unexpected.

test/cypress/support/filters.ts
```diff
@@ -0,0 +1,48 @@
...
+declare global {
+    namespace Cypress {
+        // NOTE: necessary for extending the Cypress API
+        // eslint-disable-next-line @typescript-eslint/no-empty-interface
+        export interface Chainable extends GetCypressChainableFromCommands<typeof commands> {}
+    }
+}
+
+/**
+ * @see https://cloudifysource.atlassian.net/wiki/spaces/WR/pages/1657405459/Complex+Filters+Design+Document
```
qooban, line 12: I'm not sure it will not change in the future, better would be to stick to the official user documentation, but I assume it is not ready yet, so probably - let it be for the time being.

qooban (commented):
Commented.

test/cypress/support/filters.ts
Gelio: > Using an IDE aids that as well .

Yeah, even when using an IDE, you have to hover over the function to see the signature, most of the time, or use WebStorm that displays parameter names inline. Still. when reviewing code, you don't get that hints, and IMO that's not an argument against :slightly_smiling_face: 

---

Some resource about the cons of having boolean parameters:

https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

There is very little we can do to remove the parameter altogether since it is defined by Cypress (I suppose we could just ensure it is set to `true` all the time, but that could be limiting), so some of the arguments in the article do not apply here. However, the [Use a named parameter](https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/#2-the-cheap-way-use-a-named-parameter) chapter is exactly this pattern :slightly_smiling_face: 

An interesting argument is also one related to extensibility of such API. What if we need another flag? Do we add another boolean parameter? What if you want to set only the other parameter and not this one? You would pass `undefined, true` in the parameter list, probably, or refactor existing uses of `deleteFilter` to an options argument then :smile: Then you could get into the `ajaxCall` territory:

```ts
    ajaxCall(url, method, params, data, parseResponse = true, userHeaders = {}, fileName = null, withCredentials) {
```

that is called e.g.: 

```ts
        return this.ajaxCall(url, 'get', null, null, null, null, fileName);
```

which looks very fun to read :slightly_smiling_face: 



> so using a different pattern here is unexpected.

Not sure what's unexpected here. I believe you can clearly see what is being defined, and what the parameter means :slightly_smiling_face: It is a different pattern indeed, but one that improves the quality IMO. If the rule about always following the same pattern everywhere for the sake of consistency would apply, how are we supposed to improve the codebase?

---

Having said all of the above, I do not plan on changing the code here so far :slightly_smiling_face: Looks like we have different opinions, so having a third look and opinion here would help :slightly_smiling_face: @qooban, what's your take on the matter? Would you prefer a boolean parameter

```ts
cy.deleteFilter('deployments', 'myfilter', true);
```

or an options object:

```ts
cy.deleteFilter('deployments', 'myfilter', {
  ignoreFailure: true
});
```
?

test/cypress/support/filters.ts
Gelio: I'll remove the interface and make the parameters formal :slightly_smiling_face: Thanks for your voices

test/cypress/support/filters.ts
Gelio: You convinced me in the other argument to make the parameters formal, so I'll do that.

I intend on not splitting the methods, though, as very likely you will be needing to create/delete blueprint filters, and IMO having 4 methods (and potentially more, basically `2 * the number of resources that can be filters`) is unnecessary maintenance if they only differ by hardcoding a single parameter :slightly_smiling_face:

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -40,7 +42,8 @@ describe('Deployments View widget', () => {
 
     before(() => {
         cy.activate()
-            .deleteDeployments(deploymentName, true)
+            // NOTE: will remove all deployments used in this test spec
```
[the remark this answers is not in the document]
Gelio: In the few test specs that I read, I did not notice that, but I suppose I have missed the ones that did :slightly_smiling_face: I'll remove the comment

test/cypress/support/filters.ts
Gelio: Oh, right! Thanks for the tip :slightly_smiling_face: I have noticed that we already have https://docs.cloudify.co/api/v3.1/#the-filter-resource

so I'll use that as a link

test/cypress/support/filters.ts
Gelio: Since you're both in favor of deobjectifying the parameters, let me do that :slightly_smiling_face:

Gelio:
The system tests failed, but I'm guessing that is related to the problem with SSL that we already have on `master`. They pass when I run them locally:

image

I will not rerun them until we fix the SSL issue. If the issue is not solved by the time the PR is ready to merge, I'll merge it without rerunning the tests, since they will likely fail anyway and will not provide much value :slightly_smiling_face:

test/cypress/support/filters.ts
Gelio: Since this discussion was about removing the interface and using formal parameters, and I have done that, I'll resolve this one to keep only the active ones unresolved :slightly_smiling_face:

qooban (dismissed):
LGTM 👍🏻

qooban:

test/cypress/support/filters.ts
kubama: Actually, as this PR does not make use of blueprint filters, we are talking about just 2 functions. I see no reason for adding support for blueprint filters in this PR - correct me if I'm wrong, but unless I'm not aware of something blueprint filters support is not in scope of any epics we are working on at the moment. So, after removing the options object my proposal is to drop the first parameter (resource type) and rename existing functions.

test/cypress/support/filters.ts
kubama: > If the rule about always following the same pattern everywhere for the sake of consistency would apply, how are we supposed to improve the codebase?

There are two distinct cases. In terms of implementation - sure, no need to stick to conventions if the direction of changes is good. But in terms of APIs and interfaces it's different - using mixed approaches makes the API harder to use (here by the API I mean APIs of all our custom cypress commands). I understood your points and I'm convinced, but in this case I suggest creating a tech debt ticket for refactoring existing cypress commands to use the same pattern.

test/cypress/support/filters.ts
Gelio: Alright, I'll refactor it for the sake of simplicity :+1: 

I'm afraid that in the future, there will be many specialized methods that all look the same :smile: If we ever do blueprint filtering, I will advocate for going back to this implementation rather than adding more methods

test/cypress/support/filters.ts
Gelio: Here's the ticket: https://cloudifysource.atlassian.net/browse/RD-1916 :slightly_smiling_face:

test/cypress/support/filters.ts
Gelio: Addressed :slightly_smiling_face:

Gelio:
I've pushed 1 more commit to address https://github.com/cloudify-cosmo/cloudify-stage/pull/1276#discussion_r603267773

qooban (dismissed):

kubama (approved):

Gelio:
Thanks for reviews :slightly_smiling_face: Now that the system tests should be stable, here's a run with the latest changes:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/355/pipeline

I'll wait for the results before merging the PR :slightly_smiling_face:

Gelio:
I've rebased the PR again, since there were merge conflicts. 

I thought that clicking "Replay" on a previous pipeline would take the latest commit on the branch. Looks like it just retries that past commit. Thus, the previous system tests failed.

Here's a new run:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/357/pipeline

kubama (approved):

Gelio:
The tests passed. Thus, I'm merging the PR :slightly_smiling_face:

PR #1278, opened 2021-03-30
WDR-2 Added word break for label tag

kubama:


qooban (approved):
I assume it works good. 
For the future a screenshot or two would be great to have, that will decrease reviewers time spent on the PR.

qooban:

PR #1280, opened 2021-03-30
RD-1123 Added actions column to filters widget

kubama:


widgets/filters/src/FiltersTable.tsx
```diff
@@ -16,30 +17,69 @@ interface FiltersTableProps {
...
+            <Confirm
+                open={!!filterIdToDelete}
+                onCancel={clearFilterIdToDelete}
+                onConfirm={() => {
+                    clearFilterIdToDelete();
+                    toolbox.loading(true);
+                    new FilterActions(toolbox).doDelete(filterIdToDelete).then(toolbox.refresh);
+                }}
+                content={i18n.t('widgets.filters.deleteConfirm', { filterId: filterIdToDelete })}
+            />
```
qooban, line 82: Just wondering - this will be the only place to remove the filter?
If so, then probably we can leave it here. If not, it would probably be better to extract it to separate component.

widgets/filters/src/FiltersTable.tsx
```diff
@@ -16,30 +17,69 @@ interface FiltersTableProps {
...
                 ))}
             </DataTable>
+
+            <Confirm
+                open={!!filterIdToDelete}
+                onCancel={clearFilterIdToDelete}
+                onConfirm={() => {
+                    clearFilterIdToDelete();
+                    toolbox.loading(true);
+                    new FilterActions(toolbox).doDelete(filterIdToDelete).then(toolbox.refresh);
```
qooban, line 79: `toolbox.refresh` "turns off" `toolbox.loading`?

widgets/filters/widget.css
```diff
@@ -0,0 +1,3 @@
+.filtersWidget .filtersTable {
+    table-layout: fixed;
+}
```
qooban, line 3: That's because it is not possible to assign `style` to `DataTable`? Maybe we can use `Stage.styled`?

qooban (commented):
Looks good. Commented. Thanks for having it small 👍🏻

widgets/filters/src/FiltersTable.tsx
```diff
@@ -16,30 +17,69 @@ interface FiltersTableProps {
...
+                                bordered
+                                title={i18n.t('widgets.filters.columns.actions.clone')}
+                            />
+                            <Icon
+                                name="trash"
+                                link
+                                bordered
+                                title={i18n.t('widgets.filters.columns.actions.delete')}
+                                onClick={() => setFilterIdToDelete(filter.id)}
+                            />
```
Gelio, line 70: What happens if
1. The user clicks "Delete filter"
2. The user confirms deletion
3. The user clicks "Delete filter" again, before the `doDelete` completes

I think this way we may get ourselves into some corrupted state and race conditions :slightly_smiling_face:  I would suggest disabling the delete button while the request is in progress (potentially with a spinner next to the button to show that it's this action that is in progress)

Alternatively, we may keep the confirm modal open until the delete request finishes :slightly_smiling_face:

widgets/filters/src/FiltersTable.tsx
```diff
@@ -14,32 +15,74 @@ interface FiltersTableProps {
     widget: FilterWidget;
 }
 
+const FixedLayoutDataTable = Stage.styled(Stage.Basic.DataTable)`
```
Gelio, line 18: Nice idea :+1: I did not know that would be possible

For those also wondering: https://styled-components.com/docs/basics#styling-any-component
It should work as long as the component accepts `className` :smile:

widgets/filters/src/FiltersTable.tsx
kubama: Yes

widgets/filters/src/FiltersTable.tsx
kubama: I'm not aware of any other place, so for now I'd leave it as it is.

widgets/filters/src/FiltersTable.tsx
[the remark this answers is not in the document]
kubama: Widgets have a basic error handling in place. If delete request fails an error message will be displayed inside the widget. I think it's just enough - there are no scenarios where any particular errors are expected and require dedicated handling.

widgets/filters/src/FiltersTable.tsx
kubama: Please see my related comment: https://github.com/cloudify-cosmo/cloudify-stage/pull/1280#discussion_r604772423.

Out of curiosity I tried to reproduce this scenario and I succeeded, but I had to be extremely fast 🙂. Unless there is a serious network latency or manager connection issue (in which case the whole UI would be unusable anyway) there is no way user would do this without trying hard intentionally.

I also checked what's the approach in existing widgets (except labels, which I was implementing recently, so it would bias my view). Looks like the common approach is to keep the confirm modal open until delete is completed, but without any progress indication (no spinner or disabled button). I consider my approach better, as at least it provides progress notification (`toolbox.loading`) and doesn't unnecessarily block the whole UI.

In the end, I don't think this corner case is worth addressing. Other widgets are prone to the same scenario as well, and we never had any issue with it.

widgets/filters/src/FiltersTable.tsx
qooban: OK 👍🏻

qooban (approved):
LGTM 👍🏻

widgets/filters/src/FiltersTable.tsx
Gelio: Ok, then addressing it is optional :slightly_smiling_face: Personally I would probably take some action here, but I can see how it's not required and I do not expect any changes :slightly_smiling_face: Thanks for the response and verification!

widgets/filters/src/FiltersTable.tsx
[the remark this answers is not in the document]
Gelio: Hmm... To my knowledge, there will not be any error displayed to the user, since there is no `catch` here or anywhere in the functions called in this line that would catch the error and show it. I may be mistaken, so let me know if I am :slightly_smiling_face: 

The default widget error handling does take place but for requests done in `fetchData`:
https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/actions/WidgetData.js#L45-L67

So here we risk that we close the modal, keep `toolbox.loading`, and don't show the user any message when an error occurs.

Other widgets probably don't implement error handling for mutations, but I encourage putting more effort and solving possible problems :slightly_smiling_face: 

Any actions on this comment are optional, since it's not the norm to handle mutation errors, so feel free to resolve the discussion if you're not keen on addressing it :slightly_smiling_face:

Gelio (approved):
Looks good to me too :slightly_smiling_face: Other actions on this PR are optional, yet still recommended

widgets/filters/src/FiltersTable.tsx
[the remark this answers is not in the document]
kubama: > The default widget error handling does take place but for requests done in `fetchData`

Yes, you are right, my bad.

Since this is a common issue I think it would be perfect if the widget framework itself provided a way to handle such errors. We already have `toolbox.loading` so why not `toolbox.error`? For simplicity it could use the same error message component as `fetchData` error handling. It should be relatively easy to implement - I can try adding it in the next PR if you are OK with the idea.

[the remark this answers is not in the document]
Gelio: Yeah, that sounds reasonable to me :slightly_smiling_face: Although, will it not mean that after a single failure, the user won't be able to interact with the widget anymore, as it will go into the "has error" path, and will keep showing the error message?

Errors such as these IMO could be either in the modal, or as a toast notification :slightly_smiling_face: 

This is a complex enough topic that I suggest discussing that in the team during some weekly :slightly_smiling_face:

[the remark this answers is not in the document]
kubama: I've added topic to the agenda.

PR #1283, opened 2021-03-31
RD-1846 - added permissions checking for who getting started modal is visible

gregcfy:
In this PR some condition that checks if the current user role has permissions to display getting started modal was added.
Task link: https://cloudifysource.atlassian.net/browse/RD-1846

This PR is related with: 
https://github.com/cloudify-cosmo/cloudify-manager-install/pull/1107


kubama (dismissed):

app/components/GettingStartedModal/GettingStartedModal.tsx
```diff
@@ -27,6 +30,7 @@ const castedGettingStartedSchema = gettingStartedSchema as GettingStartedSchema;
 const GettingStartedModal = () => {
     const [modalOpen, setModalOpen] = useState(() => isGettingStartedModalDisabledInLocalStorage());
```
Gelio, line 31: Optional, fun code golf:

```suggestion
    const [modalOpen, setModalOpen] = useState(isGettingStartedModalDisabledInLocalStorage);
```

Gelio, line 31: Digression:

This is how I read this line: initialize `modalOpen` to whether the modal is disabled in `localStorage`. Thus, it reads to me that when it is disabled in `localStorage`, `modalOpen = true`.

Could we change the naming to `modalClosed` or rename the function to `isGettingStartedModalEnabled`? This could be a separate ticket

Gelio (dismissed):
Looks good, although I have some non-blocking comments about the surrounding code :slightly_smiling_face:

app/components/GettingStartedModal/GettingStartedModal.tsx
```diff
@@ -1,7 +1,10 @@
 import React, { memo, useState, useMemo } from 'react';
 import i18n from 'i18next';
 import log from 'loglevel';
+import { useSelector } from 'react-redux';
 
+import { ReduxState } from '../../reducers';
```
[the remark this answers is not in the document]
gregcfy: updated

app/components/GettingStartedModal/GettingStartedModal.tsx
gregcfy: yeah, good point!

`localStorage['key']` returns `undefined` and `localStorage.getItem('key')` returns null for lack of key

fixed

gregcfy: ok, we can change it in separated PR

app/components/GettingStartedModal/localStorage.ts
```diff
@@ -1,7 +1,5 @@
-export const isGettingStartedModalDisabledInLocalStorage = (): boolean => {
-    const modalDisabled = localStorage.getItem('getting-started-modal-disabled');
-    return modalDisabled === undefined || modalDisabled !== 'true';
-};
+export const isGettingStartedModalDisabledInLocalStorage = (): boolean =>
+    localStorage.getItem('getting-started-modal-disabled') === 'true';
 
 export const disableGettingStartedModalInLocalStorage = (): void => {
     localStorage.setItem('getting-started-modal-disabled', 'true');
```
kubama, line 5: I'd suggest keeping function names shorter - given the context in which they are defined (file name) I think `InLocalStorage` suffix is not needed and actually redundant.

```suggestion
export const isGettingStartedModalDisabled = (): boolean =>
    localStorage.getItem('getting-started-modal-disabled') === 'true';

export const disableGettingStartedModal = (): void => {
    localStorage.setItem('getting-started-modal-disabled', 'true');
```

gregcfy: fixed

kubama (dismissed):

app/components/GettingStartedModal/localStorage.ts
```diff
@@ -1,8 +1,6 @@
-export const isGettingStartedModalDisabledInLocalStorage = (): boolean => {
-    const modalDisabled = localStorage.getItem('getting-started-modal-disabled');
-    return modalDisabled === undefined || modalDisabled !== 'true';
-};
+export const isGettingStartedModalDisabled = (): boolean =>
+    localStorage.getItem('getting-started-modal-disabled') === 'true';
```
Gelio, line 2: Optional:

Extracting the key and reusing it in both function would ensure there are no typos :slightly_smiling_face:

Gelio (dismissed):
Looks good, although there are merge conflicts :slightly_smiling_face:

app/components/GettingStartedModal/localStorage.ts
kubama: Actually, since the value is also a string (`'true'`) it could be extracted and reused as well, as it is prone to the same 'typo issue'.

Gelio: Yeah :smile: We could alternatively use a different value, not just `true`, as seeing a boolean used as a string could be misleading :slightly_smiling_face: To me, any truthy value would work, so we could do

```suggestion
    !!localStorage.getItem('getting-started-modal-disabled');
```

kubama: It is good, as 'yes' or '1' would work as well, for example if the value was set manually or in a different place, like a test.
It is bad, as 'no', 'false' or '0' would be considered a truthy value.
I think this is actually a common issue wherever string must be used to describe boolean value. Maybe there is something in npm that could solve it by recognizing common string values that can be used to express true or false?

gregcfy: String(true) could be some solution too

Gelio: Yeah :thinking: or using `JSON.stringify` on the value. Or, we could avoid a boolean value and use an enum https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/#3-if-it-makes-sense-use-an-enum

```ts
enum GettingStartedModalState {
  Disabled = 'disabled',
  Enabled = 'enabled',
}
```

(we could use `dismissed` instead of `disabled`), and use other key names, e.g. `getting-started-modal-dismissed-state`.

That's actually one of the ways to avoid booleans that is slightly more descriptive. 

I don't have a preference here, just putting it out there :slightly_smiling_face:

kubama: @gregcfy So, what's your take on this? You re-requested the review but this thread is still open. The comments here are optional, but it would be nice to know if you are planning any actions or not 🙂.

app/components/GettingStartedModal/localStorage.ts
```diff
@@ -1,8 +1,8 @@
-export const isGettingStartedModalDisabledInLocalStorage = (): boolean => {
+export const isGettingStartedModalDisabled = (): boolean => {
     // NOTE: quickfix, disable getting started modal to get the tests to pass on master
     return false;
```
Gelio, line 3: This should be changed to `true`, now that we are using the negation in `GettingStartedModal.tsx` to initialize the `modalOpen` state variable :slightly_smiling_face: Either we change it here, or we fix the tests to pass with the modal open :slightly_smiling_face:

Gelio (changes requested):
Having it merged like that will cause the CI to fail again :slightly_smiling_face: This needs fixing :smile: 

btw. don't forget to run the system tests for this branch prior to merging :slightly_smiling_face:

app/components/GettingStartedModal/localStorage.ts
gregcfy: I have decided to use JSON.stringify(true)

app/components/GettingStartedModal/localStorage.ts
gregcfy: changed to `true`

Gelio (approved):

PR #1284, opened 2021-03-31
RD-1758 / RD-1759 - Filter definition form - basic implementation

qooban:
### Changes
This is basic skeleton implementation for `FilterDefinitionForm` component.
Basic functionality (adding/removing) filter rules rows added.

### Components structure
image

### Video
https://user-images.githubusercontent.com/5202105/113149846-bdc93680-9233-11eb-94de-e067d88211b6.mp4

Recorded artificially (added `FilterDefinitionForm` to the Button Link widget).


webpack.config.js
```diff
@@ -19,6 +19,9 @@ module.exports = (env, argv) => {
     const context = path.join(__dirname);
     const devtool = isProduction ? undefined : 'eval-source-map';
     const outputPath = path.join(__dirname, 'dist');
+    const extensions = ['js', 'jsx', 'ts', 'tsx'];
+    const resolveExtensions = extensions.map(extension => `.${extension}`);
+    const globExtensions = `{${extensions.map(extension => `${extension}`).join(',')}}`;
```
qooban, line 24: This is not related to the PR, but I took a chance to do it at once to extract and align extension settings in webpack confg.

I hope you all will be good with it. If no, I can extract it to a separate PR.

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,34 @@
+import { useState } from 'react';
+import type { FunctionComponent } from 'react';
+
+import Form from './FormWrapper';
+import RuleRow from './RuleRow';
+import AddRuleButton from './AddRuleButton';
+
+const defaultRow = { attribute: '', operator: '', input: '' };
```
qooban, line 8: I'm not sure about the structure, yet, so don't consider it as final row definition.

widgets/common/src/filters/FormWrapper.tsx
```diff
@@ -0,0 +1,8 @@
+import { FunctionComponent } from 'react';
+
+const { Form } = Stage.Basic;
+
+// TODO: This wrapper should be unified with unsafelyTypedForm.tsx and probably added to `app/components/basic`
+export default (Form as unknown) as FunctionComponent<{ [x: string]: any }>;
+export const FormField = (Form.Field as unknown) as FunctionComponent<{ [x: string]: any }>;
+export const FormGroup = (Form.Group as unknown) as FunctionComponent<{ [x: string]: any }>;
```
qooban, line 8: I'd like our TypeScripters (@gregcfy , @Gelio ) to provide advice, here 🙂

This is a duplication of what @gregcfy introduced in `app/components/GetttingStartedModal/unsafelyTypedForm.tsx`. As I can't use it from here (`widgets/common`), I temporary created `FormWrapper`, but I want to remove it within this PR. I think we should move it somehow to `app/components/basic` or I can try implementing RD-1837 (that would probably be good excercise for me, but I'm not sure I will have enough time to do that). Please let me know what do you think, what would be the best way to have access to `Form` component (and its subcomponents, like `Form.Field`, `Form.Group`)?

kubama (dismissed):
Looks like a solid and clean starting point 👍.

I noticed there is delete icon present even if there is only single row - is this intentional? I think it makes sense to always have at least one row.

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const FiltersDefinitionForm: FunctionComponent = () => {
+    const [rows, setRows] = useState([defaultRow]);
+
+    function addRule() {
+        const newRows = [...rows, defaultRow];
```
gregcfy, line 14: const newRows = [...rows, {virtualKey: <unique_generated_key_here>, ...defaultRow}];

Gelio:
I'll take a look at the PR tomorrow :slightly_smiling_face: Meaning, happy time-off :smile:

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+        setRows(newRows);
+    }
+
+    function removeRule(id: number) {
```
gregcfy, line 18: Digression: we do not use `this` here but it is good to use arrow functions to prevent context change.

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+import Form from './FormWrapper';
+import RuleRow from './RuleRow';
+import AddRuleButton from './AddRuleButton';
+
+const defaultRow = { attribute: '', operator: '', input: '' };
+
+const FiltersDefinitionForm: FunctionComponent = () => {
+    const [rows, setRows] = useState([defaultRow]);
+
+    function addRule() {
```
gregcfy, line 13: Digression: we do not use this here but it is good to use arrow functions to prevent context change.

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+        const newRows = _.filter(rows, (_rowObject, rowId) => rowId !== id);
+        setRows(newRows);
+    }
+
+    return (
+        <Form>
+            {rows.map((_row, index) => (
+                // TODO: Change key not to use index
+                // eslint-disable-next-line react/no-array-index-key
+                <RuleRow key={index} id={index} onRemove={removeRule} />
```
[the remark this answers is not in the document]
gregcfy: <RuleRow key={_row.virtualKey} id={index} onRemove={removeRule} />

.eslintrc.temp-overrides.json
```diff
@@ -7,7 +7,7 @@
     },
     "overrides": [
         {
-            "files": ["**/*.ts", "**/*.tsx"],
```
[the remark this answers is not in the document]
Gelio: Interesting... Did you find out why the previous paths were not working? They look fine :thinking:

widgets/common/src/filters/AddRuleButton.tsx
```diff
@@ -0,0 +1,15 @@
+import type { FunctionComponent } from 'react';
+import { i18nPrefix } from './consts';
+
+interface AddRuleButtonProps {
+    onClick: () => void;
```
Gelio, line 5: Optional:

One thing we can improve here is reuse the `Button`'s `onClick` type. This way add type information about the parameters that will be accessible in JavaScript (because we do not cut out the parameters that are passed to `onClick` by the button), which aligns TS and JS :slightly_smiling_face: 

```suggestion
    onClick: ComponentProps<typeof Stage.Basic.Button>['onClick'];
```

`ComponentProps` comes from `react`

widgets/common/src/filters/RuleRemoveButton.tsx
```diff
@@ -0,0 +1,15 @@
+import type { FunctionComponent } from 'react';
+import { i18nPrefix } from './consts';
+
+interface RuleRemoveButtonProps {
+    onClick: () => void;
```
Gelio, line 5: Optional:

The same idea of reusing props applies here as in `AddRuleButton` :slightly_smiling_face:

widgets/common/src/filters/RuleRemoveButton.tsx
```diff
@@ -0,0 +1,15 @@
...
+}
+
+const RuleRemoveButton: FunctionComponent<RuleRemoveButtonProps> = ({ onClick }) => {
+    const {
+        Basic: { Button },
+        i18n
+    } = Stage;
+    return <Button aria-label={i18n.t(`${i18nPrefix}.buttons.removeRule`)} basic icon="trash" onClick={onClick} />;
```
Gelio, line 13: Nice use of the aria-label :+1: 

Optional:

The aria-label is only accessible for screen readers. How about showing a popup on hover that will explain what the trash icon button will do :slightly_smiling_face:? I believe that could provide more info to the user, and IMO that's appreciated

widgets/common/src/filters/RuleRow.tsx
```diff
@@ -0,0 +1,31 @@
+import type { FunctionComponent } from 'react';
+import RuleAttributeDropdown from './RuleAttributeDropdown';
+import RuleOperatorDropdown from './RuleOperatorDropdown';
+import RuleInput from './RuleInput';
+import RuleRemoveButton from './RuleRemoveButton';
+import { FormField, FormGroup } from './FormWrapper';
+
+interface RuleRowProps {
+    id: number;
+    onRemove: (id: number) => void;
```
Gelio, line 10: The placement of the parameter looks odd to me :smile: Will this component ever need to call it with something else than the `props.id`? I doubt it, but still, by having the API designed this way, we make it possible :slightly_smiling_face: 

What we could do is not pass the `id` here and force the parent to handle it. This way the parent can be user that the `id` will always be the `id` that is passed in the props, not something else.

```suggestion
    onRemove: () => void;
```

(or reuse the `onClick` type from `RuleRemoveButton` (see comment on `AddRuleButton` to see how to use that using the `ComponentProps` type)

widgets/common/src/filters/RuleInput.tsx
```diff
@@ -0,0 +1,7 @@
+import type { FunctionComponent } from 'react';
+
+const RuleInput: FunctionComponent = () => {
+    const { Input } = Stage.Basic;
+    return <Input type="text" />;
```
Gelio, line 5: Are there any plans to introduce some labels/placeholders/aria-labels for those inputs? At the moment, it looks quite bare :slightly_smiling_face:

widgets/common/src/filters/RuleOperatorDropdown.tsx
```diff
@@ -0,0 +1,20 @@
+import type { FunctionComponent } from 'react';
+
+const operators = [
```
Gelio, line 3: Optional:

How about adding a type annotation for the type of the values in the array, to make sure all array members are of the same shape? This way we are sure there are no typos in the objects :slightly_smiling_face: 

```suggestion
const operators: OperatorDescription[] = [
```

widgets/common/src/filters/RuleOperatorDropdown.tsx
```diff
@@ -0,0 +1,20 @@
...
+    { name: 'any_of', label: 'is one of' },
+    { name: 'not_any_of', label: 'is not one of' },
+    { name: 'is_null', label: 'key is not' },
+    { name: 'is_not_null', label: 'key is' },
+    { name: 'contain', label: 'contains' },
+    { name: 'not_contain', label: 'does not contain' },
+    { name: 'start_with', label: 'starts with' },
+    { name: 'end_with', label: 'ends with' }
```
Gelio, line 11: According to https://cloudifysource.atlassian.net/wiki/spaces/WR/pages/1657405459/Complex+Filters+Design+Document#Data-Models, different operators are allowed for labels, and different for attributes. Do you plan to split it into 2 arrays and only show the relevant operators based on the type of filter :slightly_smiling_face: ?

widgets/common/src/filters/RuleOperatorDropdown.tsx
```diff
@@ -0,0 +1,20 @@
...
+];
+
+const RuleOperatorDropdown: FunctionComponent = () => {
+    const { Dropdown } = Stage.Basic;
+    const operatorsOptions = _.map(operators, operator => ({ text: operator.label, value: operator.name }));
```
Gelio, line 16: How about avoiding lodash :slightly_smiling_face: ?

```suggestion
    const operatorsOptions = operators.map(operator => ({ text: operator.label, value: operator.name }));
```

Also, let's make sure that the objects returned match what the `Dropdown` expects as `options`:

```suggestion
    const operatorsOptions = operators.map((operator): DropdownItemProps => ({ text: operator.label, value: operator.name }));
```

btw. you can perhaps see now why I skipping parentheses around arrow function parameters is not helpful when using TS - you need to add them each time you want to either add a parameter, add type annotation for a parameter, or add a type annotation for the return type :slightly_smiling_face: Generally arrow function parameter are always enforced in TS projects, but I don't see the need to add them in this one, as that would require lots of changes :slightly_smiling_face:

widgets/common/src/filters/RuleOperatorDropdown.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+    return <Dropdown search selection name="ruleOperator" options={operatorsOptions} />;
```
Gelio, line 18: How will we target this dropdown in tests? Perhaps an aria-label would be helpful here?

widgets/common/src/filters/RuleAttributeDropdown.tsx
```diff
@@ -0,0 +1,11 @@
+import type { FunctionComponent } from 'react';
+
+const attributes = ['Label', 'Blueprint', 'Site name', 'Creator'];
+
+const RuleAttributeDropdown: FunctionComponent = () => {
+    const { Dropdown } = Stage.Basic;
+    const attributeOptions = _.map(attributes, attribute => ({ text: attribute, value: attribute }));
```
Gelio, line 7: The same comment applies here as in `RuleOperatorDropdown` :slightly_smiling_face:

widgets/common/src/filters/RuleAttributeDropdown.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+    return <Dropdown search selection name="ruleAttribute" options={attributeOptions} />;
```
Gelio, line 9: The same comment about selecting the dropdown in tests applies here as in `RuleOperatorDropdown`

widgets/common/src/filters/FormWrapper.tsx
Gelio: IMO that could stay as-is, and have it fixed in RD-1837 :slightly_smiling_face: I don't see the need to implement RD-1837 right away, so I would suggest leaving that for now, and addressing the `Form` component refactor later, when we have time for tech debt.

I would, however, encourage adding the ticket number to the `TODO` comment above, and then, in the description of RD-1837, referencing all places that have those "hacks" to refactor after having proper `Form` typing :slightly_smiling_face:

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+import AddRuleButton from './AddRuleButton';
+
+const defaultRow = { attribute: '', operator: '', input: '' };
+
+const FiltersDefinitionForm: FunctionComponent = () => {
+    const [rows, setRows] = useState([defaultRow]);
+
+    function addRule() {
+        const newRows = [...rows, defaultRow];
+        setRows(newRows);
```
Gelio, line 15: We could use the function form of the updater function, just in case :slightly_smiling_face: 

```suggestion
        setRows((rows) => ([...rows, defaultRow]));
```

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    function removeRule(id: number) {
+        const newRows = _.filter(rows, (_rowObject, rowId) => rowId !== id);
+        setRows(newRows);
```
Gelio, line 20: We could use the function form of the setter function :slightly_smiling_face: Just as I suggested for `addRule`

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,34 @@
...
+const FiltersDefinitionForm: FunctionComponent = () => {
+    const [rows, setRows] = useState([defaultRow]);
+
+    function addRule() {
+        const newRows = [...rows, defaultRow];
+        setRows(newRows);
+    }
+
+    function removeRule(id: number) {
+        const newRows = _.filter(rows, (_rowObject, rowId) => rowId !== id);
```
Gelio, line 19: I encourage using native ES `rows.filter` instead of loadsh :slightly_smiling_face:

widgets/common/src/filters/FiltersDefinitionForm.tsx
Gelio, line 8: Optional:

A type annotation would be nice :slightly_smiling_face: We could probably reuse the one that I introduce in Cypress commands for filters https://github.com/cloudify-cosmo/cloudify-stage/blob/master/test/cypress/support/filters.ts#L12-L17

Gelio: Can we not use the exact structure that the backend expects here? :thinking:

widgets/common/src/filters/FiltersDefinitionForm.tsx
Gelio, line 14: We should probably be cloning the `defaultRow` here. Otherwise, the array will contain multiple items, but each item will be the same object (`defaultRow`) :slightly_smiling_face:

Gelio (commented):
Well done using TypeScript for all files :+1:

.eslintrc.temp-overrides.json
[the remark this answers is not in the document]
kubama: I was also struggling with changes made to `.eslintrc.temp-overrides.json`. In the end it turned out that WebStorm doesn't recognize this file as eslint config file, and any changes made to it (including those made by `git pull`) do not trigger eslint config to be reloaded. Maybe this was the case here?
This is not a major thing once you know it, although still annoying - in the end I didn't check it, but maybe this file could be named in a way allowing it to be recognized by WebStorm?

widgets/common/src/filters/FormWrapper.tsx
qooban: Hmm... I'm still not convinced. I think it could happen before RD-1837 that we will need `Form` in tsx files, and then I don't like the idea of having another file with wrappers.

So I will move these temporary wrappers to `Stage.Basic` to have it in one place and make it possible to access it from the other widgets as well.

widgets/common/src/filters/RuleOperatorDropdown.tsx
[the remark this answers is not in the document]
qooban: Actually, I'll do it now.

BTW I don't consider the structure as final. It may change in the future as the operators list differs by resource. So I'm not aiming to provide final solution for that.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: I originally planned to add it later, but I'll add it within this PR. I'll use UUID.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Hmm... Could you elaborate more on that? Is there any real benefit? We use function definitions in many functional components, so if there is real reason why we should not do that, I'd like to know.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Hmm... Could you elaborate more on that? Is there any real benefit? We use function definitions in many functional components, so if there is real reason why we should not do that, I'd like to know.

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: I didn't think about it, yet. 

I see it is rendered as `<div name="ruleOperator" role="combobox" (...)>`, so I think maybe `div[name=ruleOperator]` would be OK?

widgets/common/src/filters/RuleAttributeDropdown.tsx
qooban: The same answer as in https://github.com/cloudify-cosmo/cloudify-stage/pull/1284#discussion_r607842867 :-)

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: That's probably good idea. I'll try it.

widgets/common/src/filters/FiltersDefinitionForm.tsx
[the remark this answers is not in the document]
qooban: I added proper key setting to `RuleRow`.

widgets/common/src/filters/AddRuleButton.tsx
qooban: Thanks for letting me know. Applied.

widgets/common/src/filters/RuleRemoveButton.tsx
qooban: Thanks for letting me know. Applied.

widgets/common/src/filters/RuleRemoveButton.tsx
qooban: Added `title` instead of popup. I think that is good enough.

widgets/common/src/filters/RuleRow.tsx
qooban: Of course. Removed using `id` and added typings.

widgets/common/src/filters/RuleInput.tsx
qooban: Yes, of course. This is just a placeholder. The component will differ depending on the selected attribute and operator, so I don't plan changes right now.

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: The structure is not final. It can differ between attributes, so I think I will fix it once having proper structure.

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: Yes, as mentioned in https://github.com/cloudify-cosmo/cloudify-stage/pull/1284#discussion_r607858736

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: Thanks for the suggestions. Removed lodash and added type annotation.

widgets/common/src/filters/RuleAttributeDropdown.tsx
qooban: Fixed the same way.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Oh, I wasn't aware of that form, updated 👍🏻

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Applied the same way.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Done 👍🏻

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Right. I think once I have the final structure here, I will add it.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Yep. Changed it to use spread operator for `defaultRow`.

.eslintrc.temp-overrides.json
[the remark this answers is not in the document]
qooban: Strange, but it is working right now. Probably it must have been WebStorm issue, then. I'll revert the change.

qooban:
> I noticed there is delete icon present even if there is only single row - is this intentional? I think it makes sense to always have at least one row.

Good point. I removed showing delete icon when there is only one row.

app/components/GettingStartedModal/unsafelyTypedForm.tsx
```diff
@@ -1,9 +1,4 @@
...
-
-import { Form } from '../basic';
-
-// TODO(RD-1837): remove it after forms will be changed to tsx version
-export const UnsafelyTypedForm = (Form as unknown) as FC<{ [x: string]: any }>;
+import { UnsafelyTypedForm, UnsafelyTypedFormField } from '../basic';
 
 // TODO(RD-1837): remove it after forms will be changed to tsx version
-export const UnsafelyTypedFormField = (Form.Field as unknown) as FC<{ [x: string]: any }>;
+export { UnsafelyTypedForm, UnsafelyTypedFormField };
```
qooban, line 4: I considered removing the file completely and importing `UnsafelyTypedForm*` components directly from `../basic`, but I think this is out of the scope and will introduce much more file changes not related with this PR and IMHO have small benefit, so I left this file being just "import-export proxy" 😄

test/cypress/support/filters.ts
```diff
@@ -1,4 +1,5 @@
 import { addCommands, GetCypressChainableFromCommands } from 'cloudify-ui-common/cypress/support';
+import type { FilterRule } from '../../../widgets/common/src/filters/types';
```
qooban, line 2: If you have better idea how to import this type in Cypress tests, please let me know.

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+import AddRuleButton from './AddRuleButton';
+import type { FilterRule } from './types';
+
+interface FilterRuleRow extends FilterRule {
+    id: string;
+}
+
+function getNewRow() {
+    const { uuid } = Stage.Utils;
+    const defaultRow: FilterRuleRow = { id: '', type: 'label', key: '', operator: '', values: [] };
```
qooban, line 14: This is still just a placeholder. Only `id` is populated. 

I will work on the rest of the fields to get populated in the next PRs.

qooban:
The code is ready for another review round.

There is only one functional change - remove button not present if there is only one row. 
The rest of the changes are internal changes, not customer-facing.

test/cypress/support/filters.ts
kubama: I think this case is something to be considered by @gregcfy when working on a proposal of a common place to store type definitions.

widgets/common/src/filters/FormWrapper.tsx
Gelio: Good call :slightly_smiling_face:

widgets/common/src/filters/RuleInput.tsx
Gelio: Sure thing :+1:

widgets/common/src/filters/RuleOperatorDropdown.tsx
Gelio: Sounds good :+1:

widgets/common/src/filters/RuleOperatorDropdown.tsx
Gelio: Optional:

Hmm... That is probably not so accessible, since the users don't see `name`s. `role=combobox` looks alright, although if there is more than 1, it could be ambiguous :worried: I would opt for an `aria-label`/`title`/`placeholder`/`label`, but I'll leave this remark as optional :slightly_smiling_face:

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -1,31 +1,37 @@
...
+
+    return { ...defaultRow, id: uuid() };
+}
 
 const FiltersDefinitionForm: FunctionComponent = () => {
-    const [rows, setRows] = useState([defaultRow]);
+    const { UnsafelyTypedForm: Form } = Stage.Basic;
+    const [rows, setRows] = useState([getNewRow()]);
```
Gelio, line 21: Optional:

We could pass an initializer function to `useState` to avoid recreating new arrays :slightly_smiling_face: 

```suggestion
    const [rows, setRows] = useState(() => [getNewRow()]);
```

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -1,31 +1,37 @@
...
 
     function addRule() {
-        const newRows = [...rows, defaultRow];
-        setRows(newRows);
+        setRows(() => [...rows, getNewRow()]);
```
Gelio, line 24: Unfortunately, this approach does not differ from just passing in the array immediately, without the function :smile: 

The key feature in the functional updater form is the fact that the function receives the latest value in the function's parameter, thus, it will always use the latest value, as opposed to the `rows` value captured by the closure of `addRule` :slightly_smiling_face: 

```suggestion
        setRows((latestRows) => [...latestRows, getNewRow()]);
```

The same applies to `setRows` in `removeRule` :slightly_smiling_face: 

---

This probably does not matter much now, but IMO that's good practice. One situation when this matters is if React batches handling event listeners, meaning that the user clicks multiple times without the component rerendering (perhaps due to using a low-end device), and `addRule` runs multiple times. `rows` will always be the same array, so running `addRule` 5 times will only extend the array by 1 item :smile: If we use the functional form, 5 items will be added :slightly_smiling_face: 

An article about it (for class components, but it's the same for functional ones): https://medium.com/@wisecobbler/using-a-function-in-setstate-instead-of-an-object-1f5cfd6e55d1

test/cypress/support/filters.ts
Gelio: Maybe we could leverage `Stage.Types`? Perhaps `Stage.Types.API.FilterRule`? This would mean that we need to create another namespace `Stage.Types.API` and make sure that the file that declares it is imported by cypress

IMO we can leave it be for now and wait for @gregcfy's proposal, but I also don't have anything against creating that namespace right now :smile:

widgets/common/src/filters/RuleRow.tsx
```diff
@@ -1,16 +1,16 @@
...
 import RuleAttributeDropdown from './RuleAttributeDropdown';
 import RuleOperatorDropdown from './RuleOperatorDropdown';
 import RuleInput from './RuleInput';
 import RuleRemoveButton from './RuleRemoveButton';
-import { FormField, FormGroup } from './FormWrapper';
 
 interface RuleRowProps {
-    id: number;
-    onRemove: (id: number) => void;
+    allowRemove: boolean;
```
Gelio, line 8: Optional:

Perhaps 

```suggestion
    removable: boolean;
```

would be a bit shorter :smile:

widgets/common/src/filters/RuleRow.tsx
```diff
@@ -1,16 +1,16 @@
...
+    onRemove: ComponentProps<typeof Button>['onClick'];
 }
 
-const RuleRow: FunctionComponent<RuleRowProps> = ({ id, onRemove }) => {
+const RuleRow: FunctionComponent<RuleRowProps> = ({ allowRemove = true, onRemove }) => {
```
Gelio, line 12: Do we need a default value for a required prop?

```suggestion
const RuleRow: FunctionComponent<RuleRowProps> = ({ allowRemove, onRemove }) => {
```

widgets/common/src/filters/RuleRow.tsx
```diff
@@ -22,9 +22,7 @@ const RuleRow: FunctionComponent<RuleRowProps> = ({ id, onRemove }) => {
             <FormField>
                 <RuleInput />
             </FormField>
-            <FormField>
-                <RuleRemoveButton onClick={() => onRemove(id)} />
-            </FormField>
+            <FormField>{allowRemove && <RuleRemoveButton onClick={onRemove} />}</FormField>
```
Gelio, line 25: Optional, if it is possible :smile: 

Does it matter that we render an empty `FormField`, as opposed to not rendering it at all?

```suggestion
            {allowRemove && <FormField><RuleRemoveButton onClick={onRemove} /></FormField>}
```

This will render 1 fewer element :smile:

widgets/common/src/filters/RuleAttributeDropdown.tsx
```diff
@@ -1,10 +1,17 @@
 import type { FunctionComponent } from 'react';
+import type { DropdownItemProps } from 'semantic-ui-react/dist/commonjs/modules/Dropdown/DropdownItem';
+import { i18nPrefix } from './consts';
 
-const attributes = ['Label', 'Blueprint', 'Site name', 'Creator'];
+const { i18n } = Stage;
+const attributeOptions = ['label', 'blueprint_id', 'site_name', 'created_by'].map(
```
Gelio, line 6: This is named `attributeOptions`, and the one for operators is `operatorsOptions` (both words are plural). Can we have it named consistently, either both plural, or only the _options_ plural? :smile:

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -1,31 +1,37 @@
...
-const defaultRow = { attribute: '', operator: '', input: '' };
+interface FilterRuleRow extends FilterRule {
+    id: string;
+}
+
+function getNewRow() {
+    const { uuid } = Stage.Utils;
+    const defaultRow: FilterRuleRow = { id: '', type: 'label', key: '', operator: '', values: [] };
+
+    return { ...defaultRow, id: uuid() };
```
Gelio, line 16: How about combining the lines together :smile: ?

```suggestion
    return { id: uuid(), type: 'label', key: '', operator: '', values: [] };
```

Then we probably need the type annotation on the function's return (on line 12) :slightly_smiling_face:

Gelio (commented):
Some more comments :slightly_smiling_face: Only the one about `setState` functional form matters to a non-insignificant extent :smile:

app/translations/en.json
```diff
@@ -521,6 +521,22 @@
...
+                    "operatorsLabels": {
+                        "any_of": "is one of",
+                        "not_any_of": "is not one of",
+                        "is_null": "key is not",
+                        "is_not_null": "key is",
+                        "contain": "contains",
+                        "not_contain": "does not contain",
+                        "start_with": "starts with",
+                        "end_with": "ends with"
                     }
```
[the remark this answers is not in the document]
qooban: Done.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Updated.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Thanks for clarification. I don't see big benefit in this case, but changed according to the suggestion.

widgets/common/src/filters/RuleRow.tsx
qooban: Changed.

widgets/common/src/filters/RuleRow.tsx
qooban: Removed default value setting.

widgets/common/src/filters/RuleRow.tsx
qooban: No, it's not good in this case (length of the row is changing then). I'll keep it as it is.

widgets/common/src/filters/RuleAttributeDropdown.tsx
qooban: Changed.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Don't treat it as final, it will change for sure. I'll leave it as it is.

Gelio: Ok :slightly_smiling_face:

widgets/common/src/filters/RuleRow.tsx
Gelio: Makes sense :slightly_smiling_face:

Gelio (approved):
The only open comments of mine are optional, so I'm approving :slightly_smiling_face: Good work :+1:

kubama (approved):

Gelio (approved):

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: Sure, I'll consider it when I'll be closer to real testing. I think this part will change anyway, I'm considering extracting common `Dropdown` for filters.

test/cypress/support/filters.ts
qooban: Yep, that I thought as well - better to leave it now until we have API types support.

PR #1285, opened 2021-03-31
RD-1853 Use react-query for fetching deployments

Gelio:
This PR changes the way filter rules are resolved and deployments are fetched to using [react-query](https://react-query.tanstack.com/).

The PR is actually around +278, -112 lines after excluding package.json changes. There's also some refactoring and moving code around, so the diff may appear larger than it is. I did not come up with a better way to split it into multiple PRs

## Reason for using `react-query`, and not widget's `fetchData`

The reason using widget's `fetchData` is not easy is the main functionality of RD-1853 (fetching deployments based on `deploymentId` set in the parent page):
1. We need to access the Redux store to get the parent page's context, and that is not easily done inside `fetchData`. We could use `useSelector` or `connect` but only inside a component, not inside `fetchData`
2. Resolving filter rules on every request is redundant. We need to do it only when `filterId` changes. Doing that with `fetchData` would mean some hacky caching, but is easily done when using a component

Thus, manual data fetching inside the component sounded to me like a better idea than hacking and bending `fetchData` against its will :slightly_smiling_face: `react-query` makes it that much easier that it provides many functionalities I would have to implement otherwise myself, extending the scope (RD-1852 and RD-1854 are closed because of using react-query).

### Features
1. Filter rules are only fetched when the filter actually changes, not on every polling tick
2. Deployments are refetched not only on polling ticks, but also when the user focuses the tab again, thanks to react-query (IMO that's a nice free feature)
3. Progress is displayed more granularly - the user sees a message when the UI is fetching filter rules, and when it's fetching deployments (only during the initial fetch)

    See https://react-query.tanstack.com/guides/window-focus-refetching for more information.

4. The loading spinner now appears in the table, and not in the widget's header. This is to show the spinner in the place for which the data is being loaded (after all, we are fetching the data for the table, not the details pane).
5. Error messages are displayed with a bit more context - there is information whether the request for filters or deployments failed.
6. Past requests are cached for 5 minutes. This means going back to the same grid params will return a cached response, and still do a request to update the data in the background.

### Additional changes in this PR
1. I've installed a plugin for TS that checks the content of the style blocks in styled-components. If you make a typo, e.g. `positio: relative;`, it will be caught by TypeScript. 

    Unfortunately, this does not work when using `Stage.styled` directly. You have to either destructure it: `const { styled } = Stage; styled.div`, or import the package (see point 4)
2. I've upgraded TS to the latest version at the same time.
3. I've extracted some logic related to `gridParams`. It turns out that the `gridParams` returned by `DataTable` are not the same grid params that we send to the Manager - there's some translation of the object :slightly_smiling_face: Surprising, but at least now I know
4. I've allowed importing `styled-components` in widgets (instead of using `Stage.styled`)

There are no additional tests in this PR, since there are very little behavior changes (I did not think testing the exact fetching behavior was worth it). The functionalities of the widget do not change in this PR - it's more of a refactor.

## Known issues

1. Search does not work. This is because the `/searches/deployments` API does not handle it. To be fixed by the backend in RD-1926

## Video

https://user-images.githubusercontent.com/889383/113156807-9fb30480-923a-11eb-8212-43f312916808.mp4

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/401/pipeline

Ran 2021-04-07 9:19 CEST


backend/test/handlers/SourceHandler.test.js
```diff
@@ -1,6 +1,4 @@
 const fs = require('fs-extra');
-const os = require('os');
```
Gelio, line 2: This started failing after upgrading the ESLint parser. It turns out those variables are not used in the test

widgets/deploymentsView/src/detailsPane/header.scss
```diff
@@ -1,5 +1,5 @@
 .detailsPane {
-    margin: 0 10px;
+    margin-left: 10px;
```
Gelio, line 2: This gives the component 10px of whitespace on the right :slightly_smiling_face:

Gelio:
@gregcfy you can probably also take a look at the PR, optionally, to review the usage of react-query :slightly_smiling_face:

kubama (approved):
I may be not yet fully aware of all the benefits from introducing react-query, but anyway thanks for the initiative for improvement 👍.

widgets/deploymentsView/src/widget.tsx
```diff
@@ -1,19 +1,22 @@
+/* eslint-disable react/prop-types */
```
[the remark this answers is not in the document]
Gelio: Yes, it is :sweat_smile:  Thanks, I forgot to remove those `disable` comments after rebasing on master :slightly_smiling_face: I'll do that soon

app/typings/stage-api.d.ts
```diff
@@ -22,6 +23,7 @@ declare global {
         L: typeof import('leaflet');
         log: typeof import('loglevel');
         connectToStore: typeof import('react-redux').connect;
+        ReactQuery: typeof import('react-query');
```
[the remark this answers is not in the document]
Gelio: You are right, great suggestion :+1: In fact, we can do that with all of the above too :smile: I'll refactor the above properties as well, if you don't mind, to have less repetition in the code :slightly_smiling_face:

widgets/deploymentsView/src/widget.tsx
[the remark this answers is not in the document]
Gelio: Done

app/typings/stage-api.d.ts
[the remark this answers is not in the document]
Gelio: Done

Gelio:
> I may be not yet fully aware of all the benefits from introducing react-query, but anyway thanks for the initiative for improvement .

I now understood that I did not describe in the PR description why I chose react-query :smile: I've added a section about it to the PR description ("Reason for using react-query, and not widget's fetchData") :slightly_smiling_face: I hope that slightly clarifies the reasons

Thanks for a quick review :slightly_smiling_face: 

---

Since @qooban is OOO, I will wait with merging this PR until he comes back and reviews it just to be on the same page about introducing new libraries to the project and using them :slightly_smiling_face: Thus, I will be creating new PR using this one as a base. Hopefully that won't be too much additional maintenance overhead :crossed_fingers:

app/components/Home.jsx
```diff
@@ -7,7 +7,6 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import i18n from 'i18next';
 
-import stageUtils from '../utils/stageUtils';
```
Gelio, line 10: This came up as an ESLint warning and looks like it was introduced in fc63cf776 :smile: 

I'm not sure we should be allowing `--max-warnings 10` for ESLint. We usually disable rules inline using comments and providing justification for why it's being disabled. Having warnings that are not easily visible in code review sounds worse to me than having a `eslint-disable` comment that at least we can ask questions about during code review :slightly_smiling_face: 

I'll add a topic about it for the weekly

kubama (approved):
Thanks for extending PR description, I now have a better picture.

I'm wondering what's the long-term plan for this solution, do you think it would be reasonable to gradually move away from `fetchData` in favour of `react-query`?

Gelio:
> I'm wondering what's the long-term plan for this solution, do you think it would be reasonable to gradually move away from `fetchData` in favour of `react-query`?

I'm not fully confident yet, but having experimented with `react-query` for some time, it looks to provide better UX, more functionalities, better documentation, and overall is more well-known than our `fetchData` (meaning it's friendlier for newcomers and other widget developers), so migrating from `fetchData` to `react-query` sounds reasonable :slightly_smiling_face:

qooban:
Thanks for waiting for me with merging this PR. I'm having a bit busy day packed with meetings, but I'll definitely go over the PR today.

Gelio:
@qooban No worries, take your time :slightly_smiling_face:

widgets/deploymentsView/src/widget.tsx
```diff
@@ -36,7 +36,11 @@ if (process.env.NODE_ENV === 'development' || process.env.TEST) {
         categories: [Stage.GenericConfig.CATEGORY.DEPLOYMENTS],
 
         initialConfiguration: [
-            Stage.GenericConfig.POLLING_TIME_CONFIG(10),
+            {
+                ...Stage.GenericConfig.POLLING_TIME_CONFIG(10),
+                // NOTE: polling is handled by react-query, thus, use a different ID
+                id: 'customPollingTime'
```
qooban, line 42: I'm not sure you need different `id` when you have neither `fetchData` nor `fetchUrl` defined.

widgets/deploymentsView/src/widget.tsx
```diff
@@ -72,72 +76,94 @@ if (process.env.NODE_ENV === 'development' || process.env.TEST) {
...
+                // Do that only if `fetchData` was called from `DataTable`. If it's just polling,
+                // then don't reset it (because user may be interacting with some other component)
+                if (context.getValue('deploymentId') === undefined && data.items.length > 0) {
+                    context.setValue('deploymentId', data.items[0].id);
+                }
+            },
+            refetchInterval: customPollingTime * 1000,
+            keepPreviousData: true
+        }
+    );
```
qooban, line 124: Just wondering - did you consider encapsulating common parts for each `useQuery` calls into a custom hook?

widgets/deploymentsView/src/widget.tsx
```diff
@@ -72,72 +76,94 @@ if (process.env.NODE_ENV === 'development' || process.env.TEST) {
...
+        return <Loading message={i18n.t(`${i18nMessagesPrefix}.loadingDeployments`)} />;
+    }
+    if (deploymentsResult.isError) {
+        return (
+            <ErrorMessage
+                header={i18n.t(`${i18nMessagesPrefix}.errorLoadingDeployments`)}
+                error={deploymentsResult.error as { message: string }}
+            />
+        );
+    }
```
qooban, line 148: Looks very clear what will happen with those booleans (`isLoading`, `isError`, ...). I like the way how it can be done with `react-query`.

qooban (commented):
Don't see any blockers. Added few comments/questions. Good work with `react-query` :-)

We definitely need to think about how to use it in the future in the other parts of the application. 

P.S. First time did a review from the IDE (WebStorm). Hope all  of my comments will find the right place :-)

app/utils/shared/mapGridParamsToManagerGridParams.ts
```diff
@@ -0,0 +1,43 @@
+/* eslint no-underscore-dangle: ["error", { "allow": ["_sort", "_search", "_size", "_offset"] }] */
+declare global {
+    namespace Stage.Types {
+        interface GridParams {
+            sortColumn?: string;
+            sortAscending?: boolean;
+            _search?: string;
```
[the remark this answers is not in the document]
Gelio: Yup, I would love to see consistency here, but that's the interface that comes out of `DataTable` :smile: 

https://github.com/cloudify-cosmo/cloudify-ui-components/blob/master/src/components/data/DataTable/DataTable.jsx#L103-L109

widgets/deploymentsView/src/widget.tsx
Gelio: Good point. Although, if I left it as is, then we would still run the polling logic in `WidgetDynamicContent` needlessly. The polling logic keeps executing, but it will be just noops on an interval :smile: IMO this could cause less commotion :slightly_smiling_face:

widgets/deploymentsView/src/widget.tsx
Gelio: Yeah, I did for a while, but then I thought that I don't know how the hooks will look like when this component is invoked from the drill-down view (which does not need to fetch filters), so I decided to leave it inlined and abstract it away later :slightly_smiling_face:

widgets/deploymentsView/src/widget.tsx
Gelio: I'm glad you like it :smile: I also think highly of this pattern :smile:

widgets/deploymentsView/src/widget.tsx
qooban: Got it, in that case it make sense to keep `customPollingTime`.

widgets/deploymentsView/src/widget.tsx
qooban: OK 👍🏻

qooban (approved):
LGTM 👍🏻

Gelio:
Thanks for the approvals :slightly_smiling_face: 

I'll wait for the results of https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/401/pipeline and probably merge the PR in 1h :smile:

Gelio:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/401/pipeline/56/ passed, so I'll merge the PR :slightly_smiling_face:

PR #1288, opened 2021-04-01
RD-1123 Added on delete filter usage checking, aligned REST URLs to recent changes

kubama:
When trying to delete filter that is used as a default filter in deployments view widget the following alert is shown:

image

### DIFF
--- a/backend/handler/FilterHandler.js
+++ b/backend/handler/FilterHandler.js
@@ -1,0 +1,39 @@
+const _ = require('lodash');
+const { db } = require('../db/Connection');
+
+const logger = require('./LoggerHandler').getLogger('FilterHandler');
+
+module.exports = (() => {
+    function getFilterUsage(filterId) {
+        return db.UserApp.findAll({ attributes: ['appData', 'username'] }).then(userApp => {
+            const result = [];
+            _.forEach(userApp, ({ appData, username }) => {
+                _.forEach(appData.pages, page => {
+                    function checkWidgets(widgets) {
+                        _.forEach(widgets, widget => {
+                            if (widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId) {
+                                result.push({ pageName: page.name, widgetName: widget.name, username });
+                            }
+                        });
+                    }
+
+                    _.forEach(page.layout, ({ type, content }) => {
+                        if (type === 'widgets') {
+                            checkWidgets(content);
+                        } else if (type === 'tabs') {
+                            _.forEach(content, tab => checkWidgets(tab.widgets));
+                        } else {
+                            logger.warn('Unsupported layout type:', type);
+                        }
+                    });
+                });
+            });
+
+            return result;
+        });
+    }
+
+    return {
+        getFilterUsage
+    };
+})();
--- a/backend/handler/FilterHandler.js
+++ b/backend/handler/FilterHandler.js
@@ -3,37 +3,35 @@ const { db } = require('../db/Connection');
 
 const logger = require('./LoggerHandler').getLogger('FilterHandler');
 
-module.exports = (() => {
-    function getFilterUsage(filterId) {
-        return db.UserApp.findAll({ attributes: ['appData', 'username'] }).then(userApp => {
-            const result = [];
-            _.forEach(userApp, ({ appData, username }) => {
-                _.forEach(appData.pages, page => {
-                    function checkWidgets(widgets) {
-                        _.forEach(widgets, widget => {
-                            if (widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId) {
-                                result.push({ pageName: page.name, widgetName: widget.name, username });
-                            }
-                        });
+async function getFilterUsage(filterId) {
+    const userAppsArr = await db.UserApp.findAll({ attributes: ['appData', 'username'] });
+
+    const filterUses = [];
+    _.forEach(userAppsArr, ({ appData, username }) => {
+        _.forEach(appData.pages, page => {
+            function checkWidgets(widgets) {
+                _.forEach(widgets, widget => {
+                    if (widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId) {
+                        filterUses.push({ pageName: page.name, widgetName: widget.name, username });
                     }
-
-                    _.forEach(page.layout, ({ type, content }) => {
-                        if (type === 'widgets') {
-                            checkWidgets(content);
-                        } else if (type === 'tabs') {
-                            _.forEach(content, tab => checkWidgets(tab.widgets));
-                        } else {
-                            logger.warn('Unsupported layout type:', type);
-                        }
-                    });
                 });
+            }
+
+            _.forEach(page.layout, ({ type, content }) => {
+                if (type === 'widgets') {
+                    checkWidgets(content);
+                } else if (type === 'tabs') {
+                    _.forEach(content, tab => checkWidgets(tab.widgets));
+                } else {
+                    logger.warn('Unsupported layout type:', type);
+                }
             });
-
-            return result;
         });
-    }
+    });
 
-    return {
-        getFilterUsage
-    };
-})();
+    return filterUses;
+}
+
+module.exports = {
+    getFilterUsage
+};
--- a/backend/handler/FilterHandler.js
+++ b/backend/handler/FilterHandler.js
@@ -1,6 +1,6 @@
 const _ = require('lodash');
 const { db } = require('../db/Connection');
-
+const { LAYOUT } = require('../consts');
 const logger = require('./LoggerHandler').getLogger('FilterHandler');
 
 async function getFilterUsage(filterId) {
@@ -20,9 +20,9 @@ async function getFilterUsage(filterId) {
             }
 
             _.forEach(page.layout, ({ type, content }) => {
-                if (type === 'widgets') {
+                if (type === LAYOUT.WIDGETS) {
                     checkWidgets(content);
-                } else if (type === 'tabs') {
+                } else if (type === LAYOUT.TABS) {
                     _.forEach(content, tab => checkWidgets(tab.widgets));
                 } else {
                     logger.warn('Unsupported layout type:', type);
--- a/backend/handler/FilterHandler.js
+++ b/backend/handler/FilterHandler.js
@@ -11,9 +11,9 @@ async function getFilterUsage(filterId) {
         _.forEach(appData.pages, page => {
             function checkWidgets(widgets) {
                 _.forEach(widgets, widget => {
-                    const filteUsed =
+                    const filterUsed =
                         widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId;
-                    if (filteUsed) {
+                    if (filterUsed) {
                         filterUses.push({ pageName: page.name, widgetName: widget.name, username });
                     }
                 });
--- a/backend/handler/FilterHandler.js
+++ b/backend/handler/FilterHandler.js
@@ -11,7 +11,9 @@ async function getFilterUsage(filterId) {
         _.forEach(appData.pages, page => {
             function checkWidgets(widgets) {
                 _.forEach(widgets, widget => {
-                    if (widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId) {
+                    const filteUsed =
+                        widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId;
+                    if (filteUsed) {
                         filterUses.push({ pageName: page.name, widgetName: widget.name, username });
                     }
                 });
--- a/backend/test/routes/Filters.test.js
+++ b/backend/test/routes/Filters.test.js
@@ -3,56 +3,56 @@ const mockDb = require('../mocks/mockDb');
 require('../mocks/passport');
 
 describe('/filters endpoint', () => {
-    it('allows to get filter usage data', () => {
-        const filterId = 'filterId';
-        const pageName = 'Page1';
-        const widgetName = 'Deployments View 1';
-        const tabWidgetName = 'Deployments View 2';
-        const username = 'test_user';
+    const filterId = 'filterId';
+    const pageName = 'Page1';
+    const widgetName = 'Deployments View 1';
+    const tabWidgetName = 'Deployments View 2';
+    const username = 'test_user';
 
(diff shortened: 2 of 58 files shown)


Gelio:
I won't be able to review this PR today, I'll do that on Tuesday :slightly_smiling_face: I've done some overtime yesterday due to the system test issues and I want to balance it out by finishing early today

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,39 @@
+const _ = require('lodash');
+const { db } = require('../db/Connection');
+
+const logger = require('./LoggerHandler').getLogger('FilterHandler');
+
+module.exports = (() => {
```
Gelio, line 6: Digression:

What is the benefit of using a revealing module pattern/IIFE here, instead of having all the functions be defined in the module's scope, and then just assign some to `module.exports`? I see this pattern is used in some files, and I didn't come up with its advantages, but I do see cons (more indentation :smile: )

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,39 @@
...
+    function getFilterUsage(filterId) {
+        return db.UserApp.findAll({ attributes: ['appData', 'username'] }).then(userApp => {
```
Gelio, line 8: Digression:

Do we not use `async/await` in the backend's code? It could reduce the nesting here and seems useful to me :slightly_smiling_face:

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,39 @@
...
+            const result = [];
+            _.forEach(userApp, ({ appData, username }) => {
```
Gelio, line 10: Is `userApp` an array, or an object of userApps? Having not type annotations, the variable contents are unclear to me.

I would suggest naming it `userApps`, since it looks like there are many of them as we iterate over them, and then suffixing it with either `Obj` or `Arr` or `Dict` to hint the type :slightly_smiling_face:

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,39 @@
...
+                    }
+
+                    _.forEach(page.layout, ({ type, content }) => {
+                        if (type === 'widgets') {
+                            checkWidgets(content);
+                        } else if (type === 'tabs') {
+                            _.forEach(content, tab => checkWidgets(tab.widgets));
+                        } else {
+                            logger.warn('Unsupported layout type:', type);
+                        }
```
Gelio, line 27: Digression:

Do we not have a `forAllWidgets` helper to encapsulate similar logic?

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,39 @@
+const _ = require('lodash');
+const { db } = require('../db/Connection');
+
+const logger = require('./LoggerHandler').getLogger('FilterHandler');
+
+module.exports = (() => {
+    function getFilterUsage(filterId) {
+        return db.UserApp.findAll({ attributes: ['appData', 'username'] }).then(userApp => {
+            const result = [];
```
Gelio, line 9: We could make the variable name more concrete and provide a bit of context:

```suggestion
            const filterUses = [];
```

or

```suggestion
            const filterOccurrences = [];
```

backend/test/routes/Filters.test.js
```diff
@@ -0,0 +1,66 @@
+const request = require('supertest');
+const mockDb = require('../mocks/mockDb');
+require('../mocks/passport');
+
+describe('/filters endpoint', () => {
+    it('allows to get filter usage data', () => {
```
Gelio, line 6: Optional:

Can we have a test case that verifies there are no false positives (if some filter is not used in Deployments View, then the response is `[]`)?

widgets/filters/src/FiltersTable.tsx
```diff
@@ -77,12 +78,37 @@ const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, wid
...
             />
+
+            <Alert
+                open={!isEmpty(filterUsage)}
+                onDismiss={() => {
+                    clearFilterUsage();
+                    clearFilterIdToDelete();
+                }}
+                content={
+                    <span style={{ lineHeight: '3em', fontSize: '0.9em' }}>
```
Gelio, line 102: Optional:

How about using styled-components here? :slightly_smiling_face: You could also explain if the exact values set in this style matter (e.g. they are set so the content aligns with some other elements nearby), or they're just arbitrary :smile:

widgets/filters/src/FiltersTable.tsx
```diff
@@ -77,12 +78,37 @@ const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, wid
...
+                            filterActions.doDelete(filterIdToDelete).then(toolbox.refresh);
+                        } else {
+                            setFilterUsage(resolvedFilterUsage);
+                        }
+                    });
                 }}
                 content={i18n.t('widgets.filters.deleteConfirm', { filterId: filterIdToDelete })}
             />
+
+            <Alert
```
Gelio, line 95: Will this alert always appear on top of the `Confirm` defined above?

We could increase the safety here by running `clearFilterIdToDelete` unconditionally before calling `setFilterUsage` in the `onConfirm` handler above. Then, the `Confirm` will always close and the user will simply see only the `Alert`

What do you think?

Gelio (commented):
I appreciate the tests and the use of TS :+1:

backend/handler/FilterHandler.js
kubama: No idea, I just followed the pattern believing there must be a reason behind it ;-).

backend/handler/FilterHandler.js
kubama: We do, but it's on frontend side

backend/handler/FilterHandler.js
kubama: I was making changes in this file anyway, so I decided to refactor it.

backend/handler/FilterHandler.js
kubama: Good point, refactored.

backend/handler/FilterHandler.js
kubama: Done

backend/test/routes/Filters.test.js
kubama: Done

widgets/filters/src/FiltersTable.tsx
kubama: > How about using styled-components here?

In such simple case I don't see much advantage, so I'd rather keep the styling close to where it applies.

> You could also explain if the exact values set in this style matter (e.g. they are set so the content aligns with some other elements nearby), or they're just arbitrary 😄

They are arbitrary. The default font style in `Alert` seems to be intended for single-sentence messages - the font is big and line height is not even set in line with it. Here, the message to be displayed is complex, so it looks much better with a smaller font.

widgets/filters/src/FiltersTable.tsx
kubama: > Will this alert always appear on top of the `Confirm` defined above?

Technically yes. But `Alert` is also a modal, so from user perspective it doesn't matter. It looks the same as if the `Alert` was displayed alone.

> We could increase the safety here by running `clearFilterIdToDelete` unconditionally before calling `setFilterUsage` in the `onConfirm` handler above.

That would be cleaner. Unfortunately `filterIdToDelete` is used in alert's content, so it wouldn't work.

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,37 @@
...
+
+async function getFilterUsage(filterId) {
+    const userAppsArr = await db.UserApp.findAll({ attributes: ['appData', 'username'] });
+
+    const filterUses = [];
+    _.forEach(userAppsArr, ({ appData, username }) => {
+        _.forEach(appData.pages, page => {
+            function checkWidgets(widgets) {
+                _.forEach(widgets, widget => {
+                    if (widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId) {
```
qooban, line 14: Can we extract it to a const or a function (e.g. `isFilterUsed`)? 

I assume that in the future we might have another widget here (Deployments View drill-down), so possibly there will be an array necessary.

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,37 @@
...
+                        filterUses.push({ pageName: page.name, widgetName: widget.name, username });
+                    }
+                });
+            }
+
+            _.forEach(page.layout, ({ type, content }) => {
+                if (type === 'widgets') {
+                    checkWidgets(content);
+                } else if (type === 'tabs') {
```
qooban, line 23: General: 

I'm wondering if we can have consts for it ( `widgets` and `tabs`). I see we have consts already defined in the frontend side. Unfortunately no easy way to reuse it, yet between frontend and backend. That's probably another vote for having RD-1103 implemented.

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,37 @@
+const _ = require('lodash');
+const { db } = require('../db/Connection');
+
+const logger = require('./LoggerHandler').getLogger('FilterHandler');
+
+async function getFilterUsage(filterId) {
+    const userAppsArr = await db.UserApp.findAll({ attributes: ['appData', 'username'] });
```
qooban, line 7: General:

I assume that you are expecting multiple rows as consequence of not using `managerIp` here?
Another point to fix RD-1457 as soon as possible.

widgets/filters/src/FiltersTable.tsx
```diff
@@ -77,12 +78,37 @@ const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, wid
...
                 onConfirm={() => {
-                    clearFilterIdToDelete();
-                    toolbox.loading(true);
-                    new FilterActions(toolbox).doDelete(filterIdToDelete).then(toolbox.refresh);
+                    const filterActions = new FilterActions(toolbox);
+                    filterActions.doGetFilterUsage(filterIdToDelete).then(resolvedFilterUsage => {
+                        if (isEmpty(resolvedFilterUsage)) {
+                            clearFilterIdToDelete();
+                            toolbox.loading(true);
+                            filterActions.doDelete(filterIdToDelete).then(toolbox.refresh);
```
qooban, line 86: It's probably unlikely to happen, but I'm not sure by whom will the potential error in that promise be captured? 
Shouldn't we handle it here?

widgets/filters/src/FiltersTable.tsx
qooban: Looks like `Alert` is a modal without content, but with just `Modal.Header`, so that was for sure desired for single sentence.

https://github.com/cloudify-cosmo/cloudify-ui-components/blob/537a0e5c50f6152ba2f996372cc92529dc707a96/src/components/modal/Alert/Alert.jsx#L13..L18

qooban (commented):
Looks good in general. Added few comments.

backend/handler/FilterHandler.js
kubama: Done

backend/handler/FilterHandler.js
kubama: Turned out there were already other references to these constants in the backend, so I decided to extract them for now.

backend/handler/FilterHandler.js
kubama: Actually my thinking was that there might be multiple rows as UserApp is defined per user.

widgets/filters/src/FiltersTable.tsx
kubama: Actually we discussed this on our weekly, please see RD-1967 that was created as a result.

backend/handler/FilterHandler.js
```diff
@@ -0,0 +1,39 @@
...
+    const filterUses = [];
+    _.forEach(userAppsArr, ({ appData, username }) => {
+        _.forEach(appData.pages, page => {
+            function checkWidgets(widgets) {
+                _.forEach(widgets, widget => {
+                    const filteUsed =
+                        widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId;
+                    if (filteUsed) {
+                        filterUses.push({ pageName: page.name, widgetName: widget.name, username });
+                    }
```
qooban, line 18:
```suggestion
                    const filterUsed =
                        widget.definition === 'deploymentsView' && widget.configuration.filterId === filterId;
                    if (filterUsed) {
                        filterUses.push({ pageName: page.name, widgetName: widget.name, username });
                    }
```

qooban (commented):
Just one comment.

backend/handler/FilterHandler.js
qooban: Oh, right, that is clear now. 

Funny thing will be (until we solve RD-1457) that in the cluster environment user can get multiple bullets ("Used as default filter in ...") with the same text.

widgets/filters/src/FiltersTable.tsx
qooban: OK, clear 👍🏻

qooban (approved):

backend/handler/FilterHandler.js
Gelio: I assume that we don't often need to do such iteration, so let's leave it as is :slightly_smiling_face: But I reckon it may be useful to abstract away such iteration if it happens often in the backend code too, so as to use a single implementation for both frontend and backend

widgets/filters/src/FiltersTable.tsx
Gelio: I see :smile: I'm ok with not doing anything here, so I'll let any of you resolve the thread if you think that it's minor as well :slightly_smiling_face:

widgets/filters/src/FiltersTable.tsx
Gelio: I see, thanks for the explanation :+1:

backend/handler/FilterHandler.js
Gelio: Just to clarify, the Deployments View drill-down will likely not use any predetermined filter ID :smile: It will get the filter rules from its parent (basically, `csys-obj-type = environment/service` and `csys-obj-parent = parentDeploymentId`)

Still, I see the value from extracting, good point :+1:

Gelio (approved):
:ok_hand:

PR #1291, opened 2021-04-07
RD-1123 Added add/edit/clone operations implementation (without filter rules definition form)

kubama:
All the modals lack filter definition form which will be added as part of RD-1985.
- add modal has only filter id input and creates filter with no rules
- edit modal has no content, clicking 'Save' updates the filter but same rules are used
- clone modal has only filter id input and creates new filter with rules same as in the original filter

image

image

image


Gelio:
I've spent 1.5h on code review already today, so I'll defer this PR to later today or tomorrow :slightly_smiling_face:

qooban:
I'll review this PR tomorrow morning.

app/translations/en.json
```diff
@@ -640,6 +641,15 @@
                     "clone": "Clone filter"
                 }
             },
+            "modal": {
+                "id": "Filter id",
+                "idMissing": "Please provide filter id",
```
Gelio, line 646: Optional:

I would use proper English here since it looks like a sentence

```suggestion
                "idMissing": "Please provide the filter id",
```

or

```suggestion
                "idMissing": "Please provide a filter id",
```

widgets/filters/src/FilterActions.ts
```diff
@@ -14,4 +14,14 @@ export default class FilterActions {
     doGetFilterUsage(filterId: string): Promise<FilterUsage[]> {
         return this.toolbox.getInternal().doGet(`/filters/usage/${filterId}`);
     }
+
+    doCreate(filterId: string, filterRules: []) {
```
Gelio, line 18: Optional:

The empty array type here is actually an empty tuple, meaning that this function will only accept empty arrays :smile: Not sure if that was what you're going for :slightly_smiling_face: 

Using

```suggestion
    doCreate(filterId: string, filterRules: unknown[]) {
```

would probably be a bit nicer to use :slightly_smiling_face: I understand it's only an intermediate state, so no changes necessary here

IMO it would be nice to use the `FilterRule` type added by @qooban in #1284 

```suggestion
    doCreate(filterId: string, filterRules: FilterRule[]) {
```

---

The same applies to the `doUpdate` method

widgets/filters/src/types.ts
```diff
@@ -4,6 +4,10 @@ export interface Filter {
     created_by: string;
     // eslint-disable-next-line camelcase
     created_at: string;
+    // eslint-disable-next-line camelcase
+    labels_filter_rules: [];
+    // eslint-disable-next-line camelcase
+    attrs_filter_rules: [];
```
Gelio, line 10: There are 4 `eslint-disable-next-line` comments :slightly_smiling_face: How about doing a single `eslint-disable` and then `eslint-enable`? We save 2 lines of code this way

widgets/filters/src/types.ts
```diff
@@ -4,6 +4,10 @@ export interface Filter {
     created_by: string;
     // eslint-disable-next-line camelcase
     created_at: string;
+    // eslint-disable-next-line camelcase
+    labels_filter_rules: [];
```
Gelio, line 8: The same comment about using an empty tuple applies here as in `FilterActions.ts` :slightly_smiling_face:

widgets/filters/src/FilterModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+    i18nHeaderKey: string;
+    onSubmit: (filterId: string, filterRules: []) => Promise<void>;
+    onCancel: () => void;
+}
+
+const { Form } = Stage.Basic;
+
+// TODO: RD-1837
+const UnsafeForm = (Form as unknown) as FunctionComponent<{ [x: string]: any }>;
+const UnsafeField = (Form.Field as unknown) as FunctionComponent<{ [x: string]: any }>;
```
Gelio, line 16: You can probably use the components added by @qooban in #1284 now :slightly_smiling_face:

widgets/filters/src/FilterModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+        clearErrors();
+
+        if (filterIdProp !== undefined && !filterId) {
+            setErrors({ filterId: i18n.t('widgets.filters.modal.idMissing') });
+            return;
+        }
+
+        onSubmit(
+            (filterId || initialFilter?.id) as string,
+            initialFilter ? [...initialFilter.attrs_filter_rules, ...initialFilter.labels_filter_rules] : []
```
Gelio, line 40: Optional:

How about extracting the logic that combines `attrs_filter_rules` and `labels_filter_rules` into a separate function? e.g. `getConcatenatedFilterRules`

```suggestion
            initialFilter ? getConcatenatedFilterRules(initialFilter) : []
```

widgets/filters/src/FilterModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    const { ApproveButton, CancelButton, Icon, Modal } = Stage.Basic;
+    return (
+        <Modal open onClose={onCancel}>
+            <Modal.Header>
+                <Icon name="filter" /> {i18n.t(i18nHeaderKey, { filterId: initialFilter?.id })}
+            </Modal.Header>
+
+            <Modal.Content>
+                {filterId !== undefined && (
```
Gelio, line 52: Maybe it would read a bit better if we used a separate prop to determine whether to show filter ID input, rather than relying on `filterId` being undefined. I came to that conclusion after seeing that `filterId=""` in `FilterAddModal`, which seems counter-intuitive to me ("why do we set an empty filter ID when we're adding a filter?")

I imagine the prop should be named `showFilterIdField`

Then, we could rename `filterId` to `initialFilterId`, which seems more appropriate, also taking into account how it's used (it's an uncontrolled prop) :slightly_smiling_face:

widgets/filters/src/FilterAddModal.tsx
```diff
@@ -0,0 +1,11 @@
+import { FunctionComponent } from 'react';
+import FilterModal, { FilterModalProps } from './FilterModal';
+
+type FilterAddModalProps = Pick<FilterModalProps, 'onCancel' | 'onSubmit'>;
```
Gelio, line 4: Nice type reuse :+1:

widgets/filters/src/FilterAddModal.tsx
```diff
@@ -0,0 +1,11 @@
+import { FunctionComponent } from 'react';
```
Gelio, line 1: Optional:

Could we do `import type` when possible :smile: ?

Gelio (commented):
Looks mostly good :+1: 

I like the pattern of having a shared modal with only 3 subcomponents that render it with different props :ok_hand: Reads well

widgets/filters/src/types.ts
kubama: I consider this temporary so I don't really care - I imagine once we have common place for storing REST types then this will be disabled on directory level.

widgets/filters/src/FilterActions.ts
kubama: I reused the FilterRule type.
I used direct type import from `common/src/filters/types` - probably temporary solution until we agree where to store REST types and how to access them. For now as it is just a type import I assume it is safe.

widgets/filters/src/types.ts
kubama: Addressed

widgets/filters/src/FilterModal.tsx
kubama: Right, reused

widgets/filters/src/FilterAddModal.tsx
kubama: Fixed.
Looks like WebStorm auto-import feature does not add `type`. I believe adding an eslint rule would aid it as I expect the rule to be auto-fixable.

widgets/filters/src/FilterModal.tsx
kubama: This is temporary - once rules form will be in place the list of rules to be submitted will be provided by the form.

app/translations/en.json
```diff
@@ -670,6 +671,15 @@
                     "clone": "Clone filter"
                 }
             },
+            "modal": {
+                "id": "Filter id",
+                "idMissing": "Please provide the filter id",
```
qooban, line 676: I just roughly checked how we present "id" to the user in cloudify-stage. It seems that sometimes we use "ID", sometime we use "Id". I think "ID" is the correct solution, but I'm open to any other suggestions.

widgets/filters/src/FilterModal.tsx
```diff
@@ -0,0 +1,66 @@
+import { FunctionComponent } from 'react';
+import { Filter } from './types';
+import type { FilterRule } from '../../common/src/filters/types';
```
qooban, line 3: Optional:

We can also expose `FilterRule` through `Stage.Common.Filter`, but as long as we use `import type` IIRC it should be handled without big overhead by TypeScript.

widgets/filters/src/types.ts
```diff
@@ -1,9 +1,15 @@
...
 export interface Filter {
     id: string;
     // eslint-disable-next-line camelcase
     created_by: string;
     // eslint-disable-next-line camelcase
     created_at: string;
+    // eslint-disable-next-line camelcase
+    labels_filter_rules: FilterRule[];
+    // eslint-disable-next-line camelcase
+    attrs_filter_rules: FilterRule[];
```
qooban, line 12: Maybe instead of having multiple lines with: `// eslint-disable-next-line camelcase`
we can have single line: 
```
/*eslint camelcase: ["error", {allow: ["created_by", "created_at", "labels_filter_rules", "attrs_filter_rules"]}]*/
```
or maybe: 
```
/*eslint camelcase: ["error", {properties: "never"}]*/
``` 
will work?

Ref.: https://eslint.org/docs/rules/camelcase

qooban (commented):
I like the way of reusing modals 👍🏻 
Don't see any blockers. Left some minor comments.

widgets/filters/src/FilterActions.ts
Gelio: Ok sure :slightly_smiling_face:

widgets/filters/src/types.ts
Gelio: Makes sense :+1:

widgets/filters/src/FilterModal.tsx
Gelio: Makes sense :+1:

widgets/filters/src/FilterAddModal.tsx
Gelio: I have just created RD-2000 to enforce it :slightly_smiling_face: We can tackle it some time in the future to avoid such review comments

widgets/filters/src/types.ts
Gelio: This is related to my previous comment on the same topic :smile: https://github.com/cloudify-cosmo/cloudify-stage/pull/1291#discussion_r608697896

Gelio:
I've reviewed the code again and I don't have further comments :slightly_smiling_face: I'm only looking for an answer to https://github.com/cloudify-cosmo/cloudify-stage/pull/1291#discussion_r608654273 before approving

widgets/filters/src/types.ts
qooban: Hm... Haven't seen that when I was doing the review. Probably after some code change that comment was automatically removed and it is only visible in the "Conversation" tab (which I did not focus on during the review).

widgets/filters/src/FilterModal.tsx
kubama: Good point - added explicit `showFilterIdInput` prop.

app/translations/en.json
kubama: I believe "ID" is the correct one as well. Updated.

widgets/filters/src/FilterModal.tsx
kubama: I assume `FilterRule` should be placed in the common location for REST API types. Once it is done it will be easier to decide how to access the types from app/widgets/tests. For now I consider this import as (probably) temporary solution.

widgets/filters/src/types.ts
qooban: Resolving for now.

widgets/filters/src/FilterModal.tsx
qooban: OK 👍🏻

qooban (approved):
LGTM 👍🏻

widgets/filters/src/FilterEditModal.tsx
```diff
@@ -0,0 +1,16 @@
+import { FunctionComponent } from 'react';
+import FilterModal, { FilterModalProps } from './FilterModal';
+
+type FilterEditModalProps = Omit<FilterModalProps, 'i18nHeaderKey' | 'filterId'>;
```
Gelio, line 4: Hmm... Should the `filterId` prop be there? Should it be the _suffix_ prop?

Gelio (approved):
Looks good to me, but I think that there is 1 typo :smile: You can fix it in a future PR, so I'll approve the PR anyway

widgets/filters/src/FilterEditModal.tsx
kubama: Yes, good catch. I'll fix it in the upcoming PR.

PR #1292, opened 2021-04-07
RD-1839 RD-1840 Use renamed and stable properties

Gelio:
The recently-added environments-related deployment properties are now present in the API responses, so we can safely use them, thus, they no longer need to be optional (RD-1839)

The latest execution progress properties were prefixed with `latest_execution_`, so they had to be renamed in the fixture and the code (RD-1840)

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/406/pipeline

Ran 2021-04-07 13:22 CEST

### DIFF
--- a/widgets/deploymentsView/src/types.ts
+++ b/widgets/deploymentsView/src/types.ts
@@ -32,14 +32,13 @@ export interface Deployment {
     latest_execution_status: LatestExecutionStatus;
     deployment_status: DeploymentStatus;
     installation_status: InstallationStatus;
-    // TODO(RD-1839): make properties required after they are stable in the backend
-    environment_type?: string;
+    environment_type: string;
     latest_execution_total_operations: number;
     latest_execution_finished_operations: number;
-    sub_services_count?: number;
-    sub_services_status?: SubdeploymentStatus;
-    sub_environments_count?: number;
-    sub_environments_status?: SubdeploymentStatus;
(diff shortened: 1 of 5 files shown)


widgets/deploymentsView/src/types.ts
```diff
@@ -32,14 +32,13 @@ export interface Deployment {
...
-    sub_services_status?: SubdeploymentStatus;
-    sub_environments_count?: number;
-    sub_environments_status?: SubdeploymentStatus;
+    environment_type: string;
+    latest_execution_total_operations: number;
+    latest_execution_finished_operations: number;
+    sub_services_count: number;
+    sub_services_status: SubdeploymentStatus | null;
+    sub_environments_count: number;
+    sub_environments_status: SubdeploymentStatus | null;
```
Gelio, line 41: The statuses can be `null` when there are no children

qooban (approved):

kubama (approved):

Gelio:
The system tests passed, and I assume @qooban is still approving the PR (the approval was dismissed because I pushed a commit to remove an unnecessary import), meaning I should be good to go with both approvals :slightly_smiling_face: Thus, I'm merging the PR

qooban:
Sure, no issues 👍🏻 🙂

PR #1293, opened 2021-04-07
RD-1619 Changing pages layout for CaaS

alexmolev:


alexmolev:
The new layout for CaaS to simplify the navigation for the users.

image

qooban (approved):
Looking at: RD-1619 got some questions:
* Does "Should contain Local Blueprints only" mean that you want also to remove Composer Link widget?
* Does "Should contain Deployments only" mean that you want also to remove Create Deployment button widget and Resource Filter widget?

If the answer for both questions is no, then it looks good.

PR #1296, opened 2021-04-08
RD-1758 / RD-1760 - Add basic support for labels rules and free-text support for other rules

qooban:
### Changes

* Make `Stage.Common.Filters.Form` usable component with simple interface: 
```
{
    initialFilters: FilterRule[];
    onChange: (filterRules: FilterRule[]) => void;
    toolbox: Stage.Types.Toolbox;
}
```
where `FilterRule` is Cloudify API accepted filter rule (see: https://docs.cloudify.co/api/staging/dev/#filters).
* Add support for labels rules with `is_null` and `is_not_null` operators
* Add partial support for labels rules with `any_of`, `not_any_of` operators (partial, because you cannot specify multiple label values)
* Add free text support for other rules (blueprint, site name, creator)

In this PR there is no support for multiple things. I have created few issues to handle that:
* Multiple selection dropdown with auto completion for label values (RD-2006)
* Labels dropdowns styling (RD-2007)
* Error handling (e.g. empty rule values) and sending info to the parent (RD-2011)
* ValueDropdown & KeyDropdown should not allow to provide free text in labels filters (RD-2012)

### Video

https://user-images.githubusercontent.com/5202105/114146924-73714500-9918-11eb-94c9-222337518eb3.mp4


qooban:
@Gelio @kubama 

I know we all are extremely busy right now, so I don't expect you to review PR in draft state, but... If you - by chance - find some free cycles today, you can take a look. 

Just, be aware that it is not yet finalized, I will continue tomorrow morning, eg. I didn't manage to add proper description to the PR today. On the other hand I don't expect big code changes to come into this PR. There are many leftovers I am aware of, I'm going to update/create other issues to handle it and probably provide some more explanations using PR description / GitHub comments.

Gelio:
I doubt having free time today, especially with the upcoming monthly meeting :slightly_smiling_face: I'll review this one tomorrow, though

widgets/common/src/filters/types.ts
```diff
@@ -2,6 +2,43 @@
 export interface FilterRule {
     key: string;
     values: string[];
-    operator: string;
+    operator: RuleOperator;
     type: 'label' | 'attribute';
```
kubama, line 6: Would it be beneficial to introduce enum here?

widgets/common/src/filters/RuleInput.tsx
```diff
@@ -1,7 +1,56 @@
...
+
+interface RuleOperatorDropdownProps {
+    onChange: (value: string[]) => void;
+    operator: RuleOperator;
+    ruleType: RuleType;
+    toolbox: Stage.Types.Toolbox;
+    values: string[];
+}
+
+const RuleInput: FunctionComponent<RuleOperatorDropdownProps> = ({ onChange, operator, ruleType, toolbox, values }) => {
```
kubama, line 16: Wouldn't `RuleValueInput` be a more appropriate name for this component?

widgets/common/src/filters/RuleInput.tsx
```diff
@@ -1,7 +1,56 @@
...
     const { Input } = Stage.Basic;
-    return <Input type="text" />;
+    const keyDropdownRef = useRef();
+    const [labelKey, labelValue] = values;
+
+    useEffect(() => {
+        if (
+            ruleType === RuleType.Label &&
+            (operator === LabelsOnlyRuleOperator.IsNull || operator === LabelsOnlyRuleOperator.IsNotNull)
```
kubama, line 24: This somehow duplicates condition that is already present in L30 and L40 - if so, maybe the `onChange` call could be moved to `LabelValueDropdown` and called on component unmount?

widgets/common/src/filters/types.ts
```diff
@@ -2,6 +2,43 @@
...
+}
+
+export enum RuleType {
+    Label = 'label',
+    Blueprint = 'blueprint_id',
+    SiteName = 'site_name',
+    Creator = 'created_by'
+}
+
+export enum CommonRuleOperator {
```
kubama, line 20: Looking at how it is used I don't think we actually need an exported separate enum for common rule operators - instead it could be composed into `LabelsOnlyRuleOperator` and `AttributesOnlyRuleOperator` enums.

widgets/common/src/filters/RuleOperatorDropdown.tsx
```diff
@@ -1,28 +1,50 @@
...
 import type { DropdownItemProps } from 'semantic-ui-react/dist/commonjs/modules/Dropdown/DropdownItem';
+
 import { i18nPrefix } from './consts';
+import { RuleType, LabelsRuleOperator, AttributesRuleOperator } from './types';
+import type { RuleOperator } from './types';
+
+interface RuleOperatorDropdownProps {
+    ruleType: RuleType;
+    onChange: (value: RuleOperator) => void;
+    operator: RuleOperator;
```
kubama, line 12: Wouldn't it be more expected to name this prop `value`?

widgets/common/src/filters/RuleTypeDropdown.tsx
```diff
@@ -0,0 +1,40 @@
...
+    return (
+        <Dropdown
+            clearable={false}
+            search
+            selection
+            selectOnNavigation
+            name="ruleType"
+            options={ruleTypeOptions}
+            value={ruleType}
+            onChange={(_event: SyntheticEvent, data: DropdownProps) => onChange(data.value as RuleType)}
```
kubama, line 36: This looks like something very common - don't we somewhere have this value extraction logic as utility function that could be reused here?

kubama (commented):
I've added some comments.

I like enums introduction in this PR - reviewing it surely broadened my TS perspective a bit 🙂.

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -1,24 +1,52 @@
...
+
+    const emptyRow: FilterRuleRow = {
+        id: '',
+        type: 'label',
+        key: '',
+        operator: CommonRuleOperator.AnyOf,
+        values: []
+    };
+
+    return { ...emptyRow, id: uuid() };
```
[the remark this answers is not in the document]
qooban: Originally it was outside of the function, but as L21 is the only use of `emptyRow`, I put it inside the function.
Do you think it'd be better to change it?

widgets/common/src/filters/types.ts
```diff
@@ -2,6 +2,43 @@
...
     key: string;
     values: string[];
-    operator: string;
+    operator: RuleOperator;
     type: 'label' | 'attribute';
 }
+
+export interface FilterRuleRow extends FilterRule {
+    id: string;
+}
```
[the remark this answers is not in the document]
qooban: Sorry, I think I have too little TS understanding, so I don't get it. Can you explain it a bit more?

Is it to have something like:
```
export interface FilterRuleRow {
    id: string;
    rule: FilterRule;
}
```
?

By conversion logic you mean tranforming `FilterRule` into `FilterRuleRow` and the other way round?
`FilterRuleRow` -> `FilterRule` is a one-liner: `omit(filterRuleRow, 'id')`
`FilterRule` -> `FilterRuleRow` is a one-liner, too: `{ ...filterRule, id: uuid() }`
So I think they are quite simple.

widgets/common/src/filters/types.ts
```diff
@@ -2,6 +2,43 @@
...
+
+export const RuleOperator = { ...CommonRuleOperator, ...LabelsOnlyRuleOperator, ...AttributesOnlyRuleOperator };
+export type RuleOperator = CommonRuleOperator | LabelsOnlyRuleOperator | AttributesOnlyRuleOperator;
+
+export const LabelsRuleOperator = { ...CommonRuleOperator, ...LabelsOnlyRuleOperator };
+export type LabelsRuleOperator = CommonRuleOperator | LabelsOnlyRuleOperator;
+
+export const AttributesRuleOperator = { ...CommonRuleOperator, ...AttributesOnlyRuleOperator };
+export type AttributesRuleOperator = CommonRuleOperator | AttributesOnlyRuleOperator;
```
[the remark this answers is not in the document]
qooban: I found that way to allow providing a type and "enum"-like const.
Ref.: https://stackoverflow.com/questions/48478361/how-to-merge-two-enums-in-typescript - section "1. Using string enums"

I think now that I don't need all of them. I'll try to refactor it a bit.

widgets/common/src/filters/types.ts
```diff
@@ -2,6 +2,43 @@
 export interface FilterRule {
     key: string;
     values: string[];
-    operator: string;
+    operator: RuleOperator;
     type: 'label' | 'attribute';
 }
```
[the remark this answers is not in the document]
qooban: Hmm... Maybe, I'm not very fluent in TS, so I cannot judge it properly. 

I have only one concern, wouldn't it introduce more complexity and at the same time provide very little value?

widgets/common/src/filters/types.ts
qooban: I was thinking about that as well. I'll add it.

widgets/common/src/filters/RuleInput.tsx
qooban: Yeah, make sense, I'll change it.

widgets/common/src/filters/RuleInput.tsx
qooban: I'll try to implement it, I'm not sure it will work this way.

widgets/common/src/filters/types.ts
qooban: Right, I'll remove export.

widgets/common/src/filters/FiltersDefinitionForm.tsx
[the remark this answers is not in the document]
kubama: Not necessarily. I was just surprised by creating an object with empty `id` and then cloning it immediately and assigning a new `id` value. I think it would make perfect sense to create it just once with proper `id` value set at hand.

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: If so, then in other components like `RuleTypeDropdown` and `RuleInput` dropdown we should also change the main prop to `value`. I did that at the beginning, but I ended up thinking it's more clear to see the actual name of the filter field as a prop. Let me know if I should still consider the change.

widgets/common/src/filters/RuleTypeDropdown.tsx
qooban: I'll see if I manage to do so. I cannot promise (due to tight schedule).

qooban:
I'm going to change type names and polish the changes, so I'll notify you once the PR is ready for the reviews by sending proper requests.

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
kubama: > Is it to have something like:
> 
> ```
> export interface FilterRuleRow {
>     id: string;
>     rule: FilterRule;
> }
> ```
> 
> ?

Yes, this is what I meant.

The conversion logic is indeed simple, but it could be even simpler and at the same time more robust. The `FilterRuleRow` to `FilterRule` conversion could be then as simple as `filterRuleRow.filter` - in this case it wouldn't depend on what decorator properties are defined - it would work correctly without the need for being updated if more decorator properties than just `id` were added.

This is an optional/cosmetic thing.

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
kubama: Ok, I'm resolving this thread then.

widgets/common/src/filters/FiltersDefinitionForm.tsx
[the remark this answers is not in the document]
qooban: Ah, yeah, of course. Now I get it. I'll change it.

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
kubama: The value wouldn't be that great, but I think the solution wouldn't be complex either. Still, I'm leaving it up to you.

Gelio:
> I'm going to change type names and polish the changes, so I'll notify you once the PR is ready for the reviews by sending proper requests.

Alright then, I'll wait with the review in this case :slightly_smiling_face:

widgets/common/src/filters/RuleOperatorDropdown.tsx
kubama: Since it is non standard for me it was not clear at the first sight. Since we are dealing with an input component I was expecting to see `value` prop. The `operator` prop was confusing to me - as opposed to `value` my thinking was that it is a prop that is a rather static configuration option. "Why do I have to specify an operator if it's an operator selection input?" I ended up drilling down to the implementation to find out what's going on. Summing up, I'm still opting for the change.

qooban: Changed.

widgets/common/src/filters/RuleTypeDropdown.tsx
qooban: I didn't find anything and I think we can defer extracting it.

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
qooban: To be honest, I don't know with my current knowledge how to do it properly. If you provide me ready solution I can do it, but IMHO we can defer it to the later phase of filters development, to make it possible to finalize basic part till 13th Apr.

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
qooban: Done.

widgets/common/src/filters/RuleInput.tsx
qooban: I'll be modifying LabelValueDropdown in the next PRs (eg. RD-2006), so I think I can consider it then.

widgets/common/src/filters/RuleTypeDropdown.tsx
kubama: Sure 👍. Thanks for checking though.

widgets/common/src/filters/RuleInput.tsx
kubama: Ok, I'm resolving this thread then.

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
kubama: My try would be:
```
interface LabelFilterRule {
    key: string;
    values: string[];
    operator: LabelsRuleOperator;
    type: 'label';
}

interface AttributeFilterRule {
    key: 'blueprint_id' | 'site_name' | 'created_by';
    values: string[];
    operator: AttributesRuleOperator;
    type: 'attribute';
}

export type FilterRule = LabelFilterRule | AttributeFilterRule;
```

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -150,7 +151,12 @@ describe('Deployments View widget', () => {
         before(() => {
             cy.deleteDeploymentsFilter(filterId, { ignoreFailure: true })
                 .createDeploymentsFilter(filterId, [
-                    { type: 'label', key: 'precious', values: ['yes'], operator: 'any_of' }
+                    {
```
Gelio, line 154: Thanks :smile:

widgets/common/src/filters/types.ts
```diff
@@ -2,6 +2,46 @@
...
+export enum FilterRuleRowType {
+    Blueprint = 'blueprint_id',
+    SiteName = 'site_name',
+    Creator = 'created_by',
+    Label = 'label'
+}
+
+export const FilterRuleOperators = { ...CommonRuleOperator, ...LabelsOnlyRuleOperator, ...AttributesOnlyRuleOperator };
+export const LabelsFilterRuleOperators = { ...CommonRuleOperator, ...LabelsOnlyRuleOperator };
+export const AttributesFilterRuleOperators = { ...CommonRuleOperator, ...AttributesOnlyRuleOperator };
```
Gelio, line 47: Clever way to extend enums :smile:

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
Gelio: Good suggestion :+1: I was about to comment about the same thing. The implementation from @kubama looks solid :slightly_smiling_face: 

This feature is TS is called a discriminated union type (https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions), with the `type` property being the discriminator in this case :slightly_smiling_face:

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -28,10 +55,25 @@ const FiltersDefinitionForm: FunctionComponent = () => {
         setRows(latestRows => latestRows.filter(row => row.id !== id));
     }
 
+    function updateRule(id: string, newRule: FilterRule) {
+        setRows(latestRows => latestRows.map(row => (row.id === id ? { id: row.id, rule: newRule } : row)));
```
Gelio, line 59: Nice use of the updater function :+1:

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -3,22 +3,49 @@ import type { FunctionComponent } from 'react';
...
-const FiltersDefinitionForm: FunctionComponent = () => {
-    const { UnsafelyTypedForm: Form } = Stage.Basic;
-    const [rows, setRows] = useState(() => [getNewRow()]);
+function getFilterRule(filterRuleRow: FilterRuleRow): FilterRule {
+    return filterRuleRow.rule;
+}
+
+interface FiltersDefinitionFormProps {
+    initialFilters: FilterRule[];
+    onChange: (filterRules: FilterRule[]) => void;
```
Gelio, line 35: :tada: 

Very cool design :+1: Having the `id`s be handled internally sounds IMO like a very nice pattern :+1:

widgets/common/src/filters/FiltersDefinitionForm.tsx
```diff
@@ -28,10 +55,25 @@ const FiltersDefinitionForm: FunctionComponent = () => {
...
+
     return (
         <Form>
             {rows.map(row => (
-                <RuleRow key={row.id} removable={rows.length > 1} onRemove={() => removeRule(row.id)} />
+                <RuleRow
+                    key={row.id}
+                    rule={getFilterRule(row)}
+                    removable={rows.length > 1}
+                    onChange={(value: FilterRule) => updateRule(row.id, value)}
```
Gelio, line 73: Optional:

Since `onChange` is statically typed, TS can infer the type of `value`, so we can drop it, unless it's there for readability purposes :slightly_smiling_face: 

```suggestion
                    onChange={value => updateRule(row.id, value)}
```

Also, we could name the `value` a bit more concretely :slightly_smiling_face: 

```suggestion
                    onChange={rule => updateRule(row.id, rule)}
```

widgets/common/src/filters/RuleRowTypeDropdown.tsx
```diff
@@ -0,0 +1,39 @@
...
+    onChange: (value: FilterRuleRowType) => void;
+    value: FilterRuleRowType;
+}
+
+const ruleRowTypeOptions = (() => {
+    const { i18n } = Stage;
+    return Object.values(FilterRuleRowType).map(
+        (ruleType): DropdownItemProps => ({
+            text: i18n.t(`${i18nPrefix}.ruleTypesLabels.${camelCase(ruleType)}`),
+            value: ruleType as FilterRuleRowType
```
Gelio, line 18: Optional:

`value` accepts any `string`, so the type assertion (that tells TS that it's only one of the possible enum values) is unnecessary :slightly_smiling_face: 

```suggestion
            value: ruleType
```

widgets/common/src/filters/RuleOperatorDropdown.tsx
```diff
@@ -1,28 +1,52 @@
...
+    ruleType: FilterRuleRowType;
+    onChange: (value: FilterRuleOperator) => void;
+    value: FilterRuleOperator;
+}
+
+function getDropdownOptions(operators: string[]) {
+    const { i18n } = Stage;
+    return operators.map(
+        (operator): DropdownItemProps => ({
+            text: i18n.t(`${i18nPrefix}.operatorsLabels.${camelCase(operator as string)}`),
```
Gelio, line 19: Optional:

The type assertion here seems unnecessary to me since `operator` is already a `string` :slightly_smiling_face: 

```suggestion
            text: i18n.t(`${i18nPrefix}.operatorsLabels.${camelCase(operator)}`),
```

widgets/common/src/filters/RuleOperatorDropdown.tsx
```diff
@@ -1,28 +1,52 @@
...
-        value: operator
-    })
-);
-
-const RuleOperatorDropdown: FunctionComponent = () => {
+const RuleOperatorDropdown: FunctionComponent<RuleOperatorDropdownProps> = ({ ruleType, onChange, value }) => {
     const { Dropdown } = Stage.Basic;
 
-    return <Dropdown search selection name="ruleOperator" options={operatorsOptions} />;
+    const [options, setOptions] = useState([] as DropdownItemProps[]);
```
Gelio, line 28: Digression:

I've been always using the following:

```suggestion
    const [options, setOptions] = useState<DropdownItemProps[]>([]);
```

That handles a bit more options, e.g. if you do not provide the default `[]` value, then `options` will be `DropdownItemProps[] | undefined` :slightly_smiling_face: 

Both are valid, I just wanted to show another way to do the same thing, one that can save you some typing in the future :slightly_smiling_face:

widgets/common/src/filters/RuleOperatorDropdown.tsx
```diff
@@ -1,28 +1,52 @@
...
+
+    useEffect(() => {
+        const operators = Object.values(
+            ruleType === FilterRuleRowType.Label ? LabelsFilterRuleOperators : AttributesFilterRuleOperators
+        );
+        const dropdownOptions = getDropdownOptions(operators);
+
+        setOptions(dropdownOptions);
+    }, [ruleType]);
```
Gelio, line 37: This looks like a poor man's `useMemo` :smile: 

```suggestion
    const options = useMemo(() => {
        const operators = Object.values(
            ruleType === FilterRuleRowType.Label ? LabelsFilterRuleOperators : AttributesFilterRuleOperators
        );
        return getDropdownOptions(operators);
    }, [ruleType]);
```

widgets/common/src/filters/RuleValueInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const RuleValueInput: FunctionComponent<RuleOperatorDropdownProps> = ({
+    onChange,
+    operator,
+    ruleType,
+    toolbox,
+    values
+}) => {
+    const { Input } = Stage.Basic;
+    const keyDropdownRef = useRef();
```
Gelio, line 24: Optional:

We could specify the type of the `current` property:

```suggestion
    const keyDropdownRef = useRef<HTMLElement>();
```

widgets/common/src/filters/RuleValueInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+                    <LabelValueDropdown
+                        labelKey={labelKey}
+                        onChange={newValue => onChange([labelKey, newValue])}
+                        toolbox={toolbox}
+                        value={labelValue}
+                    />
+                )}
+            </>
+        );
+    }
```
Gelio, line 57: Optional:

We could probably extract this part into a separate component , which would most likely make it a bit more readable :smile:

Gelio (approved):
Looks good from my perspective :slightly_smiling_face: Solid work!

IMO all the comments can be addressed in the future, as they are minor :slightly_smiling_face:

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Thanks 🙂

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Thank you 🙂 

One of the intentions here was to make the parent able to work on the list of rules accepted by REST endpoint without any additional conversions.

widgets/common/src/filters/FiltersDefinitionForm.tsx
qooban: Make sense, I will change it according to your second suggestion.

widgets/common/src/filters/RuleRowTypeDropdown.tsx
qooban: True. Leftover from my original take to iterate over type values. I'll change that.

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: True. Leftover from my original take to iterate over type values. I'll change that.

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: Nice, I wasn't aware of it 👍🏻  Will change it.

widgets/common/src/filters/RuleValueInput.tsx
qooban: Right, will add it.

widgets/common/src/filters/RuleOperatorDropdown.tsx
qooban: Good idea, I'll change it.

widgets/common/src/filters/RuleValueInput.tsx
qooban: That's definitely good idea. I'll address it within RD-2006 or RD-2007.

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
qooban: Hmm... I tried it, but I failed on TS errors. I think they are coming from changing `interface FilterRule` into `type FilterRule` or something else.

I'd suggest to leave it for now and do that in the later phase of the filters development, WDYT?

test/cypress/integration/widgets/deployments_view_spec.ts
qooban: No prob 🙂

widgets/common/src/filters/types.ts
qooban: Thanks 🙂

qooban:
I think that PR is in a shape ready for your final reviews 🙂

widgets/common/src/filters/types.ts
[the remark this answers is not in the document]
kubama: OK

kubama (approved):

Gelio (approved):
Looks alright :ok_hand:

PR #1297, opened 2021-04-08
RD-1123 Added cypress test for filters management widget

kubama:
This PR includes `FilterEditModalProps` type fix as a follow-up to previous PR review.

This PR adds `cypress-get-table` cypress extension which makes table content validations easier. It's not perfect but it's extremely easy to use - it abstracts table structure by allowing to reference table cells by row number and column header. As such it encourages better, more thorough table content checking.

System tests run is queued (build 416).

### DIFF
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -1,4 +1,4 @@
-import { FilterRule } from '../../../../widgets/common/src/filters/types';
+import type { FilterRule } from '../../../../widgets/common/src/filters/types';
 
 describe('Filters widget', () => {
     before(() => {
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -1,0 +1,116 @@
+import { FilterRule } from '../../../../widgets/common/src/filters/types';
+
+describe('Filters widget', () => {
+    before(() => {
+        cy.usePageMock('filters').activate().mockLogin();
+    });
+
+    const filterName = 'filters_test_filter';
+    const filterRules: FilterRule[] = [
+        { type: 'attribute', key: 'blueprint_id', values: ['bpid'], operator: 'any_of' },
+        { type: 'label', key: 'precious', values: ['yes'], operator: 'any_of' }
+    ];
+
+    beforeEach(() => {
+        cy.deleteDeploymentsFilters(filterName).createDeploymentsFilter(filterName, filterRules).refreshPage();
+        cy.get('input[placeholder="Search..."]').type(filterName);
+        cy.get('.loading').should('not.exist');
+    });
+
+    it('should list existing filters', () => {
+        cy.get('table')
+            .getTable()
+            .should(tableData => {
+                expect(tableData).to.have.length(1);
+                expect(tableData[0]['Filter name']).to.eq(filterName);
+                expect(tableData[0].Creator).to.eq('admin');
+                expect(tableData[0].Created).not.to.be.null;
+            });
+    });
+
+    it('should allow to add new filter', () => {
+        cy.contains('Add').click();
+
+        cy.contains('Save').click();
+        cy.contains('Please provide the filter ID');
+
+        const newFilterName = `${filterName}_added`;
+        cy.contains('.field', 'Filter ID').find('input').type(newFilterName);
+        // TODO: RD-1985 Define some filter rules once rules form is available here
+        cy.interceptSp('PUT', `/filters/deployments/${newFilterName}`).as('createRequest');
+        cy.contains('Save').click();
+        cy.wait('@createRequest').then(({ request }) => {
+            expect(request.body).to.deep.equal({
+                filter_rules: []
+            });
+        });
+
+        cy.get('.modal').should('not.exist');
+        cy.contains(newFilterName);
+
+        cy.get('table')
+            .getTable()
+            .should(tableData => {
+                expect(tableData).to.have.length(2);
+                expect(tableData[1]['Filter name']).to.eq(newFilterName);
+                expect(tableData[1].Creator).to.eq('admin');
+                expect(tableData[1].Created).not.to.be.null;
+            });
+    });
+
+    it('should allow to edit existing filter', () => {
+        cy.get('.edit').click();
+        cy.contains(`Edit filter '${filterName}'`);
+
+        // TODO: RD-1985 Change some filter rules once rules form is available here
+        cy.interceptSp('PATCH', `/filters/deployments/${filterName}`).as('updateRequest');
+        cy.contains('Save').click();
+        cy.wait('@updateRequest').then(({ request }) => {
+            expect(request.body).to.deep.equal({
+                filter_rules: filterRules
+            });
+        });
+
+        cy.get('.modal').should('not.exist');
+    });
+
+    it('should allow to clone existing filter', () => {
+        cy.get('.clone').click();
+        cy.contains(`Clone filter '${filterName}'`);
+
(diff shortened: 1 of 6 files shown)


test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -0,0 +1,116 @@
+import { FilterRule } from '../../../../widgets/common/src/filters/types';
```
Gelio, line 1: Let's go for an `import type` when importing stuff from other top-level directories, just to be on the safe side :smile:

test/cypress/support/commands.ts
```diff
@@ -37,8 +38,10 @@ const getCommonHeaders = () => ({
 declare global {
     namespace Cypress {
         // NOTE: necessary for extending the Cypress API
-        // eslint-disable-next-line @typescript-eslint/no-empty-interface
-        export interface Chainable extends GetCypressChainableFromCommands<typeof commands> {}
+        export interface Chainable extends GetCypressChainableFromCommands<typeof commands> {
+            // typing for cypress-get-table
+            getTable: () => Cypress.Chainable;
```
Gelio, line 43: Optional:

We could enhance the type safety here by specifying the subject type:

```suggestion
            getTable: () => Cypress.Chainable<Record<string, any>[]>;
```

test/cypress/support/commands.ts
```diff
@@ -37,8 +38,10 @@ const getCommonHeaders = () => ({
 declare global {
     namespace Cypress {
         // NOTE: necessary for extending the Cypress API
-        // eslint-disable-next-line @typescript-eslint/no-empty-interface
-        export interface Chainable extends GetCypressChainableFromCommands<typeof commands> {}
+        export interface Chainable extends GetCypressChainableFromCommands<typeof commands> {
+            // typing for cypress-get-table
```
Gelio, line 42: Optional:

We could use a JSDoc comment here, so the comment appears whenever you hover over `getTable` in tests. We could also enhance the comment by adding a link to the library itself, making it easier to get to :slightly_smiling_face: 

```suggestion
            /**
             * Returns the table data
             *
             * @see https://www.npmjs.com/package/cypress-get-table
             */
```

image

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    const filterName = 'filters_test_filter';
+    const filterRules: FilterRule[] = [
+        { type: 'attribute', key: 'blueprint_id', values: ['bpid'], operator: 'any_of' },
+        { type: 'label', key: 'precious', values: ['yes'], operator: 'any_of' }
+    ];
+
+    beforeEach(() => {
+        cy.deleteDeploymentsFilters(filterName).createDeploymentsFilter(filterName, filterRules).refreshPage();
+        cy.get('input[placeholder="Search..."]').type(filterName);
+        cy.get('.loading').should('not.exist');
```
Gelio, line 17: Optoinal:

How about reusing https://github.com/cloudify-cosmo/cloudify-stage/blob/a95d27c7ac29d5589adf8ee2801da8a37015c4a0/test/cypress/support/commands.ts#L49-L52

```suggestion
        cy.waitUntilPageLoaded();
```

Gelio (commented):
Nicely done :ok_hand: I've left some rather small comments :slightly_smiling_face:

The `cypress-get-table` looks awesome :smile: I would have benefited from it writing tests for Environments, so I can see the value :+1: I like the addition

> System tests run is queued (build 416).

Thanks for running the system tests and providing the build number :slightly_smiling_face: A cherry on top would be to have the link to the system tests included directly in the PR, so one would not have to look for the job manually :smile:

Gelio:
I see you've also encountered RD-2010 in the system tests :worried:

kubama:
> A cherry on top would be to have the link to the system tests included directly in the PR, so one would not have to look for the job manually 😄

Unfortunately, when the job is queued then direct link to it returns 404, that's why I supplied build number instead. But if we are all aware of that then no problem - I'll provide a complete link next time, even if the build is queued at the time of writing.

Gelio:
> Unfortunately, when the job is queued then direct link to it returns 404, that's why I supplied build number instead. But if we are all aware of that then no problem - I'll provide a complete link next time, even if the build is queued at the time of writing.

I think that's ok :slightly_smiling_face: btw. links to jobs that are waiting in the queue are working fine in the Blue Ocean UI of Jenkins :smile: I have recently found out :slightly_smiling_face: So after queueing a job, you can switch to Blue Ocean UI and get a valid link, just FYI :slightly_smiling_face:

test/cypress/integration/widgets/filters_spec.ts
kubama: Right, done. Yet another point for implementing RD-2000 ASAP.

test/cypress/support/commands.ts
kubama: Cool, done.

test/cypress/integration/widgets/filters_spec.ts
kubama: Can't be done - `.loader` is the main, full screen spinner visible before the app is loaded, while here we are waiting for `.loading`, which is the small in-widget spinner.

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -0,0 +1,116 @@
...
+
+describe('Filters widget', () => {
+    before(() => {
+        cy.usePageMock('filters').activate().mockLogin();
+    });
+
+    const filterName = 'filters_test_filter';
+    const filterRules: FilterRule[] = [
+        { type: 'attribute', key: 'blueprint_id', values: ['bpid'], operator: 'any_of' },
+        { type: 'label', key: 'precious', values: ['yes'], operator: 'any_of' }
```
qooban, line 11: After https://github.com/cloudify-cosmo/cloudify-stage/pull/1296 it probably would need to be changed to use enums.

For example:
- `FilterRuleType.Attribute` instead of `attribute`
- `FilterRuleOperators.AnyOf` instead of `any_of`

With my current knowledge I cannot judge what is better - string or enums - but taking into account my experience with other languages I prefer having enum/const instead of strings.

qooban (approved):
`cypress-get-table` looks like a cool addition 👍🏻

test/cypress/integration/widgets/filters_spec.ts
kubama: Here we are dealing with string constants defined by external system (manager REST API). Enums abstracts actual string literals, so it's good as their values are out of our control and may change which in fact already happened recently (start_with was renamed to starts_with or so). So I think it's the way to go.
String literal unions are sufficient in cases where we are in control of the values. I consider them enums, but without the abstraction layer over their values.
Well, this is at least my current, probably simplistic understanding 🙂.

kubama:
> I see you've also encountered [RD-2010](https://cloudifysource.atlassian.net/browse/RD-2010) in the system tests 😟

If there is no timeline for fixing it I would still like to merge this PR, my work is having more and more in common with what's @qooban providing, so I'd like to keep integrating.

kubama:
> I see you've also encountered [RD-2010](https://cloudifysource.atlassian.net/browse/RD-2010) in the system tests 😟

If there is no timeline for fixing it I would still like to merge this PR, my work is having more and more in common with what's @qooban providing, so I'd like to keep integrating.

test/cypress/integration/widgets/filters_spec.ts
Gelio: Oh, I see :slightly_smiling_face: Thanks, I did not notice the different class name :sweat_smile:

Gelio (approved):
Looks good :+1:

PR #1298, opened 2021-04-09
RD-2002 Extract some shared code for Deployments View & TS refactors

Gelio:
In this PR you will find:
1. Extracting some code that will be shared between the top-level and the drilled-down Deployments View widgets to `widgets/common/src/deploymentsView`
2. Refactoring `Stage.Common` to be a namespace, and not an interface.

    Having it be a namespace allows creating nested namespaces, e.g. `Stage.Common.DeploymentsView.Types`, that can expose TS types. `Stage.Common` interface could only expose JS values, not types. I found it very limiting as I could not import the shared types from `widgets/common/src/deploymentsView/types.ts`

3. Refactoring some files that accidentally declare global variables (they had no imports/exports, so any variables defined in them were considered by TypeScript to be global). The change here was to add `export {};` for the file to be considered a module again.

I will raise a separate PR with the 2nd part of RD-2002

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/426/pipeline

Ran 2021-04-12 8:38 CEST


Gelio:
I have rebased the branch on `master` since the base branch was merged :slightly_smiling_face:

widgets/common/src/filters/RuleRow.tsx
```diff
@@ -6,7 +6,7 @@ import RuleRemoveButton from './RuleRemoveButton';
 
 interface RuleRowProps {
     removable: boolean;
-    onRemove: ComponentProps<typeof Button>['onClick'];
+    onRemove: ComponentProps<typeof RuleRemoveButton>['onClick'];
```
qooban, line 9: Thanks 🙂

widgets/composerLink/src/widget.jsx
```diff
@@ -17,6 +12,8 @@ Stage.defineWidget({
     categories: [Stage.GenericConfig.CATEGORY.BUTTONS_AND_FILTERS],
 
     render() {
+        const { Button } = Stage.Basic;
```
qooban, line 15: 👍🏻

widgets/deploymentsView/src/widget.tsx
```diff
@@ -1,29 +1,21 @@
 import { find } from 'lodash';
 import { FunctionComponent, useMemo, useState } from 'react';
 import { useQuery } from 'react-query';
+// NOTE: workaround for Cypress TS project to pick up the common types
+import type {} from '../../common/src/deploymentsView';
```
qooban, line 5: Oh, that's something new to me (`import type {}`). So that gets all exported types from `widgets/common/src/deploymentsView/types.ts`?

widgets/common/src/UploadBlueprintModal.jsx
```diff
@@ -1,6 +1,9 @@
 const { BlueprintActions } = Stage.Common;
 const i18nPrefix = 'widgets.common.blueprintUpload';
 
+// NOTE: prevents leaking variables as global in TypeScript
+export {};
```
qooban, line 5: Oh, so TypeScript automatically sees those consts defined above (`BlueprintActions`, `i18nPrefix`) as available globals unless we export something, right?

qooban (commented):
Looks good 👍🏻 

Added some minor comments, especially for me to understand TS better.

test/jest/widgets/common/DeploymentActions.test.ts
```diff
@@ -5,15 +5,15 @@ describe('(Widgets common) DeploymentActions', () => {
...
         Stage.Common = {
             PollHelper() {
                 this.wait = wait;
             },
             ExecutionActions() {
                 this.doGetExecutions = doGetExecutions;
             }
-        };
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        } as any;
```
kubama, line 16: Is there no other type we could use here to prevent need for eslint-disable? Like generic object type or so?

widgets/common/src/Consts.ts
```diff
@@ -22,11 +22,13 @@ const consts = {
...
-declare namespace Stage {
-    interface Common {
-        Consts: typeof consts;
+declare global {
+    namespace Stage.Common {
+        const Consts: typeof consts;
     }
 }
+// NOTE: prevents exposing `consts` as a global variable in TS
+export {};
```
kubama, line 31: Maybe there is eslint rule to catch files which are not modules?

widgets/common/src/Consts.ts
```diff
@@ -22,11 +22,13 @@ const consts = {
...
 
-declare namespace Stage {
-    interface Common {
-        Consts: typeof consts;
+declare global {
+    namespace Stage.Common {
+        const Consts: typeof consts;
     }
 }
+// NOTE: prevents exposing `consts` as a global variable in TS
```
kubama, line 30: Not sure about others, but I'm already aware what's the purpose of `export {}` - maybe we don't need this comment for each occurrence?

widgets/deploymentsView/src/widget.tsx
Gelio: `import type {}` is what it says on the tin - evaluate the file, but do not explicitly import anything :smile: While this may seem like a no-op, global types are affected.

We have to remember that this file is used in 2 projects:
1. The `widgets` project.
    No problem here, as the widgets project includes all the files in the `widgets` directory. `Stage.Common.DeploymentsView` is defined in one of the files in `widgets/common/src`, which is included in this project's compilation, so all is good
2. The `test/cypress` project.
    This is because I am importing the `DeploymentsViewWidgetConfiguration` in Cypress tests. This means that the cypress tests import this file, and try to do type-checking on it. However, there was nothing in this file that defined `Stage.Common.DeploymentsView`, and this caused to compilation to fail, as that was an unknown type :slightly_smiling_face: There was nothing that let the Cypress project know that it should import also the `common/src/deploymentsView` file to know that `Stage.Common.DeploymentsView` is defined. This imports adds that missing link :slightly_smiling_face: 

To sum up, the problem appears when this file is imported in the Cypress tests, as there is nothing in this file that would point to `widgets/common/src/deploymentsView` where `Stage.Common.DeploymentsView` is defined :smile: 

That would be solved by TS references, but that's the workaround for now

widgets/common/src/UploadBlueprintModal.jsx
Gelio: Yup, exactly :slightly_smiling_face: If there is no import/export in a file, TypeScript assumes that this file is a global script, meaning that it is evaluated in the global context where all declarations are available :slightly_smiling_face:

This is why you could e.g. use `Button` in any widget file and not have any TS error, because there was some file that had `const { Button } = Stage;` :smile: 

I fixed all the TS files that polluted the global namespace (the `isolatedModules` TS compiler flag guards it), but we still have JS files that define `const`s which TS treats as available globally :worried:

test/jest/widgets/common/DeploymentActions.test.ts
Gelio: I took some time to explore it, but did not find any better solution that does not involve `any` :worried: If you find a solution, let me know and I'll be happy to use it here

widgets/common/src/Consts.ts
Gelio: There is a TS compiler flag - `isolatedModules`. I enabled it in this PR, so TS files will have to be modules, and you will get compilation errors otherwise :slightly_smiling_face: 

Unfortunately, JS files can still be treated as regular scripts, and since `checkJs` is not enabled, we have to be careful. For example, I can use `UploadLabels` in widgets code without explicitly importing it even though it's not a global variable:
image

widgets/common/src/Consts.ts
Gelio: Ok, makes sense :slightly_smiling_face: It is ok if I leave these be, and not insert any more in the future?

widgets/common/src/Consts.ts
kubama: Cool, thanks.

widgets/common/src/Consts.ts
kubama: Sure thing - resolving this thread.

kubama (approved):

widgets/common/src/UploadBlueprintModal.jsx
qooban: Got it, thank you for the explanation 👍🏻

widgets/deploymentsView/src/widget.tsx
qooban: Got it, thank you for the explanation 👍🏻

widgets/common/src/filters/RuleRow.tsx
qooban: There was a small conflict here (after https://github.com/cloudify-cosmo/cloudify-stage/pull/1296 was merged to master), so I resolved it. Hope you don't mind 🙂

qooban (approved):

widgets/common/src/filters/RuleRow.tsx
Gelio: Not at all! Thank you for doing that, this saves me a few minutes :+1: :smile:

Gelio:
Thanks for the reviews and @qooban for resolving the merge conflict for me :smile: The nightly CI seems to be functional, so I'll run the system tests:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/426/pipeline

I'll merge the PR after the tests (hopefully) pass

Gelio:
The only failure was code coverage (76.9% < required 77%). I assume we can accept it on a similar note to RD-1999 (and this PR only moves stuff around, changing no functionality)

I'll merge the PR :slightly_smiling_face:

PR #1303, opened 2021-04-12
RD-2002 Add drilled-down Deployments View widget

Gelio:
The PR adds a new `deploymentsViewDrilledDown` widget that is rendered in the drilled-down pages of Environments. The changes should be transparent to the user - there are no functional changes. The new widget is very similar to the existing widget, but:
1. Has limited configuration
2. Always filters by parent deployment
3. Does not fetch filter rules (they will come from the context in RD-2004)

Despite having a large number of changes on paper, most of the code for `widgets/common/src/deploymentsView/DeploymentsView.tsx` comes from the preexisting widget code. You may want to use [the git's `--color-moved` config](https://github.blog/2018-04-05-git-217-released/#coloring-moved-code) to reduce the diff and show which lines were moved as-is :smile: I'd say around 200 lines were moved, and not added/deleted :slightly_smiling_face: 

I have also found a way to expose `FilterRule` from `widgets/common/src/filters` as `Stage.Common.Filters.Rule`, which IMO is a better way to use that type than importing the type using a long `../..` path :slightly_smiling_face:

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/436/pipeline

Ran 2021-04-13 9:35 CEST

### DIFF
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -73,8 +73,8 @@ describe('Deployments View widget', () => {
         cy.wait('@deployments');
     };
 
-    // NOTE: matches both deploymentsViewWidget and deploymentsViewDrilledDownWidget
-    const getDeploymentsViewWidget = () => cy.get('.widget').filter('[class*="deploymentsView"]').find('.widgetItem');
+    const getDeploymentsViewWidget = () =>
+        cy.get('.widget').filter('.deploymentsViewWidget, .deploymentsViewDrilledDownWidget').find('.widgetItem');
     const getDeploymentsViewTable = () => getDeploymentsViewWidget().get('.gridTable');
     const getDeploymentsViewDetailsPane = () => getDeploymentsViewWidget().get('.detailsPane');
 
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -34,7 +34,8 @@ describe('Deployments View widget', () => {
         'blueprintSources',
         'nodes',
         'executions',
-        'deploymentActionButtons'
+        'deploymentActionButtons',
+        'deploymentsViewDrilledDown'
     ];
     /** Column numbers as they appear in the table */
     const columnNumbers = {
@@ -72,7 +73,8 @@ describe('Deployments View widget', () => {
         cy.wait('@deployments');
     };
 
-    const getDeploymentsViewWidget = () => cy.get('.deploymentsViewWidget .widgetItem');
+    // NOTE: matches both deploymentsViewWidget and deploymentsViewDrilledDownWidget
+    const getDeploymentsViewWidget = () => cy.get('.widget').filter('[class*="deploymentsView"]').find('.widgetItem');
     const getDeploymentsViewTable = () => getDeploymentsViewWidget().get('.gridTable');
     const getDeploymentsViewDetailsPane = () => getDeploymentsViewWidget().get('.detailsPane');
 
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -1,4 +1,3 @@
-import type { FilterRule } from '../../../../widgets/common/src/filters/types';
 import { FilterRuleOperators, FilterRuleType } from '../../../../widgets/common/src/filters/types';
 
 describe('Filters widget', () => {
@@ -7,7 +6,7 @@ describe('Filters widget', () => {
     });
 
     const filterName = 'filters_test_filter';
-    const filterRules: FilterRule[] = [
+    const filterRules: Stage.Common.Filters.Rule[] = [
         { type: FilterRuleType.Attribute, key: 'blueprint_id', values: ['bpid'], operator: FilterRuleOperators.AnyOf },
         { type: FilterRuleType.Label, key: 'precious', values: ['yes'], operator: FilterRuleOperators.AnyOf }
     ];
--- a/test/cypress/support/filters.ts
+++ b/test/cypress/support/filters.ts
@@ -1,5 +1,4 @@
 import { addCommands, GetCypressChainableFromCommands } from 'cloudify-ui-common/cypress/support';
-import type { FilterRule } from '../../../widgets/common/src/filters/types';
 import { waitUntilEmpty } from './resource_commons';
 
 declare global {
@@ -11,7 +10,7 @@ declare global {
 }
 
 const commands = {
-    createDeploymentsFilter: (id: string, rules: FilterRule[]) =>
+    createDeploymentsFilter: (id: string, rules: Stage.Common.Filters.Rule[]) =>
         cy.cfyRequest(`/filters/deployments/${id}`, 'PUT', null, { filter_rules: rules }),
 
     deleteDeploymentsFilter: (filterId: string, { ignoreFailure }: { ignoreFailure?: boolean } = {}) =>
--- a/test/cypress/tsconfig.json
+++ b/test/cypress/tsconfig.json
@@ -14,6 +14,7 @@
         "sourceMap": false,
         "types": ["cypress"]
     },
-    "include": ["**/*"],
+    // Some Stage.Common types used in tests are defined in widgets/common
+    "include": ["**/*", "../../widgets/common/**/*"],
     "files": ["../../app/typings/stage-api.d.ts"]
 }
--- a/widgets/common/src/deploymentsView/DeploymentsView.tsx
+++ b/widgets/common/src/deploymentsView/DeploymentsView.tsx
@@ -6,6 +6,7 @@ import { i18nMessagesPrefix } from './common';
 import type { SharedDeploymentsViewWidgetConfiguration } from './configuration';
 import DetailsPane from './detailsPane';
 import { DeploymentsTable } from './table';
+import { FilterRuleOperators, FilterRuleType } from '../filters/types';
 
 export interface DeploymentsViewProps {
     widget: Stage.Types.Widget<SharedDeploymentsViewWidgetConfiguration>;
@@ -123,11 +124,11 @@ const useFilteringByParentDeployment = ({ filterByParentDeployment }: { filterBy
     return {
         filterable: true,
         parentDeploymentRule: {
-            type: 'label',
+            type: FilterRuleType.Label,
             key: 'csys-obj-parent',
-            operator: 'any_of',
+            operator: FilterRuleOperators.AnyOf,
             values: [parentDeploymentId]
-        }
+        } as Stage.Common.Filters.Rule
     } as const;
 };
 
--- a/widgets/common/src/deploymentsView/DeploymentsView.tsx
+++ b/widgets/common/src/deploymentsView/DeploymentsView.tsx
@@ -107,15 +107,7 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
 
 const useFilteringByParentDeployment = ({ filterByParentDeployment }: { filterByParentDeployment: boolean }) => {
     const drilldownContext = ReactRedux.useSelector((state: Stage.Types.ReduxState) => state.drilldownContext);
-    const parentDeploymentId = useMemo(() => {
-        if (drilldownContext.length < 2) {
-            return undefined;
-        }
-
-        const parentPageContext = drilldownContext[drilldownContext.length - 2].context;
-
-        return parentPageContext?.deploymentId as string | undefined;
-    }, [drilldownContext]);
+    const parentDeploymentId = useMemo(() => getParentDeploymentId(drilldownContext), [drilldownContext]);
 
     if (!filterByParentDeployment) {
         return { filterable: true } as const;
@@ -137,4 +129,22 @@ const useFilteringByParentDeployment = ({ filterByParentDeployment }: { filterBy
             values: [parentDeploymentId]
         }
     } as const;
+};
+
+const getParentDeploymentId = (drilldownContext: Stage.Types.ReduxState['drilldownContext']) => {
+    /**
+     * NOTE: drilldownContext contains contexts for each page, starting with
+     * the top-level page, and ending with the current page's initial context.
+     *
+     * Thus, the current page's parent context will be the penultimate entry in the array.
+     *
+     * It may happen, that the array only contains a single item (if we are on the top-level page).
+     */
+    if (drilldownContext.length < 2) {
+        return undefined;
+    }
+
+    const parentPageContext = drilldownContext[drilldownContext.length - 2].context;
+
+    return parentPageContext?.deploymentId as string | undefined;
 };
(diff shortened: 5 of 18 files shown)


test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -1,4 +1,3 @@
-import type { FilterRule } from '../../../../widgets/common/src/filters/types';
 import { FilterRuleOperators, FilterRuleType } from '../../../../widgets/common/src/filters/types';
```
Gelio, line 1: Using types from `Stage` works fine, but objects that are compiled to JS (e.g. enums) are not automatically made global (`defineCommon` does not make them global automatically :worried: ), so I could not easily expose these enums as globals and use in the tests

test/cypress/tsconfig.json
```diff
@@ -14,6 +14,7 @@
         "sourceMap": false,
         "types": ["cypress"]
     },
-    "include": ["**/*"],
+    // Some Stage.Common types used in tests are defined in widgets/common
+    "include": ["**/*", "../../widgets/common/**/*"],
```
Gelio, line 18: This allows us to delete the hacks of using `import type {}` just to get the common global types visible in the `test/cypress` project

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -0,0 +1,140 @@
+import React, { FunctionComponent, useMemo, useState } from 'react';
```
Gelio, line 1: Much of the code in this file is moved from `deploymentsView/src/widget.tsx`

widgets/common/src/deploymentsView/configuration.ts
```diff
@@ -0,0 +1,49 @@
+import { i18nPrefix } from './common';
```
Gelio, line 1: Much of the code in this file is moved from `deploymentsView/src/widget.tsx`

widgets/common/src/deploymentsView/index.ts
```diff
@@ -1,27 +1,35 @@
...
 
 declare global {
-    namespace Stage.Common.DeploymentsView {
-        const Table: typeof DTable;
-        const Common: typeof DCommon;
-        const DetailsPane: typeof DDetailsPane;
-        namespace Types {
-            export { Deployment, DeploymentsResponse, SharedDeploymentsViewWidgetConfiguration };
+    namespace Stage.Common {
+        // eslint-disable-next-line @typescript-eslint/no-namespace
```
Gelio, line 9: Perhaps it makes sense to disable the `no-namespace` rule altogether, as it results in false-positives in the `declare global` context :worried:

Gelio:
The system tests failed due to RD-2026 (an issue with the tests setup)

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -72,7 +73,8 @@ describe('Deployments View widget', () => {
         cy.wait('@deployments');
     };
 
-    const getDeploymentsViewWidget = () => cy.get('.deploymentsViewWidget .widgetItem');
+    // NOTE: matches both deploymentsViewWidget and deploymentsViewDrilledDownWidget
+    const getDeploymentsViewWidget = () => cy.get('.widget').filter('[class*="deploymentsView"]').find('.widgetItem');
```
kubama, line 77: Wouldn't `'.deploymentsViewWidget, .deploymentsViewDrilledDownWidget'` work? This way the code would be self-descriptive without the need for the comment.

widgets/common/src/filters/index.ts
```diff
@@ -8,6 +9,10 @@ const Filters = {
 const FiltersAlias = Filters;
 declare global {
     namespace Stage.Common {
+        // eslint-disable-next-line @typescript-eslint/no-namespace
```
kubama, line 12: Looks like we are disabling this rule frequently. Candidate for global disable?

widgets/deploymentsView/src/widget.tsx
```diff
@@ -149,69 +73,20 @@ const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({ widget, tool
...
-    }
-    if (deploymentsResult.isError) {
-        return (
-            <ErrorMessage
-                header={i18n.t(`${i18nMessagesPrefix}.errorLoadingDeployments`)}
-                error={deploymentsResult.error as { message: string }}
-            />
+    if (filterRulesResult.isIdle) {
+        throw new Error(
+            'Fetching filter rules is disabled without explicitly handling this scenario. Those rules should always be fetched.'
```
kubama, line 78: I have no idea what this message means 🙂.

widgets/common/src/filters/index.ts
Gelio: Yup, I've added a point about it to the weekly :slightly_smiling_face:

widgets/deploymentsView/src/widget.tsx
Gelio: The problem that this addresses comes from the fact TS assumes that `filterRulesResult.data` can be `undefined` (and there's an error on line 88), because we have only checked for the `isLoading` and `isError` statuses. There is also `isSuccess` and `isIdle`. Since I added this check for `isIdle`, TS knows that the only possible status after this `if` is `success`, and `filterRulesResult.data` must be an array. 

Thus, I decided to check for `isIdle`, instead of doing a type assertion `as FilterRule[]` on line 88, since that will provide a more descriptive message (this one you're commenting on)

I don't have certainty whether you are aware of what react-query's `isIdle` is doing, so just to be on the same page, I'll assume not and mention the docs :slightly_smiling_face:  

As it says in https://react-query.tanstack.com/reference/useQuery, 
image

Given the explanation from the docs, I came up with this error message (if this error occurs, then someone disabled fetching the query and did not handle how the UI should behave then) :slightly_smiling_face: If you have a better way to handle it, let me know

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Thanks for the tip, seems to work fine :slightly_smiling_face: Done

test/cypress/support/filters.ts
```diff
@@ -11,7 +10,7 @@ declare global {
 }
 
 const commands = {
-    createDeploymentsFilter: (id: string, rules: FilterRule[]) =>
+    createDeploymentsFilter: (id: string, rules: Stage.Common.Filters.Rule[]) =>
```
qooban, line 13: That seems useful (having access to types from `Stage.Common` in cypress tests).

widgets/common/src/deploymentsView/configuration.ts
```diff
@@ -0,0 +1,49 @@
...
+        name: Stage.i18n.t(`${i18nPrefix}.configuration.fieldsToShow.name`),
+        placeHolder: Stage.i18n.t(`${i18nPrefix}.configuration.fieldsToShow.placeholder`),
+        items: deploymentsViewColumnIds.map(columnId => ({
+            name: deploymentsViewColumnDefinitions[columnId].name,
+            value: columnId
+        })),
+        default: deploymentsViewColumnIds.filter(columnId => columnId !== 'environmentType'),
+        type: Stage.Basic.GenericField.MULTI_SELECT_LIST_TYPE
+    },
+    Stage.GenericConfig.PAGE_SIZE_CONFIG(100),
```
qooban, line 31: Is it intentional to have the page size set to such big number (in comparison to what we had till now in tabular widgets - 5 or 10 and IIRC max. 20 items)?

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -0,0 +1,140 @@
...
+        {
+            enabled: filteringByParentDeploymentResult.filterable,
+            onSuccess: data => {
+                const context = toolbox.getContext();
+                // TODO(RD-1830): detect if deploymentId is not present in the current page and reset it.
+                // Do that only if `fetchData` was called from `DataTable`. If it's just polling,
+                // then don't reset it (because user may be interacting with some other component)
+                if (context.getValue('deploymentId') === undefined && data.items.length > 0) {
+                    context.setValue('deploymentId', data.items[0].id);
+                }
```
qooban, line 54: Are we fully safe with these extractions: `data.items.length` and `data.items[0].id`?

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const drilldownContext = ReactRedux.useSelector((state: Stage.Types.ReduxState) => state.drilldownContext);
+    const parentDeploymentId = useMemo(() => {
+        if (drilldownContext.length < 2) {
+            return undefined;
+        }
+
+        const parentPageContext = drilldownContext[drilldownContext.length - 2].context;
+
+        return parentPageContext?.deploymentId as string | undefined;
+    }, [drilldownContext]);
```
qooban, line 118: WDYT about extracting this magical `2` used in relation to `drilldownContext.length`? Having a const and/or provide an explanation for it, would be useful for the code readers IMO 🙂

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    return {
+        filterable: true,
+        parentDeploymentRule: {
+            type: 'label',
+            key: 'csys-obj-parent',
+            operator: 'any_of',
+            values: [parentDeploymentId]
+        }
```
qooban, line 138: I'm wondering shouldn't we have a const/function for getting such rule. But you have better context for that - is it the only place where we use that rule?

Also, WDYT about using proper already defined consts for `type` and `operator` values?

qooban (commented):
The change looks pretty good 👍🏻 
I like the way you allowed using `Stage.Common` in Cypress.
Added just few minor comments.

test/cypress/support/filters.ts
Gelio: :smile: I'm glad I found a way that is acceptable :smile:

widgets/common/src/deploymentsView/configuration.ts
Gelio: Yes, that's a requirement from Ofer.

In https://cloudifysource.atlassian.net/wiki/spaces/WR/pages/1703280709/Environments#Environments-page---Tabular-view, there's 

> The environments list should display a large number of items per page

and in the email "Re: Environments UI story list draft", there's:

> The list should include paging. The default number of items per page will be 50. Options should include (10, 20, 50, 100)

Now that I double-checked, looks like 50 should be the default value, so I'll change it to 50 :slightly_smiling_face:

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: `data.items[0].id` is safe if `data.items.length` is safe, and I have not seen a response where `data.items` was not present or was not an array :slightly_smiling_face:

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: Good point :slightly_smiling_face: I'll adjust the code here

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: This is the only place, so I don't see a huge benefit in extracting it to a function/const, but using the consts makes sense - I'll try to do that :slightly_smiling_face:

widgets/common/src/deploymentsView/configuration.ts
Gelio: I've changed the number to 50 :slightly_smiling_face:

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: I've extracted the logic to a separate function and added an explanatory comment :slightly_smiling_face: Let me know if that's clear enough and if not, what change would be helpful

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: I've used the enums :slightly_smiling_face:

widgets/deploymentsView/src/widget.tsx
kubama: Thanks for a thorough explanation. Not only I know the context now, but also learned about react-query 👍.
It's a cosmetic thing, however I can imagine that seeing this error in the console (or seeing this code without knowing the context) would still leave me confused - what is the referred "this scenario"? I would change the message to something more standard like "Idle query state handling not implemented". Further explanation (the reason why it's not implemented) could be given in a comment.

Gelio: No problem :slightly_smiling_face: I'm glad I could help :smile: 

Makes sense, I'll add some more information in the comment and change the error message

Gelio: I've added a comment and reworded the error message, let me know what you think :slightly_smiling_face:

kubama: It's perfect now 👍. Thanks.

kubama (approved):

widgets/common/src/deploymentsView/DeploymentsView.tsx
qooban: 👍🏻

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -0,0 +1,151 @@
...
+     * It may happen, that the array only contains a single item (if we are on the top-level page).
+     */
+    if (drilldownContext.length < 2) {
+        return undefined;
+    }
+
+    const parentPageContext = drilldownContext[drilldownContext.length - 2].context;
+
+    return parentPageContext?.deploymentId as string | undefined;
+};
```
qooban, line 151: Optional:

Thanks for extracting that. I originally thought about having a const for `2`. The comment: "Thus, the current page's parent context will be the penultimate entry in the array." seems to fill the need.

Still I would consider having something like:
```
const minDrilldownContextLength = 2;
```

qooban (approved):

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: I didn't find a suitable name for the const, though :thinking: `minDrilldownContextLength` is not really true, since it can contain a single item, and it's still a valid `drilldownContext`. 2 in this case is the number of items in the case of being a drilled-down page. `minDrilldownContextLengthInDrilldownPage` sounds like a mouthful :slightly_smiling_face: I hope the current solution will suffice, as I will merge the PR without spending more time on it to unblock other PRs

Gelio:
Thanks for reviews :slightly_smiling_face: The system tests failed only due to coverage (https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/436/pipeline), so I'll merge the PR as soon as the multibranch job passes

PR #1304, opened 2021-04-12
RD-1758 / RD-2012 & RD-2006 - Extend support for type: 'label'

qooban:
### Changes
* RD-2012 Adapted `KeyDropdown` and `ValueDropdown` to allow only selection and not text input (read-only mode)
* RD-2006 Adapted `ValueDropdown` to allow multiple selection
* Adapted `LabelValueInput` component to use `readOnly` and `multiple` props from `KeyDropdown` and `ValueDropdown` components

### Video
https://user-images.githubusercontent.com/5202105/114397998-92353d00-9b9f-11eb-8682-06a6cacd403a.mp4

### Merge strategy

This PR builds on top of https://github.com/cloudify-cosmo/cloudify-stage/pull/1302 so we need to have the base merged first, then I'll rebase this one to master.


kubama (approved):

widgets/common/src/filters/inputs/LabelValueInput.tsx
```diff
@@ -23,17 +23,23 @@ const LabelValueInput: FunctionComponent<LabelValueInputProps> = ({
...
-        // TODO(RD-2012): Adapt label key and value dropdowns to disallow free text
         // TODO(RD-2007): Add better styling
         <>
-            <LabelKeyDropdown innerRef={keyDropdownRef} onChange={onKeyChange} toolbox={toolbox} value={labelKey} />
+            <LabelKeyDropdown
+                innerRef={keyDropdownRef}
+                onChange={onKeyChange}
+                toolbox={toolbox}
+                value={labelKey}
+                readOnly
```
Gelio, line 33: In my opinion, `readOnly` is misleading :smile: As mentioned in [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly), `readonly` for HTML elements prevents mutating the field.

In this case, the user can mutate the field (add/remove values). They just cannot add entirely new values. Thus, IMO, a more suitable name for the prop would be `onlyKnownLabels` or `preventCreatingLabels`

The same goes for the `LabelValueDropdown` component :slightly_smiling_face: 

Moreover, the property keys used in `en.json` should also be adjusted, since `readOnly` and `readWrite` may be misleading there too (I suggest `readOnly` -> `onlyKnown` and `readWrite` -> `knownOrNew`)

widgets/common/src/labels/CommonDropdown.jsx
```diff
@@ -11,21 +11,26 @@ function useDebouncedSetValue(value, setValue, deps) {
...
+    const {
+        inputValue,
+        invalidCharacterTyped,
+        submitChange,
+        resetInput,
+        unsetInvalidCharacterTyped
+    } = useLabelInput(onChange, { readOnly });
 
-    useDebouncedSetValue(baseFetchUrl ? addSearchToUrl(baseFetchUrl, inputValue) : '', setFetchUrl, [
+    useDebouncedSetValue(readOnly ? baseFetchUrl : addSearchToUrl(baseFetchUrl, inputValue), setFetchUrl, [
```
Gelio, line 30: General:

Will the autocomplete still work? If the user has a lot of labels, do we fetch all of them in the initial request and then do UI filtering based on what the user typed to narrow-down the search results that appear in the dropdown?

I didn't check, but if we only fetch the first page of labels, then the autocomplete may be broken :worried:

Gelio (commented):
Clever changes, looks like small changes can yield nice features in this case :+1:

qooban:
@Gelio @kubama Thanks for your reviews. I'll address the comments tomorrow morning and ask you then for re-review.

widgets/common/src/filters/inputs/LabelValueInput.tsx
qooban: I got your point. I actually was expecting a comment in this area as I was not sure about the name I used. I should have created author's comment.

Anyways, before I introduce a change, I'd like to discuss the actual prop name. I think the better would be `allowKnownOnly` instead of `onlyKnownLabels` (which in my understanding refers more to a label which is key-value pair, not a single key or single value). It will fit in all cases: `KeyDropdown`, `ValueDropdown`, `CommonDropdown` components and `useLabelInput` hook. I like props named `allow*` as they are commonly used in Semantic-UI-React as well as in our app, so I think they should be recognized easily. What do you think?

Gelio: `allowKnownOnly` makes sense :+1:

qooban: Changed.

widgets/common/src/labels/CommonDropdown.jsx
qooban: Yeah, it will not work. I need to re-think auto-completion feature in this case. Will write once I have proper solution.

qooban: I created separate PR for handling auto-completion: https://github.com/cloudify-cosmo/cloudify-stage/pull/1311 , so I think we can resolve this conversation and continue in the new PR.

qooban:
Except rebasing to master branch I introduced two changes:

- https://github.com/cloudify-cosmo/cloudify-stage/pull/1304/commits/7e9eb4c26bf6f0475dda3b3368a9b50481671b71
- https://github.com/cloudify-cosmo/cloudify-stage/pull/1304/commits/25ed5a8e8f0ede451ed75b7b0d985d93b441a35b

kubama (approved):

widgets/common/src/labels/CommonDropdown.jsx
Gelio: Sure, thanks :+1:

Gelio (approved):

qooban:
I scheduled Stage system tests execution for this one:
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/443
but I see we've got quite long queue, so it won't be soon.

@kubama I think I won't be available before tests are finished. If you need this PR earlier than tomorrow, feel free to merge it once tests are fine.
@Gelio I'll continue working on the auto-completion support as part of https://github.com/cloudify-cosmo/cloudify-stage/pull/1311

qooban:
Or actually I think I can merge it right away. Tested locally:
image
and there's very little chance that this PR affects any other tests.

@Gelio @kubama Let me know if you have any objections.

Gelio:
If @kubama needs this code today, let's merge it today :slightly_smiling_face: If not, I would rather not create precedence for merging without running tests :smile: If we're not in dire straits and if it's not critical, I suggest sticking to the rules

qooban:
I see job execution has just started:
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/443/console
and failed due to: `No space left on device` 
I won't be in soon, so it's up to you when you merge it.

> If we're not in dire straits and if it's not critical, I suggest sticking to the rules

🙂 Maybe I missed something, but do we have any rule about running system tests? I thought it's optional and only a good practice, rather than strict rule.

Gelio:
> I thought it's optional and only a good practice, rather than strict rule.

You are right, not a strict rule :slightly_smiling_face: I misremembered the time we were initially discussing it

qooban:
OK. As long as we don't have stable system test environment (we all see how many issues we can encounter) I will be against having strict rule for running system tests before merging PRs.

Another run scheduled: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/447/

As I already said - I'm OK with merging it before getting final results as there is very little chance this change affects anything other that `deployment_actions_spec`. Do as you wish. If it won't be merged today, I'll probably merge it tomorrow morning.

kubama:
System tests passed, build failed due to test coverage - I'm merging it so I could proceed with system tests on my dev branch.

PR #1295, opened 2021-04-08
RD-2001 Add core drill-down functionality to Deployments View

Gelio:
This PRs is adding support for the basic drill-down functionalities in the Deployments View widget.

This is a partial implementation. The drill-down functionality itself is working, but there is still work to be done to get it working correctly and according to the designs. Take a look at the subtasks for RD-1226 for a list of upcoming tasks.

## Video

https://user-images.githubusercontent.com/889383/114018951-1f118b00-986e-11eb-894c-ca56f6be45e4.mp4

The structure of the deployments in the video is the same as in the tests:

```
        //                app-env
        //                  - -
        //                -/   \-
        //               /       \-
        //             -/          \-
        //            /              \
        //         db-env           web-app
        //          ---
        //        -/   \-
        //      -/       \
        //   db-1        db-2
```

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/417/pipeline

Ran 2021-04-08 15:49 CEST

PR #1305, opened 2021-04-12
RD-2004 Filter deployments shown in drill-down view

Gelio:
This PR adds filtering the deployments by their type (service or environment) according to which drill-down button was clicked in the parent page.

Internally, the parent page injects `filterRules` with the correct `csys-obj-type` rule into the drilled-down page's context.

Additionally, the drilled-down widget will shown a more descriptive error message when it's used on a top-level page (unexpectedly) :slightly_smiling_face: 

## Video

https://user-images.githubusercontent.com/889383/114423295-54dca980-9bb7-11eb-84cc-5c5f724317c6.mp4

The numbers on the drill-down buttons are incorrect and will be addressed in RD-2003

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/445/pipeline

Ran 2021-04-13 14:56 CEST


widgets/deploymentsViewDrilledDown/src/widget.tsx
```diff
@@ -18,17 +18,58 @@ Stage.defineWidget<never, never, DrilledDownWidgetConfiguration>({
     initialConfiguration: sharedConfiguration,
 
     render(widget, _data, _error, toolbox) {
-        // TODO(RD-2004): get filter rules from context
+        return <DrilledDownDeploymentsViewWidget widget={widget} toolbox={toolbox} />;
```
Gelio, line 21: I had to extract a component to be able to use the `useSelector` hook

kubama (approved):

Gelio:
I had to add one more commit after discussing a problem with filtering with Ofer, Yoni, Mohammed, and Alex. Copying the commit message with explanation:


    Due to a limitation of the filtering API, Services are expected to
    always have a `csys-obj-type=service` label.

    This is because previously, if services optionally have this label,
    there is no way to filter out deployments that have
    `csys-obj-type=service` OR `csys-obj-type is null`. OR operations are
    not allowed in the filtering API.

    Relevant backend changes to mark deployments as services by default will
    be done in the near future.

widgets/common/src/deploymentsView/detailsPane/DrilldownButtons.tsx
```diff
@@ -60,10 +60,17 @@ const DrilldownButton: FunctionComponent<DrilldownButtonProps> = ({
     const icon = type === 'services' ? subservicesIcon : subenvironmentsIcon;
 
     const drilldownToSubdeployments = () => {
+        const deploymentTypeRule: Stage.Common.Filters.Rule = {
+            // NOTE: cannot use enums as they are not exported
+            type: 'label' as any,
+            key: 'csys-obj-type',
+            operator: 'any_of' as any,
```
qooban, line 67: You can export it, I think, as it seems there is use case here.

widgets/deploymentsViewDrilledDown/src/widget.tsx
```diff
@@ -18,17 +18,58 @@ Stage.defineWidget<never, never, DrilledDownWidgetConfiguration>({
...
+}) => {
+    const {
+        DeploymentsView,
+        Common: { i18nMessagesPrefix, filterRulesContextKey }
+    } = Stage.Common.DeploymentsView;
+    const filterRules: Stage.Common.Filters.Rule[] | undefined = toolbox.getContext().getValue(filterRulesContextKey);
+    const { ErrorMessage } = Stage.Basic;
+
+    const drilldownContext = ReactRedux.useSelector((state: Stage.Types.ReduxState) => state.drilldownContext);
+    const isTopLevelPage = drilldownContext.length < 2;
```
qooban, line 42: That's probably another place where the const for magic `2` would be useful 🙂 

Ref.: https://github.com/cloudify-cosmo/cloudify-stage/pull/1303#discussion_r612270518

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -331,9 +333,9 @@ describe('Deployments View widget', () => {
         const deployments: Record<DrilldownDeploymentName, SystemLabel[]> = {
             'app-env': [{ 'csys-obj-type': 'environment' }],
             'db-env': [{ 'csys-obj-type': 'environment' }, { 'csys-obj-parent': getDeploymentFullName('app-env') }],
-            'db-1': [{ 'csys-obj-parent': getDeploymentFullName('db-env') }],
-            'db-2': [{ 'csys-obj-parent': getDeploymentFullName('db-env') }],
-            'web-app': [{ 'csys-obj-parent': getDeploymentFullName('app-env') }]
+            'db-1': [{ 'csys-obj-type': 'service' }, { 'csys-obj-parent': getDeploymentFullName('db-env') }],
```
qooban, line 336: As I see now in the tests we have `SystemLabel` type defined, so we are safe with providing strings like "csys-obj-type".

Do we have something like that in the source code?

qooban (commented):
Looks good. Added few comments.

Regarding:
> This is because previously, if services optionally have this label,
there is no way to filter out deployments that have
`csys-obj-type=service` OR `csys-obj-type is null`. OR operations are
not allowed in the filtering API.

as far as I understand there is OR operation available in filters, but probably not with `null` as one of the accepted values?

You can have a rule:
```
 {
  "key": "environment",
  "values": ["aws", "gcp"],
  "operator": "any_of",
  "type": "label"
 }
```
but probably not:
```
 {
  "key": "environment",
  "values": ["aws", null],
  "operator": "any_of",
  "type": "label"
 }
```

Am I correct?

Gelio:
I've only rebased the PR on `master`, no functional changes so far

widgets/common/src/deploymentsView/detailsPane/DrilldownButtons.tsx
Gelio: Since that's all within the `common` directory, I will instead just import the enums directly, and not expose the enums on `Stage`. I hope you're ok with that :slightly_smiling_face:

Gelio:
> as far as I understand there is OR operation available in filters, but probably not with `null` as one of the accepted values?
> 
> You can have a rule:
> 
> ```
>  {
>   "key": "environment",
>   "values": ["aws", "gcp"],
>   "operator": "any_of",
>   "type": "label"
>  }
> ```
> 
> but probably not:
> 
> ```
>  {
>   "key": "environment",
>   "values": ["aws", null],
>   "operator": "any_of",
>   "type": "label"
>  }
> ```
> 
> Am I correct?

Yup, that's correct. I was not as precise :smile: Using `['environment', 'service']` as the `values` array works fine. This is what the response is when trying to add `null` to the array in the filter (`['environment', null]`):

```json
{
    "err_filter_rule": {
        "key": "csys-obj-type",
        "operator": "any_of",
        "type": "label",
        "values": [
            "environment",
            null
        ]
    },
    "err_reason": "Labels must be a list of 1-entry dictionaries: [{<key1>: <value1>}, {<key2>: [<value2>, <value3>]}, ...]",
    "error_code": "invalid_filter_rule",
    "message": "The filter rule {'key': 'csys-obj-type', 'values': ['environment', None], 'operator': 'any_of', 'type': 'label'} is not in the right format. Labels must be a list of 1-entry dictionaries: [{<key1>: <value1>}, {<key2>: [<value2>, <value3>]}, ...]",
    "server_traceback": null
}
```

To conclude, OR is supported inside a single rule, but not between rules :slightly_smiling_face:

widgets/deploymentsViewDrilledDown/src/widget.tsx
Gelio: Let's find a suitable name that would not be overly long at the same time and more confusing than just the number :smile: I addressed your previous candidate in the comment you referenced, and my candidates are:

1. `minDrilldownContextLengthForDrilldownPage`
2. `onlyTopLevelPageDrilldownContextLength` (it would be set to `1` then, and I would use `<= 1` instead of `< 2`)

I'm not convinced about either one :smile:

qooban: Hmm... Maybe `minDrilldownPageContextLength` or `minContextLengthForDrilldownPage`?

widgets/common/src/deploymentsView/detailsPane/DrilldownButtons.tsx
qooban: Yes, perfectly fine.

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Nope, we don't :slightly_smiling_face: To add it and get the partial type safety we would have to implement https://github.com/cloudify-cosmo/cloudify-stage/pull/1296#discussion_r610444415. Adding full type safety requires some high-level TS magic that would verify that `csys-`-prefixed label keys must be from a known set :smile: 

However, my impression from https://github.com/cloudify-cosmo/cloudify-stage/pull/1295/files#r610383069 was that you were not too keen on adding them to the app code ("For tests it's fine") :smile: 

Anyway, I hope that's optional at this point :slightly_smiling_face:

widgets/deploymentsViewDrilledDown/src/widget.tsx
Gelio: `minDrilldownPageContextLength` - to me, this describes the context (an object), not the `drilldownContext` (array) :worried: 

`minContextLengthForDrilldownPage` - same, IMO it describes the `context`, not the `drilldownContext`

If that helps, I can extract 2 functions:
1. `isTopLevelPage`
2. `getParentPageContext`
that would be close to each other, and have a description next to them that would describe the use of `2`. IMO that would be more readable than trying to name the `2` with a lengthy name :slightly_smiling_face:

Let me know what you think about such a solution

Gelio: To save time, I've implemented my suggestion about extracting 2 functions, I hope that's ok :slightly_smiling_face: 

There's little time left today for completing stuff for 5.3.0, so I would appreciate if we postpone the discussion if you do not agree yet

widgets/common/src/deploymentsView/detailsPane/DrilldownButtons.tsx
Gelio: Done :slightly_smiling_face:

kubama (approved):

qooban (approved):

test/cypress/integration/widgets/deployments_view_spec.ts
qooban: Yep, let's keep it as it is.

Gelio:
I've hit

> No space left on device

in https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/441/pipeline/ :worried:

Gelio:
The tests https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/445/pipeline/ failed only due to coverage. Thus, I'm merging the PR

PR #1309, opened 2021-04-13
RD-2033 Add images for Deployments View widgets

Gelio:
This PR adds images for both variants of the Deployments View widget


qooban (approved):

qooban (commented):
Consider using https://compresspng.com/ it can reduce the size of the image.

Gelio:
Yeah, good point. I didn't think of that :slightly_smiling_face:  I'll do that :smile:

qooban (approved):

PR #1311, opened 2021-04-13
RD-1758 / RD-1764 - Allow auto-completion and adding custom values in DynamicDropdown & update related dropdowns

qooban:
`DynamicDropdown` did not have support for auto-completion (by that I mean refetching the dropdown data using `_search` query parameter). The only case for which we had auto-completion working were labels components: `KeyDropdown` and `ValueDropdown`. The second missing thing is that `DynamicDropdown` did not provide support for allowing custom values additions, which is necessary in case of filter rules value inputs having auto-completion on.

### Changes
* Added support for auto-completion in `DynamicDropdown`
  * created custom hook `useFetchTrigger` to encapsulate triggering delayed fetch
  * refactored using `searchQuery` in `DynamicDropdown`
  * Updated `KeyDropdown` and `ValueDropdown` components to use it
* Added support for allowing custom values addition in `DynamicDropdown`
  * Added `allowAdditions` support in `DynamicDropdown` 
  * Updated `AttributeValueInput` component accordingly
  * Removed `allowKnownOnly` prop and added `allowAdditions` prop to label dropdowns

### Video

https://user-images.githubusercontent.com/5202105/115511828-96d5b180-a281-11eb-8d4d-2846408c0649.mp4

### TODO
* `DynamicDropdown` tests for auto-completion and custom values additions (will be part of https://github.com/cloudify-cosmo/cloudify-stage/pull/1313 )


widgets/common/src/DynamicDropdown.jsx
```diff
@@ -99,10 +107,17 @@ export default function DynamicDropdown({
         }
     }, [value]);
 
+    const delayMs = 500;
+    const delayedSetShouldLoadMore = useCallback(
+        _.debounce(() => {
```
Gelio, line 112: Optional:

We're already importing from `lodash`, so we could also import `debouce` from there and not use the global variable :smile:

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -76,9 +84,9 @@ export default function DynamicDropdown({
     }
 
     function refreshData() {
-        setOverrideOptionsAfterFetch();
-        setHasMore(true);
+        if (!multiple) setOverrideOptionsAfterFetch();
```
Gelio, line 87: What's the reason we want to only do it for a single item selection? :thinking:

Gelio (commented):
I don't fully fathom the PR :smile: Is the main change moving the denounce logic to `DynamicDropdown` and including `_search` in the query parameters?

widgets/common/src/DynamicDropdown.jsx
qooban: Yep, I'll change that.

qooban:
> Is the main change moving the denounce logic to DynamicDropdown and including _search in the query parameters?

Yes, that is main change. 
Change consequences:
1. Add auto-completion to all uses of `DynamicDropdown` which IMHO is big gain
2. Allow out-of-the-box auto-completion support for multiple selection in case of labels

widgets/common/src/DynamicDropdown.jsx
qooban: Looks like it is not needed anymore (after moving auto-completion to `DynamicDropdown`). It was used before by label's dropdowns only. Removed `overrideOptionsAfterFetch` state variable.

widgets/common/src/DeployBlueprintModal.jsx
```diff
@@ -203,7 +203,7 @@ class DeployBlueprintModal extends React.Component {
 
     selectBlueprint(id) {
         if (!_.isEmpty(id)) {
-            this.setState({ loading: true, loadingMessage: t('deploymentInputs.loading') });
+            this.setState({ loading: true, loadingMessage: t('inputs.deploymentInputs.loading') });
```
qooban, line 206: When running `filter_spec` test (which uses extensively `DynamicDropdown` components) I found that one, so fixed it.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,10 +1,38 @@
...
 import './DynamicDropdown.css';
 
 let instanceCount = 0;
 
+/**
+ * Creates two `useUpdateEffect` hooks, one calls fetchTrigger with debouncing
+ * on delayedFetchDeps change and the second calls immediately fetchTrigger
+ * on immediateFetchDeps change
+ *
+ * @param {function(immediateFetch: boolean)} fetchTrigger function to be called to trigger data fetching
```
qooban, line 15: I used JSDoc notation to make it work in WebStorm. I now that in TS we can do it differently.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -41,14 +70,19 @@ export default function DynamicDropdown({
     const [isLoading, setLoading] = useState(false);
     const [overrideOptionsAfterFetch, setOverrideOptionsAfterFetch, resetOverrideOptionsAfterFetch] = useBoolean();
 
-    function unsetLoadingAndShouldLoadMore() {
-        setLoading(false);
-        setShouldLoadMore(false);
```
qooban, line 46: It was just moved below, closer to the actual use, inside `loadMore` function.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -57,30 +91,29 @@ export default function DynamicDropdown({
...
-                .finally(unsetLoadingAndShouldLoadMore);
+                .finally(onFetchFinished);
         } else {
             const nextPage = currentPage + 1;
 
             toolbox
                 .getManager()
-                .doGet(fetchUrl, { _sort: valueProp, _size: pageSize, _offset: nextPage * pageSize })
+                .doGet(fetchUrl, {
+                    _search: searchQuery,
```
qooban, line 101: I'm not sure, maybe we should expose a prop in `DynamicDropdown` and let user decide if he wants to use auto-completion? And not do that for all cases?

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -99,10 +132,15 @@ export default function DynamicDropdown({
         }
     }, [value]);
 
-    useUpdateEffect(() => {
-        refreshData();
-        setShouldLoadMore(true);
-    }, [fetchUrl]);
+    useFetchTrigger(
+        resetOptions => {
+            if (resetOptions) setOverrideOptionsAfterFetch();
```
qooban, line 137: I had to revert `setOverrideOptionsAfterFetch` (which I removed few commits ago). It make sense to override options in case of `fetchUrl` change.

qooban:
System tests: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/479

qooban:
All the planned functional changes are in right now, so I don't expect much extension apart from Cypress test updates.

Gelio:
Thanks for the notification, I'll review the PR tomorrow :slightly_smiling_face:

kubama (approved):
I found nothing I'd like to comment, looks good to me so I'm approving 👍.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -27,82 +58,110 @@ export default function DynamicDropdown({
...
+            const optionsToAdd = [];
             _.castArray(value).forEach(singleValue => {
                 if (!_.find(options, { [valueProp]: singleValue })) {
-                    setOptions([{ [valueProp]: singleValue, implicit: true }, ...options]);
+                    optionsToAdd.push({ [valueProp]: singleValue, implicit: true });
                 }
             });
+            if (optionsToAdd.length > 0) {
+                setOptions([...optionsToAdd, ...options]);
+            }
```
qooban, line 153: This is a bug-fix. Implicit options addition was not working properly for dropdowns configured to handle multiple values.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,44 @@
...
+const defaultFetchState = { hasMore: true, currentPage: -1, shouldLoadMore: false };
+
+/**
+ * Creates two `useUpdateEffect` hooks to call `fetchTrigger` function with debouncing.
+ * First hook calls `fetchTrigger` with `reset` argument set to false on `withoutResetFetchDeps` change.
+ * The second calls `fetchTrigger` with `reset` argument set to true on `withResetFetchDeps` change
+ *
+ * @param {function(reset: boolean)} fetchTrigger function to be called to trigger data fetching
+ * @param {React.DependencyList} withoutResetFetchDeps list of dependencies for delayed fetch without reset
+ * @param {React.DependencyList} withResetFetchDeps list of dependencies for delayed fetch with reset
```
Gelio, line 18: Nice idea adding the types at least in JSDoc :slightly_smiling_face: 

Optional:

Instead of providing the types in JSDoc, we could convert the file to TS, and add `@ts-nocheck` on the top, so we can use TS in this file. This would also allow registering the component in `Stage.Common` namespace :smile:

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,44 @@
...
 
 let instanceCount = 0;
+const defaultFetchState = { hasMore: true, currentPage: -1, shouldLoadMore: false };
+
+/**
+ * Creates two `useUpdateEffect` hooks to call `fetchTrigger` function with debouncing.
+ * First hook calls `fetchTrigger` with `reset` argument set to false on `withoutResetFetchDeps` change.
+ * The second calls `fetchTrigger` with `reset` argument set to true on `withResetFetchDeps` change
+ *
+ * @param {function(reset: boolean)} fetchTrigger function to be called to trigger data fetching
```
Gelio, line 16: I'm not sure what's this syntax. I've never seen it :smile: Can you tell me what it is?

I've always used TS-compatible syntax in JSDoc

```suggestion
 * @param {(reset: boolean) => void} fetchTrigger function to be called to trigger data fetching
```

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,44 @@
...
+ * @param {React.DependencyList} withoutResetFetchDeps list of dependencies for delayed fetch without reset
+ * @param {React.DependencyList} withResetFetchDeps list of dependencies for delayed fetch with reset
+ */
+function useFetchTrigger(fetchTrigger, withoutResetFetchDeps, withResetFetchDeps) {
+    const { useUpdateEffect } = Stage.Hooks;
+    const delayMs = 500;
+    const delayedFetchTrigger = useCallback(
+        debounce(reset => fetchTrigger(reset), delayMs),
+        []
```
Gelio, line 25: `fetchTrigger` should probably appear in the dependency list :slightly_smiling_face: Ohterwise, we'll be calling the initial `fetchTrigger` function even it it changes during the life of the component :slightly_smiling_face: 

IMO this requires a bit more logic.
1. We should probably use `useRef` to always have the handle to the latest `delayedFetchTrigger` function inside `useUpdateEffect`s, when the function changes
2. If the `fetchTrigger` function is changed, which function should be called? Should we still call the old function, or the new one? This is an open question, I don't know :smile:

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,44 @@
...
 import VisibilitySensor from 'react-visibility-sensor';
 import './DynamicDropdown.css';
 
 let instanceCount = 0;
+const defaultFetchState = { hasMore: true, currentPage: -1, shouldLoadMore: false };
+
+/**
+ * Creates two `useUpdateEffect` hooks to call `fetchTrigger` function with debouncing.
+ * First hook calls `fetchTrigger` with `reset` argument set to false on `withoutResetFetchDeps` change.
+ * The second calls `fetchTrigger` with `reset` argument set to true on `withResetFetchDeps` change
```
Gelio, line 14: Isn't it the implementation described in a comment :smile: ? I see a risk in it becoming outdated, and also the readers have to read the code twice and see if the implementation matches the comment :slightly_smiling_face:

IMO we could remove this description :slightly_smiling_face:

Gelio: Alternatively, you could explain what _reset_ means, because just by reading the hook's code, I did not get its purpose :thinking:

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,44 @@
...
+ *
+ * @param {function(reset: boolean)} fetchTrigger function to be called to trigger data fetching
+ * @param {React.DependencyList} withoutResetFetchDeps list of dependencies for delayed fetch without reset
+ * @param {React.DependencyList} withResetFetchDeps list of dependencies for delayed fetch with reset
+ */
+function useFetchTrigger(fetchTrigger, withoutResetFetchDeps, withResetFetchDeps) {
+    const { useUpdateEffect } = Stage.Hooks;
+    const delayMs = 500;
+    const delayedFetchTrigger = useCallback(
+        debounce(reset => fetchTrigger(reset), delayMs),
```
Gelio, line 24: Optional:

Boolean parameters are confusing to me and IMO make the API harder to read :slightly_smiling_face: Just by reading the code, we have to trust that `false` means `reset`.

I suggest refactoring it to an object, which has a boolean property. This way there is an explicit contract and less knowledge required when using the hook

```suggestion
        debounce(reset => fetchTrigger({ reset }), delayMs),
```

or

```suggestion
        debounce(reset => fetchTrigger({ shouldReset : reset }), delayMs),
```

widgets/common/src/DynamicDropdown.jsx
Gelio: If auto-completion should work for all cases at the moment, then IMO we can go the easy route and not implement unnecessary options :smile: We could always add it later :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,44 @@
...
+    useUpdateEffect(() => {
+        delayedFetchTrigger(true);
+    }, withResetFetchDeps);
+}
 
 export default function DynamicDropdown({
     fetchUrl,
     onChange,
     toolbox,
+    allowAdditions = false,
```
Gelio, line 41: General:

I'm not certain about the name of the prop. I understand it matches the Semantic's `Dropdown`, yet I find it slightly confusing. When using a `DynamicDropdown`, it fetches data from the network. Seeing `allowAdditions` on it leads me to think that it will somehow create new entities in the API, which is (as far as I checked) not the case :smile: 

Thus, I would encourage revisiting the name of the prop. `onlyKnownValues` that we used for labels make more sense to me here :slightly_smiling_face: Other suggestions are also welcome

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -27,82 +58,110 @@ export default function DynamicDropdown({
...
+    const { useBoolean, useEventListener } = Stage.Hooks;
 
     const [id] = useState(() => {
         instanceCount += 1;
         return `dynamicDropdown${instanceCount}`;
     });
     const [options, setOptions] = useState([]);
-    const [hasMore, setHasMore] = useState(true);
-    const [currentPage, setCurrentPage] = useState(-1);
+    const [fetchState, setFetchState] = useState({ ...defaultFetchState, shouldLoadMore: prefetch });
```
Gelio, line 68: The state management for this variable is getting more and more complex, and is ATM scattered around the file :slightly_smiling_face: IMO it would be more readable and easier to understand to encapsulate it in a `useReducer` with a very explicit set of acceptable actions that would be handled in a single place :slightly_smiling_face: This way we have all the code related to the fetching state management in a single place, and just dispatch actions, redux-style

widgets/common/src/labels/ValueDropdown.jsx
```diff
@@ -27,13 +26,13 @@ ValueDropdown.propTypes = {
...
-    allowKnownOnly: PropTypes.bool,
+    allowAdditions: PropTypes.bool,
     value: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)])
 };
 
 ValueDropdown.defaultProps = {
     labelKey: '',
     multiple: false,
-    allowKnownOnly: false,
+    allowAdditions: false,
```
Gelio, line 36: I did not go deep enough in the code to be sure, but `allowKnownOnly` and `alllowAdditions` are opposite of each other, if I'm not mistaken. Should we not, in this case, change the default values to `true`? :smile: 

With `allowKnownOnly = true` we don't allow user-defined values, and with `allowAdditions = true` we do. So, `allowAdditions = !allowKnownOnly` to have the same behavior.

The same goes for all changes of `allowKnownOnly` in this PR

Gelio (commented):
This is a partial review :smile: I did not read through the `DynamicDropdown`'s code yet as I hope to simplify the state management with `useReducer` before I review further, which will allow me do a full review more efficiently :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
qooban: It's valid JSDoc syntax. I think this can be left as it is.

See my comment: https://github.com/cloudify-cosmo/cloudify-stage/pull/1311#discussion_r614049310 for a bit more explanation.

widgets/common/src/DynamicDropdown.jsx
qooban: Hmm... But in that case the function will not change during the life of the component, so is it worth to handle it for that specific case? This hook is internal and used only here.

widgets/common/src/DynamicDropdown.jsx
Gelio: :thinking: Does this mean that `(immediateFetch: boolean) => void` as a JSDoc annotation does not work in WebStorm?

widgets/common/src/DynamicDropdown.jsx
qooban: I'll improve the description.

widgets/common/src/DynamicDropdown.jsx
qooban: I don't see a point in having an object here, in that case I would stick with single boolean argument. 
I will change it to `shouldReset` if that reads better.

widgets/common/src/DynamicDropdown.jsx
qooban: Hmm... I don't agree. I think `allowAdditions` make sense here. It refers to the fact that we can add options to the dropdown itself. I wouldn't think that you can create new entities in the API. How dropdown would be able to call any 'write' APIs as we always pass just `fetchUrl` (which IMHO clearly says that the intention is just to fetch data)?

widgets/common/src/DynamicDropdown.jsx
qooban: This is very good idea. I was not aware of `useReducer`. Will change it.

widgets/common/src/labels/ValueDropdown.jsx
qooban: No, I don't think so. I mentioned in the changes list: "Removed `allowKnownOnly` prop and added `allowAdditions` prop to label dropdowns" to show that it is not only name change. 

This was changed mainly to handle the new requirement: we need to allow additions and not block dropdowns to only use known only. `ValueDropdown` previously did not allow additions. Custom values in Labels component (where only single values are allowed and we need to have value validation) are handled there differently (see `useLabelInput` hook to see that actual value of that component is taken from the `searchQuery`).

widgets/common/src/DynamicDropdown.jsx
qooban: It would be great to have use TS in that file. I had initial attempt to do so, but I failed. TS started reporting errors in other files which uses `DynamicDropdown`. Due to limited time, I don't want to go in that direction right now.

widgets/common/src/DynamicDropdown.jsx
qooban: Yes.

Untitled

qooban:
@Gelio @kubama I will appreciate very much prioritizing reviews on that. I'd like to finally switch to tests (RD-1775) and error handling (RD-2011) soon.

kubama (approved):

qooban:
System tests started: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/482/

qooban:
System tests failed due to... Another environment issue (reported RD-2096).

Another run: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/484/ (seems like it went further, so hopefully will end without any issues)

qooban:
System tests passed (except one known failure - Topology widget test), so waiting for final reviews and as I mentioned in the description tests will come in: https://github.com/cloudify-cosmo/cloudify-stage/pull/1313 (which is already started).

widgets/common/src/DynamicDropdown.jsx
Gelio: Good to know :smile: Thanks for explaining

widgets/common/src/DynamicDropdown.jsx
Gelio: I see, no problemo :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
Gelio: I fear the situation when the hook is extracted as a more general-purpose and causes errors :slightly_smiling_face: If you think that's justified, a comment for the `fetchTrigger` parameter about the fact that the hook is indifferent to the `fetchTrigger` parameter and will keep calling the initial function, would be great :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
Gelio: Maybe I didn't describe it well enough. For me, the argument is having an explicit API contract and fewer things to manually validate. If you have a boolean, who knows what is its meaning. You have to go to the function definition and check.

```
const fetchTrigger = (dontReset) => {...};

useFetchTrigger(fetchTrigger, ...);
```

It compiles, and reads well. Booleans match. But the expected behavior is exactly opposite :slightly_smiling_face: An API that uses objects reduces the overhead and the need for manual validation, because the object that is passed around is `{ shouldReset: boolean }`, and it's obvious what the meaning of `true` for that parameter is.

Anyway, I'll make the comment optional, change at your with :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
Gelio: After some thought and analysis, I came to the same conclusion. Still, I had to spend time analyzing a prop that could either be named a bit clearer, or accompanied by a JSDoc comment explaining that :slightly_smiling_face: 

If you're against changing the name of the prop, how about adding a comment?

widgets/common/src/DynamicDropdown.jsx
Gelio: Also, adding a comment that `fetchTrigger` is explicitly not used in the dependency array would be great :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -147,24 +232,33 @@ export default function DynamicDropdown({
                 value={getDropdownValue()}
                 id={id}
                 name={name}
-                onChange={(event, field) => onChange(!_.isEmpty(field.value) ? field.value : null)}
-                onSearchChange={(e, data) => setSearchQuery(data.searchQuery)}
+                allowAdditions={allowAdditions}
+                onAddItem={(event, data) => {
+                    if (allowAdditions) setOptions([{ [valueProp]: data.value }, ...options]);
```
Gelio, line 237: Is there a point in checking if additions are allowed here? Does Semantic UI still fire `onAddItem` even if `allowAdditions={false}`?

Optional:

```suggestion
                    setOptions([{ [valueProp]: data.value }, ...options]);
```

widgets/common/src/hooks/useLabelInput.js
```diff
@@ -10,7 +10,7 @@ function useLabelInput(onChange, { allowKnownOnly = false, initialValue = '' })
         inputValue,
         invalidCharacterTyped,
         submitChange: (event, data) => {
-            if (allowKnownOnly) {
+            if (allowAdditions) {
```
Gelio, line 13: I find it hard to come up with a reason why the logic below should happen if we allow additions. Should it not be called e.g. `validationDisabled` or `allowAnyValue`, because that's how the property is used, as far as I see :thinking:

widgets/common/src/hooks/useLabelInput.js
```diff
@@ -29,7 +29,7 @@ function useLabelInput(onChange, { allowKnownOnly = false, initialValue = '' })
         resetInput: () => {
             resetInputValue();
             onChange(initialValue);
-            if (!allowKnownOnly) unsetInvalidCharacterTyped();
+            if (!allowAdditions) unsetInvalidCharacterTyped();
```
Gelio, line 32: Optional:

How about unsetting the flag anyway?

```suggestion
            unsetInvalidCharacterTyped();
```

The behavior should be the same as now, and the code gets simpler :slightly_smiling_face:

widgets/common/src/labels/CommonDropdown.jsx
```diff
@@ -44,17 +26,18 @@ export default function CommonDropdown({ innerRef, baseFetchUrl, onChange, toolb
...
                 fetchUrl={fetchUrl}
                 itemsFormatter={items => _.map(items, item => ({ id: item }))}
                 onBlur={unsetInvalidCharacterTyped}
                 onChange={newValue => submitChange(null, { value: newValue })}
-                onSearchChange={allowKnownOnly ? undefined : submitChange}
-                searchQuery={allowKnownOnly ? undefined : inputValue}
-                selectOnNavigation={allowKnownOnly}
+                onSearchChange={allowAdditions ? undefined : submitChange}
+                searchQuery={allowAdditions ? undefined : inputValue}
+                selectOnNavigation={allowAdditions}
```
Gelio, line 38: General:

I did not read it long enough. Is it relevant to provide those values based on `allowAdditions`? The same goes for `value`.

widgets/common/src/labels/KeyDropdown.jsx
```diff
@@ -1,19 +1,20 @@
...
-            placeholder={i18n.t(`widgets.common.labels.keyPlaceholder.${allowKnownOnly ? 'knownOnly' : 'knownOrNew'}`)}
+            fetchUrl="/labels/deployments"
+            noResultsMessage={value && !allowAdditions ? i18n.t('widgets.common.labels.newKey') : undefined}
+            placeholder={i18n.t('widgets.common.labels.keyPlaceholder')}
             name="labelKey"
             tabIndex={0}
             onChange={onChange}
             toolbox={toolbox}
-            allowKnownOnly={allowKnownOnly}
+            additionLabel={`${i18n.t('widgets.common.labels.newKey')} `}
```
Gelio, line 16: What's this prop? I did not find it in `CommonDropdown` :worried:

widgets/common/src/labels/ValueDropdown.jsx
```diff
@@ -1,22 +1,21 @@
...
+            fetchUrl={labelKey ? `/labels/deployments/${labelKey}` : ''}
+            noResultsMessage={value && !allowAdditions ? i18n.t('widgets.common.labels.newValue') : undefined}
+            placeholder={i18n.t('widgets.common.labels.valuePlaceholder')}
             name="labelValue"
             tabIndex={labelKey ? 0 : -1}
             onChange={onChange}
             toolbox={toolbox}
             multiple={multiple}
-            allowKnownOnly={allowKnownOnly}
+            additionLabel={`${i18n.t('widgets.common.labels.newValue')} `}
```
Gelio, line 17: Is this prop needed :smile: ?

Gelio (commented):
Looks mostly good, I've gone through all the files this time :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
qooban: Hmm... Not sure what should I write here. If you're for adding a comment, please provide what would you expect.

For me it's quite easy to get how it works. You just go from the prop definition to the prop use (which is directly passed to `Dropdown`) and if you still need explanation, you just go to https://react.semantic-ui.com/modules/dropdown/ and you can read that `allowAdditions`:
> Allow user additions to the list of options (boolean). Requires the use of `selection`, `options` and `search`.

widgets/common/src/labels/KeyDropdown.jsx
qooban: It's `Dropdown` component prop.
Ref.: https://react.semantic-ui.com/modules/dropdown/

widgets/common/src/labels/ValueDropdown.jsx
qooban: Yes. It's `Dropdown` component prop.
Ref.: https://react.semantic-ui.com/modules/dropdown/

widgets/common/src/labels/CommonDropdown.jsx
qooban: Yes, `LabelsInput` component requires different behavior.

widgets/common/src/hooks/useLabelInput.js
qooban: What your suggesting was already part of this PR. I was debating with myself, but as it is very minor I'm leaving it for now.

widgets/common/src/DynamicDropdown.jsx
qooban: Make sense, will remove it.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,70 @@
...
+        case fetchActionType.PREPARE_FOR_NEXT_PAGE_FETCH:
+            return {
+                currentPage: action.currentPage,
+                hasMore: action.hasMore,
+                shouldLoadMore: false
+            };
+        case fetchActionType.TRIGGER_FETCH:
+            return { ...state, shouldLoadMore: true };
+        case fetchActionType.END_FETCH:
+            return { ...defaultFetchState, hasMore: false };
```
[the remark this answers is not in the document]
qooban: Hmm... So should I specify all the values from `defaultFetchState` here where just `hasMore` differs from the expected state after the action?

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -27,82 +84,110 @@ export default function DynamicDropdown({
...
     const [options, setOptions] = useState([]);
-    const [hasMore, setHasMore] = useState(true);
-    const [currentPage, setCurrentPage] = useState(-1);
+    const [fetchState, dispatchFetchAction] = useReducer(fetchReducer, {
+        ...defaultFetchState,
+        shouldLoadMore: prefetch
+    });
     const [searchQuery, setSearchQuery] = useState('');
-    const [shouldLoadMore, setShouldLoadMore] = useState(prefetch);
     const [isLoading, setLoading] = useState(false);
```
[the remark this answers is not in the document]
qooban: I considered that, but as it is optional I would leave it for now.

Digression: I think at some point `DynamicDropdown` should be refactored, but with the changes from this PR it is already much closer to have better shape IMHO.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,70 @@
...
+const fetchActionType = {
+    PREPARE_FOR_INITIAL_FETCH: 'prepareForInitialFetch',
+    PREPARE_FOR_NEXT_PAGE_FETCH: 'prepareForNextPageFetch',
+    TRIGGER_FETCH: 'triggerFetch',
+    END_FETCH: 'endFetch'
+};
+function fetchReducer(state, action) {
+    switch (action.type) {
+        case fetchActionType.PREPARE_FOR_INITIAL_FETCH:
+            return { ...defaultFetchState };
```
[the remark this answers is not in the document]
qooban: My understanding is that `prefetch` is one time use only when component is mounted.

I used originally "RESET" here, but IMHO `PREPARE_FOR_INITIAL_FETCH` is more accurate. Alternatively it could be `PREPARE_FOR_FIRST_PAGE_FETCH`.

widgets/common/src/hooks/useLabelInput.js
qooban: I will change to `allowAnyValue`.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,70 @@
 // NOTE: Disabling react/require-default-props as default values are provided in component's definition
 /* eslint-disable react/require-default-props */
+import { useCallback, useReducer } from 'react';
+import { debounce, isFunction } from 'lodash';
 import VisibilitySensor from 'react-visibility-sensor';
 import './DynamicDropdown.css';
 
 let instanceCount = 0;
+const defaultFetchState = { hasMore: true, currentPage: -1, shouldLoadMore: false };
```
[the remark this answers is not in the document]
qooban: Will do.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -1,14 +1,70 @@
...
+ * @param {React.DependencyList} withoutResetFetchDeps list of dependencies for delayed `fetchTrigger` call with
+`shouldReset` argument set to false
+ * @param {React.DependencyList} withResetFetchDeps list of dependencies for delayed `fetchTrigger` call with
+ * `shouldReset` argument set to true
+ */
+function useFetchTrigger(fetchTrigger, withoutResetFetchDeps, withResetFetchDeps) {
+    const { useUpdateEffect } = Stage.Hooks;
+    const delayMs = 500;
+    const delayedFetchTrigger = useCallback(
+        debounce(shouldReset => fetchTrigger(shouldReset), delayMs),
```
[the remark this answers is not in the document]
qooban: Will change it.

widgets/common/src/DynamicDropdown.jsx
qooban: Changed.

widgets/common/src/DynamicDropdown.jsx
qooban: Please don't be afraid 🙂 

I can imagine that when I would have two hooks inside `DynamicDropdown` you would comment that it could be extracted to a custom hook. When there's a custom hook which is really strictly connected to `DynamicDropdown` and planned to only be used here you are commenting that you are afraid of a situation when it will be extracted... Ehh...

qooban:
@Gelio I answered all your comments. Please take a look and let's merge it on Friday when I'm back. 

System tests scheduled: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/485/

widgets/common/src/DynamicDropdown.jsx
Gelio: Ok, if you believe the existing approach is more readable than introducing some other name, then let's leave it as is. IMO the prop is not as straightforward

widgets/common/src/labels/ValueDropdown.jsx
Gelio: I see, so refactoring the name combined with changing the default behavior makes sense :+1:

widgets/common/src/DynamicDropdown.jsx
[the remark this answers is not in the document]
Gelio: If you're settling on the `PREPARE_FOR` scheme, then `FIRST_PAGE` is closer to the truth than `INITIAL` :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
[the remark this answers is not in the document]
Gelio: I'm not saying it's unjustified, but it's surprising to me at first glance :slightly_smiling_face: I suppose it makes sense with this kind of state management

widgets/common/src/labels/KeyDropdown.jsx
Gelio: > `CommonDropdown` handle all `Dropdown` props using `rest`

That's the part that I missed, thanks

Gelio (approved):
Looks good to me, although I do not fully understand all the logic in `CommonDropdown` & `DynamicDropdown`, so my confidence there is moderate :slightly_smiling_face:

kubama (approved):

widgets/common/src/DynamicDropdown.jsx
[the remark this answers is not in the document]
qooban: Yep, as that seems like a very simple change, not changing the functionality, I can do that within test PR https://github.com/cloudify-cosmo/cloudify-stage/pull/1313 .

qooban:
Thanks for the reviews. Tests are green (except the one which is failing since 21st - Topology test), so I'm merging the PR.

PR #1312, opened 2021-04-14
RD-1758 / RD-1774 & RD-1762 - Add support for type: 'attribute' and operators: 'contains', 'not_contains', 'starts_with', 'ends_with'

qooban:
### Changes
* Added `MultipleStringValuesInput` component allowing to type in multiple string values
* Extracted `isAnyOperator` common function
* Used `MultipleStringValuesInput` component in `AttributeValueInput` to allow full support for type: `attribute` and operators: `contains`, `not_contains`, `starts_with`, `ends_with`
* Extracted `getTranslation` and `getPlaceholderTranslation` common functions

Tests will come from: https://github.com/cloudify-cosmo/cloudify-stage/pull/1313

### Video
https://user-images.githubusercontent.com/5202105/114687646-42c44d80-9d14-11eb-839f-b22029a912c1.mp4

### DIFF
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -532,7 +532,7 @@
                         "blueprintId": "Select blueprints",
                         "siteName": "Select site names",
                         "createdBy": "Select usernames",
-                        "multipleStrings": "Type in texts"
+                        "multipleStrings": "Type in values"
                     },
                     "ruleTypesLabels": {
                         "label": "Label",
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -667,6 +667,7 @@
                     "services": "Services"
                 },
                 "buttons": {
+                    "detailsFetchingError": "Could not fetch deployment details",
                     "environments": {
                         "label": "Subenvironments",
                         "title": "Drill down to subenvironments"
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -531,7 +531,8 @@
                     "inputsPlaceholders": {
                         "blueprintId": "Select blueprints",
                         "siteName": "Select site names",
-                        "createdBy": "Select usernames"
+                        "createdBy": "Select usernames",
+                        "multipleStrings": "Type in texts"
                     },
                     "ruleTypesLabels": {
                         "label": "Label",
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -17,6 +17,12 @@ describe('Filters widget', () => {
         cy.get('.loading').should('not.exist');
     });
 
+    function typeAttributeRuleValue(value: string) {
+        cy.get('[name=ruleValue]')
+            .click()
+            .within(() => cy.get('input').type(`${value}{enter}`).blur());
+    }
+
     function checkExistingRules() {
         cy.get('.fields:eq(0)').within(() => {
             cy.contains('Blueprint');
@@ -37,7 +43,7 @@ describe('Filters widget', () => {
         cy.get('.fields:eq(0)').within(() => {
             cy.contains('is one of').click();
             cy.contains('contains').click();
-            cy.get('.input input').type(newBlueprintRuleValue);
+            typeAttributeRuleValue(newBlueprintRuleValue);
         });
     }
 
@@ -86,7 +92,8 @@ describe('Filters widget', () => {
 
         cy.get('.modal').within(() => {
             cy.contains('.field', 'Filter ID').find('input').type(newFilterName);
-            cy.get('.fields:eq(0) .input input').type(blueprintId);
+
+            cy.get('.fields:eq(0)').within(() => typeAttributeRuleValue(blueprintId));
             cy.contains('Add new rule').click();
             cy.get('.fields:eq(1)').within(() => {
                 cy.get('[name=ruleRowType]').click();
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -2,7 +2,7 @@ import { FilterRuleOperators, FilterRuleType } from '../../../../widgets/common/
 
 describe('Filters widget', () => {
     before(() => {
-        cy.usePageMock('filters').activate().mockLogin();
+        cy.usePageMock(['filters', 'onlyMyResources']).activate().mockLogin();
     });
 
     const filterName = 'filters_test_filter';
@@ -193,5 +193,13 @@ describe('Filters widget', () => {
 
         cy.contains('OK').click();
         cy.get('.modal').should('not.exist');
+    });
+
+    it('should support "Only my resources" setting', () => {
+        cy.interceptSp('GET', new RegExp('/filters/deployments\\?.*created_by=admin.*')).as('getRequest');
+
+        cy.contains('Show only my resources').click();
+
+        cy.wait('@getRequest');
     });
 });
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -18,9 +18,7 @@ describe('Filters widget', () => {
     });
 
     function typeAttributeRuleValue(value: string) {
-        cy.get('[name=ruleValue]')
-            .click()
-            .within(() => cy.get('input').type(`${value}{enter}`).blur());
+        cy.get('[name=ruleValue]').click().find('input').type(`${value}{enter}`).blur();
     }
 
     function checkExistingRules() {
--- a/widgets/common/src/filters/RuleRowTypeDropdown.tsx
+++ b/widgets/common/src/filters/RuleRowTypeDropdown.tsx
@@ -2,7 +2,7 @@ import { camelCase } from 'lodash';
 import type { FunctionComponent } from 'react';
 import type { DropdownItemProps } from 'semantic-ui-react/dist/commonjs/modules/Dropdown/DropdownItem';
 
-import { i18n, i18nPrefix } from './consts';
+import { getTranslation } from './common';
 import { FilterRuleRowType } from './types';
 
 interface RuleRowTypeDropdownProps {
@@ -13,7 +13,7 @@ interface RuleRowTypeDropdownProps {
 const ruleRowTypeOptions = (() => {
     return Object.values(FilterRuleRowType).map(
         (ruleType): DropdownItemProps => ({
-            text: i18n.t(`${i18nPrefix}.ruleTypesLabels.${camelCase(ruleType)}`),
+            text: getTranslation(`ruleTypesLabels.${camelCase(ruleType)}`),
             value: ruleType
         })
     );
--- a/widgets/common/src/filters/inputs/AttributeValueInput.tsx
+++ b/widgets/common/src/filters/inputs/AttributeValueInput.tsx
@@ -2,7 +2,7 @@ import type { FunctionComponent } from 'react';
 
 import DynamicDropdown from '../../DynamicDropdown';
 import type { CommonAttributeValueInputProps } from './types';
-import { isAnyOfOrNotAnyOfOperator } from './common';
+import { isAnyOperator } from './common';
 import { i18n, i18nPlaceholdersPrefix } from '../consts';
 import MultipleStringValuesInput from './MultipleStringValuesInput';
 
@@ -21,7 +21,7 @@ const AttributeValueInput: FunctionComponent<AttributeValueInputProps> = ({
     toolbox,
     value
 }) => {
-    if (isAnyOfOrNotAnyOfOperator(operator)) {
+    if (isAnyOperator(operator)) {
         return (
             <DynamicDropdown
                 name="ruleValue"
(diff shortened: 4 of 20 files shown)


Gelio:
I have meetings till the end of the day today, so I'll read through the PR tomorrow :smile:

app/translations/en.json
```diff
@@ -531,7 +531,8 @@
                     "inputsPlaceholders": {
                         "blueprintId": "Select blueprints",
                         "siteName": "Select site names",
-                        "createdBy": "Select usernames"
+                        "createdBy": "Select usernames",
+                        "multipleStrings": "Type in texts"
```
kubama, line 535: How about "Type in values" or "Enter values"?

widgets/common/src/filters/inputs/common.ts
```diff
@@ -0,0 +1,7 @@
+import type { FilterRuleOperator } from '../types';
+import { FilterRuleOperators } from '../types';
+
+// eslint-disable-next-line import/prefer-default-export
+export function isAnyOfOrNotAnyOfOperator(operator: FilterRuleOperator) {
```
kubama, line 5: Quite hard to read and even a little bit hard to understand function name. How about shortening to `isAnyOperator`? Or, given the context in which it is used `isExactValueOperator`?

widgets/common/src/filters/inputs/AttributeValueInput.tsx
```diff
@@ -1,9 +1,10 @@
-import { castArray } from 'lodash';
 import type { FunctionComponent } from 'react';
 
 import DynamicDropdown from '../../DynamicDropdown';
-import { LabelsFilterRuleOperators } from '../types';
 import type { CommonAttributeValueInputProps } from './types';
+import { isAnyOfOrNotAnyOfOperator } from './common';
+import { i18n, i18nPlaceholdersPrefix } from '../consts';
```
kubama, line 6: Looks like importing `i18n` and `i18nPlaceholdersPrefix` or `i18nPrefix` is a common pattern. Instead I'd consider exporting a function (or actually two functions) that would add a prefix and delegate to `i18n.t`. It could be placed in `common.ts` and `consts.ts` would be not needed.

app/translations/en.json
qooban: I was debating with myself what to put here. I like the first one, so changed it to "Type in values".

widgets/common/src/filters/inputs/common.ts
qooban: I was sure that somebody will point that out 🙂 
I like the first suggestion, changed.

widgets/common/src/filters/inputs/AttributeValueInput.tsx
qooban: Good idea. That involved a bit bigger change, so many files updated.

NOTE: There are 2 `common.ts` files right now. I'm OK with having two as they do their job very precisely.

kubama (approved):

qooban:
Started system test execution: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/456/

Waiting for review from @Gelio and I hope to merge this PR today.

Gelio:
I'm beginning with the review :eyes: I hope to be faster than Jenkins :racehorse:

widgets/common/src/filters/RuleRowTypeDropdown.tsx
```diff
@@ -13,7 +13,7 @@ interface RuleRowTypeDropdownProps {
 const ruleRowTypeOptions = (() => {
     return Object.values(FilterRuleRowType).map(
         (ruleType): DropdownItemProps => ({
-            text: i18n.t(`${i18nPrefix}.ruleTypesLabels.${camelCase(ruleType)}`),
+            text: getTranslation(`ruleTypesLabels.${camelCase(ruleType)}`),
```
Gelio, line 16: I think I like this `getTranslation` pattern more than exporting the prefix :smile: Looks shorter :slightly_smiling_face: 

Good suggestion @kubama :+1:

widgets/common/src/filters/inputs/common.ts
```diff
@@ -0,0 +1,11 @@
+import type { FilterRuleOperator } from '../types';
+import { FilterRuleOperators } from '../types';
+import { getTranslation } from '../common';
+
+export function isAnyOperator(operator: FilterRuleOperator) {
+    return ([FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf] as FilterRuleOperator[]).includes(operator);
```
Gelio, line 6: Optional:

Using `as` is less safe than a type annotation for a variable. For example, the following works:

```diff
-    return ([FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf] as FilterRuleOperator[]).includes(operator);
+    return ([FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf, 'abc'] as FilterRuleOperator[]).includes(operator);
```

Whereas, the following reports an error:

```ts
    const anyOperators: FilterRuleOperator[] = [FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf, 'abc'];
    return anyOperators.includes(operator);
```

image

Thus, I suggest refactoring this to the following:

```suggestion
    const anyOperators: FilterRuleOperator[] = [FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf];
    return anyOperators.includes(operator);
```

This is optional here, but have it in mind in the future :smile: When it's not too much hassle to create a variable, do that to be more type-safe :slightly_smiling_face: 

A bit more info: https://stackoverflow.com/a/54734992/4874344

In short, `as` is potentially unsafe, as it will accept some type-mismatches :slightly_smiling_face: It won't accept anything, e.g. `[] as string` would not work, since the types are very different, but it's worth keeping in mind that `as` has more "power"

widgets/common/src/filters/inputs/common.ts
```diff
@@ -0,0 +1,11 @@
+import type { FilterRuleOperator } from '../types';
+import { FilterRuleOperators } from '../types';
+import { getTranslation } from '../common';
+
+export function isAnyOperator(operator: FilterRuleOperator) {
```
Gelio, line 5: Optional:

I know it's been discussed already, but I still find the name not matching the behavior :smile: _is any operator_ to me sounds like a function that would check if some string is an operator :smile: 

I don't have a great name here, but perhaps we could not name it by what it does mechanically, but rather by what's the intent when calling this function. I see when it's called, we display `DynamicDropdown`. Thus, how about `shouldAutocompleteKnownValues`? :smile:

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const MultipleStringValuesInput: FunctionComponent<MultipleStringValuesInputProps> = ({
+    name,
+    onChange,
+    placeholder,
+    value
+}) => {
+    const { Dropdown } = Stage.Basic;
+    const options = useMemo(() => value.map(element => ({ text: element, value: element })), [value]);
```
Gelio, line 17: Optional:

Is there a type we could use for the return type of the mapping function to get a better error message if it does not match the expected type? I think we could do

```suggestion
    const options = useMemo(() => value.map((element): DropdownItemProps => ({ text: element, value: element })), [value]);
```

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
```diff
@@ -0,0 +1,35 @@
+import { FunctionComponent, useMemo } from 'react';
+
+interface MultipleStringValuesInputProps {
+    name: string;
+    onChange: (newValue: string[]) => void;
+    placeholder: string;
+    value: string[];
```
Gelio, line 7: Optional:

I prefer using plurals for arrays :smile: 

```suggestion
    values: string[];
```

Gelio (approved):
Left minor comments, did not run it in the UI, but overall, looks good :smile: I like the translation-function pattern :+1: Seems better than exposing the prefixes :slightly_smiling_face:

widgets/common/src/filters/inputs/common.ts
qooban: Thanks for the description, appreciate that. Applied your suggestion.

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
qooban: Sure, make sense. Applied.

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
qooban: Hmm... If we would do it so here, then I think we should do it as well in the other places:
image

As this will be a bit bigger change, I would leave it for now and maybe consider changing it later.

widgets/common/src/filters/inputs/common.ts
qooban: Hmm... I'm not sure about the suggested name..

Another option would to change it to `isCommonRuleOperator` as `AnyOf` and `NotAnyOf` are defined inside `CommonRuleOperator` type.

Then we could probably even more change the function to have something like:
```
export function isCommonRuleOperator(operator: FilterRuleOperator) {
    return Object.keys(CommonRuleOperator).includes(operator);
}
```
(for that we would need to `export enum CommonRuleOperator` from `filters/types.ts`)

Anyways, for this PR I would leave it as it is. If we all agree to any specific change, I'd like to apply it in the later phase.

Gelio: I see how we could leverage `CommonRuleOperator`, but I'm not sure if it will always stay like that - if we add another `CommonRuleOperator`, will it automatically mean that we use the `DynamicDropdown` for it? Probably not :smile: Thus, I would be cautious going in that direction

I don't have a better suggestion than I gave in the initial comment, so we can ignore the matter for now :slightly_smiling_face:

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
Gelio: I see :smile: We can ignore the matter here, but generally, does it matter that we always make changes in a single step? I don't see it as problematic that we would have it a prop named `values` and we would pass `data.value` to it :slightly_smiling_face: I find it easier to make small improvements when it's possible, even if that means that we break some existing pattern (that we can agree can be improved)

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
```diff
@@ -1,4 +1,5 @@
 import { FunctionComponent, useMemo } from 'react';
+import { DropdownItemProps } from 'semantic-ui-react';
```
Gelio, line 2: Optional: `import type` could help, but let's save time, we can do without it here and merge the PR with this small blunder :smile:

Gelio (approved):

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
qooban: Oops, yes, we need RD-2000 🙂

Changed.

qooban:
Previous test execution failed on Filters widget test. I applied the fix within https://github.com/cloudify-cosmo/cloudify-stage/pull/1312/commits/732b70da9335cba256231f667fc1f94eccb21361 . More test changes are already started and will be applied later this or next week within: https://github.com/cloudify-cosmo/cloudify-stage/pull/1313

I don't want to wait for another test execution, so I tested it locally:
image

widgets/common/src/filters/inputs/common.ts
qooban: OK, so let's ignore it.

Gelio (dismissed):
Good thing we've detected the `filters_spec` issue before merging :smile:

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
qooban: I'm not sure if I understand the message entirely. In general more often I'm leaning towards having consistent application, instead of perfectly optimized parts. Maybe that is something to discuss in the weekly 🙂 Maybe that's just the preferences difference which we can live with 🙂

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -17,6 +17,12 @@ describe('Filters widget', () => {
         cy.get('.loading').should('not.exist');
     });
 
+    function typeAttributeRuleValue(value: string) {
+        cy.get('[name=ruleValue]')
+            .click()
+            .within(() => cy.get('input').type(`${value}{enter}`).blur());
```
kubama, line 23:
```suggestion
            .find('input').type(`${value}{enter}`).blur();
```

qooban: I know that's not the the first time you remind me about it 🙂 
Changed. `filters_spec` tests passed.

qooban:
Did one small change related to https://github.com/cloudify-cosmo/cloudify-stage/pull/1312#discussion_r613938545 .
`filters_spec` tests passed locally, again, so I'm waiting for final approvals and I'm going to merge it then.

widgets/common/src/filters/inputs/MultipleStringValuesInput.tsx
Gelio: Sure, I've added a point about it :slightly_smiling_face:

Gelio (approved):

kubama (approved):

PR #1313, opened 2021-04-14
RD-1758 / RD-1775 - Add filter rule form tests

qooban:
This PR adds filter rule form tests.

#### Planned tests
- [x] type "label" with operators "any_of" and "not_any_of"
- [x] type "label" with operators "is_null" and "is_not_null"
- [x] type "attribute" with operators "any_of" and "not_any_of"
- [x] type "attribute" with operators "contains", "not_contains", "starts_with" and "ends_with"
- [ ] auto-completion in `DynamicDropdown` component
- [x] allow addition in `DynamicDropdown` component
- [x] filter rules error handling in filter modals
- [ ] swagger endpoint verification (keys, operators, etc.) - [endpoint details](https://github.com/cloudify-cosmo/cloudify-manager/pull/2771)

#### All current tests
image

#### Tests added
image

### Open items
* `filters_spec.ts` file has >700 LOCs. Do you think we should split it into two files?


widgets/common/src/DynamicDropdown.jsx
```diff
@@ -258,8 +259,7 @@ export default function DynamicDropdown({
                                     <Loading message="" />
                                 </VisibilitySensor>
                             ),
-                            key: 'loader',
-                            text: searchQuery
+                            key: 'loader'
```
qooban, line 262: This was done to make loader to be independent from `searchQuery`.
That allowed to save one HTTP request.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -23,24 +23,24 @@ function useFetchTrigger(fetchTrigger, withoutResetFetchDeps, withResetFetchDeps
...
 
     useUpdateEffect(() => {
-        delayedFetchTrigger(true);
+        delayedFetchTrigger({ shouldReset: true });
     }, withResetFetchDeps);
 }
 
 const fetchActionType = {
-    PREPARE_FOR_INITIAL_FETCH: 'prepareForInitialFetch',
+    PREPARE_FOR_FIRST_PAGE_FETCH: 'prepareForFirstPageFetch',
```
qooban, line 35: This change is just naming change discussed in https://github.com/cloudify-cosmo/cloudify-stage/pull/1311#discussion_r617545146 .

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -180,6 +180,7 @@ export default function DynamicDropdown({
     useFetchTrigger(
         ({ shouldReset: shouldResetOptions }) => {
             if (shouldResetOptions) setOverrideOptionsAfterFetch();
+            dispatchFetchAction({ type: fetchActionType.PREPARE_FOR_FIRST_PAGE_FETCH });
```
qooban, line 183: Small bug-fix: In both cases - after `searchQuery` change and `fetchUrl` change - we should start fetching from the first page.

widgets/common/src/DynamicDropdown.jsx
```diff
@@ -23,24 +23,24 @@ function useFetchTrigger(fetchTrigger, withoutResetFetchDeps, withResetFetchDeps
     const delayedFetchTrigger = useCallback(debounce(fetchTrigger, delayMs), []);
```
qooban, line 24: General:

I changed `setOptions` calls to use function instead of value to be more reliable.

qooban:
System tests scheduled: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/515/

qooban:
FYI there are two more test cases to add (I just realized that we wanted also to verify swagger, so added it to the list). They should be quite short, so I already asked you to review this PR (as it already is quite big). 

I can add those two test cases within separate PR. Feel free to suggest what is your preferred option.

Gelio:
Looks like a hefty PR :smile: I'll take a look at it tomorrow, hopefully it won't take a day :smile:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
 
         cy.wait('@getRequest');
     });
+
+    describe('should handle errors', () => {
```
kubama, line 273: Looks like these tests don't modify the app state. Isn't the global `beforeEach` an overkill in this case? Maybe the whole spec could be reorganized, so that current state resetting setup is not performed for tests that don't really need it?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+        type FilterModalError = {
+            name: string;
+            prepare: () => void;
+            verify: () => void;
+            clean: () => void;
+        };
```
kubama, line 279: I'm not sure if the pattern with `FilterModalError` array is a robust solution. Wouldn't it be better if each validation had it's own test with a proper setup independent of the previous test result?

kubama (commented):
I did a partial review. I'll continue in the morning.

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -17,6 +24,30 @@ describe('Filters widget', () => {
...
+    function saveFilter() {
+        cy.contains('Save').click();
+    }
+
+    function addNewRule() {
+        cy.contains('Add new rule').click();
+    }
+
+    function removeLastRule() {
+        withinTheLastRuleRow(() => cy.get('button[aria-label="Remove rule"]').click());
```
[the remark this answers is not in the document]
qooban: It has `title` defined. OK, switched to use `title` here.

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -25,14 +56,22 @@ describe('Filters widget', () => {
         return cy.contains('.field', 'Filter ID').find('input');
     }
 
+    function withinTheLastRuleRow(fn: (currentSubject: JQuery<HTMLElement>) => void) {
```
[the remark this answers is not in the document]
qooban: OK, changed.

test/cypress/integration/widgets/filters_spec.ts
qooban: Yeah, looking at it now I think I will switch to normal way of splitting the tests in Cypress. Originally it made more sense.

test/cypress/integration/widgets/filters_spec.ts
qooban: As I wrote in the description - `filters_spec` is becoming quite big, so I'll try to reorganize it first, but maybe I'll end up with splitting Filters widget test into two separate files.

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+            }
+        ];
+
+        ruleRowTests.map(ruleRowTest => {
+            const { name, testFilterName, testFilterRules } = ruleRowTest;
+            return it(name, () => {
+                openAddFilterModal();
+                populateFilterRuleRows(testFilterName, testFilterRules);
+                saveAndVerifyFilter(testFilterName, testFilterRules);
+            });
```
kubama, line 698: Similarly to my previous comment on `beforeEach` - I don't think full state reset is needed before each of these tests, maybe the spec could be restructured to avoid it?

Gelio:
Do you plan on addressing the leftover TODO items in this PR? :smile: If they are independent, I would suggest raising a separate PR for them, since we've already started reviewing this one :slightly_smiling_face: This way we get it merged faster

> filters_spec.ts file has >700 LOCs. Do you think we should split it into two files?

Not sure :smile: If it's all related, then I reckon it may as well stay. However, I see that most tests for filters form could be component tests since they are related to error validation :smile: I'm not saying we should be refactoring them (especially that we don't have a good component-testing solution ATM), but I would suggest such a direction in the future :smile: 

This is a very general comment, as I'm not sure how we would go about mocking the network in such tests. Anyway, worth considering in the future, as it could help keep the system test suites smaller :slightly_smiling_face:

widgets/common/src/DynamicDropdown.jsx
Gelio: Hmm... I don't get it :smile: How does removing the `text` from an option save an HTTP request?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -17,6 +24,30 @@ describe('Filters widget', () => {
...
+        cy.contains('Save').click();
+    }
+
+    function addNewRule() {
+        cy.contains('Add new rule').click();
+    }
+
+    function removeLastRule() {
+        withinLastRuleRow(() => cy.get('button[title="Remove rule"]').click());
+    }
```
Gelio: Nice idea to have such functions :+1: 

Optional:

A pattern that could shorten the code could be wrapping the functions in an object:

```ts
const filterModal = {
  openAdd: () => ...,
  openClone: () => ...,
  ...
  removeLastRule: () => ...
};
```

This way it would be clear that those actions (e.g. `addNewRule`) are in the context of the `filterModal` :slightly_smiling_face: 

Maybe only some of them (e.g. the last 3) should make it into the `filterModal` object - I'll leave that decision to you :smile: The rest (`openAdd`, `openClone`, `openEdit`) could remain regular functions, or could be grouped in a `filters` object

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -25,14 +56,22 @@ describe('Filters widget', () => {
         return cy.contains('.field', 'Filter ID').find('input');
     }
 
+    function withinLastRuleRow(fn: (currentSubject: JQuery<HTMLElement>) => void) {
+        cy.get('.fields:last-of-type').within(fn);
+    }
+
+    function withinNthRuleRow(rowNumber: number, fn: (currentSubject: JQuery<HTMLElement>) => void) {
+        cy.get(`.fields:eq(${rowNumber})`).within(fn);
+    }
```
Gelio, line 59: Maybe instead of exposing functions that accept a callback and call `within` themselves, we could expose functions that return that row, and the consumer would call `within` themselves. This way the code that uses the function would look like:

```ts
getNthRuleRow(1).within(...);
```

The benefits are:
1. No need to explicitly type the `curentSubject` and the callback `fn` - less code
2. Consumers can call other functions, not only `within`

Gelio: Also, we could expose those functions on the `filterModal` object that I suggested above :smile:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -109,31 +148,24 @@ describe('Filters widget', () => {
...
 
         cy.get('.modal').within(() => {
-            cy.contains('Save').click();
-            cy.contains('Please provide the filter ID');
-
-            getFilterIdInput().type('csys-invalid');
-            cy.contains('Save').click();
-            cy.contains('All filters with a `csys-` prefix are reserved for internal use');
-
-            getFilterIdInput().clear().type(newFilterName);
```
Gelio: Do we not need this validation anymore? :thinking:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -17,6 +24,30 @@ describe('Filters widget', () => {
...
+    function openAddFilterModal() {
+        cy.contains('Add').click();
+    }
+
+    function openCloneFilterModal() {
+        cy.get('.clone').click();
+    }
+
+    function openEditFilterModal() {
+        cy.get('.edit').click();
```
Gelio, line 30: Optional:

Is there a better selector we could use, e.g. `title` or `aria-label`?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -187,25 +219,19 @@ describe('Filters widget', () => {
...
         cy.get('.modal').within(() => {
             cy.contains(`Clone filter '${filterName}'`);
 
             getFilterIdInput().should('have.value', `${filterName}_clone`);
-
-            getFilterIdInput().clear().type('csys-invalid');
-            typeAttributeRuleValue('test');
-            cy.contains('Save').click();
-            cy.contains('All filters with a `csys-` prefix are reserved for internal use');
-
```
Gelio: Is it no longer needed?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+
+            openCloneFilterModal();
+            cy.get('.modal').within(() => {
+                const cloneModalErrors = [...cloneModalOnlyErrors, ...commonModalErrors];
+                cloneModalErrors.forEach(verifyErrorHandling);
+            });
+        });
+    });
+
+    describe('should allow to define all kinds of filter rules', () => {
```
Gelio: That's the thing that came to my mind when reading this title :smile: 

image

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+        const testPrefix = 'filters_test_form';
+        const blueprintId = `${testPrefix}_blueprint`;
+        const deploymentId = `${testPrefix}_deployment`;
+
+        before(() => {
+            cy.deleteDeployments(testPrefix)
+                .deleteBlueprints(testPrefix)
+                .uploadBlueprint('blueprints/empty.zip', blueprintId)
```
Gelio, line 374: Optional:

I know we have `exampleBlueprintUrl` defined in https://github.com/cloudify-cosmo/cloudify-stage/blob/55d1338950d20806d5b454abf1fc2fb76291a8eb/test/cypress/support/resource_urls.ts#L2

Maybe it would be worth it to also extract this path to the file above to avoid using the same string multiple times in the app?

image

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+            });
+        }
+
+        function selectRuleAttributeValues(
+            ruleRowType: FilterRuleRowType,
+            values: string[],
+            newValues = [] as string[],
+            withAutocomplete = false
+        ) {
+            if (values.length > 0) {
```
Gelio, line 414: If we invert the condition and use an early return, the indentation for the code will be smaller :slightly_smiling_face: 

```ts
if (values.length === 0) {
    return;
}

// code
```

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+                    };
+                    cy.get('div[name="ruleValue"]')
+                        .click()
+                        .within(() => {
+                            values.forEach(value => {
+                                if (withAutocomplete) {
+                                    const searchEndpointRegExp = RegExp(
+                                        `${searchEndpoint[ruleRowType]}.*_search=${value}`
+                                    );
+                                    cy.interceptSp('GET', searchEndpointRegExp).as(`valueSearch_${value}`);
```
Gelio, line 430: Shouldn't we set up the intercept before clicking the dropdown to prevent race conditions?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+                                }
+                                cy.get('input').type(`${value}`);
+                                if (withAutocomplete) cy.wait(`@valueSearch_${value}`);
+                                if (newValues.includes(value)) cy.contains('[role="option"]', 'Add ').click();
+                                else cy.get(`div[option-value="${value}"]`).click();
```
Gelio, line 435: For some reason this code looks very dense and hard to read without the braces and newlines as vertical separators :smile: Could we use braces for the `if` bodies, or separate `if`s by newlines?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+            withinLastRuleRow(() => {
+                cy.get('div[name="ruleOperator"]').click();
+                cy.get(`div[option-value="${operator}"]`).click();
+            });
+        }
+
+        function selectRuleAttributeValues(
+            ruleRowType: FilterRuleRowType,
+            values: string[],
+            newValues = [] as string[],
```
Gelio, line 411: I didn't understand what's the use of `values` vs `newValues` :worried: Could you describe it, perhaps right in the code?

Alternatively, we can make the code/variable names clearer :smile: 

I see we're typing `values` in the search field, and then picking `newValues` from it. Maybe we should rename the parameters to

```suggestion
            valuesToType: string[],
            newValuesToAdd = [] as string[],
```

test/cypress/integration/widgets/filters_spec.ts
qooban: Yes, I'll change it.

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+            }
+        }
+
+        function selectRuleLabelKey(key: string, newKey = false) {
+            withinLastRuleRow(() => {
+                cy.interceptSp('GET', `/labels/deployments?_search=${key}`).as(`keySearch_${key}`);
+                cy.get('div[name="labelKey"]').within(() => {
+                    cy.get('input').type(key);
+                    cy.wait(`@keySearch_${key}`);
+                    if (newKey) cy.contains('[role="option"]', 'New key ').click();
```
Gelio, line 449: Is the space at the end of `New key` significant?

```suggestion
                    if (newKey) cy.contains('[role="option"]', 'New key').click();
```

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+                    else cy.get(`div[option-value="${key}"]`).click();
+                    cy.get(`input.search`).should('not.have.value');
+                });
+            });
+        }
+
+        function selectRuleLabelValues(values: string[], newValues = [] as string[]) {
+            if (values.length > 0) {
```
Gelio, line 457: We could use an early return here :slightly_smiling_face:

test/cypress/integration/widgets/filters_spec.ts
Gelio: Another option would be to avoid accepting 2 separate arrays, and accept an array of objects of the following shape instead:

```ts
{
  value: string;
  isNew?: boolean;
}
```

Then, we co-locate whether some value should be new or it's an existing value :smile:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+                    cy.get('input').type(key);
+                    cy.wait(`@keySearch_${key}`);
+                    if (newKey) cy.contains('[role="option"]', 'New key ').click();
+                    else cy.get(`div[option-value="${key}"]`).click();
+                    cy.get(`input.search`).should('not.have.value');
+                });
+            });
+        }
+
+        function selectRuleLabelValues(values: string[], newValues = [] as string[]) {
```
Gelio, line 456: The same comment about values vs newValues applies here as for `selectRuleAttributeValues`

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+            const anyOperators: FilterRuleOperator[] = [FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf];
+            return anyOperators.includes(operator);
+        }
+        function isFreeTextValueOperator(operator: FilterRuleOperator) {
+            const freeTextOperator: FilterRuleOperator[] = [
+                FilterRuleOperators.Contains,
+                FilterRuleOperators.NotContains,
+                FilterRuleOperators.StartsWith,
+                FilterRuleOperators.EndsWith
+            ];
```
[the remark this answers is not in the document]
qooban: I did it differently - removed const definition.

test/cypress/integration/widgets/filters_spec.ts
Gelio: Nevermind, the request will file after typing `value` regardless :smile:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+                                else cy.get(`div[option-value="${value}"]`).click();
+                                cy.get(`.label[value="${value}"]`).should('exist');
+                            });
+                        });
+                });
+            }
+        }
+
+        function isLabelValueOperator(operator: FilterRuleOperator) {
+            const anyOperators: FilterRuleOperator[] = [FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf];
```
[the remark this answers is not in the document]
qooban: Could be, but I did it differently - removed const definition.

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -243,4 +269,433 @@ describe('Filters widget', () => {
...
+                        .click()
+                        .within(() => {
+                            values.forEach(value => {
+                                if (withAutocomplete) {
+                                    const searchEndpointRegExp = RegExp(
+                                        `${searchEndpoint[ruleRowType]}.*_search=${value}`
+                                    );
+                                    cy.interceptSp('GET', searchEndpointRegExp).as(`valueSearch_${value}`);
+                                }
+                                cy.get('input').type(`${value}`);
```
Gelio, line 432: If I'm not mistaken, we could use

```suggestion
                                cy.get('input').type(value);
```

The same applies in some other `type` calls :slightly_smiling_face:

test/cypress/integration/widgets/filters_spec.ts
qooban: I refactored test structure, please take a look.

qooban:
@Gelio @kubama 

I changed `filters_spec` test structure:
image
as per response to @kubama 's comments.

Please use:
image
to hide not relevant changes.

qooban:
@Gelio 
> Do you plan on addressing the leftover TODO items in this PR? smile If they are independent, I would suggest raising a separate PR for them, since we've already started reviewing this one slightly_smiling_face This way we get it merged faster

No, I'll leave it for the separate PR.

> Not sure smile If it's all related, then I reckon it may as well stay. However, I see that most tests for filters form could be component tests since they are related to error validation smile I'm not saying we should be refactoring them (especially that we don't have a good component-testing solution ATM), but I would suggest such a direction in the future smile

> This is a very general comment, as I'm not sure how we would go about mocking the network in such tests. Anyway, worth considering in the future, as it could help keep the system test suites smaller slightly_smiling_face

I've got two comments:
1. Yes, component testing would help in this case. We definitely need to go in that direction.
2. Having such big files (even if they contain only related content) is not good idea IMHO, they are hard to maintain and read.

Anyways, for now I think I'll not split the file.

test/cypress/integration/widgets/filters_spec.ts
qooban: Done.

test/cypress/integration/widgets/filters_spec.ts
qooban: Done.

Gelio (commented):
The PR was exhausting to review for me :worried: 500 lines of code is a lot. I would appreciate reviewing it in chunks, perhaps section by section or feature by feature, when it comes to tests added. Then, we could also decide early whether the pattern of having "test cases" and helper functions that handle them is right :smile: 

Anyway, this is a full review :slightly_smiling_face: I won't be reviewing the PR again until tomorrow

---

I did not review the last 2 commits. So much has changed, and I am already tired from reviewing this PR that's on the larger side :slightly_smiling_face: 

Generally, when the PR author aims to make significant changes in the PR, I suggest notifying the reviewers in some way, so they do not use their time reviewing code that is already being refactored :slightly_smiling_face: At the moment, I have spent 1h and 15 mins reviewing the PR, and with the file being refactored, it's probably additional 30 minutes to go over the changes :smile:

kubama (dismissed):
I definitely like the new spec structure. It's clean and allows faster, lighter, more suitable setups. Great job 👍.

qooban:
@Gelio 

> The PR was exhausting to review for me worried 500 lines of code is a lot. I would appreciate reviewing it in chunks, perhaps section by section or feature by feature, when it comes to tests added. Then, we could also decide early whether the pattern of having "test cases" and helper functions that handle them is right smile

That's something we can discuss. IMHO the worst thing in general in RD-1758 was that we didn't take all things into account beforehand (story has now 37 points and that is root cause for most of the problems we have right now). Ideally we would have test per case, but as the requirements are changing and sub-tasks are not stable + there is a pressure to finalize the story, I don't see a good solution in such mess.

Also, maybe we should discuss how to review the test code. IMHO Cypress test code is: 1. easier to review than source code, 2. not customer-facing, so no UI tests needed. Personally I'm not reviewing test code with the same focus on details as the source code, but I'm not forcing that should be a rule for all of us. 

> Generally, when the PR author aims to make significant changes in the PR, I suggest notifying the reviewers in some way, so they do not use their time reviewing code that is already being refactored slightly_smiling_face At the moment, I have spent 1h and 15 mins reviewing the PR, and with the file being refactored, it's probably additional 30 minutes to go over the changes smile

Sorry for that. I was not aware when you'll be reviewing the code. Two commits bring changes in the test structure (most of the lines changed within those commits are "white-space" changes). Anyways, I'll try to avoid such situations in the future.

widgets/common/src/DynamicDropdown.jsx
qooban: Change of the `searchQuery` triggered this component (`VisibilitySensor`) to re-render and that called `TRIGGER_FETCH` action.

test/cypress/integration/widgets/filters_spec.ts
qooban: Yes, it was moved to error handling section.

test/cypress/integration/widgets/filters_spec.ts
qooban: Yes, it was moved to error handling section.

test/cypress/integration/widgets/filters_spec.ts
qooban: 🙂 Title changed to be more accurate and provide full sentence with test cases.

test/cypress/integration/widgets/filters_spec.ts
qooban: Make sense, but I don't want to add more to this PR, so I would leave it for the future.

test/cypress/integration/widgets/filters_spec.ts
qooban: I'll add a bit more spacing there.

test/cypress/integration/widgets/filters_spec.ts
qooban: I would like to leave it, it clearly maps to the dropdown setting. Empty space is expected, then comes the key, but as it is inside `<b>...</b>` simply having ``'New key ${newKey}`` did not work.

test/cypress/integration/widgets/filters_spec.ts
qooban: Will change it.

test/cypress/integration/widgets/filters_spec.ts
qooban: Sure, will change it.

test/cypress/integration/widgets/filters_spec.ts
qooban: Good idea with passing array of objects, will implement.

test/cypress/integration/widgets/filters_spec.ts
qooban: Will apply accordingly.

test/cypress/integration/widgets/filters_spec.ts
qooban: Will change it.

test/cypress/integration/widgets/filters_spec.ts
qooban: Actually yes, I'll use `title`.

test/cypress/integration/widgets/filters_spec.ts
qooban: This is nice idea, but I think I'd like to leave it as it is for now not to make more big changes.

test/cypress/integration/widgets/filters_spec.ts
qooban: Hmm... I'm not sure the suggested solution is better. We already used such pattern in other tests and as long as there is no need to get the element, I think it is better.

qooban:
I'm not planning more changes today in this PR. 

I hope I answered all of the comments. In case of re-review I suggest to use `Hide whitespaces changes` (that will save you seeing ~100 line changes). 

Any more optional changes can be implemented within the next PR containing two missing test cases (see PR description).

kubama (dismissed):

qooban:
The most recent full tests execution passed without any errors.
Ref.: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/521

Code coverage for filters looks good:

File |   | Statements |   | Branches |   | Functions |   | Lines
-- | -- | -- | -- | -- | -- | -- | -- | --
widgets/common/src/filters |   | 94.44% | 85/90 | 95.83% | 23/24 | 90% | 36/40 | 96.34% | 79/82
widgets/common/src/filters/inputs |   | 100% | 23/23 | 100% | 4/4 | 100% | 10/10 | 100% | 21/21
widgets/filters/src |   | 100% | 74/74 | 100% | 31/31 | 100% | 26/26 | 100% | 74/74

Ref.: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/521/Overall_20coverage/

test/cypress/integration/widgets/filters_spec.ts
Gelio: I did not expect changes here :smile: It just sounded funny when I read it :slightly_smiling_face:

test/cypress/integration/widgets/filters_spec.ts
Gelio: Oh, so it does matter :slightly_smiling_face: I didn't figure it out just by reading the test's code, but I guess it does not matter much

Gelio:
> Also, maybe we should discuss how to review the test code. IMHO Cypress test code is: 1. easier to review than source code, 2. not customer-facing, so no UI tests needed. Personally I'm not reviewing test code with the same focus on details as the source code, but I'm not forcing that should be a rule for all of us.

I agree and have a similar view on reviewing tests :slightly_smiling_face: The 1h I spent was with that mindset :smile:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -63,184 +101,612 @@ describe('Filters widget', () => {
...
+                cy.get('div[name="ruleValue"]')
+                    .click()
+                    .within(() => {
+                        values.forEach(({ value, isNew }) => {
+                            if (withAutocomplete) {
+                                const searchEndpointRegExp = RegExp(`${searchEndpoint[ruleRowType]}.*_search=${value}`);
+                                cy.interceptSp('GET', searchEndpointRegExp).as(`valueSearch_${value}`);
+                            }
+
+                            cy.get('input').type(`${value}`);
```
Gelio, line 434: Ha! I've found something to improve in the next PR

```suggestion
                            cy.get('input').type(value);
```

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -63,184 +101,612 @@ describe('Filters widget', () => {
...
+                            if (withAutocomplete) cy.wait(`@valueSearch_${value}`);
+
+                            if (isNew) cy.contains('[role="option"]', 'Add ').click();
+                            else cy.get(`div[option-value="${value}"]`).click();
```
Gelio, line 438: Reads much easier to me :ok_hand:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -63,184 +101,612 @@ describe('Filters widget', () => {
...
+            cy.log('Filter rules form population verification');
+            cy.getSearchInput().clear().type(filterId);
+            cy.get('.loading').should('not.exist');
+            openEditFilterModal();
+            cy.get('.modal').within(() => verifyRulesForm(testRules));
+        }
+
+        interface ExtendedFilterRule extends FilterRule {
+            newKey?: boolean;
+            newValues?: string[];
```
Gelio, line 573: Could we apply a pattern of removing `newValues` array and using a `isNew` property on `values`, similarly to what was already done for `RuleValueObject`?

In order to do that, we will probably need to use the `Omit` type (`Omit<FilterRule, 'value'> & { value: ... }`)

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -63,184 +101,612 @@ describe('Filters widget', () => {
...
+                            else cy.get(`div[option-value="${value}"]`).click();
+
+                            cy.get(`.label[value="${value}"]`).should('exist');
+                        });
+                    });
+            });
+        }
 
-        cy.wait('@usageRequest');
+        function selectRuleLabelKey(key: string, isNew = false) {
```
Gelio, line 446: General:

All these helper functions IMO are candidates to be extracted, maybe to a separate file, maybe to some other scope (perhaps a IIFE that returns an object in the top-level module scope). This way the actual tests will be shorter

Gelio (approved):
This time I put a lot less effort into reviewing the PR. Looks good :+1: All the comments I have left could be implemented in future PRs or never :smile:

test/cypress/integration/widgets/filters_spec.ts
qooban: Oh, I thought I already change that. Will fix in the next PR definitely 👍🏻

test/cypress/integration/widgets/filters_spec.ts
qooban: Hmm... Yeah, I was considering that. I was afraid won't it make the type too complex? 
That's the question from TS-beginner perspective, so probably not if you are suggesting that. I'll reconsider it in the next PR.

test/cypress/integration/widgets/filters_spec.ts
qooban: I'll keep that in mind for the next PR.

qooban:
Thanks for the reviews and approvals. I'm going to merge this PR once CI passes.

qooban:
The last CI run failed on size limit check:
```
[2021-04-29T11:47:29.860Z]   Application package
[2021-04-29T11:47:29.860Z]   Package size limit has exceeded by 96 B
[2021-04-29T11:47:29.860Z]   Size limit:   13946060 B
[2021-04-29T11:47:29.860Z]   Size:         13946156 B
[2021-04-29T11:47:29.860Z]   Loading time: 272.4 s    on slow 3G
```

Must have been very tight (just 96 B over the limit 🙂), as in this PR I don't introduce much new code.
Updated size-limit within https://github.com/cloudify-cosmo/cloudify-stage/pull/1313/commits/abc92d0245858be3c030432f2ebb777e4a458143 . 

@Gelio @kubama Please re-review once CI pass.

kubama (approved):

test/cypress/integration/widgets/filters_spec.ts
Gelio: That would not be overly complex from my side :smile: Using such type operations is more of an intermediate TS feature, so I can see how that's new and could be seen as complex, but if I was to write this code, I would apply it, especially that it also has readability benefits in terms of JS code :slightly_smiling_face: 

Overall I encourage learning such type operators as `Omit`, `Exclude`, `Partial`, `Extract`, `Required`, etc. Those are quite short but give a lot of flexibility :slightly_smiling_face:

.size-limit
```diff
@@ -20,7 +20,7 @@
     },
     {
         "name": "Application package",
-        "limit": "13.3 MB",
+        "limit": "13.35 MB",
```
Gelio, line 23: Thanks for using the same size I did in #1335 :smile: Maybe we avoid a conflict this way :crossed_fingers:

Gelio (approved):

PR #1315, opened 2021-04-15
RD-2007 - Improve styling for Label type inputs

qooban:
### Before
image

### After
image

### DIFF
--- a/widgets/common/src/filters/RuleRow.tsx
+++ b/widgets/common/src/filters/RuleRow.tsx
@@ -18,6 +18,8 @@ interface RuleRowProps {
 const defaultOperator = FilterRuleOperators.AnyOf;
 const defaultValues: string[] = [];
 const defaultOperatorAndValues = { operator: defaultOperator, values: defaultValues };
+const dropdownFieldsWidth = 5;
+const removeButtonFieldWidth = 1;
 
 const RuleRow: FunctionComponent<RuleRowProps> = ({ onChange, onRemove, removable, rule, toolbox }) => {
     const { UnsafelyTypedFormField: FormField, UnsafelyTypedFormGroup: FormGroup } = Stage.Basic;
@@ -46,13 +48,13 @@ const RuleRow: FunctionComponent<RuleRowProps> = ({ onChange, onRemove, removabl
 
     return (
         <FormGroup widths="equal">
-            <FormField>
+            <FormField width={dropdownFieldsWidth}>
                 <RuleRowTypeDropdown onChange={onRuleTypeChange} value={ruleType} />
             </FormField>
-            <FormField>
+            <FormField width={dropdownFieldsWidth}>
                 <RuleOperatorDropdown onChange={onOperatorChange} value={operator} ruleType={ruleType} />
             </FormField>
-            <FormField>
+            <FormField width={dropdownFieldsWidth}>
                 <RuleValueInput
                     onKeyChange={onKeyChange}
                     onValuesChange={onValuesChange}
@@ -61,7 +63,7 @@ const RuleRow: FunctionComponent<RuleRowProps> = ({ onChange, onRemove, removabl
                     toolbox={toolbox}
                 />
             </FormField>
-            <FormField>{removable && <RuleRemoveButton onClick={onRemove} />}</FormField>
+            <FormField width={removeButtonFieldWidth}>{removable && <RuleRemoveButton onClick={onRemove} />}</FormField>
         </FormGroup>
     );
 };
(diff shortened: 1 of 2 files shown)


qooban:
I'm not planning to run full system tests execution. This change affects only `filters_spec`. Executed locally, all tests passed.

kubama:
I have noticed a minor issue:
image

kubama:
Looks like label/attribute and operator dropdowns are now even wider. I think it's a waste of space as there is no option that would require these dropdowns to be that wide. On the other hand the key/value inputs may contain variable length, multiple items. So it would be good if they were wider.

qooban:
@kubama 

> I have noticed a minor issue:
I don't think it is related to this PR. I have seen it before and I'm almost sure it is long with us. I suspect it comes from Semantic-UI-React.
See for example: https://codesandbox.io/embed/semantic-ui-react-example-xghj2 :

https://user-images.githubusercontent.com/5202105/114867185-c2c1e480-9df4-11eb-8666-f0b73eb02d9c.mp4



> Looks like label/attribute and operator dropdowns are now even wider. I think it's a waste of space as there is no option that would require these dropdowns to be that wide. On the other hand the key/value inputs may contain variable length, multiple items. So it would be good if they were wider.

Good point 👍🏻 Makes perfect sense, so I adapted the widths a bit and got something like:

image

What do you think?

widgets/common/src/filters/RuleRow.tsx
```diff
@@ -61,7 +63,7 @@ const RuleRow: FunctionComponent<RuleRowProps> = ({ onChange, onRemove, removabl
                     toolbox={toolbox}
                 />
             </FormField>
-            <FormField>{removable && <RuleRemoveButton onClick={onRemove} />}</FormField>
+            <FormField width={removeButtonFieldWidth}>{removable && <RuleRemoveButton onClick={onRemove} />}</FormField>
```
Gelio, line 67: Optional:

To me, seeing `width={1}` would be as understandable as the current solution, but would provide me with more immediate information and not having to jump to some other place to know the value :smile: 

When it comes to styling, I don't see this as a _magic value_, so IMO hardcoding it here, if it's used once, sounds reasonable :slightly_smiling_face: it's optional, though

Gelio: The same goes for `7` used as `valueInputFieldWidth` :smile:

Gelio (dismissed):
Looks good from my side :+1:

widgets/common/src/filters/RuleRow.tsx
qooban: Yep, make sense not to over-complicate it. Originally I had just two consts and then thought it's a bit better, but in that case (3 different consts) it makes it more difficult to read. Changed.

kubama:
> I don't think it is related to this PR. I have seen it before and I'm almost sure it is long with us. I suspect it comes from Semantic-UI-React. See for example: https://codesandbox.io/embed/semantic-ui-react-example-xghj2 :

I have seen it for the first time, but if this is the case then it's of course out of scope of this PR.

> What do you think?

Of course it's better, but it looks like type and operator dropdowns are still wider than needed. Is there any reason for not making them as narrow as possible? In the example screenshot you've given it could make the blueprint value input render without the scroll bar, so I think there is much to gain.

qooban:
> I have seen it for the first time, but if this is the case then it's of course out of scope of this PR.

In this matter I'm 100% sure it is not the first time I see it.
Created: RD-2058 for that and added a comment in https://github.com/Semantic-Org/Semantic-UI-React/issues/3720 .

> Of course it's better, but it looks like type and operator dropdowns are still wider than needed. Is there any reason for not making them as narrow as possible? In the example screenshot you've given it could make the blueprint value input render without the scroll bar, so I think there is much to gain.

I already tried to do so.

With:
```
            <FormField width={3}>
                <RuleRowTypeDropdown onChange={onRuleTypeChange} value={ruleType} />
            </FormField>
            <FormField width={3}>
                <RuleOperatorDropdown onChange={onOperatorChange} value={operator} ruleType={ruleType} />
            </FormField>
```

I get:

image

I suspect that maybe `Dropdown` have a minimal width set somehow, so it would not be so easy to avoid that. If you have a solution for that, please let me know.

kubama:
> I suspect that maybe `Dropdown` have a minimal width set somehow, so it would not be so easy to avoid that. If you have a solution for that, please let me know.

I don't have a solution. In this case I guess we have no option other than to leave it as it is.

kubama (approved):

qooban:
As this one is small and I applied already @Gelio 's comments, I'll merge it with one approval.
Thanks for the reviews and the discussion around it 👍🏻

PR #1316, opened 2021-04-15
WDR-13 Added filters list refetch after successful edit

kubama:
System tests build: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/459/

### DIFF
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -152,7 +152,7 @@ describe('Filters widget', () => {
 
         cy.get('.modal').should('not.exist');
         cy.log('Verify filters list is refetched immediatelly');
-        cy.wait('@filtersRequest', { timeout: 1000 });
+        cy.wait('@filtersRequest', { requestTimeout: 1000 });
     });
 
     it('should allow to clone existing filter', () => {
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -145,11 +145,14 @@ describe('Filters widget', () => {
             modifyBlueprintRule();
 
             cy.interceptSp('PATCH', `/filters/deployments/${filterName}`).as('rulesRequest');
+            cy.interceptSp('GET', `/filters/deployments`).as('filtersRequest');
             cy.contains('Save').click();
             checkRequestRules();
         });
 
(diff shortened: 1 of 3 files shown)


test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -145,11 +145,14 @@ describe('Filters widget', () => {
...
 
             cy.interceptSp('PATCH', `/filters/deployments/${filterName}`).as('rulesRequest');
+            cy.interceptSp('GET', `/filters/deployments`).as('filtersRequest');
             cy.contains('Save').click();
             checkRequestRules();
         });
 
         cy.get('.modal').should('not.exist');
+        cy.log('Verify filters list is refetched immediatelly');
+        cy.wait('@filtersRequest', { timeout: 1000 });
```
qooban, line 155: Are you sure about it? Wouldn't it make our test flaky in case of network delays?

widgets/filters/src/FilterModal.tsx
```diff
@@ -39,7 +39,12 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
             return;
         }
 
-        onSubmit(filterId, filterRules).catch(setMessageAsError);
+        onSubmit(filterId, filterRules)
+            .then(() => {
+                toolbox.loading(true);
+                toolbox.refresh();
+            })
```
qooban, line 46: Maybe I'm wrong, but IIRC you implemented recently just a `toolbox.refresh` call somewhere else and when I asked about lack of `toolbox.loading` you said that it is triggered automatically. Am I missing something or we can avoid calling `toolbox.loading(true)` here?

qooban (commented):
Commented.

widgets/filters/src/FilterModal.tsx
kubama: My understanding was that `toolbox.loading(false)` is called after the refresh, but after checking `toolbox.loading(true)` is also called at the beginning. So I'm removing `toolbox.loading(true)` from here.

test/cypress/integration/widgets/filters_spec.ts
kubama: Good point, I changed it to `requestTimeout` so `responseTimeout` is+ not changed.

qooban (approved):

PR #1317, opened 2021-04-15
RD-1824 Prevent going to parent page on deployment delete

Gelio:
Deleting a deployment in the Deployments View widget no longer triggers a redirect to the parent page.

The behavior is changed only when the _Deployment action buttons_ are displayed in the Deployments View widget. I have confirmed that redirecting to the parent page still happens when deleting a deployment from the drilled-down view of the old Deployments widget :slightly_smiling_face: (where it makes sense)

One caveat with deleting deployments is that the deployment is still present in the LIST request that follows the delete request. This is because the backend responds to the DELETE request and then deletes the deployment in the background, asynchronously, so the user will still see the deployment being deleted. The test still passes, since the deployment is gone with the next polling request. I've created RD-2057 to address it in the backend

The PR also contains some TS migrations of files I touched or were in the vicinity :slightly_smiling_face: Looks like git did not detect some of them as renames :worried: Reviewing commit-by-commit looks to be reasonable here

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/460/pipeline

Ran 2021-04-15 13:54 CEST

### DIFF
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -352,12 +352,19 @@ describe('Deployments View widget', () => {
             });
         });
 
-        it('should support the drill-down workflow', () => {
+        const useEnvironmentsWidget = () => {
             useDeploymentsViewWidget({
                 configurationOverrides: {
                     filterId: 'csys-environment-filter'
                 }
             });
+        };
+        const getSubenvironmentsButton = () => cy.get('button').contains('Subenvironments');
+        const getSubservicesButton = () => cy.get('button').contains('Services');
+        const getBreadcrumbs = () => cy.get('.breadcrumb');
+
+        it('should support the drill-down workflow', () => {
+            useEnvironmentsWidget();
 
             getDeploymentsViewTable().within(() => {
                 cy.log('Only top-level environments should be visible');
@@ -367,10 +374,6 @@ describe('Deployments View widget', () => {
                 cy.contains('db-2').should('not.exist');
                 cy.contains('web-app').should('not.exist');
             });
-
-            const getSubenvironmentsButton = () => cy.get('button').contains('Subenvironments');
-            const getSubservicesButton = () => cy.get('button').contains('Services');
-            const getBreadcrumbs = () => cy.get('.breadcrumb');
 
             getDeploymentsViewDetailsPane().within(() => {
                 getSubservicesButton().contains('1');
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -359,8 +359,8 @@ describe('Deployments View widget', () => {
                 }
             });
         };
-        const getSubenvironmentsButton = () => cy.get('button').contains('Subenvironments');
-        const getSubservicesButton = () => cy.get('button').contains('Services');
+        const getSubenvironmentsButton = () => cy.contains('button', 'Subenvironments');
+        const getSubservicesButton = () => cy.contains('button', 'Services');
         const getBreadcrumbs = () => cy.get('.breadcrumb');
 
         it('should support the drill-down workflow', () => {
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -430,6 +430,31 @@ describe('Deployments View widget', () => {
                 cy.contains('db-env').should('not.exist');
             });
         });
+
+        it('should allow deleting a deployment without redirecting to the parent page', () => {
+            const tempDeploymentId = `${specPrefix}_temp_deployment_to_remove`;
+            const parentDeploymentId = getDeploymentFullName('app-env');
+            cy.deployBlueprint(blueprintName, tempDeploymentId).setLabels(tempDeploymentId, [
+                { 'csys-obj-type': 'service' },
+                { 'csys-obj-parent': parentDeploymentId }
+            ]);
+
+            useEnvironmentsWidget();
+
+            getDeploymentsViewDetailsPane().within(() => getSubservicesButton().click());
+
+            getBreadcrumbs().contains(parentDeploymentId);
+
+            getDeploymentsViewTable().within(() => cy.contains(tempDeploymentId).click());
+
+            cy.log('Delete the deployment');
+            getDeploymentsViewDetailsPane().contains('Deployment actions').click();
+            cy.get('.popup').contains('Delete').click();
+            cy.get('.modal').contains('Yes').click();
+
+            getDeploymentsViewTable().within(() => cy.contains(tempDeploymentId).should('not.exist'));
+            getBreadcrumbs().contains(parentDeploymentId);
+        });
     });
 
     it('should display an error message when using the drilled-down widget on a top-level page', () => {
--- a/widgets/common/src/DeploymentActionsModals.tsx
+++ b/widgets/common/src/DeploymentActionsModals.tsx
@@ -7,13 +7,15 @@ interface DeploymentActionsModalsProps {
     deploymentId: string;
     onHide: () => void;
     toolbox: Stage.Types.Toolbox;
+    redirectToParentPageAfterDelete: boolean;
 }
 
 const DeploymentActionsModals: FunctionComponent<DeploymentActionsModalsProps> = ({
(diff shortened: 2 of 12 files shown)


qooban:
I'll review this PR tomorrow morning.

widgets/common/src/DeploymentActionsModals.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    const commonProps = { deploymentId, open: true, onHide, toolbox };
+
+    switch (activeAction) {
+        case actions.manageLabels:
+            return <ManageLabelsModal {...commonProps} />;
+        case actions.install:
+            return <ExecuteDeploymentModal {...commonProps} workflow="install" />;
+        case actions.uninstall:
+            return <ExecuteDeploymentModal {...commonProps} workflow="uninstall" />;
```
kubama, line 43:
```suggestion
        case actions.install:
        case actions.uninstall:
            return <ExecuteDeploymentModal {...commonProps} workflow={activeAction} />;
```

widgets/common/src/DeploymentActionsModals.tsx
```diff
@@ -0,0 +1,81 @@
...
+    deploymentId: PropTypes.string.isRequired,
+    onHide: PropTypes.func.isRequired,
+    // NOTE: `as any` assertion since Toolbox from PropTypes and TS slightly differ
+    toolbox: Stage.PropTypes.Toolbox.isRequired as any,
+    redirectToParentPageAfterDelete: PropTypes.bool.isRequired
+};
+
+declare global {
+    namespace Stage.Common {
+        // eslint-disable-next-line import/prefer-default-export
```
kubama, line 73: Digression:

I'm seeing this rule being disabled frequently in different contexts. Maybe another candidate for global disable?

widgets/deploymentActionButtons/src/widget.tsx
```diff
@@ -0,0 +1,71 @@
...
+    fetchData(_widget, toolbox, { id }) {
+        if (!id) {
+            return Promise.resolve({ id: '', workflows: [] });
+        }
+
+        const { DeploymentActions } = Stage.Common;
+        const actions = new DeploymentActions(toolbox);
+
+        toolbox.loading(true);
+        return actions.doGetWorkflows(id).finally(() => toolbox.loading(false));
```
kubama, line 44: I know it's existing code, but it happened to catch my eye - I think calling `toolbox.loading(true)` and then `toolbox.loading(false)` is not needed here as these are called anyway in the context from which `fetchData` function is called. Anyway, this is optional comment.

widgets/deploymentActionButtons/src/widget.tsx
```diff
@@ -0,0 +1,71 @@
...
+    },
+
+    fetchParams(_widget, toolbox): { id: string | null } {
+        const deploymentId = toolbox.getContext().getValue('deploymentId');
+        // Deployment Actions Buttons widget does not support multiple actions, thus picking only one deploymentId
+        const firstDeploymentId: string | undefined | null = _.castArray(deploymentId)[0];
+
+        return { id: firstDeploymentId };
+    },
```
kubama, line 53: I'm wondering how this works - according to its type `firstDeploymentId` can be `undefined`, but still `{ id: firstDeploymentId }` is of type `{ id: string | null }`?

widgets/deploymentActionButtons/src/widget.tsx
```diff
@@ -31,19 +33,19 @@ Stage.defineWidget({
         return actions.doGetWorkflows(id).finally(() => toolbox.loading(false));
     },
 
-    fetchParams(widget, toolbox) {
+    fetchParams(_widget, toolbox): { id: string | null } {
         const deploymentId = toolbox.getContext().getValue('deploymentId');
         // Deployment Actions Buttons widget does not support multiple actions, thus picking only one deploymentId
-        const firstDeploymentId = _.castArray(deploymentId)[0];
+        const firstDeploymentId: string | undefined | null = _.castArray(deploymentId)[0];
```
qooban, line 50: You can also import `castArray` from `lodash` and use it here.

qooban (approved):
Looks good 👍🏻 
One minor comment from my side.

widgets/common/src/DeploymentActionsModals.tsx
Gelio: Definitely, thanks for bringing it up, it makes sense :smile: In fact, I'm in favor of always using named exports, and dropping default exports. My points can be summed up by https://basarat.gitbook.io/typescript/main-1/defaultisbad :smile: 

I'll add 2 points to the weekly's agenda

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -352,12 +352,19 @@ describe('Deployments View widget', () => {
...
-        it('should support the drill-down workflow', () => {
+        const useEnvironmentsWidget = () => {
             useDeploymentsViewWidget({
                 configurationOverrides: {
                     filterId: 'csys-environment-filter'
                 }
             });
+        };
+        const getSubenvironmentsButton = () => cy.get('button').contains('Subenvironments');
+        const getSubservicesButton = () => cy.get('button').contains('Services');
```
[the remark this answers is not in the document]
Gelio: Very nice, thanks for the tip :+1:

widgets/common/src/DeploymentActionsModals.tsx
Gelio: Looks reasonable :+1: That's copy-pasted code, I'm not sure whether there was a reason before to have 2 separate cases, but definitely makes sense

It would be nice to have it fully type-safe by migrating `ExecuteDeploymentModal` and `DeploymentActionsMenu` to TS, but I guess we can live with it for now

widgets/deploymentActionButtons/src/widget.tsx
Gelio: I believe you are correct, that also caught my eye. I'll double-check and remove the calls if it is indeed redundant

widgets/deploymentActionButtons/src/widget.tsx
Gelio: Oh yeah... Definitely, something weird's going on here. I should have reused the `WidgetParams` type on L47 and L50, but I'm curious why it's working without a problem :thinking: 

Looks like TypeScript is discarding the `undefined` variant from the union of type for `firstDeploymentId`:
image

If I use the `as` type assertion instead of a type annotation for `firstDeploymentId`, this rightfully reports an error:

image

Anyway, I'll reuse the `WidgetParams` type so it always matches

widgets/deploymentActionButtons/src/widget.tsx
Gelio: Yes, definitely :smile: I've fallen prey to the thing I'm warning against :smile: Thanks for catching that

widgets/deploymentActionButtons/src/widget.tsx
Gelio: Yup, to confirm, `loading` is set in https://github.com/cloudify-cosmo/cloudify-stage/blob/5b0998d41b082f68add800973900815f19f4b0aa/app/components/WidgetDynamicContent.jsx#L177 which in the end call those 2: https://github.com/cloudify-cosmo/cloudify-stage/blob/5b0998d41b082f68add800973900815f19f4b0aa/app/components/WidgetDynamicContent.jsx#L105-L114

And then it is hidden by https://github.com/cloudify-cosmo/cloudify-stage/blob/5b0998d41b082f68add800973900815f19f4b0aa/app/components/WidgetDynamicContent.jsx#L206 which then calls https://github.com/cloudify-cosmo/cloudify-stage/blob/5b0998d41b082f68add800973900815f19f4b0aa/app/components/WidgetDynamicContent.jsx#L116-L122

kubama (approved):

Gelio:
Thanks for the review :smile: I've addressed all comments, so I'm looking for a re-review :slightly_smiling_face: 

I'm surprised, as the system tests actually passed, meaning the coverage is finally above the threshold :smile: 

I don't plan on rerunning the system tests again after having addressed the comments - I've just rerun the deployments view spec locally, and it is still working :slightly_smiling_face:

qooban (approved):

PR #1318, opened 2021-04-16
RD-1123 Added support for system filters

kubama:
image


kubama:
System tests build: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/466/

widgets/filters/src/FiltersTable.tsx
```diff
@@ -56,35 +56,49 @@ const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, wid
...
                             <Icon
                                 name="clone"
                                 link
                                 bordered
                                 title={i18n.t('widgets.filters.columns.actions.clone')}
                                 onClick={() => setFilterToClone(filter)}
                             />
+                            <Icon
+                                name="edit"
+                                link={!filter.is_system_filter}
```
Gelio, line 82: General:

It is important that we toggle the `link` attribute too, not only `disabled`?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -67,7 +71,28 @@ describe('Filters widget', () => {
                 expect(tableData[0]['Filter name']).to.eq(filterName);
                 expect(tableData[0].Creator).to.eq('admin');
                 expect(tableData[0].Created).not.to.be.null;
+                expect(tableData[0].Type).to.eq('user');
+            });
+
+        const systemFilterName = 'csys-environment-filter';
+        cy.get('input[placeholder="Search..."]').clear().type(systemFilterName);
```
Gelio, line 78: Optional:

I've noticed that the selector for `search` appears in the tests multiple times (I've seen 3), and is hardcoded each time. Perhaps it would be useful to extract it into the `support` directory, or add a command for typing in the search phrase?

image

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -67,7 +71,28 @@ describe('Filters widget', () => {
...
+                expect(tableData[0].Creator).to.eq('admin');
+                expect(tableData[0].Created).not.to.be.null;
+                expect(tableData[0].Type).to.eq('system');
             });
+
+        const disabledIconTitle = "System filter can't be edited or deleted";
+        cy.get('.edit').should('have.class', 'disabled');
+        cy.get('.edit').should('have.prop', 'title', disabledIconTitle);
+        cy.get('.trash').should('have.class', 'disabled');
+        cy.get('.trash').should('have.prop', 'title', disabledIconTitle);
```
Gelio, line 95: I believe it would be worth it to verify that the Clone icon is enabled :slightly_smiling_face:

Gelio (approved):
Looks solid :+1:

widgets/filters/src/FiltersTable.tsx
kubama: Yes, with `disabled` only the icon does not react on click, but it still changes its style on hover, which suggests that it actually is clickable. Disabling the `link` prop solves the issue.

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -81,15 +106,18 @@ describe('Filters widget', () => {
...
-        cy.contains('Save').click();
-        cy.contains('Please provide the filter ID');
+        cy.get('.modal').within(() => {
+            cy.contains('Save').click();
+            cy.contains('Please provide the filter ID');
 
-        const newFilterName = `${filterName}_added`;
+            getFilterIdInput().type('csys-invalid');
+            cy.contains('Save').click();
+            cy.contains('All filters with a `csys-` prefix are reserved for internal use');
```
qooban, line 118: This message comes from the REST API right now, it's not our popup (at least for now), right?

widgets/filters/src/types.ts
```diff
@@ -10,6 +10,8 @@ export interface Filter {
     labels_filter_rules: FilterRule[];
     // eslint-disable-next-line camelcase
     attrs_filter_rules: FilterRule[];
+    // eslint-disable-next-line camelcase
+    is_system_filter: boolean;
```
qooban, line 14: Optional:

Maybe worth to have camelcase disabled for the whole `interface Filter` block?

widgets/filters/src/FiltersTable.tsx
```diff
@@ -56,35 +56,49 @@ const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, wid
                 searchable
```
qooban, line 56: Optional: 

Adding comment here as it is not possible to add comments to the binary files (or at least I don't see how to do it).

I know it's not a big gain in terms of size, but you can optionally use https://compresspng.com/ to make `widget.png` a ~50% smaller:
image

qooban (commented):
Added few minor/optional comments.

widgets/filters/src/FiltersTable.tsx
Gelio: Good to know, thanks :+1:

test/cypress/integration/widgets/filters_spec.ts
kubama: Yes. In this case I wanted to test if REST errors that occur during save operation are reported in the UI.

widgets/filters/src/types.ts
kubama: It's a temporary solution - this will be disabled on directory level once REST API types are moved to a common location.

widgets/filters/src/FiltersTable.tsx
```diff
@@ -56,35 +56,49 @@ const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, wid
...
                 sortable
             >
-                <DataTable.Column width="33%" label={i18n.t('widgets.filters.columns.name')} name="id" />
-                <DataTable.Column width="33%" label={i18n.t('widgets.filters.columns.creator')} name="created_by" />
-                <DataTable.Column width="33%" label={i18n.t('widgets.filters.columns.created')} name="created_at" />
+                <DataTable.Column width="60%" label={i18n.t('widgets.filters.columns.name')} name="id" />
```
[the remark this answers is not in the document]
kubama: Done

test/cypress/integration/widgets/filters_spec.ts
kubama: Done

test/cypress/integration/widgets/filters_spec.ts
kubama: Done

widgets/filters/src/FiltersTable.tsx
kubama: Done

qooban (approved):

Gelio (approved):

qooban:
@kubama I know this PR has already been merged, but I was thinking about few things:

1. How is the filter type handled on the backend side? Is there an endpoint ready for setting the filter type? Probably there's no such way as it was meant to be set only during installation, but maybe you know more. Probably it's completely blocked on the REST API level, right?

2. Regarding presentation I was thinking that maybe better way to show the information if the filter is system or not (`is_system_filter` value from the API) would be something similar to what we do with for example `Is Active` or `Is Admin` in Users Management widget or `Hidden Value` in Secrets Store Management widget. We could use checkbox (can be disabled if there is no option to change the value from the UI side) and name column "System Filter" or "System" or "Is System". In general IMHO that's probably better way for presenting table column with boolean values. At least we would get UX consistency across the application.

kubama:
> 1. How is the filter type handled on the backend side? Is there an endpoint ready for setting the filter type? Probably there's no such way as it was meant to be set only during installation, but maybe you know more. Probably it's completely blocked on the REST API level, right?

There is no possibility to create system filter via REST API. There is also no way to switch the type. So the flag is read-only.

> 2. Regarding presentation I was thinking that maybe better way to show the information if the filter is system or not (`is_system_filter` value from the API) would be something similar to what we do with for example `Is Active` or `Is Admin` in Users Management widget or `Hidden Value` in Secrets Store Management widget. We could use checkbox (can be disabled if there is no option to change the value from the UI side) and name column "System Filter" or "System" or "Is System". In general IMHO that's probably better way for presenting table column with boolean values. At least we would get UX consistency across the application.

If this would be more consistent then yes, it's a good idea. We can still introduce the change.

PR #1319, opened 2021-04-19
RD-1123 Changed Type column to boolean System column

kubama:
image

I'm not planning system tests execution - I run the updated test locally and it passes.

### DIFF
--- a/widgets/filters/src/FiltersTable.tsx
+++ b/widgets/filters/src/FiltersTable.tsx
@@ -29,7 +29,7 @@ function tColumn(columnKey: string) {
 
 const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, widget }) => {
     const { i18n } = Stage;
-    const { Alert, Button, Confirm, DataTable, Icon, List } = Stage.Basic;
+    const { Alert, Button, Checkbox, Confirm, DataTable, Icon, List } = Stage.Basic;
     const { Time } = Stage.Utils;
     const { useResettableState, useRefreshEvent, useBoolean } = Stage.Hooks;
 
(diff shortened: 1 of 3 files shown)


widgets/filters/src/FiltersTable.tsx
```diff
@@ -63,15 +63,15 @@ const FiltersTable: FunctionComponent<FiltersTableProps> = ({ data, toolbox, wid
...
                 <DataTable.Column width="112px" />
                 {data.filters.map(filter => (
                     <DataTable.Row key={filter.id}>
                         <DataTable.Data>{filter.id}</DataTable.Data>
                         <DataTable.Data>{filter.created_by}</DataTable.Data>
                         <DataTable.Data>{Time.formatTimestamp(filter.created_at)}</DataTable.Data>
-                        <DataTable.Data>
-                            {i18n.t(`widgets.filters.type.${filter.is_system_filter ? 'system' : 'user'}`)}
+                        <DataTable.Data className="center aligned">
+                            <Checkbox checked={filter.is_system_filter} disabled />
```
qooban, line 74: Optional/General 🙂 

Looking at the search:
image
that could probably be extracted to a separate component (sth like `CenteredData` or `CenterAlignedData`), but:
1. I'm not sure `DataTable` would be able to pick it up and handle it properly (not sure if it does not require to have `DataTable.Data` just under `Data.TableRow` component)
2. It's out of the scope of this PR

qooban (approved):
All good 👍🏻

widgets/filters/src/FiltersTable.tsx
kubama: Personally I would rather try to add a new prop to `DataTable.Data` that would set these classes.

qooban: Yep, make sense. Created RD-2083.

PR #1321, opened 2021-04-20
RD-2089 Add deployments map placeholder

Gelio:
The purpose of the PR is to prepare the structure for the Deployments Map in the Deployments View widget. The actual map will be implemented in RD-2090

## Changes

1. Change the scrollbar behavior in the Deployments View widget

    Each "component" (details, table, map) has its own area in a grid. Details and table have their own scrollbars

2. Add a button to toggle the map
3. Add map widget configuration
4. Add tests for the added functionalities'

## Video

https://user-images.githubusercontent.com/889383/115364630-dfc73080-a1c3-11eb-92e1-f3a193c523fe.mp4

## System tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/475/

Ran 2021-04-20 10:31 CEST

### DIFF
--- a/widgets/common/src/deploymentsView/layout.ts
+++ b/widgets/common/src/deploymentsView/layout.ts
@@ -18,9 +18,8 @@ export const DeploymentsViewHeaderContainer = styled.div`
     justify-content: flex-end;
 `;
 
-export const DeploymentsMapContainer = styled.div`
-    // TODO(RD-1225): consume height from widget configuration
-    height: 400px;
+export const DeploymentsMapContainer = styled.div<{ height: number }>`
+    height: ${props => props.height}px;
     grid-area: map;
 `;
 
--- a/widgets/common/src/deploymentsView/layout.ts
+++ b/widgets/common/src/deploymentsView/layout.ts
@@ -3,12 +3,28 @@ import styled from 'styled-components';
 export const DeploymentsViewContainer = styled.div`
     height: 100%;
     display: grid;
+    grid-template-areas:
+        'header header'
+        'map map'
+        'table details';
+    grid-template-rows: min-content min-content minmax(250px, 1fr);
     grid-template-columns: minmax(50%, 1fr) 1fr;
+`;
+
+export const DeploymentsViewHeaderContainer = styled.div`
+    grid-area: header;
+`;
+
+export const DeploymentsMapContainer = styled.div`
+    // TODO(RD-1225): consume height from widget configuration
+    height: 400px;
+    grid-area: map;
 `;
 
 export const DeploymentsTableContainer = styled.div`
     height: 100%;
     overflow: auto;
+    grid-area: table;
 
     .gridTable {
         height: 100%;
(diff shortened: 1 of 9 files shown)


widgets/common/src/deploymentsView/layout.ts
```diff
@@ -0,0 +1,48 @@
+import styled from 'styled-components';
+
+export const DeploymentsViewContainer = styled.div`
+    height: 100%;
+    display: grid;
+    grid-template-areas:
+        'header header'
+        'map map'
+        'table details';
+    grid-template-rows: min-content min-content minmax(250px, 1fr);
```
Gelio, line 10: Note that the bottom row (table & details) will always be 250px high, or more. This is to prevent them from being very thin in situations when the map takes most of the widget's height

kubama (commented):
The code looks high quality as usual, but I have a usability doubt. When the map is collapsed there is an empty space above the search box. This IMHO looks quite artificial and it's also a waste of space. Additionally it's not in line with the original mock-up, where the search box is placed right at the top of the widget, at the same level as the map button. Also, the mock-up with the map displayed also shows the search box above the map. Is this discrepancy intentional?

Gelio:
@kubama 

Thanks :smile: You're bringing up good points that I have not addressed

> When the map is collapsed there is an empty space above the search box. This IMHO looks quite artificial and it's also a waste of space. Additionally it's not in line with the original mock-up, where the search box is placed right at the top of the widget, at the same level as the map button. Also, the mock-up with the map displayed also shows the search box above the map. Is this discrepancy intentional?

All of these seem to be addressed by RD-1228. Unfortunately, it's not as easy, since it requires refactoring `DataTable` to accept a `searchPhrase` from a parent component, and not use its own search box. That is additional work, that has been postponed due to time limitations and the map being of more priority :slightly_smiling_face: 

I am aware that the placement of the map toggle button is not ideal, and it looks unusual, being a single button in the top bar, but I cannot do much without RD-1228. However, with RD-1768 we will probably be adding a _Filter_ button to the top bar, to the right of the map toggle, which would make the top bar a bit more justified, usable, and in line with the mockup :slightly_smiling_face: 

Let me know if you have other suggestions on the placement of the button, keeping in mind that _Filter_ button is incoming. Any other concerns are also worth to bring up :slightly_smiling_face:

kubama (approved):
If RD-1228 is on horizon then all good, I don't have a better idea as for now.

Gelio:
The only failure in system tests is in `topology_spec` and seems unrelated to this PR. It also failed for me when running the tests locally. I have created RD-2092 as it will likely happen in the upcoming nightly tests.

I don't see that as a blocker to the PR, I assume that the map is working fine since the `deployments_view_spec` passed :slightly_smiling_face:

qooban (approved):
I didn't go very deep in the review, but it looks good 👍🏻

Gelio:
Thanks for swift reviews, guys :slightly_smiling_face: I'll merge the PR then

PR #1322, opened 2021-04-21
RD-1948 - Add checkboxes to User manager widget that disables / enables Getting Started Modal for different users

gregcfy:
This PR adds enable / disable `Getting Started Modal` checkboxes into `User manager` widget list.
From now it should be possible to disable and enable modal for different users from the widget level. According to this feature: local storage will be not used from now to store the current modal status.
Task: https://cloudifysource.atlassian.net/browse/RD-1948
image


widgets/userManagement/src/UsersTable.jsx
```diff
@@ -322,6 +348,7 @@ export default class UsersTable extends React.Component {
                     <DataTable.Column label="Last login" name="last_login_at" width="18%" />
                     <DataTable.Column label="Admin" width="10%" />
                     <DataTable.Column label="Active" name="active" width="10%" />
+                    <DataTable.Column label="Getting started modal" name="show_getting_started" width="10%" />
```
kubama, line 351: I think 'modal' is an implementation detail which doesn't need to be included in the column name. Additionally, looking at the screen shot, this column is very wide but it contains just a checkbox. Usability suffers as the 'Last login' is now to narrow and its content gets line break. This should be somehow addressed. My first take would be to try to include line break in the column name ('Getting\nstarted'). This way the whole column would be narrower.

kubama (commented):
Some general comments:
1. I can see this PR contains something more than what the title suggests - IIUC it also changes the way the modal flag is stored, from local storage to manager user profile setting. I understand it was necessary but it seems to be out of scope of the referenced ticket. I think such change should be done as part of separate ticket, or scope of the referenced ticket should have been extended to be clear what changes it covers. A mention in PR description would also help to make it clear.
2. This PR should include an update to the widget cy test, so it covers new functionality.
3. Are you planning to add another reviewer?

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -322,6 +348,7 @@ export default class UsersTable extends React.Component {
...
                     <DataTable.Column label="# Groups" width="10%" />
                     <DataTable.Column label="# Tenants" width="10%" />
```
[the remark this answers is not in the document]
gregcfy: Ad 1. Task + PR description updated

Ad 2. I was thinking to add all tests in different PR, currently, I am working on https://cloudifysource.atlassian.net/browse/RD-2077

Ad 3. It is a small PR - the second reviewer added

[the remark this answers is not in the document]
gregcfy: i18n added

widgets/userManagement/src/UsersTable.jsx
gregcfy: muli line column headers support added

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -287,6 +315,19 @@ export default class UsersTable extends React.Component {
             });
     }
 
+    renderMultilineText = text => {
+        const lines = text.split('\n');
+        if (lines.length > 0) {
```
Gelio, line 320: This will always be `true` because `.split` always returns an array with at least 1 item (even if `text` was an empty string) :slightly_smiling_face: Thus, no need to make the if

app/translations/en.json
```diff
@@ -699,6 +699,17 @@
                 }
             }
         },
+        "userManagement": {
+            "columns": {
+                "username": "Username",
+                "last_login_at": "Last login",
```
Gelio, line 705: We generally use camelcase in the `en.json` file :slightly_smiling_face: Could you use it for the keys you've added?

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -364,6 +418,24 @@ export default class UsersTable extends React.Component {
                                         />
                                     )}
                                 </DataTable.Data>
+                                <DataTable.Data className="center aligned">
+                                    <Checkbox
+                                        checked={item.show_getting_started}
+                                        disabled={!this.hasAdminRole()}
```
Gelio, line 424: The button should also be disabled if the request is in flight. Otherwise, we get into weird states if we don't control how many requests are in flight, and if each request managest the state of `toolbox.loading` :slightly_smiling_face: 

Thus, either here or in a new ticket, let's make sure to disable the button if there is a request in progress, or control multiple simulatenous requests in a smarter way :slightly_smiling_face:

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -364,6 +418,24 @@ export default class UsersTable extends React.Component {
...
+                                        onChange={() =>
+                                            item.show_getting_started
+                                                ? this.showModal(
+                                                      MenuAction.SET_GETTING_STARTED_MODAL_DISABLED_ACTION,
+                                                      item
+                                                  )
+                                                : this.showModal(
```
Gelio, line 431: There is no modal, is there? Why not just call `setGettingStartedModalEnabled` and skip adding the `MenuAction` items and modifying the `showModal` method :smile:

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -364,6 +418,24 @@ export default class UsersTable extends React.Component {
...
+                                                      MenuAction.SET_GETTING_STARTED_MODAL_ENABLED_ACTION,
+                                                      item
+                                                  )
+                                        }
+                                        onClick={e => e.stopPropagation()}
```
Gelio, line 436: Is the line necessary? If yes, why :slightly_smiling_face: ? I don't know, it's a general question

app/components/GettingStartedModal/useModalState.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useManager } from './common/managerHooks';
+
+type UserResponse = {
+    // eslint-disable-next-line camelcase
+    show_getting_started: boolean;
+};
+
+const useModalState = () => {
+    const manager = useManager();
+    const { loading, response } = useFetch<UserResponse>(manager, `/users/${manager.getCurrentUsername()}`);
```
Gelio, line 14: Optional:

Did you consider using react-query?

app/components/GettingStartedModal/useModalState.ts
```diff
@@ -0,0 +1,35 @@
...
+    const [modalOpen, setModalOpen] = useState(false);
+    useEffect(() => {
+        if (response?.show_getting_started) {
+            setModalOpen(true);
+        }
+    }, [loading, response]);
```
Gelio, line 20: Do we need `loading` in the dependency array? It seems unnecessary to me :slightly_smiling_face:

Gelio (commented):
Remember to run system tests to this PR when you'll not be expecting more changes :slightly_smiling_face:

app/components/GettingStartedModal/useModalState.ts
```diff
@@ -0,0 +1,35 @@
...
+    const closeModal = async (disabled: boolean) => {
+        try {
+            if (disabled) {
+                // TODO(RD-1874): use common api for backend requests
+                await manager.doPost(`/users/${manager.getCurrentUsername()}`, null, { show_getting_started: false });
```
Gelio, line 25: There will be no loading indicator while this request is in flight, right? I suppose this may confuse the user and they will click the button multiple times, expecting the modal to finally close :slightly_smiling_face:

app/components/GettingStartedModal/GettingStartedModal.tsx
```diff
@@ -93,10 +93,10 @@ const GettingStartedModal = () => {
         EventBus.trigger('secrets:refresh');
         setInstallationProcessing(false);
     };
-    const handleModalClose = () => {
-        setModalOpen(false);
+    const handleModalClose = async () => {
+        await modalState.closeModal(modalDisabledChecked);
         if (modalDisabledChecked) {
-            disableGettingStartedModal();
+            EventBus.trigger('plugins:refresh');
```
Gelio, line 99: Do we want to do it only when the modal should be disabled? :thinking: In other words, is the `if` necessary?

Gelio: Also, IMO, this `trigger` should happen after the plugins are installed, not when the modal is being closed :smile: It is unnecessary work for plugins to be refreshed when the user just opened and is now closing the modal, without doing any additional actions

app/components/GettingStartedModal/GettingStartedModal.tsx
```diff
@@ -28,7 +28,7 @@ import type {
 const castedGettingStartedSchema = gettingStartedSchema as GettingStartedSchema;
 
 const GettingStartedModal = () => {
-    const [modalOpen, setModalOpen] = useState(() => !isGettingStartedModalDisabled());
+    const modalState = useModalState();
```
Gelio, line 31: Can we name it somewhat differently to suggest that it's the modal's _open_ state?

```suggestion
    const modalOpenState = useModalOpenState();
```

app/components/GettingStartedModal/useModalState.ts
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const useModalState = () => {
+    const manager = useManager();
+    const { loading, response } = useFetch<UserResponse>(manager, `/users/${manager.getCurrentUsername()}`);
+    const [modalOpen, setModalOpen] = useState(false);
+    useEffect(() => {
+        if (response?.show_getting_started) {
+            setModalOpen(true);
+        }
```
Gelio, line 19: Is it expected for users to suddenly see a _getting started_ modal? Suppose the request takes a long time, and the user is already interacting with the app. Then, a large _Getting Started_ modal appears in their face :smile:

I suppose we could improve it by delaying app rendering until the information about the getting started modal's open state is loaded. What do you think?

This does not have to be done here, can be a separate ticket :slightly_smiling_face:

Gelio (commented):
btw. nice job adding TS where you could :slightly_smiling_face: did you try to convert some existing files to TS too?

app/translations/en.json
```diff
@@ -699,6 +699,17 @@
...
+                "isAdmin": "Admin",
+                "active": "Active",
+                "show_getting_started": "Getting\nstarted",
```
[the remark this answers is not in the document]
gregcfy: ok, finally changed to css

widgets/userManagement/src/MenuAction.jsx
```diff
@@ -20,6 +20,10 @@ export default class MenuAction extends React.Component {
 
     static SET_ADMIN_USER_ROLE_ACTION = 'set-admin-role';
 
+    static SET_GETTING_STARTED_MODAL_ENABLED_ACTION = 'set-getting-started-modal-enabled';
+
+    static SET_GETTING_STARTED_MODAL_DISABLED_ACTION = 'set-getting-started-modal-disabled';
```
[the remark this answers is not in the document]
gregcfy: fixed :)

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -318,12 +359,25 @@ export default class UsersTable extends React.Component {
...
                     noDataMessage={NO_DATA_MESSAGE}
                 >
-                    <DataTable.Column label="Username" name="username" width="37%" />
-                    <DataTable.Column label="Last login" name="last_login_at" width="18%" />
-                    <DataTable.Column label="Admin" width="10%" />
-                    <DataTable.Column label="Active" name="active" width="10%" />
-                    <DataTable.Column label="# Groups" width="10%" />
-                    <DataTable.Column label="# Tenants" width="10%" />
+                    <DataTable.Column
+                        label={this.renderMultilineText(t('columns.username'))}
```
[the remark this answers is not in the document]
gregcfy: fixed

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -9,6 +9,8 @@ import TenantModal from './TenantModal';
 import UserDetails from './UserDetails';
 import UserPropType from './props/UserPropType';
 
+const { i18n } = Stage;
+const t = (key, options) => i18n.t(`widgets.userManagement.${key}`, options);
```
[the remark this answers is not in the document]
gregcfy: `options` removed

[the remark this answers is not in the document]
gregcfy: fixed

widgets/userManagement/src/UsersTable.jsx
gregcfy: ticket created: https://cloudifysource.atlassian.net/browse/RD-2100

widgets/userManagement/src/UsersTable.jsx
gregcfy: just, I was keeping the format of the code - there are many actions but without modals

widgets/userManagement/src/UsersTable.jsx
gregcfy: yes, checkbox click causes row expansion, it is used with other checkboxes too

app/components/GettingStartedModal/useModalState.ts
gregcfy: fixed

app/components/GettingStartedModal/GettingStartedModal.tsx
gregcfy: if it is not called, used 'Users widget' keeps unchecked 'Getting started' even current state is set to false

app/components/GettingStartedModal/GettingStartedModal.tsx
gregcfy: fixed

widgets/userManagement/src/UsersTable.jsx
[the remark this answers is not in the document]
Gelio: You answered one point, but ignored the second point. Is there a reason? If you only address a comment partially, I would appreciate if you kept the discussion open :slightly_smiling_face:

widgets/userManagement/src/UsersTable.jsx
Gelio: Oh, you're right! Then, in this case, how about renaming the `showModal` method to `invokeAction` or something else that does not suggest there will be a modal? :smile:

widgets/userManagement/src/UsersTable.jsx
Gelio: Good to know :slightly_smiling_face: Perhaps a comment in the code would provide context for future readers

app/components/GettingStartedModal/GettingStartedModal.tsx
Gelio: Weird, I don't see the `Users widget` reacting to `plugins:refresh`. In fact, I only see `PluginsTable` reacting to that event :smile: 

image

Do you know what I am missing?

This just makes it more confusing to me :slightly_smiling_face: Did you consider my suggestion about moving the `.trigger` to the place in the code that actually invokes plugin installation, so it does not happen each time the modal is closed, but only when some plugins were installed?

app/components/GettingStartedModal/useModalState.ts
gregcfy: good idea 👍

gregcfy: Ticket created: https://cloudifysource.atlassian.net/browse/RD-2103

app/components/GettingStartedModal/GettingStartedModal.tsx
gregcfy: changed to `EventBus.trigger('users:refresh');` that fires refresh action only if users list is used somewhere

app/components/GettingStartedModal/useModalState.ts
gregcfy: yes, I have created a ticket to block multiple that kind of operations, other checkboxes have bugs in blocking logic, it should be fixed together.
Ticket: https://cloudifysource.atlassian.net/browse/RD-2100

app/components/GettingStartedModal/useModalState.ts
gregcfy: I have created ticket: https://cloudifysource.atlassian.net/browse/RD-2104

widgets/userManagement/src/UsersTable.jsx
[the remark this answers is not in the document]
gregcfy: finally `columnT` used as name :)

widgets/userManagement/src/UsersTable.jsx
gregcfy: changed

widgets/userManagement/src/UsersTable.jsx
gregcfy: comment added

app/translations/en.json
gregcfy: column names are 1 to one to names received from backend

gregcfy: fixed

widgets/userManagement/src/UsersTable.jsx
gregcfy: yes, mistake related with few changes - now there is css only

widgets/userManagement/src/UsersTable.jsx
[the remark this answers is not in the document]
Gelio: I would use `columnsT`, since the part of the `en.json` it refers to is `.columns.` :smile: but okk, changing it to `columnsT` can be optional :slightly_smiling_face:

app/components/GettingStartedModal/useModalState.ts
Gelio: Sounds good :+1: Thanks

app/components/GettingStartedModal/GettingStartedModal.tsx
Gelio: Makes sense now :+1:

app/components/GettingStartedModal/useModalOpenState.ts
```diff
@@ -0,0 +1,35 @@
...
+        if (response?.show_getting_started) {
+            setModalOpen(true);
+        }
+    }, [response]);
+    const closeModal = async (disabled: boolean) => {
+        try {
+            if (disabled) {
+                // TODO(RD-1874): use common api for backend requests
+                await manager.doPost(`/users/${manager.getCurrentUsername()}`, null, { show_getting_started: false });
+            }
```
Gelio, line 26: Optional:

Adding the `EventBus.trigger('users:refresh')` call here would make more sense, IMO :smile:

widgets/userManagement/src/UsersTable.jsx
```diff
@@ -9,6 +9,8 @@ import TenantModal from './TenantModal';
 import UserDetails from './UserDetails';
 import UserPropType from './props/UserPropType';
 
+const columnT = key => Stage.i18n.t(`widgets.userManagement.columns.${key}`);
```
Gelio, line 12: Opitonal:

I would name it `columnsT`, since the part of `en.json` it refers to is `.columns.`, not `.column.` :slightly_smiling_face: 

```suggestion
const columnsT = key => Stage.i18n.t(`widgets.userManagement.columns.${key}`);
```

Gelio (approved):
Looks good to me :+1: I appreciate you taking extra effort to extract column names to `en.json`

Only minor comments left, so I'm approving

Gelio:
@gregcfy What's your take on system tests for this PR? Please consider running them or justify why you skip it :smile: (as we agreed on the process during the weekly meetings)

widgets/userManagement/src/UsersTable.jsx
gregcfy: it returns the translation for a single column, maybe it could be better to keep `columnT`

app/components/GettingStartedModal/useModalOpenState.ts
gregcfy: I didn't put it here because I got the idea it can be confusing to call hook and affect on the UI.

app/components/GettingStartedModal/GettingStartedModal.tsx
```diff
@@ -93,10 +93,10 @@ const GettingStartedModal = () => {
...
         setInstallationProcessing(false);
     };
-    const handleModalClose = () => {
-        setModalOpen(false);
+    const handleModalClose = async () => {
+        await modalOpenState.closeModal(modalDisabledChecked);
         if (modalDisabledChecked) {
-            disableGettingStartedModal();
+            EventBus.trigger('users:refresh');
         }
```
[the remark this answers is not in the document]
gregcfy: I didn't put it here because I got the idea it can be confusing to call hook and affect on the UI.

kubama (dismissed):

widgets/userManagement/src/UsersTable.jsx
Gelio: Ok, if you see it that way, sure, it's ok to leave it unchanged :slightly_smiling_face:

app/components/GettingStartedModal/useModalOpenState.ts
Gelio: I see it differently :slightly_smiling_face: You're not calling the hook, you're calling the `closeModal` function of the hook, that introduces changes anyway :slightly_smiling_face: Who knows what side-effects it has?

I wanted to have the related side-effects close together - if there is a clear action-reaction (as with "users changed on the backend -> refresh the users"), I would love to see the code lines as close to each other as possible, to see that they are related. Looking at the current code, seeing "when the modal closes, refresh the users if the modal should be disabled", I was wondering - yeah, but why?

If the line was moved here, it would read "when users are updated in the backend, refresh the users" - that makes much more sense to me. I care about that relationship more than the fact that users may be refreshed after calling `closeModal` :slightly_smiling_face:

Gelio (approved):
The code looks good to me. I have left only optional comments :slightly_smiling_face: I encourage implementing the one about moving `EventBus.trigger`, but if you're short on time or don't fully agree, let's discard it

I see my messages about the system tests have been ignored. @gregcfy can you respond with the status of system tests for this PR **before** merging it?

app/components/GettingStartedModal/useModalOpenState.ts
gregcfy: moved into hook

Gelio:
Looks ok :slightly_smiling_face: I'm slightly fed up with the lack of responses about system tests, so I'm not approving until I get a response :slightly_smiling_face:

gregcfy:
> Looks ok  I'm slightly fed up with the lack of responses about system tests, so I'm not approving until I get a response

system tests passed except `widgets/topology_spec.js` that is related with different ticket
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/496/

Gelio (approved):
Thanks for updating on the system tests status :+1:

PR #1323, opened 2021-04-21
RD-1758 / RD-2011 - Add error handling to FilterDefinitionForm

qooban:
### Changes
* Added basic error handling for filter definition form
* Small refactors:
  * Changed name from `FilterDefinitionForm` to `RulesForm`
  * Improved rule value input fields' placeholders

### Known issues
* For Label rules with operator type "is one of" or "is not one of" in case of missing key or value, always both (key and value dropdown) are marked red (even when there is only value missing). Fixing that would require passing error field down to the dropdown component through `RuleValueInput` > `LabelValueInput` > `KeyDropdown` / `ValueDropdown` > `CommonDropdown`, so I think it is not worth to do it. Business value is tiny IMHO.

### Tests

Tests are planned to be added in https://github.com/cloudify-cosmo/cloudify-stage/pull/1313 .

### Video

https://user-images.githubusercontent.com/5202105/115878752-33967b80-a449-11eb-9e7c-27eedaec9557.mp4


qooban:
System tests scheduled: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/498/

Gelio:
I've planned a review of this PR for Monday :slightly_smiling_face:

widgets/common/src/filters/common.ts
```diff
@@ -1,5 +1,12 @@
...
+import { FilterRuleOperators } from './types';
+
 const { i18n } = Stage;
 const i18nPrefix = 'widgets.common.filters.form';
 
-// eslint-disable-next-line import/prefer-default-export
 export const getTranslation = (key: string) => i18n.t(`${i18nPrefix}.${key}`);
+
+export function isAnyOperator(operator: FilterRuleOperator) {
+    const anyOperators: FilterRuleOperator[] = [FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf];
```
kubama, line 10: Looks like this const can be moved outside of the function scope

widgets/common/src/filters/inputs/common.ts
```diff
@@ -1,12 +1,6 @@
-import type { FilterRuleOperator } from '../types';
-import { FilterRuleOperators } from '../types';
 import { getTranslation } from '../common';
 
-export function isAnyOperator(operator: FilterRuleOperator) {
-    const anyOperators: FilterRuleOperator[] = [FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf];
-    return anyOperators.includes(operator);
-}
```
kubama, line 8: Seeing that this function had to be moved from one `common.ts` to another and that there are just few lines in each of these files I'm wondering if it actually makes sense to have separate `common.ts` file per directory level.

kubama (commented):
+1 for integrating this change with filters widget at once, thanks!

Gelio:
I've got a meeting till 10:30 and I'll read the PR right after that :eyes:

Gelio:
> * For Label rules with operator type "is one of" or "is not one of" in case of missing key or value, always both (key and value dropdown) are marked red (even when there is only value missing). Fixing that would require passing error field down to the dropdown component through `RuleValueInput` > `LabelValueInput` > `KeyDropdown` / `ValueDropdown` > `CommonDropdown`, so I think it is not worth to do it. Business value is tiny IMHO.

I've got an idea here for your consideration. We could create a bug in JIRA for this behavior and immediately close it, describing why it's closed without addressing it. If we get a question about this behavior in the future, we could link to the bug (and reopen it, if it turns out that someone wants it done) instead of having to dig into this case and explain why it's hard :slightly_smiling_face:

My suggestion is optional :slightly_smiling_face:

widgets/common/src/filters/RulesForm.tsx
```diff
@@ -1,49 +1,57 @@
...
 import { FilterRuleType, FilterRuleOperators, FilterRuleRowType } from './types';
+import { isAnyOperator } from './common';
+
+function hasError({ type, key, operator, values }: FilterRule) {
+    const ruleHasValuesDefined = values?.length > 0;
+    const ruleHasKeyDefined = !!key;
+    const isLabelRuleInvalid = !ruleHasKeyDefined || (isAnyOperator(operator) && !ruleHasValuesDefined);
+    const isAttributeRuleInvalid = !ruleHasValuesDefined;
+
+    return type === FilterRuleType.Label ? isLabelRuleInvalid : isAttributeRuleInvalid;
```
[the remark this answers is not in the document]
qooban: Applied.

widgets/common/src/filters/common.ts
qooban: Yep, moved it outsided.

widgets/common/src/filters/inputs/common.ts
qooban: Yeah, after some thinking, I see more value in having single `common.ts`. Changed.

qooban:
@Gelio 
> I've got an idea here for your consideration. We could create a bug in JIRA for this behavior and immediately close it, describing why it's closed without addressing it. If we get a question about this behavior in the future, we could link to the bug (and reopen it, if it turns out that someone wants it done) instead of having to dig into this case and explain why it's hard 

Sure, I have created RD-2132 in Filters UI Enhancements epic. I think I leave it open for now.

widgets/common/src/filters/RulesForm.tsx
```diff
@@ -1,49 +1,59 @@
...
+        const noKeyDefined = !key;
+        return noKeyDefined || (isAnyOperator(operator) && noValuesDefined);
+    }
+
+    return noValuesDefined;
+}
 
 function getNewRow(): FilterRuleRow {
     const { uuid } = Stage.Utils;
+    const emptyRule = {
```
Gelio, line 23: Optional:

There's a missing type annotation here

```suggestion
    const emptyRule: FilterRule = {
```

widgets/common/src/filters/RulesForm.tsx
```diff
@@ -55,18 +65,21 @@ const FiltersDefinitionForm: FunctionComponent<FiltersDefinitionFormProps> = ({
...
+    useEffect(() => {
+        onChange(rows.map(getFilterRule), rows.filter(row => row.hasError).length > 0);
     }, [rows]);
 
     return (
         <>
             {rows.map(row => (
                 <RuleRow
                     key={row.id}
+                    error={markErrors && !!row.hasError}
```
Gelio, line 82: `row.hasError` is already a boolean - do we need the `!!`? :thinking:

widgets/filters/src/FilterModal.tsx
```diff
@@ -29,13 +30,22 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
         initialFilter ? [...initialFilter.attrs_filter_rules, ...initialFilter.labels_filter_rules] : []
     );
     const { errors, setErrors, clearErrors, setMessageAsError } = useErrors();
+    const [filterRulesInvalid, setFilterRulesInvalid] = useState<boolean>(false);
```
Gelio, line 33: Optional:

The generic parameter can be inferred from the type of the default value :slightly_smiling_face: 

```suggestion
    const [filterRulesInvalid, setFilterRulesInvalid] = useState(false);
```

widgets/filters/src/FilterModal.tsx
```diff
@@ -58,7 +70,15 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
                     <UnsafelyTypedFormField label={i18n.t('widgets.filters.modal.rules')}>
-                        <RulesForm initialFilters={filterRules} toolbox={toolbox} onChange={setFilterRules} />
+                        <RulesForm
+                            initialFilters={filterRules}
+                            toolbox={toolbox}
+                            onChange={(newFilterRules, hasErrors) => {
+                                setFilterRules(newFilterRules);
+                                setFilterRulesInvalid(hasErrors);
+                            }}
+                            markErrors={markRulesFormErrors}
```
Gelio, line 80: Can we not pass `true` here, instead of checking the `errors` object? The form checks `rule.hasError` anyway, so it will always only show errors for rows that actually have errors :smile: Am I missing something why `isEmpty(errors)` is necessary?

Gelio (approved):
Looks mostly correct :+1: Some minor comments left

widgets/common/src/filters/RulesForm.tsx
qooban: Right, no need for `!!` in this case, fixed.

widgets/filters/src/FilterModal.tsx
qooban: Right, good point, removed `<boolean>`.

widgets/common/src/filters/RulesForm.tsx
qooban: Thx, added type annotation.

app/translations/en.json
```diff
@@ -741,7 +741,10 @@
             "modal": {
                 "id": "Filter ID",
                 "rules": "Filter rules",
-                "idMissing": "Please provide the filter ID",
+                "errors": {
+                    "idMissing": "Please provide the filter ID",
+                    "filterRulesInvalid": "Please provide all the values in filter rules section"
```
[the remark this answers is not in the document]
qooban: I'm not sure here. Currently we check only if values are present or not. There are no other checks.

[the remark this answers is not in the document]
Gelio: In this case, the one I suggested may future-proof us, but may also be unnecessary :slightly_smiling_face: I'll leave it up to you to decide

widgets/filters/src/FilterModal.tsx
qooban: `errors` variable is filled only after submitting the form. It is emptied when user dismisses the error message (clicks X icon in the top right):
https://user-images.githubusercontent.com/5202105/116059256-f0265200-a680-11eb-9e7a-586bb125e426.mp4

That's why I created a variable `markRulesFormErrors` which is set to true when there are `errors` set and set to false when there are no `errors`.

kubama (approved):

app/translations/en.json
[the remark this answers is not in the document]
qooban: I'll prefer to leave it for now. We'll adapt it if needed.

qooban:
System tests scheduled: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/507

widgets/filters/src/FilterModal.tsx
Gelio: Oh, ok, makes sense now :smile: Since it was not clear to me from the beginning, maybe we could (optionally) describe why this pattern is used here :slightly_smiling_face:

Gelio (dismissed):

qooban:
There was one failure in the last system tests execution.
I fixed that in https://github.com/cloudify-cosmo/cloudify-stage/pull/1323/commits/34219adfd48be27e4c003847c415186fe7b7add6 .

I'm not going to re-run all tests. Filters are fine right now:
image

qooban:
This PR is ready for final approvals, so waiting for the green light and I'll merge it.

Gelio (approved):
Looks good :+1:

qooban:
As there was just `filters_spec` test fix added since last double approve, I'm merging with just a single green light. Thanks!

PR #1324, opened 2021-04-22
RD-1768 Added filter button to deployments widget

kubama:
TODO: Add system tests.

The button is placed next to the map button (the map button styling was slightly updated to match the new button and the rest of the application):
image

Button when filter selected:
image

Filter selection modal:
image


app/translations/en.json
```diff
@@ -648,7 +648,15 @@
...
-              }
+              },
+                "filter": {
+                    "button": "Filter",
+                    "modal": {
+                        "header": "Select a filter",
+                        "submit": "OK",
+                        "noFilterError": "Please select a filter"
+                    }
+                }
```
Gelio, line 659: Looks misaligned to me :slightly_smiling_face:

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -21,28 +21,53 @@ export interface DeploymentsViewProps {
     toolbox: Stage.Types.Toolbox;
 
     filterByParentDeployment: boolean;
-    filterRules: Stage.Common.Filters.Rule[];
-    fetchingRules: boolean;
+    defaultFilterId?: string | undefined;
```
Gelio, line 24: Maybe instead of `default` we could use `initial`. IMO that better describes the intention of the prop

```suggestion
    initialFilterId?: string | undefined;
```

btw. do we need both `?` and `| undefined`? I think just `?` works well, as far as I have checked locally
```suggestion
    initialFilterId?: string;
```

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -21,28 +21,53 @@ export interface DeploymentsViewProps {
...
+    additionalFilterRules?: Stage.Common.Filters.Rule[];
```
Gelio, line 25: Optional:

IMO it's worth describing what _additional_ means in this context. Perhaps we could add a JSDoc comment with something along the lines of:

> Those rules will be always appended to the rules from filterId or that were chosen by the user

widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+            {/* TODO(RD-1225): enable the map in production */}
+            {!production && (
+                <Button
+                    labelPosition="left"
+                    icon="map"
+                    active={mapOpen}
+                    onClick={toggleMap}
+                    title={mapT(mapOpen ? 'closeMap' : 'openMap')}
+                    content={mapT('label')}
+                />
```
Gelio, line 46: Thanks for aligning the appearance of the button :+1:

widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+            {filterId ? (
+                <Button.Group color="blue">
+                    <Button
+                        icon="filter"
+                        labelPosition="left"
+                        content={filterId}
+                        onClick={openFilterModal}
+                        style={{ whiteSpace: 'nowrap', maxWidth: 200, textOverflow: 'ellipsis', overflow: 'hidden' }}
+                    />
+                    <Button icon="delete" onClick={() => handleFilterChange(undefined)} />
```
Gelio, line 57: Optional:

How about some `title`/`aria-label` to describe what the icon does?

widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const [filterModalOpen, openFilterModal, closeFilterModal] = Stage.Hooks.useBoolean();
+    const [filterId, setFilterId] = useState<string>();
+
+    const { Button } = Stage.Basic;
+
+    function handleFilterChange(newFilterId: string | undefined) {
+        setFilterId(newFilterId);
+        onFilterChange(newFilterId);
+        closeFilterModal();
+    }
```
Gelio, line 33: How do we handle the situation when the user selects some rules, clicks "Apply", but does not save the filter explicitly? I believe such a scenario was planned for the Filters button from the beginning :thinking: 

image

(just hitting the "Apply" button)

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -21,28 +21,53 @@ export interface DeploymentsViewProps {
...
-    fetchingRules
+    additionalFilterRules = [],
+    defaultFilterId
 }) => {
     const manager = toolbox.getManager();
     const [gridParams, setGridParams] = useState<Stage.Types.ManagerGridParams>();
+    const [userFilterId, setUserFilterId] = useState<string>();
+
+    const filterRulesResult = (() => {
+        const filterId = userFilterId ?? defaultFilterId;
```
Gelio, line 40: Nice use of `??` :ok_hand:

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -21,28 +21,53 @@ export interface DeploymentsViewProps {
...
-    filterRules,
-    fetchingRules
+    additionalFilterRules = [],
+    defaultFilterId
 }) => {
     const manager = toolbox.getManager();
     const [gridParams, setGridParams] = useState<Stage.Types.ManagerGridParams>();
+    const [userFilterId, setUserFilterId] = useState<string>();
+
+    const filterRulesResult = (() => {
```
Gelio, line 39: Awesome pattern with IIFE :smile: Is it to limit the scope and not leak variables?

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -21,28 +21,53 @@ export interface DeploymentsViewProps {
...
 
+    const filterRules = filterRulesResult.data ?? [];
     const filteringByParentDeploymentResult = useFilteringByParentDeployment({ filterByParentDeployment });
     const finalFilterRules = useMemo(() => {
         if (!filteringByParentDeploymentResult.parentDeploymentRule) {
-            return filterRules;
+            return [...filterRules, ...additionalFilterRules];
         }
 
         return [...filterRules, filteringByParentDeploymentResult.parentDeploymentRule];
```
Gelio, line 69: For consistency, should we not also append `additionalFilterRules` here?

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const { i18n } = Stage;
+    const { ApproveButton, CancelButton, Icon, Modal, UnsafelyTypedForm, UnsafelyTypedFormField } = Stage.Basic;
+    // @ts-ignore Property 'DynamicDropdown' does not exist on type 'typeof Common'
+    const { DynamicDropdown } = Stage.Common;
+    const { useResettableState } = Stage.Hooks;
+
+    const { errors, setErrors, clearErrors } = Stage.Hooks.useErrors();
+    const [filterId, setFilterId, resetFilterId] = useResettableState<string | undefined>(filterIdProp);
+
+    useEffect(resetFilterId, [filterIdProp]);
```
Gelio, line 29: Nice and concise :+1: 

btw. should we also not reset the errors here?

Gelio (commented):
I like the overall approach :+1: Nice work! 

Are you planning on adding the tests in this PR, or in a separate one?

widgets/common/src/deploymentsView/header/index.tsx
Gelio: Nevermind, I see RD-1768 is just about selecting a predefined filter. I reckon other tasks in RD-2067 will require a larger refactor :smile: Looks good here

app/translations/en.json
kubama: Fixed

widgets/common/src/deploymentsView/DeploymentsView.tsx
kubama: > Maybe instead of `default` we could use `initial`. IMO that better describes the intention of the prop

As for me:
- initial - just a starting point
- default - value that is used when there is no explicitly specified value, it's common to revert to default

So I think 'default' suites better.

> btw. do we need both `?` and `| undefined`? I think just `?` works well, as far as I have checked locally

Yes, updated.

widgets/common/src/deploymentsView/DeploymentsView.tsx
kubama: Done

widgets/common/src/deploymentsView/DeploymentsView.tsx
kubama: > Is it to limit the scope and not leak variables?

This was my intention.

widgets/common/src/deploymentsView/DeploymentsView.tsx
kubama: Looks like you've found a bug 🙂. Fixed.

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: > should we also not reset the errors here?

It's not necessary - when the modal is being cancelled the errors are already cleared, and when it is submitted it doesn't contain any errors (it couldn't be submitted otherwise).

widgets/common/src/deploymentsView/header/index.tsx
kubama: I've added a title.

kubama:
> Are you planning on adding the tests in this PR, or in a separate one?

If I got approval after second review round I'd like to merge it and provide tests in separate PR.

kubama:
System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/490/

kubama:
I consider system tests run a pass (only topology spec failed - known issue).

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: You're right about the default vs initial dilemma :+1: Thanks for the explanation

widgets/common/src/deploymentsView/header/FilterModal.tsx
Gelio: I see, so the implicit assumption is that `filterIdProp` will be constant when the modal is open - sounds fine to me :slightly_smiling_face:

Gelio (approved):
:ok_hand:

qooban (approved):
Looks good. I don't have bandwidth to go deep in the review. No UI tests done from my side.
Added some general comments and approved.

widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+    mapOpen: boolean;
+    toggleMap: () => void;
+    onFilterChange: (filterId: string | undefined) => void;
+    toolbox: Stage.Types.Toolbox;
+}
+
+const headerT = (suffix: string) => Stage.i18n.t(`${i18nPrefix}.header.${suffix}`);
+const mapT = (suffix: string) => headerT(`map.${suffix}`);
+
+const production = process.env.NODE_ENV === 'production' && !process.env.TEST;
```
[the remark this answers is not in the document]
kubama: I think we don't, but this is a pre-existing solution that is applied to the map feature.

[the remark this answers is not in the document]
Gelio: I remove it in #1325. Anyway, I did not want to add a "Map" button that displays a "Hey, I am a map" in production - I thought that would be silly, and adding such a check was easy enough, given that the env variables were ready :slightly_smiling_face:

PR #1325, opened 2021-04-22
RD-2090 Add Deployments Map without popups

Gelio:
This PR adds an actual map in the _Deployments View_ widget. The map displays a marker for each deployment that has a site with the location filled in. It fetches site data using react-query.

The map right now is quite similar to the one in the _Sites_ widget. It will change, though, with the upcoming tasks:
1. RD-2093 (handling deployment selection on the map)
2. RD-1526 (adding info in the popup)
3. RD-1527 (marker clustering)

I have extracted parts of `SitesMap` to `Stage.Common.Map` to reuse some of the code. I have also migrated a few files to TS. Both of these reactors add up to roughly 50% of the changes in this PR. I did not find a better way to split the code in this PR into multiple PRs without making the splits artificial.

I will add system tests for the Deployments Map (RD-2108) in a separate PR to keep this one shorter :slightly_smiling_face: 

## Known issues

1. RD-2107 Multiple deployments in the same site are displayed as standalone markers, and appear on top of each other

## Video

https://user-images.githubusercontent.com/889383/115715723-b8659480-a378-11eb-8a6f-6c68291c8891.mp4

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/495/pipeline

Ran 2021-04-23 11:38 CEST


widgets/common/src/deploymentsView/map/container.tsx
```diff
@@ -0,0 +1,64 @@
...
+    toolbox: Stage.Types.Toolbox;
+    widgetDimensions: ComponentProps<typeof DeploymentsMap>['widgetDimensions'];
+}
+
+const DeploymentsMapContainer: FunctionComponent<DeploymentsMapContainerProps> = ({
+    deployments,
+    toolbox,
+    widgetDimensions
+}) => {
+    // TODO: extract fetching to a new component for using `useMemo` only when data is available
```
kubama, line 19: When do you plan to address this TODO?

widgets/common/src/deploymentsView/map/container.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    if (sitesResult.isError || mapAvailableResult.isError) {
+        const { ErrorMessage } = Stage.Basic;
+
+        return <ErrorMessage header={mapT('errorLoadingSites')} error={sitesResult.error as { message: string }} />;
+    }
+
+    if (sitesResult.isIdle || mapAvailableResult.isIdle) {
+        throw new Error('Idle state for fetching map data is not handled');
+    }
```
kubama, line 49: Digression: Looks like this is becoming a common pattern. I don't have a good and simple idea how this could be extracted into a common piece due to the returns. Any thoughts on this?

widgets/common/src/deploymentsView/map/map.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const mapRef = useRef<Map | null>(null);
+
+    useEffect(() => Stage.Common.Map.invalidateSizeAfterDimensionsChange(mapRef), [
+        widgetDimensions.height,
+        widgetDimensions.width,
+        widgetDimensions.maximized
+    ]);
+
+    // TODO: Create markers and display the map
```
kubama, line 29: Hm... "display the map"? Isn't it what this PR adds?

widgets/common/src/deploymentsView/header.tsx
```diff
@@ -8,16 +8,9 @@ interface DeploymentsViewHeaderProps {
...
-const production = process.env.NODE_ENV === 'production' && !process.env.TEST;
-
 const DeploymentsViewHeader: FunctionComponent<DeploymentsViewHeaderProps> = ({ mapOpen, toggleMap }) => {
     const { Button, Icon } = Stage.Basic;
 
-    // TODO(RD-1225): enable the map in production
-    if (production) {
-        return null;
-    }
-
```
kubama, line 20: Looks like we will have a conflict here, should be very easy to resolve though.

widgets/common/src/map/options.ts
```diff
@@ -0,0 +1,21 @@
+import type { LatLngBounds, MapOptions } from 'leaflet';
+import { siteToLatLng, SiteWithPosition } from './site';
```
kubama, line 2: Import type? 🙂

widgets/common/src/deploymentsView/map/container.tsx
Gelio, line 19: This should be removed, I'll do that soon

widgets/common/src/deploymentsView/map/map.tsx
Gelio, line 29: This should be removed :smile:

widgets/deployments/src/DeploymentsList.jsx
```diff
@@ -132,6 +132,7 @@ export default class DeploymentsList extends React.Component {
                         deploymentId={deploymentId}
                         onHide={this.hideActionModal}
                         toolbox={toolbox}
+                        redirectToParentPageAfterDelete={false}
```
Gelio, line 135: A bugfix - the prop was required, but not provided explicitly here. Thanks @qooban for spotting it

widgets/common/src/map/DefaultTileLayer.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+const DefaultTileLayer: FunctionComponent = () => {
+    const { urlTemplate } = Stage.Common.Consts.leaflet;
+    const url = useMemo(() => Stage.Utils.Url.url(urlTemplate), [urlTemplate]);
+
+    const attribution = ReactRedux.useSelector(
+        (state: Stage.Types.ReduxState): string => state.config.app.maps.attribution
+    );
+
+    return <Stage.Basic.Leaflet.TileLayer attribution={attribution} url={url} />;
```
Gelio, line 11: Extracted from `SitesMap`

widgets/common/src/map/options.ts
```diff
@@ -0,0 +1,21 @@
...
+        bounds = L.latLngBounds(sites.map(siteToLatLng)).pad(0.05);
+    } else {
+        mapOptions.center = siteToLatLng(sites[0]);
+        mapOptions.zoom = initialZoom;
+    }
+
+    return {
+        options: mapOptions,
+        bounds
+    };
```
Gelio, line 20: Extracted from `SitesMap`

widgets/common/src/map/widget-dimensions.ts
```diff
@@ -0,0 +1,38 @@
...
+export function invalidateSizeAfterDimensionsChange(mapRef: RefObject<Map | null>) {
+    // Widget properties change doesn't affect immediately DOM changes,
+    // so it's necessary to wait a bit till DOM is updated.
+
+    const refreshAfterDimensionChangeTimeout = 500;
+    const timeoutId = setTimeout(() => {
+        mapRef.current?.leafletElement.invalidateSize();
+    }, refreshAfterDimensionChangeTimeout);
+
+    return () => clearTimeout(timeoutId);
```
Gelio, line 37: Extracted from `SitesMap`

kubama (dismissed):
Added some comments, but really nothing I would expect to be changed, so I'm approving at once.

widgets/common/src/deploymentsView/map/container.tsx
Gelio: I didn't review the code well enough before pushing :smile: I'll remove it (ref https://github.com/cloudify-cosmo/cloudify-stage/pull/1325#discussion_r618409745)

widgets/common/src/deploymentsView/map/map.tsx
Gelio: Yeah yeah, I should have removed it before I pushed the PR, but I did not :smile: sorry (ref https://github.com/cloudify-cosmo/cloudify-stage/pull/1325#discussion_r618410292)

widgets/common/src/deploymentsView/header.tsx
Gelio: Yup, also what I spotted :smile:

widgets/common/src/map/options.ts
Gelio: I'm afraid it's not possible as `siteToLatLng` is used as a function :worried:

widgets/common/src/deploymentsView/map/container.tsx
Gelio: Good point :+1: Unfortunately, I also don't see a nice way of extracting this logic, since it sometimes relies on a single query, sometimes on more.

I guess we could do:

```ts
const resultNode = handleReactQueryState([sitesResult, mapAvailableResult]); // would throw error for `isIdle`
if (resultNode) {
  return resultNode; // handles `isLoading` and `isError`
}
```

but I'm not sure if TS would catch that both queries would then be in `isSuccess` state :smile: We would also pass the messages for `error` and `loading` states to that function.

Let me know what do you think about this pattern :slightly_smiling_face:

Gelio:
@kubama Thanks for a quick review :+1: I'll remove the TODOs I mistakingly left in the code tomorrow :smile:

Gelio:
I see the system tests failed https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/488/pipeline. Probably because the map requires some special license/permissions. I'll verify tomorrow

widgets/common/src/map/options.ts
kubama: Can't we split it?
```suggestion
import { siteToLatLng } from './site';
import type { SiteWithPosition } from './site';
```

widgets/common/src/deploymentsView/map/container.tsx
kubama: How about providing a wrapper component that could be used like this:

```
<Wrapper results={[sitesResult, mapAvailableResult]} errorMessage=... loadingMessage=...>
   <ComponentToBeRenderedWhenDataIsLoaded/>
</Wrapper>
```

So, depending on the result(s) state it would render loader, error message or child component. It could still throw an error for idle.

widgets/common/src/map/options.ts
Gelio: Yes, we could, but I don't really see the point of doing that :slightly_smiling_face: Using `import type` is nice for a single import line, as there is no additional cost then, but in my view, having 2 lines just so one can be `import type` is additional unnecessary overhead in the code :smile: 

I am open to discussing that on the weekly, if you'd like to enforce this as a pattern :slightly_smiling_face:

widgets/common/src/deploymentsView/map/container.tsx
Gelio: That's another way, but how do we pass the results back to the component? We could force `children` to be a function, like that:

```tsx
<Wrapper results={[sitesResult, mapAvailableResult]} errorMessage=... loadingMessage=...>
   {([sites, mapAvailable]) => (<ComponentToBeRenderedWhenDataIsLoaded sites={sites} mapAvailable={mapAvailable} />)}
</Wrapper>
```

What do you think? :smile:

kubama: My first thought was that it's overcomplicated as the child component could receive result (or result.data) directly, but in this case it would need to handle lack of the data internally. So in the end I like your proposal, as the child component would be only instantiated when the data is ready.

widgets/common/src/map/options.ts
kubama: You were pushing strongly to use `import type` where applicable, now when there a minimal cost of adding an extra line to use it you are saying that you don't see a point and it would be an unnecessary overhead. I really don't understand.

Gelio: My rule of thumb for `import type` is to behave as if [`importsNotUsedAsValues`](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) was set to `error` - if we only import types, then use `import type`. Otherwise, it's fine to have an import statement that imports both JS values and types. This was my opinion from the beginning :slightly_smiling_face: I do not remember a case when I recommended splitting an import into 2 to have `import` and `import type` :slightly_smiling_face:

widgets/common/src/deploymentsView/map/container.tsx
Gelio: I'm glad you like it. I'll add a topic about it to today's weekly to agree on it as a team :slightly_smiling_face: 

---

> My first thought was that it's overcomplicated as the child component could receive result (or result.data) directly, but in this case it would need to handle lack of the data internally. So in the end I like your proposal, as the child component would be only instantiated when the data is ready.

My proposal is not based on React's or JS limitations, but on TS one :smile:  Having

```ts
<Wrapper results={[sitesResult, mapAvailableResult]} errorMessage=... loadingMessage=...>
  <ComponentToBeRenderedWhenDataIsLoaded sites={sitesResult.data} mapAvailable={mapAvailableResult.data} />
</Wrapper>
```

would be fine from JS/React perspective - it's ok to **create** an element for which some props are `undefined` (in the case that `sitesResult.data` is `undefined`), as long as the component is not rendered by the `Wrapper`. See [this CodeSandbox](https://codesandbox.io/s/musing-flower-eh5w3?file=/src/App.tsx) for an example. You can experiment with `loading` and `data`

As you can see, if `loading = true` and `data = null`, there is no JS error, and the `Wrapper` renders a fallback, without ever rendering the `MyComponent`. Then you set `loading = true` and keep `data = null`, only then is `MyComponent` rendered, and there is a runtime error, since `null.toExponential` is problematic :slightly_smiling_face: 

However, TypeScript does not know that the `MyComponent` won't be rendered, and complains whenever `data = null`. This is why having a `children` render prop solves the problem, since `sites` and `mapAvailable` will be the extracted `.data` property, and the extraction of `.data` would happen in `Wrapper`, that has already checked that `.data` is available on the results :smile: Basically, TS type narrowing at work

widgets/common/src/map/options.ts
kubama: So what's the benefit of using `import type` if adding an additional line to have it is an "unnecessary overhead"? From this standpoint I understand that the value is close to zero.
Adding a `type` keyword is also an overhead, yet you consider it worth it?

Gelio: The `type` keyword is one word, a new import statement is much longer :slightly_smiling_face: I see the overhead to benefit ratio as rather small when only adding `type` to an existing import, and much higher when having to split an import into 2 - one for types, one for JS :smile: I cannot imagine doing that across the project manually if we wanted to be consistent :slightly_smiling_face: 

---

Besides my opinion on the overhead to benefit ratios, one hard argument against having multiple import statements for the same file is hard to manage when using auto-imports:

https://user-images.githubusercontent.com/889383/115846377-1f8b5380-a422-11eb-8b33-c472320e6978.mp4

All auto-imports are added to the first import statement :slightly_smiling_face:

Gelio:
Map tests failed in the previous pipeline for a correct reason - there was a bug in that the case when there were 0 sites was not covered :slightly_smiling_face: I have fixed with the last commit. I have ran only the `deployments_view_spec` to validate it:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/493/

It may happen that I need to use the `valid_spire_license` anyway.

I will run another round of system tests once I get clarification about the license I should use :slightly_smiling_face: I will re-request reviews then. I don't believe much code to be changed from now on anyway, so the PR is stable to review

qooban:
Sorry to say, but I don't have bandwidth to review this PR today, so if you'd like to merge it today, please rely on @kubama review only. If you'd like to wait till Monday/Tuesday, I can provide review early next week.

Gelio:
`deployments_view_spec` passed with the latest code changes, so I'm scheduling a full test run:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/495/pipeline

---

@qooban Thanks for the info :slightly_smiling_face: No worries, take your time. Merging this PR is not a priority at the moment, so it can wait till Monday

widgets/common/src/map/options.ts
kubama: Having an extra import line is a "much higher overhead"? My question is still valid: what's the benefit of using import type if adding an additional line to have it is an "unnecessary overhead"? If the value is so low, why bother using `import type` at all?

> I cannot imagine doing that across the project manually if we wanted to be consistent [...] All auto-imports are added to the first import statement

You are suggesting dropping a language feature since your IDE doesn't handle it nicely? 🙂
To be honest, sounds fair, but... I also reported that WebStorm doesn't add `type` when auto-importing and it needs to be added manually. Looks like it wasn't considered to be an argument 🙁.

Summing up, since the value of using `type` is very low and it's troublesome in different IDEs I encourage dropping the rule to use, treat it as a cosmetic thing and stop commenting it.

Gelio: > My question is still valid: what's the benefit of using import type if adding an additional line to have it is an "unnecessary overhead"? If the value is so low, why bother using `import type` at all?

Not sure what you're expecting here, I don't have hard data.

1. In my view, adding a single word to achieve a possibly faster compilation is low overhead, especially that it's _opt-in_ at the moment (and I was leaving _optional_ comments, as far as I remember).
2. Using `import type` does not cause problems later on when you're using the _auto import_ feature of TS (the `import type` is automatically transformed to a regular `import` when importing a JS value), so it does not introduce additional work later on when working on the file.
3. Having TS-language-server related IDEs use `import type` automatically is a matter of enabling the TS config option we already agreed on (I wrote about it in https://github.com/cloudify-cosmo/cloudify-stage/pull/1302#discussion_r613819740).
4. Having multiple imports for the same file is not something I saw widely used (I only saw it be required in Svelte when using TS). Using `import type`, on the other hand, is used more often, AFAIK

> Summing up, since the value of using type is very low and it's troublesome in different IDEs I encourage dropping the rule to use, treat it as a cosmetic thing and stop commenting it.

2 points here :slightly_smiling_face: 
1. As I remember, I always added _optional_ comments to use `import type`, meaning that I did not expect it, only suggested it :slightly_smiling_face:
2. RD-2000, which enforces usage of `import type` when possible, as mentioned in quite a few PR reviews. I thought we agreed to implement it. Using `import type` when an import only contains types will make the diff in RD-2000 smaller. Thus, I see value in still using `import type` in cases where RD-2000 will require fixing the code (which is exactly the mindset I explained in https://github.com/cloudify-cosmo/cloudify-stage/pull/1325#discussion_r619019112)

If, despite the points above, you are not willing to use `import type` yourself (in cases there's a single import with only types) and do not want to receive optional comments about it in your PRs, let me know :slightly_smiling_face:

Gelio:
Only the topology did not pass in the latest run :slightly_smiling_face: However, there are conflicts now since #1324 was merged. I'll resolve them soon and just run the tests locally

Gelio:
I've resolved the conflict (removed the `production` variable):
image

Tests passed locally afterward:

image


With that, I'm looking for reviews once again :slightly_smiling_face:

widgets/common/src/map/options.ts
kubama: So the `type` import only affects compilation time? I wonder what's the benefit for a single occurrence? And if commenting it will ever pay off in terms of time required for writing the comment and then addressing it.

> Using import type when an import only contains types will make the diff in RD-2000 smaller

Hard to call it a benefit, really.

I'm still opting for dropping the rule (until RD-2000 is implemented).

Gelio: > And if commenting it will ever pay off in terms of time required for writing the comment and then addressing it.

You are right :+1: 

> Hard to call it a benefit, really.

What would you call it :smile: ? Having fewer changes to review is an objective benefit :smile: 


> I'm still opting for dropping the rule (until [RD-2000](https://cloudifysource.atlassian.net/browse/RD-2000) is implemented).

What rule? I did not know there was a rule :smile:  I thought I am just adding optional comments, which I did not expect to be implemented before approving some PR :slightly_smiling_face:

kubama: I'm not willing to continue this thread. My final thoughts are:
- I don't see a real benefit of using `type` imports. I also have nothing against using it. It's troublesome in some IDEs which is an additional reason why I think it's OK if it's missing. Easier RD-2000 implementation is not an argument to me - there is no benefit, just moving work from one place to another.
- Because of the previous point I suggest stop commenting on lack of `type` import - even if the comment is optional it costs some time to write it, and even if the PR author is not willing to address it it should be answered to indicate that it was at least read (unless we are OK with ignoring optional comments). Lets value each others time.

kubama (dismissed):

widgets/common/src/map/options.ts
Gelio: Ok, sure thing, I'll skip adding such comments in your PRs :slightly_smiling_face:

widgets/common/src/deploymentsView/map/container.tsx
Gelio: Whoops, I forgot to add the topic to today's weekly. I've added it to the next one

widgets/common/src/NoDataMessage.tsx
```diff
@@ -0,0 +1,44 @@
...
+const NoDataMessage: FunctionComponent<NoDataMessageProps> = props => {
+    const { Message } = Stage.Basic;
+    const { MessageContainer } = Stage.Basic;
+    return (
+        <MessageContainer wide margin="30px auto">
+            <Message>
+                {props.children ||
+                    (((props.error as Record<string, any>) || {}).name === 'SyntaxError'
+                        ? `The widget content cannot be displayed because configured URL does not point to a valid ${props.repositoryName} repository JSON data. Please check widget's configuration.`
+                        : `The widget content cannot be displayed because there is no connection to ${props.repositoryName} repository. Please check network connection and widget's configuration.`)}
```
qooban, line 22: Optional:

What do you think about moving it to `en.json`?

widgets/common/src/NoDataMessage.tsx
```diff
@@ -0,0 +1,44 @@
...
+type NoDataMessageProps =
+    | {
+          children: ReactElement;
+          repositoryName: never;
+          error: never;
+      }
+    | { error?: { name: string }; repositoryName: string };
+
+// NOTE: cannot destructure props due to using a type union
+/* eslint-disable react/destructuring-assignment */
```
qooban, line 12: So props destructuring cannot be done in both - function signature and function body?

widgets/common/src/deploymentsView/map/container.tsx
```diff
@@ -0,0 +1,63 @@
...
+    toolbox,
+    widgetDimensions
+}) => {
+    const sitesResult = useQuery(
+        'all-sites',
+        (): Promise<Stage.Types.PaginatedResponse<Stage.Common.Map.Site>> =>
+            toolbox.getManager().doGet('/sites', {
+                _include: 'name,latitude,longitude',
+                _get_all_results: true
+            })
```
qooban, line 25: Optional:

I'm just wondering maybe we should have something like `SiteActions` for such `doGet`'s in `common/src`? 

Currently there is `SiteActions.js` file inside `widgets/sites`, so I don't have good solution (as moving `SiteActions.js` to `common/src` would probably be a bit artificial as all endpoints are used locally in the Sites widgets).

widgets/common/src/map/DefaultTileLayer.tsx
qooban, line 11: I'm wondering what kind of logic you use to decide whether you want to use destructuring, 
like: `const { urlTemplate } = Stage.Common.Consts.leaflet;`
or avoid destructuring, 
like: `return <Stage.Basic.Leaflet.TileLayer...`
?

widgets/common/src/map/site.ts
```diff
@@ -0,0 +1,19 @@
...
+    name: string;
+    latitude: number | null;
+    longitude: number | null;
+}
+
+export interface SiteWithPosition {
+    name: string;
+    latitude: number;
+    longitude: number;
+}
```
qooban, line 11: **Just** wondering as I'm not TS expert - there is no way to somehow connect each other that way that one type will extend/modify the other?

widgets/common/src/map/site.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function isSiteWithPosition(site: Site): site is SiteWithPosition {
+    return !!site.latitude && !!site.longitude;
```
qooban, line 14: What if `site.latitude === 0` or `site.longitude === 0`? 🙂 

(I know there's probably very low probability it can happen, but still London/Greenwich is at lng = 0 🙂 )

widgets/common/src/deploymentsView/map/map.tsx
```diff
@@ -0,0 +1,79 @@
...
+            style={{ height: '100%' }}
+        >
+            <DefaultTileLayer />
+            {deploymentSitePairs.map(({ deployment, site }) => (
+                <DeploymentSiteMarker deployment={deployment} site={site} key={`${deployment.id}-${site.name}`} />
+            ))}
+        </MapComponent>
+    );
+};
+export default DeploymentsMap;
```
qooban, line 48: Is it intentional that you name the file `map.tsx` where you have `export default DeploymentsMap`?

General:

I thought that the common pattern is to name the file the same as the component exported.

widgets/common/src/deploymentsView/map/map.tsx
```diff
@@ -0,0 +1,79 @@
...
+        .map((deployment): DeploymentSitePair => ({ deployment, site: sitesLookupTable[deployment.site_name] }))
+        // NOTE: additional filtering, since the site may not have a position, and thus not be in the lookup table
+        .filter(({ site }) => !!site);
+
+const deploymentStatusToIconColorMapping: Record<DeploymentStatus, Stage.Common.MarkerIconColor> = {
+    [DeploymentStatus.Good]: 'blue',
+    [DeploymentStatus.InProgress]: 'yellow',
+    [DeploymentStatus.RequiresAttention]: 'red'
+};
+const DeploymentSiteMarker: FunctionComponent<DeploymentSitePair> = ({ deployment, site }) => {
```
qooban, line 69: I'm wondering if the order is right (`DeploymentSiteMarker` defined after using it in `DeploymentsMap`). 
Compiler is doing the work probably, but shouldn't we provide the correct order for the reader?

qooban (commented):
Looks good in general 👍🏻 I added some comments.

Did quick UI tests:
1. (minor) Maybe we should have margins dependant on the zoom level? (that way the marker icon will be visible)
image

2. (not related to the PR) I found that maximize button does not work properly when we have smaller screen (hidden sidebar):
https://user-images.githubusercontent.com/5202105/116040375-3e7d2600-a66c-11eb-89df-fc3d886d1b13.mp4
Reported RD-2128.

test/cypress/support/commands.ts
```diff
@@ -49,6 +49,15 @@ declare global {
     }
 }
 
+/** See `fixtures/license` directory */
+type License =
+    | 'expired_trial_license'
+    | 'invalid_license'
+    | 'tampered_paying_license'
+    /** Use valid_spire_license for testing maps */
```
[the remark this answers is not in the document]
Gelio: Good catch :+1: I added it before clarifying that the Spire license is not necessary to test maps. I'll remove it

widgets/common/src/NoDataMessage.tsx
Gelio: Sure thing :+1: That's moved code, but I may as well do it, since I'll be making other changes too :slightly_smiling_face:

widgets/common/src/NoDataMessage.tsx
Gelio: I double-checked and TypeScript seems to be happy with me destructuring a union here :slightly_smiling_face: I'll use destructuring

widgets/common/src/map/DefaultTileLayer.tsx
Gelio: In this case:
* `urlTemplate` is used twice, so destructuring results in less repetition
* `TileLayer` is only used once (as a self-closing tag, so it does not even have to be repeated), so having it on a single line is short and concise for me :slightly_smiling_face:

    If there was another import from `Stage.Basic.Leaflet` that was necessary in this file, I would probably destructure that other thing and `TileLayer` together, but since there is not, the code looks the way it does now

widgets/common/src/map/site.ts
Gelio: I could use TS mapped types here:

```ts
type SiteWithPosition = {
    [K in keyof Site]: NonNullable<Site[K]>;
};
```

but I thought that's harder to read than just repeating the properties. Let me know if you'd prefer the `type` solution

widgets/common/src/map/site.ts
Gelio: Good catch :+1: I did not think about 0 being falsy when copying the code :smile: I'll use `typeof` instead

widgets/common/src/deploymentsView/map/map.tsx
Gelio: Sure, I'll rename it to `DeploymentsMap`. The reason I left is as map is to avoid having long names and use the context of the current directory - we are in the `deploymentsView` directory, and this is just a `map` in that directory :smile:

widgets/common/src/deploymentsView/map/map.tsx
Gelio: IMO that's a matter of preference, not correctness :smile: It depends if you prefer reading the code from the top-down or from the bottom-up :slightly_smiling_face: I prefer top-down - starting with the main/most important component (the map in this case), and leaving the lower-level details down below :slightly_smiling_face: 

AFAIK, no compilation is necessary for this order to work. You can use symbols declared lower inside of functions that are declared higher in the code :slightly_smiling_face: An example from the console, without any compilation:
image

Gelio:
@qooban 

> Did quick UI tests:
Thanks!

> 1. (minor) Maybe we should have margins dependant on the zoom level? (that way the marker icon will be visible)
>    image

There is some padding added in `getMapOptions` that I copy-pasted from the Sites Map, but I can see how it's too low. I will play around with it and give it a few minutes, but if I fail to get it right, I hope it would be ok to address it in a separate ticket later on and focus on work of more priority :smile: I'll let you know how it goes

> 2. (not related to the PR) I found that maximize button does not work properly when we have smaller screen (hidden sidebar):
>    https://user-images.githubusercontent.com/5202105/116040375-3e7d2600-a66c-11eb-89df-fc3d886d1b13.mp4
>    Reported [RD-2128](https://cloudifysource.atlassian.net/browse/RD-2128).
Good catch, thanks for reporting it :+1:

Gelio:
@qooban I hope I have responded to all the comments :slightly_smiling_face: 

I did not end up fixing the markers being partially cut off - I have created a bug to address it in the future: RD-2133

I'm looking for more reviews :slightly_smiling_face: I'm not going to rerun the system tests

test/cypress/support/commands.ts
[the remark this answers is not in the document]
Gelio: Removed

widgets/common/src/NoDataMessage.tsx
Gelio: Done

widgets/common/src/deploymentsView/map/container.tsx
qooban: Agree 👍🏻

widgets/common/src/map/DefaultTileLayer.tsx
qooban: Got it.

widgets/common/src/map/site.ts
qooban: I see, let's leave it as it is, then.

qooban (approved):

Gelio:
Thanks for a re-review :slightly_smiling_face: I assume that @kubama is still fine with the changes, since he approved it before, so I'll merge the PR as soon as the CI passes

Gelio:
The build failed due to a Jenkins issue (RD-2134). I have force-pushed the latest commit to force Jenkins to rerun the pipeline. I'm looking for a re-approval :slightly_smiling_face: No code changed

qooban (approved):

Gelio:
Now the CI failed due to `.size-limit` :angry: I hope it will pass now :smile:

qooban (approved):

Gelio:
`EOFException` again... :facepalm: I've force-pushed to retrigger CI

qooban (approved):
You've got no luck. Definitely something to point out on retro.

BTW Maybe we can merge https://github.com/cloudify-cosmo/cloudify-stage/pull/1323 first then? :-)

Gelio:
> You've got no luck. Definitely something to point out on retro.

Yes :smile: 

> BTW Maybe we can merge #1323 first then? :-)

Sure, I approved it, so go ahead :slightly_smiling_face:

Gelio:
Great, the CI finally passed :smile: I'll merge the PR

PR #1329, opened 2021-04-26
RD-1768 Added test for filter button

kubama:


test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -222,6 +223,23 @@ describe('Deployments View widget', () => {
...
+
+            cy.contains('Filter').click();
+
+            cy.get('.modal').within(() => {
+                cy.get('input').type(`${filterId}{enter}`);
+                cy.contains('OK').click();
+            });
+
+            cy.contains(deploymentNameThatMatchesFilter);
+            cy.contains(deploymentName).should('not.exist');
```
qooban, line 241: Maybe we can also check if proper HTTP request is sent with the selected `filterId`?

qooban (commented):
Added one comment.

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -222,6 +223,23 @@ describe('Deployments View widget', () => {
 
             cy.contains(/with ID .* was not found/);
         });
+
+        it('should take the selected filter into account when displaying deployments', () => {
+            useDeploymentsViewWidget({});
```
Gelio, line 228: Optional:

```suggestion
            useDeploymentsViewWidget();
```

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -222,6 +223,23 @@ describe('Deployments View widget', () => {
...
+
+            cy.contains(deploymentNameThatMatchesFilter);
+            cy.contains(deploymentName);
+
+            cy.contains('Filter').click();
+
+            cy.get('.modal').within(() => {
+                cy.get('input').type(`${filterId}{enter}`);
```
Gelio, line 236: Can we use some other more specific selector instead of just `input`? Perhaps a `name`, or some label? This could result to be flaky if there are other inputs in the form :slightly_smiling_face:

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio, line 241: The first assertion will likely pass before the response is received, meaning it will always pass, since the table displays the deployments from the initial request.

How about either:
1. Reversing the order, so we know that the table displays the data from the 2nd request?
2. Waiting for the loader to disappear?

I suggest the first solution, as it's easier

kubama: Done, although in this case we are posting filter rules, not filter id.

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: Done

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: Good point, I reversed the order

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: At the moment there is nothing more specific that could be used as there is only one (thus unlabelled) form input in the modal. Input label would be the most natural selector to use, but it will be introduced once this modal is extended. As for now I'd rather leave it as it is.

kubama:
Since this PR only adds one test and alters another one I consider a local test execution a sufficient check of system tests stability. There is no way anything outside of the `deployments_view_spec.ts` could be broken .

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -182,25 +181,28 @@ describe('Deployments View widget', () => {
     });
 
     describe('with filters', () => {
+        const deploymentNameThatMatchesFilter = `${specPrefix}precious_deployment`;
         const filterId = 'only-precious';
+        const filterRules = [
```
Gelio, line 186: Optional:

Could we add a type annotation?

```suggestion
        const filterRules: Stage.Common.Filters.Rule[] = [
```

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: I see, ok :+1:

Gelio (approved):
Clear and exhaustive :+1:

Gelio (approved):

kubama:
I consider this PR small so I'm merging with 1 approval.

PR #1330, opened 2021-04-26
RD-2093 Handle Deployment Map selection

Gelio:
## Changes

1. Use `Tooltip` instead of `Popup` to display information about a marker.

    It should have been a tooltip right from the beginning :smile: It can't be a popup, since clicking the marker should select the deployment, not open the popup.

2. Add a circle on the map underneath the selected deployment to highlight it
3. Select the correct deployment while clicking a marker on the map
4. Add tests for the above

## Next steps

1. RD-1526 - adding more info in the tooltip
2. RD-2135 - auto-panning the map so the selected deployment marker is scrolled into view
3. RD-1527 - map marker clustering

## Video

https://user-images.githubusercontent.com/889383/116108565-4d87c680-a6b4-11eb-8350-1a0619c66a60.mp4

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/520/pipeline

Ran 2021-04-28 10:55 CEST

The deployments view spec passed locally:
image


test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -11,7 +11,7 @@ describe('Deployments View widget', () => {
     const specPrefix = 'deployments_view_test_';
     const blueprintName = `${specPrefix}blueprint`;
     const deploymentName = `${specPrefix}deployment`;
-    const siteName = 'Olsztyn';
+    const exampleSiteName = 'Olsztyn';
```
Gelio, line 14: I have changed the name to be more specific since I could not use `siteName` as a variable name in the map tests anymore

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -526,9 +526,57 @@ describe('Deployments View widget', () => {
...
         const getMarkerByImageSrcSuffix = (srcSuffix: string) =>
             cy.get(`.leaflet-marker-pane img[src$="${srcSuffix}"]`);
+        const withinMarkerTooltip = (
+            getMarker: () => Cypress.Chainable,
+            callback: (currentSubject: JQuery<HTMLElement>) => void,
+            { force = false }: { force?: boolean } = {}
+        ) => {
+            getMarker().trigger('mouseover', { force });
+            getDeploymentsMapTooltip().within(callback);
+            getMarker().trigger('mouseout', { force });
```
Gelio, line 578: There is no `.hover` for Cypress :worried: In this case, I had to use such triggers

Gelio:
I've hit `EOFException` (RD-2134) in https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/511/pipeline/56/ :worried: 

Here's a rerun:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/514/pipeline

.size-limit
```diff
@@ -7,7 +7,7 @@
     },
     {
         "name": "Vendor JS bundle",
-        "limit": "4.22 MB",
+        "limit": "4.23 MB",
```
Gelio, line 10: This increase is likely due to exposing all of `react-leaflet`.

The actual size increase was 9.08 KB:

https://jenkins.cloudify.co/blue/organizations/jenkins/cloudify-stage/detail/RD-2093-handle-deployment-map-selection/8/pipeline

```
[2021-04-26T15:37:29.888Z]   Main JS bundle
[2021-04-26T15:37:29.888Z]   Size limit:   326 KB
[2021-04-26T15:37:29.888Z]   Size:         325.27 KB
[2021-04-26T15:37:29.888Z]   Loading time: 6.6 s     on slow 3G
[2021-04-26T15:37:29.888Z]   Running time: 196 ms    on Snapdragon 410
[2021-04-26T15:37:29.888Z]   Total time:   6.8 s
[2021-04-26T15:37:29.888Z]   
[2021-04-26T15:37:29.889Z]   Vendor JS bundle
[2021-04-26T15:37:29.889Z]   Package size limit has exceeded by 9.08 KB
[2021-04-26T15:37:29.889Z]   Size limit:   4.22 MB
[2021-04-26T15:37:29.889Z]   Size:         4.23 MB
[2021-04-26T15:37:29.889Z]   Loading time: 86.7 s  on slow 3G
[2021-04-26T15:37:29.889Z]   Running time: 2.4 s   on Snapdragon 410
[2021-04-26T15:37:29.889Z]   Total time:   89 s
[2021-04-26T15:37:29.889Z]   
[2021-04-26T15:37:29.889Z]   Widgets
[2021-04-26T15:37:29.889Z]   Size limit:   2.6 MB
[2021-04-26T15:37:29.889Z]   Size:         2.59 MB
[2021-04-26T15:37:29.889Z]   Loading time: 53 s    on slow 3G
[2021-04-26T15:37:29.889Z]   
[2021-04-26T15:37:29.889Z]   Application package
[2021-04-26T15:37:29.889Z]   Package size limit has exceeded by 8.85 KB
[2021-04-26T15:37:29.889Z]   Size limit:   13.3 MB
[2021-04-26T15:37:29.889Z]   Size:         13.31 MB
[2021-04-26T15:37:29.889Z]   Loading time: 272.6 s  on slow 3G
```

Gelio:
The system tests failed due to an error in `cfy_manager/components/restservice/scripts/create_tables_and_add_defaults.py` script. Looks like the same problem that occurred in the nightly (which will likely be reported soon by @gregcfy). I'll not be rerunning the system tests for now since I do not see the point as it will fail

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -526,9 +526,57 @@ describe('Deployments View widget', () => {
...
+            });
+        });
+
+        const getDeploymentsMapTooltip = () => cy.get('.leaflet-tooltip');
         const getMarkerByImageSrcSuffix = (srcSuffix: string) =>
             cy.get(`.leaflet-marker-pane img[src$="${srcSuffix}"]`);
+        const withinMarkerTooltip = (
+            getMarker: () => Cypress.Chainable,
+            callback: (currentSubject: JQuery<HTMLElement>) => void,
+            { force = false }: { force?: boolean } = {}
```
kubama, line 574: I can see why it was handy to name this parameter `force`, but looking at the function name and other parameters it's mysterious what this flag actually does, one need to dig into the implementation to understand. How about naming it in a more descriptive way?

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -600,16 +616,74 @@ describe('Deployments View widget', () => {
...
             });
+
+            const getSelectedMarker = () => cy.get('path.test__map-selected-marker');
+
+            cy.getSearchInput().type(mapDeploymentsPrefix);
+
+            function selectDeploymentInTableAndVerifyMapSelection(name: string) {
+                getDeploymentsViewTable().within(() => {
+                    cy.contains(name).click();
+                });
```
kubama, line 648:
```suggestion
                getDeploymentsViewTable().contains(name).click();
```

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -600,16 +616,74 @@ describe('Deployments View widget', () => {
...
+                            withinMarkerTooltip(
+                                () => cy.wrap(marker),
+                                element => {
+                                    currentDeploymentName = element.text();
+                                }
+                            );
+                            cy.wrap(marker).click();
+                            getDeploymentsViewTable().within(() => {
+                                cy.get('tr.active').contains(currentDeploymentName);
+                            });
```
kubama, line 684:
```suggestion
                            getDeploymentsViewTable().find('tr.active').contains(currentDeploymentName);
```

widgets/common/src/deploymentsView/map/DeploymentsMap.tsx
```diff
@@ -40,18 +58,19 @@ const DeploymentsMap: FunctionComponent<DeploymentsMapProps> = ({ deployments, s
         >
             <DefaultTileLayer />
             {deploymentSitePairs.map(({ deployment, site }) => (
-                <DeploymentSiteMarker deployment={deployment} site={site} key={`${deployment.id}-${site.name}`} />
+                <DeploymentSiteMarker
+                    deployment={deployment}
+                    site={site}
+                    key={`${deployment.id}-${site.name}`}
```
kubama, line 64: This would be an edge case, but `-` is a valid character in both deployment id as well as site name. Thus there is a risk of key duplication. To be safe the values should be joined with a character that can't be present in one of the values.

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Hmm... Good point :+1: I don't have a single good name :worried: I have the following suggestions that came to my mind:

1. `forceOpen`
2. `ignoreMarkerHidden`
3. `ignoreMarkerNotVisible`
4. `ignoreMarkerObstructed`
5. `ignoreMarkerNotInView`

Is one of these suitable for you? If not, do you have another suggestion?

widgets/common/src/deploymentsView/map/DeploymentsMap.tsx
Gelio: Right, right! I have forgotten to change it, I'll come up with something else :slightly_smiling_face: Perhaps a `\n`?

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Good suggestion :+1: It would normally work, but I tried that and found out it does not work. The problem is that `currentDeploymentName` is set dynamically in a closure just above this code, which happens asynchronously.

When `.contains(currentDeploymentName)` is called inside `within`, the `currentDeploymentName` is already set in the previous closure. When we use `currentDeploymentName` outside of `within`, in the parent scope, the `currentDeploymentName` is not set, so we call essentially:

```ts
getDeploymentsViewTable().find('tr.active').contains('');
```

I can add a comment about the `currentDeploymentName` variable so it is clearer, hopefully :slightly_smiling_face:

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: I think the last option is the best 👍

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Good point, thanks! :+1: Sorry for not committing the suggestion so you have a contribution - I applied the suggestion manually, as I wanted to batch it in a single commit with the one below, but it turns out the one below is not possible, so I just left the one I already committed myself :slightly_smiling_face:

widgets/common/src/deploymentsView/map/DeploymentsMap.tsx
Gelio: I've changed it to a `\n`. It works well, the tests pass

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Great, I have renamed `force` to `ignoreMarkerNotInView` then :slightly_smiling_face:

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: I think this kind of dependency should be handled by `then`. If `currentDeploymentName` was passed through the command chain then the correct handling would be enforced. It would also be self-descriptive without the need for the comment.

Gelio: Very probable and I agree :slightly_smiling_face: I tried doing that before creating the PR, but I did not know how to do it efficiently and effectively, given the existing `withinMarkerTooltip` helper. The problems I have noticed:
1. The structure of `withinMarkerTooltip` would have to greatly change, perhaps to something along the lines of the following (which I am not sure would work, I did not check):
    ```ts
            getMarker().trigger('mouseover', { force: ignoreMarkerNotInView });
            getDeploymentsMapTooltip()
                .within(callback)
                .then(v => {
                    getMarker().trigger('mouseout', { force: ignoreMarkerNotInView });
                    return cy.wrap(v);
                });
    ```
2. `callback` would have to return some data. However, `within` does not change the Cypress's subject, so there would have to be some additional logic (I do not know what logic) to propagate the value returned by the `callback` to the next `then`

If you know how to do it, could you commit it yourself? Since I have already spent some time on it and failed, it would be a time sink for me to try to fiddle it with again :slightly_smiling_face: If the implementation is easy and I just bluntly missed it, let me know

app/components/basic/index.ts
```diff
@@ -73,7 +73,7 @@ import {
     VisibilityIcon
 } from 'cloudify-ui-components';
 
-import { Map, TileLayer, Marker, Popup as LeafletPopup } from 'react-leaflet';
+import * as Leaflet from 'react-leaflet';
```
qooban, line 76: Do you think we need all the exported data from `react-leaflet`?

qooban (commented):
Looks good 👍🏻 
Just added some questions.

widgets/common/src/deploymentsView/map/marker.tsx
```diff
@@ -0,0 +1,76 @@
...
+        );
+    }
+
+    return (
+        <BareDeploymentSiteMarker status={deployment.deployment_status} position={position} onClick={onClick}>
+            {tooltip}
+        </BareDeploymentSiteMarker>
+    );
+};
+export default DeploymentSiteMarker;
```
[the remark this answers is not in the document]
Gelio: Sure thing, thanks for adding it yourself :slightly_smiling_face:

widgets/common/src/deploymentsView/map/marker.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const icon = Stage.Common.createMarkerIcon(deploymentStatusToIconColorMapping[status]);
+
+    return (
+        <Marker icon={icon} position={position} riseOnHover onclick={onClick}>
+            {children}
+        </Marker>
+    );
+};
+
+const tooltipOffset = L.point(0, -40);
```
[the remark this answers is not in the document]
Gelio: It does not strictly match the icon's height, I just settled on the first number that looked fine :smile: I could use `41`, since that's the height of the icon. Then, I could add an explanatory comment. Would you like that?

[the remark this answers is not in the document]
qooban: If `41` is the height of the icon and you can use it maybe we can have something like:
```suggestion
const markerIconHeight = 41;
const tooltipOffset = L.point(0, -markerIconHeight);
```
That way we will not need a comment IMHO 🙂 WDYT?

app/components/basic/index.ts
qooban: OK 👍🏻

Gelio: I have reverted the change and only exposed some components. I have also reverted the `.size-limit` change, maybe it will pass without increasing it unnecessarily. Let's see what the CI tells us :smile:

widgets/common/src/deploymentsView/map/marker.tsx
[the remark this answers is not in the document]
Gelio: I have used `41`px now and added an explanatory comment

app/components/basic/index.ts
Gelio: Looks like the CI passed with the current size limits :slightly_smiling_face:

widgets/common/src/deploymentsView/map/marker.tsx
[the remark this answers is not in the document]
Gelio: Yeah, your solution is better :smile: Nice one, I'll use it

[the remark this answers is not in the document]
Gelio: Done

qooban (approved):
Looks good to me 👍🏻

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: Looks like all `withinMarkerTooltip` usages except the one here are using a callback `() => cy.contains(name)`. If so, I would change the function signature to take the name as an input instead of the callback. `within` in L577 would be no longer necessary and `withinMarkerTooltip` could then directly return a chainable with subject set to the value we need here.

Gelio: With the code in this PR, I suppose we could accept `name` instead of the callback in `withinMarkerTooltip`. It will not be viable with RD-1526, since I will have to verify a lot more information than just the name, meaning that the callback passed to `withinMarkerTooltip` will be much larger. Thus, I don't see how I could go forward without the callback :slightly_smiling_face: 

Anyway, even if I did the change and only accepted the name, how would this (L684) look like? What would we pass as the `name` here? It is not yet known, because we don't know the order of the tooltips in the DOM :worried: 

As you can hopefully see, the problem is not trivial, so I encourage you to try to make some changes yourself :slightly_smiling_face: This way we avoid the back-and-forth of suggesting some changes that may not be viable in the long run :smile: Or, we could just accept the current implementation

kubama: I'm not happy with the current implementation, but after giving it a try I was not able to find a solution too. Lets leave it then.

app/components/basic/index.ts
kubama: I'm not sure what was the issue here. Even if unused components are exported they are grouped in the `Leaflet` object, so they do not provide a mess or pollute the top level export, so I don't see any problem here. The solution with * import is simpler, less error prone and easier to maintain.

qooban: IMHO it's better to be specific and explicit in this case.

When I looked at `react-leaflet`, I found that export word is used 81 times in the main module file. 
What's the reason for having unused components and at the same time increase bundle size?
I can agree that `import *` is simpler, but how it makes the code to be less error prone and easier to maintain?

BTW @Gelio We need also to update https://docs.cloudify.co/latest/developer/writing_widgets/widgets-components/#react-leaflet . Can be done once the map epic is finished.

kubama (approved):

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Thanks for taking the time to tackle the issue :+1:

app/components/basic/index.ts
Gelio: I believe some middle ground would be to use `import * as Leaflet`, but then only expose some components from the Leaflet API on `Stage`. I believe webpack's tree-shaking would cut out the unused components :slightly_smiling_face: Anyway, that's a matter for another time, when someone else will be requiring some more leaflet components :smile: 

> BTW @Gelio We need also to update https://docs.cloudify.co/latest/developer/writing_widgets/widgets-components/#react-leaflet . Can be done once the map epic is finished.

Good to know :+1: I was not aware. I'll create a task to update the docs **and** insert a comment here (when exposing `Leaflet` in `Stage.Basic`) that the docs have to be updated when the public-facing API changes :slightly_smiling_face:

Gelio:
Thanks for reviews and approvals :smile: I've scheduled system tests: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/518/pipeline

I'll wait for the results and then merge the PR if there will be no red-flags :slightly_smiling_face:

Gelio:
Yay, an `EOFException` again :angry: 

Another system tests run: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/520/pipeline

Gelio:
The only CI error is the unrelated `filters_spec` error. Knowing that, I'll merge the PR as soon as the CI passes :slightly_smiling_face:

PR #1331, opened 2021-04-27
RD-2079 Added bulk actions button

kubama:
This change is trivial but I'm submitting it as a separate PR as it's an integration point for all bulk actions related stories.

Screenshot:

image

Clicked:

image

I'm not planning system tests execution on the branch.

### DIFF
--- a/widgets/common/src/deploymentsView/header/index.tsx
+++ b/widgets/common/src/deploymentsView/header/index.tsx
@@ -22,7 +22,9 @@ const DeploymentsViewHeader: FunctionComponent<DeploymentsViewHeaderProps> = ({
     const [filterModalOpen, openFilterModal, closeFilterModal] = Stage.Hooks.useBoolean();
     const [filterId, setFilterId] = useState<string>();
 
-    const { Button } = Stage.Basic;
+    const { Button, Dropdown } = Stage.Basic;
+    // @ts-ignore Properties does not exist on type 'typeof Dropdown'
+    const { Menu, Item } = Dropdown;
 
     function handleFilterChange(newFilterId: string | undefined) {
         setFilterId(newFilterId);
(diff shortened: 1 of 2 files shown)


widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -63,6 +65,11 @@ const DeploymentsViewHeader: FunctionComponent<DeploymentsViewHeaderProps> = ({
                     onClick={openFilterModal}
                 />
             )}
+            <Dropdown button text={headerT('bulkActions.button')}>
```
qooban, line 68: This will probably be the first time we use such `Dropdown` (with `button` prop set to `true`).
I see it's a bit different than what we used so far for dropdown menus (in most cases I saw `PopupMenu`). Am I right?

I'm not saying it's bad or it's good. Maybe that approach would be simpler and better, will see. 
But for sure the look'n'feel is a bit different. Maybe we should just add `pointing` prop to have alignment with other dropdown menus?

qooban (commented):
Just one comment.

widgets/common/src/deploymentsView/header/index.tsx
kubama: The approach is slightly different, but the scenario is different as well. As you pointed out so far we didn't use dropdown with `button` prop. Here we are using it to be in line with other buttons in the widget header - at the moment there are map and filter buttons. In this case I think the regular dropdown look'n'feel suites better and is closer to a common sense.
Please have a look how it would be with the `pointing` attribute:
image
I find the distance between the dropdown button and the menu not needed and actually awkward looking, but I don't insist.

qooban: Let's leave it for now, we can verify it with the product later on. The change should be quick if necessary.

qooban (approved):

widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -22,7 +22,9 @@ const DeploymentsViewHeader: FunctionComponent<DeploymentsViewHeaderProps> = ({
     const [filterModalOpen, openFilterModal, closeFilterModal] = Stage.Hooks.useBoolean();
     const [filterId, setFilterId] = useState<string>();
 
-    const { Button } = Stage.Basic;
+    const { Button, Dropdown } = Stage.Basic;
+    // @ts-ignore Properties does not exist on type 'typeof Dropdown'
```
Gelio, line 26: We could have used `@ts-expect-error`, as that one will let us know that this place should be improved in case we fix the types for the `Dropdown` component so `Menu` and `Item` are available in TS :slightly_smiling_face:

PR #1332, opened 2021-04-27
Rd 2077 add simple e2e tests fix for getting started modal that makes tests passing

gregcfy:
This PR fixes problem of the not working e2e test when Getting Started Modal is enabled.
The test changes `show_getting_started` flag to `false` when resuest for user options is called (`/users/{user_name_here}`).

Ticket: https://cloudifysource.atlassian.net/browse/RD-2077


kubama (commented):
Looks like you added explicit `disableGettingStarted` in all places where it is needed. But in fact we need the modal enabled only for the dedicated test and disabled in all other cases, right? Wouldn't it be reasonable to make the disabled state default? It could be disabled by default in `usePageMock` or `mockLogin` without the need for disabling it explicitly in each test. It would make this PR considerably smaller.

Gelio:
I agree with @kubama  - good point :smile: I'll wait with reviewing the PR until that is addressed, so as not to waste time reviewing code that will be mostly gone :slightly_smiling_face:

test/cypress/integration/tenants_spec.js
```diff
@@ -40,6 +40,7 @@ describe('Tenants menu', () => {
         user.tenants.forEach(tenant => cy.addUserToTenant(user.username, tenant.name, tenant.role));
 
         cy.intercept('/console/widgets/list', []);
+        cy.disableGettingStarted();
```
kubama, line 43: This spec uses `mockLogin` so why do we need explicit disable here?

test/cypress/support/commands.ts
```diff
@@ -203,6 +206,7 @@ const commands = {
     ) => {
         const widgetIdsArray = _.castArray(widgetIds);
         const widgetIdsToLoad = [...widgetIdsArray, 'filter', 'pluginsCatalog', ...additionalWidgetIdsToLoad];
+        cy.disableGettingStarted();
```
kubama, line 209: Do we need it here if we have it in `mockLogin`? Do we have any test that uses `usePageMock` but not `mockLogin`?

test/cypress/integration/auth_spec.ts
```diff
@@ -1,5 +1,5 @@
 describe('Authentication', () => {
-    beforeEach(() => cy.clearCookie('XSRF-TOKEN'));
+    beforeEach(() => cy.clearCookie('XSRF-TOKEN').disableGettingStarted());
```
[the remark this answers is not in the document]
gregcfy: fixed

test/cypress/integration/widgets/filter_spec.js
```diff
@@ -44,6 +44,7 @@ describe('Filter', () => {
 
         before(() =>
             cy
+                .disableGettingStarted()
```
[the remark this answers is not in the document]
gregcfy: when `disableGettingStarted()` is not called here fiter_spec test `fails` because the Modal covers page.
image

test/cypress/integration/widgets/outputs_capabilities_spec.js
```diff
@@ -21,6 +21,7 @@ describe('Outputs/Capabilities', () => {
     describe('presents data and export button for', () => {
         before(() => {
             setUpBlueprint('outputs');
+            cy.disableGettingStarted();
```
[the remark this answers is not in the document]
gregcfy: when `disableGettingStarted()` is not called here `fiter_spec` test fails because the Modal covers page.

test/cypress/integration/tenants_spec.js
gregcfy: fixed

test/cypress/integration/widgets/filter_spec.js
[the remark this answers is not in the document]
kubama: What's special about this spec that the modal appears despite of the disable in `mockLogin`?

test/cypress/integration/widgets/outputs_capabilities_spec.js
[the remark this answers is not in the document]
kubama: What's special about this spec that the modal appears despite of the disable in `mockLogin`?

test/cypress/support/commands.ts
gregcfy: yes, it is required, there are some tests that crushed, e.g. edit_mode_spec.js

gregcfy: it is caused because login action is executed only once before all tests, later page is refreshed/re-rendered that causes displaying modal again

kubama: Looks like to common point is the `refreshTemplate` command. If it causes the modal to reappear then it should also perform the disable internally. This should remove the need for explicit disable in edit_mode_spec and the two widget tests I was commenting on.

kubama: If this proves to be true then we should no longer need the disable in `usePageMock`, which is not the right place for it as it is not performing any action that causes the modal to appear.

test/cypress/integration/login_spec.js
```diff
@@ -1,4 +1,6 @@
 describe('Login', () => {
+    beforeEach(() => cy.disableGettingStarted());
```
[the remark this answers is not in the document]
gregcfy: fixed

test/cypress/support/commands.ts
gregcfy: moved to `refreshTemplate`

test/cypress/integration/widgets/outputs_capabilities_spec.js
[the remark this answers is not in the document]
gregcfy: login action was called in before and later cy.refreshTemplate() action was called.

fixed

test/cypress/integration/widgets/filter_spec.js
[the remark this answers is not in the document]
gregcfy: removed, currently refresh action disables modal

test/cypress/integration/widgets/outputs_capabilities_spec.js
kubama, line 24: Do we still need this?

gregcfy: fixed/removed

kubama (dismissed):

gregcfy:
all tests passed except one that is fixed in https://github.com/cloudify-cosmo/cloudify-stage/pull/1313

https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/522/

test/cypress/support/commands.ts
```diff
@@ -316,6 +323,12 @@ const commands = {
         return cy
             .exec(`./node_modules/.bin/babel --config-file ${babelConfigPath} ${scriptPath}`)
             .then(commandResult => commandResult.stdout);
+    },
+
+    disableGettingStarted: () => {
+        cy.interceptSp('GET', `/users/`, req => {
+            req.reply({ show_getting_started: false });
+        });
```
Gelio, line 331: Optional:

I suggest using a bit more declarative and IMO simpler approach:

```suggestion
        cy.interceptSp('GET', `/users/`, {
            body: { show_getting_started: false }
        });
```

test/cypress/support/commands.ts
```diff
@@ -153,6 +153,8 @@ const commands = {
             }
         });
 
+        cy.disableGettingStarted();
```
Gelio, line 156: Shouldn't this be called before calling `cy.visit` above, so we have certainty that the request is intercepted?

gregcfy: fixed

test/cypress/support/commands.ts
gregcfy: fixed

Gelio (approved):

kubama (approved):

PR #1333, opened 2021-04-27
RD-2081 - Add Executions page

qooban:
image


qooban:
System tests scheduled: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/516/

kubama (commented):
In the story description it reads:
> Enhancement: change the test to avoid using a hardcoded list of pages, and verify the displayed pages against the pages JSON returned from the API

Did you consider this enhancement?

qooban:
@kubama 
Good point. I forgot about it. Updated test. Local execution passes:
image

Full system tests execution failed with one test error (Filters widget, so not related to this PR) => I won't probably run another full execution.

test/cypress/support/templates.js
```diff
@@ -2,6 +2,8 @@ Cypress.Commands.add('getPages', () => cy.stageRequest('/console/templates/pages
 
 Cypress.Commands.add('getTemplates', () => cy.stageRequest('/console/templates'));
 
+Cypress.Commands.add('getBuiltInTemplate', name => cy.stageRequest(`/console/appData/templates/${name}.json`));
```
Gelio, line 5: Optional:

The file is still in JS, so the commands will not work in any TS files :smile: Did you consider converting it to TS?

Then, we could add a stricter type for `name` - the valid values would be `main-default` and `main-sys_admin`, which I believe is a nice benefit :slightly_smiling_face:

test/cypress/integration/templates/template_management_spec.js
```diff
@@ -12,55 +12,24 @@ describe('Template Management', () => {
...
-        });
-        cy.get('.horizontal > :nth-child(3)').within(() => {
-            for (let i = 0; i < tenants.length; i += 1) {
-                cy.get(`.divided > :nth-child(${i + 1})`).should('have.text', tenants[i]);
-            }
+
+    const verifyTemplateRow = (id, pages, roles, tenants) => {
+        cy.contains('tr', id).as('templateRow');
+        cy.get('@templateRow').within(() => {
+            roles.forEach(role => cy.get(`td:nth-of-type(2)`).should('contain.text', role));
```
Gelio, line 19: We usually use `contains(role)` instead of `should('contain.text', role)`. Is there a reason we use this new way here :thinking: ?

Gelio: If not, we could (optional) probably shorten it to:

```suggestion
            roles.forEach(role => cy.contains(`td:nth-of-type(2)`, role);
```

(I did not check it myself)

test/cypress/integration/templates/template_management_spec.js
```diff
@@ -12,55 +12,24 @@ describe('Template Management', () => {
         pages: ['catalog', 'blueprints']
     };
     const users = [defaultUser];
-    const builtInTemplates = [
```
Gelio, line 15: Lots of nice improvements in this area :+1: awesome work

test/cypress/integration/templates/template_management_spec.js
```diff
@@ -12,55 +12,24 @@ describe('Template Management', () => {
...
         });
-        cy.get(`${templateRow} > :nth-child(1)`).click();
+
+        cy.get('@templateRow').click();
+        cy.get('.horizontal > :nth-child(1)').within(() =>
+            pages.forEach((page, index) => cy.get(`.divided > :nth-child(${index + 1})`).should('have.text', page))
+        );
+        cy.get('.horizontal > :nth-child(3)').within(() =>
+            tenants.forEach((tenant, index) =>
+                cy.get(`.divided > :nth-child(${index + 1})`).should('have.text', tenant === '*' ? 'all' : tenant)
```
Gelio, line 28: The same comments about the use of `should('have.text'` apply here as do above :smile:

test/cypress/integration/templates/template_management_spec.js
```diff
@@ -197,10 +170,9 @@ describe('Template Management', () => {
 
         cy.log('Verify template changes');
         verifyTemplateRow(
-            builtInTemplates.length + 1,
             'Another Template',
             ['deployment', 'plugins', 'tmm'],
-            'manager, operations',
+            ['manager', 'operations'],
```
Gelio, line 175: Was that a bug :smile: ?

kubama (dismissed):

Gelio (dismissed):
The screenshot looks pixel-perfect with the one in the ticket :smile: I suggested some optional improvements. I don't mind leaving the code as-is :slightly_smiling_face:

test/cypress/support/templates.js
qooban: I didn't consider it, but that's good idea. Done.

test/cypress/integration/templates/template_management_spec.js
qooban: Thanks 🙂

test/cypress/integration/templates/template_management_spec.js
qooban: Hmm... I think this is matter of preference. 
I'm a bit more in favor of having constructs like `cy.get(...).should(...)` instead of just `cy.contains(...)` without explicit assertion method. Shorter length IMHO is not bigger advantage than readability and clarity.

test/cypress/integration/templates/template_management_spec.js
qooban: Not exactly. Previously single string was used to be passed to `verifyTemplateRow`, but from the API we always get list of strings, so I changed `verifyTemplateRow` to work with the API structure and then just updated other uses of `verifyTemplateRow`.

test/cypress/integration/templates/template_management_spec.js
qooban: Hmmm... Could be, but I'm not convinced it's better (see my comment: https://github.com/cloudify-cosmo/cloudify-stage/pull/1333#discussion_r621972646 ).

kubama (approved):

qooban:
As @Gelio don't mind leaving code as-is, I'm merging this small PR with one approval.

test/cypress/integration/templates/template_management_spec.js
Gelio: I see, sure :+1: I don't mind. I thought that we were consistently using `cy.contains`, hence the comments

PR #1335, opened 2021-04-29
RD-1526 Add more information in marker tooltip

Gelio:
This PR adds displaying more deployment information in the map's tooltips, according to the mockups

I have extracted the components responsible for rendering the deployment statuses to reuse the code.

## Screenshot

image

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/525/pipeline

Ran 2021-04-29 12:28 CEST

### DIFF
--- a/.size-limit
+++ b/.size-limit
@@ -20,7 +20,7 @@
     },
     {
         "name": "Application package",
-        "limit": "13.3 MB",
+        "limit": "13.35 MB",
         "gzip": false,
         "running": false,
         "path": ["dist", "!dist/**/*.gz"]
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -566,8 +566,13 @@ describe('Deployments View widget', () => {
         });
 
         const getDeploymentsMapTooltip = () => cy.get('.leaflet-tooltip');
-        const getMarkerByImageSrcSuffix = (srcSuffix: string) =>
-            cy.get(`.leaflet-marker-pane img[src$="${srcSuffix}"]`);
+        enum MarkerImageSuffix {
+            Red = 'red.png',
+            Yellow = 'yellow.png',
+            Blue = 'blue.png'
+        }
+        const getMarkerByImageSuffix = (markerImageSuffix: MarkerImageSuffix) =>
+            cy.get(`.leaflet-marker-pane img[src$="${markerImageSuffix}"]`);
         const withinMarkerTooltip = (
             getMarker: () => Cypress.Chainable,
             callback: (currentSubject: JQuery<HTMLElement>) => void,
@@ -591,7 +596,7 @@ describe('Deployments View widget', () => {
             getDeploymentsViewMap().within(() => {
                 cy.get('.leaflet-marker-icon').should('have.length', 1);
                 withinMarkerTooltip(
-                    () => getMarkerByImageSrcSuffix('red.png'),
+                    () => getMarkerByImageSuffix(MarkerImageSuffix.Red),
                     () => cy.contains(getSiteDeploymentName(siteNames.london))
                 );
             });
@@ -619,7 +624,7 @@ describe('Deployments View widget', () => {
                 const getTooltipSubenvironments = () => cy.get('i.icon.object.group').parent();
                 const getTooltipSubservices = () => cy.get('i.icon.cube').parent();
                 withinMarkerTooltip(
-                    () => getMarkerByImageSrcSuffix('yellow.png'),
+                    () => getMarkerByImageSuffix(MarkerImageSuffix.Yellow),
                     () => {
                         cy.contains('hello-world-one');
                         cy.contains('Cloudify-Hello-World');
@@ -634,7 +639,7 @@ describe('Deployments View widget', () => {
                 );
 
                 withinMarkerTooltip(
-                    () => getMarkerByImageSrcSuffix('red.png'),
+                    () => getMarkerByImageSuffix(MarkerImageSuffix.Red),
                     () => {
                         cy.contains('deployments_view_test_deployment');
                         cy.contains(blueprintName);
@@ -652,7 +657,7 @@ describe('Deployments View widget', () => {
                 );
 
                 withinMarkerTooltip(
-                    () => getMarkerByImageSrcSuffix('blue.png'),
+                    () => getMarkerByImageSuffix(MarkerImageSuffix.Blue),
                     () => {
                         cy.contains('one-in-warsaw');
                         cy.contains('Cloudify-Hello-World');
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -677,7 +677,7 @@ describe('Deployments View widget', () => {
                             withinMarkerTooltip(
                                 () => cy.wrap(marker),
                                 element => {
-                                    currentDeploymentName = element.text();
+                                    currentDeploymentName = element.find('h4').text();
                                 }
                             );
                             cy.wrap(marker).click();
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -616,19 +616,51 @@ describe('Deployments View widget', () => {
             getDeploymentsViewMap().within(() => {
                 cy.get('.leaflet-marker-icon').should('be.visible').and('have.length', 3);
 
+                const getTooltipSubenvironments = () => cy.get('i.icon.object.group').parent();
+                const getTooltipSubservices = () => cy.get('i.icon.cube').parent();
                 withinMarkerTooltip(
                     () => getMarkerByImageSrcSuffix('yellow.png'),
-                    () => cy.contains('hello-world-one')
+                    () => {
+                        cy.contains('hello-world-one');
+                        cy.contains('Cloudify-Hello-World');
+                        cy.contains(siteNames.london);
+                        cy.contains('In progress').parent().find('i.orange.spinner');
+                        getTooltipSubenvironments().contains('5');
+                        getTooltipSubservices().within(() => {
+                            cy.contains('80');
+                            cy.get('[aria-label="In progress"]');
+                        });
+                    }
                 );
 
                 withinMarkerTooltip(
                     () => getMarkerByImageSrcSuffix('red.png'),
-                    () => cy.contains('deployments_view_test_deployment')
+                    () => {
+                        cy.contains('deployments_view_test_deployment');
+                        cy.contains(blueprintName);
+                        cy.contains(siteNames.olsztyn);
+                        cy.contains('Requires attention').parent().find('i.red.exclamation');
+                        getTooltipSubenvironments().within(() => {
+                            cy.contains('1');
+                            cy.get('[aria-label="In progress"]');
+                        });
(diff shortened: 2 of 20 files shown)


widgets/common/src/deploymentsView/StatusIcon.tsx
```diff
@@ -0,0 +1,54 @@
+import type { FunctionComponent } from 'react';
+import type { IconProps } from 'semantic-ui-react';
+
+import { i18nPrefix } from './common';
+import { DeploymentStatus, SubdeploymentStatus } from './types';
+
+/**
```
Gelio, line 7: Most of the code in this file is extracted from `table/columns.tsx`

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.jsx
```diff
@@ -65,7 +65,11 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
                     setGraphData(tasksGraph);
                     clearError();
                 })
-                .catch(({ message, status }) => {
+                .catch(({ message, status, isCanceled }) => {
+                    if (isCanceled) {
```
Gelio, line 69: A bugfix for the "cannot update the state of an unmounted component" that has been showing up when quickly switching deployments in the table

Gelio:
@qooban @kubama I would appreciate if we delivered this PR until the end of the week, as I'm on holiday next week, and this is my final commitment before the holiday :slightly_smiling_face: If you could review today, or early tomorrow, so we can get it done by the end of the day today, that would be great

Also, if anything major can be postponed to other tasks, I would prefer it that way to make sure this lands in time :slightly_smiling_face:

.size-limit
```diff
@@ -20,7 +20,7 @@
     },
     {
         "name": "Application package",
-        "limit": "13.3 MB",
+        "limit": "13.35 MB",
```
Gelio, line 23: Before this change:

```
[2021-04-29T10:46:33.605Z]   Application package
[2021-04-29T10:46:33.605Z]   Package size limit has exceeded by 2.14 KB
[2021-04-29T10:46:33.605Z]   Size limit:   13946060 B
[2021-04-29T10:46:33.605Z]   Size:         13948251 B
```

widgets/common/src/deploymentsView/map/marker/tooltip.tsx
```diff
@@ -0,0 +1,70 @@
...
+const markerIconHeight = 41;
+const tooltipOffset = L.point(0, -markerIconHeight);
+
+const TooltipContentContainer = styled.div`
+    display: flex;
+`;
+
+const StatusesContainer = styled.div`
+    margin-left: 1em;
+    // NOTE: 7px = leaflet's tooltip padding&border
```
qooban, line 20: Optional:

Maybe instead of a comment we can have a const for that?

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.jsx
qooban: Good catch! 👍🏻

widgets/common/src/deploymentsView/StatusIcon.tsx
```diff
@@ -0,0 +1,54 @@
...
+);
+
+const subdeploymentStatusToIconMapping: Record<SubdeploymentStatus, StatusIconName | undefined> = {
+    [SubdeploymentStatus.InProgress]: 'inProgress',
+    [SubdeploymentStatus.Good]: undefined,
+    [SubdeploymentStatus.Failed]: 'requiresAttention',
+    [SubdeploymentStatus.Pending]: undefined
+};
+export const SubdeploymentStatusIcon: FunctionComponent<{ status: SubdeploymentStatus | null }> = ({ status }) => (
+    // NOTE: handle possible `null`s
```
qooban, line 52: Optional:

Not sure the comment is necessary here. If you'd like to keep the comment, I would consider describing why we can have null here and not that `status ?? SubdeploymentStatus.Good` is for handling possible `null`'s.

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -616,19 +616,51 @@ describe('Deployments View widget', () => {
             getDeploymentsViewMap().within(() => {
                 cy.get('.leaflet-marker-icon').should('be.visible').and('have.length', 3);
 
+                const getTooltipSubenvironments = () => cy.get('i.icon.object.group').parent();
+                const getTooltipSubservices = () => cy.get('i.icon.cube').parent();
                 withinMarkerTooltip(
                     () => getMarkerByImageSrcSuffix('yellow.png'),
```
qooban, line 622: Optional:

Maybe we can extract all used image sources suffixes to consts or add proper type for `getMarkerByImageSrcSuffix` argument?

qooban (approved):
Looks good 👍🏻 I added only few optional comments.

kubama (dismissed):
Looks good! 👍

qooban:
Did quick UI tests as well - nothing to worry right now, but two points to consider for the future:

1. Tooltip is always presented on the top of the marker. If the marker is close to the top border of the map, then the tooltip won't be visible.
2. The list of things described in the popup may not be clear when the user is first seeing it. Seeing the following words "hello-world, Cracow, Good" in a column they may not associate easily each word with it's type. Maybe we can just add an icon on the left side of each word indicating the type of the information? 

Ref.: 
https://user-images.githubusercontent.com/5202105/116547231-ea37a780-a8f2-11eb-9155-a103772cbfcf.mp4

Gelio:
The system tests passed (aside from the known `filters_spec`), so I won't be rerunning them after addressing the PR comments, as I consider them to be passing :slightly_smiling_face: I'll only rerun the deployments view spec locally

widgets/common/src/deploymentsView/StatusIcon.tsx
Gelio: Makes sense :+1: I was not sure whether to leave the comment or not, but I suppose this place is not the right one. I'll remove it here and add a comment in https://github.com/cloudify-cosmo/cloudify-stage/blob/dd058b402faa8a7bd4372dbbcc5d8067dc639ba9/widgets/common/src/deploymentsView/types.ts#L39-L41 about why the values can be `null` (when there are no subdeployments of that kind)

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Yeah, makes sense :+1: I'll add an enum for that

Gelio:
@qooban 

> 1. Tooltip is always presented on the top of the marker. If the marker is close to the top border of the map, then the tooltip won't be visible. Maybe it would be possible to adapt tooltip to be presented on the different side of the marker or to take place outside of the map canvas?

I'm glad you commented on that so I can explain the options.

By default, without `direction` or `offset` set in https://github.com/cloudify-cosmo/cloudify-stage/blob/dd058b402faa8a7bd4372dbbcc5d8067dc639ba9/widgets/common/src/deploymentsView/map/marker/tooltip.tsx#L44 the tooltips does handle the distance to the map edge, so it tries to be smart and try to be visible:

https://user-images.githubusercontent.com/889383/116550345-f1f94b00-a8f6-11eb-8e59-d6b1aa83096a.mp4

Unfortunately, the `auto` option makes it so that the tooltip appears on the left/right, not top/bottom :worried: https://leafletjs.com/reference-1.7.1.html#tooltip To me, that looks worse than displaying the tooltip on top, so I hardcoded the `direction` (there is no `auto` for vertical directions :worried:). 

I did not find an easy way to get the tooltip to appear on top or bottom depending on the available space. If we were to add it, it would have to be customly written, probably by listening to the `tooltipopen` event and then somehow changing the direction. I did not find suitable, easy to use methods for that in the API, so worst-case scenario is that we would have to drop `react-leaflet` here and use `leaflet` manually - not great :smile: 

I have tried the other solution you suggested (making the tooltip overflow the map and appear on top of other elements), but that also seems to be not doable due to the map container creating a new [_stacking context_](https://www.joshwcomeau.com/css/stacking-contexts/), and the tooltip being an ancestor of the map container. In short, the contained tooltip cannot break out of the map container 
image
For that to work, we would have to stop using leaflet's tooltips and write a custom solution that would bind to the surrounding DOM elements so it can appear on top :smile:

Both ways seem to be inferior, and I did not come up with another way to do it :slightly_frowning_face: Unless there is pressure from the product team or the customers to change it, I would leaving it be

> 2\. The list of things described in the popup may not be clear when the user is first seeing it. Seeing the following words "hello-world, Cracow, Good" in a column they may not associate easily each word with it's type. Maybe we can just add an icon on the left side of each word indicating the type of the information?

Good point! I will bring it up with @ofercloudify

widgets/common/src/deploymentsView/map/marker/tooltip.tsx
Gelio: Nice one :smile: I was hesitant to create a regular JS variable and use it in the tagged template literal, but as far as I have checked the final effect is the same:

image

kubama (approved):

Gelio:
I've addressed all the comments and the tests passed locally:

image

I can wait till tomorrow with the merge, so I'll wait for @qooban to take a look if the comment resolution is satisfactory :smile:

qooban (approved):
All looks good 👍🏻 Thanks for addressing the comments.

widgets/common/src/deploymentsView/StatusIcon.tsx
qooban: Thanks. Looks good 👍🏻

qooban:
Ad. 1

> Both ways seem to be inferior, and I did not come up with another way to do it slightly_frowning_face Unless there is pressure from the product team or the customers to change it, I would leaving it be

Make sense. Thanks for checking the different options.
I checked how it works in Sites Map widget and I found that we do auto-panning on marker click:

https://user-images.githubusercontent.com/5202105/116653531-b0fd4700-a987-11eb-9a24-f62ffc75aeb0.mp4

Probably for hovering it would not be so straightforward and UX would probably be worse than now, but maybe something to consider in the future.

Anyways, as you said "unless there is pressure from the product team or the customers to change it" we should leave it as it is.

Ad. 2

> Good point! I will bring it up with @ofercloudify

Thanks 👍🏻

Gelio:
@qooban 

> Probably for hovering it would not be so straightforward and UX would probably be worse than now, but maybe something to consider in the future.

Yeah, auto-panning is supported out-of-the-box for popups, since the mouse position does not matter after opening & auto-panning the popup. As you said, the UX would probably be worse (or actually I don't see how we could do it well) because the tooltip is open as long as the mouse is hovering the marker. If we auto-pan the map once the tooltip is opened, the marker position would change and the mouse could not longer be hovering the marker, meaning the tooltip would close :smile: 

I'm pretty sure you had this behavior in mind, but let me know if you had some other idea, as I'm curious whether there's some other solution :slightly_smiling_face: 

---

Thanks for reviews, I'll merge the PR after the CI passes

qooban:
@Gelio 

> I'm pretty sure you had this behavior in mind, but let me know if you had some other idea, as I'm curious whether there's some other solution slightly_smiling_face

Yes, this is exactly the behavior I had in mind. And unfortunately I don't have another solution right now.

Gelio:
Sure, thanks for answering :+1:

PR #1341, opened 2021-04-30
RD-2166 Add filters swagger endpoint test

qooban:
This PR adds small test checking that filter rule attributes, operators and types used in the UI are supported by the API.

I'm not planning to run full system test as this is just addition of one test.

All filters tests are fine:
image


kubama:
I noticed you added 'should handle all filter parameters'. However there is no such test case in the attached screen shot, there is 'should use only filter parameters supported by the API" though. So, is there a missing commit in the PR or it's just the screen shot not in line with the actual implementation?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -720,4 +723,44 @@ describe('Filters widget', () => {
...
+            };
+
+            const supportedByUi: SupportedByObject = {
+                [RuleParameters.RuleAttributes]: getEnumValues(FilterRuleAttributes),
+                [RuleParameters.RuleAttributesOperators]: getEnumValues(AttributesFilterRuleOperators),
+                [RuleParameters.RuleLabelsOperators]: getEnumValues(LabelsFilterRuleOperators),
+                [RuleParameters.RuleTypes]: getEnumValues(FilterRuleType)
+            };
+
+            getEnumValues(RuleParameters).forEach(parameter => {
```
Gelio, line 759: We could simplify it :smile: 

```suggestion
            Object.values(RuleParameters).forEach(parameter => {
```

and then we don't need `as RuleParameters` below :slightly_smiling_face:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -720,4 +723,44 @@ describe('Filters widget', () => {
...
+        }
+        type SupportedByObject = Record<RuleParameters, string[]>;
+
+        function getEnumValues(enumType: any) {
+            return Object.keys(enumType).map(enumValue => enumType[enumValue]);
+        }
+
+        cy.cfyRequest('/searches/deployments.help.json').then(response => {
+            const {
+                operations: [operations]
```
Gelio, line 742: Digression:

Wow, the additional wrapping operations in a singleton array is really offputting :smile: But I've checked, it's correct :worried: 

@YoniItzhak Is it intended to wrap the allowed operations in an array with a single element? Maybe we could unwrap the object :smile: ?

image

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -720,4 +723,44 @@ describe('Filters widget', () => {
...
+            RuleAttributes = 'attributes',
+            RuleAttributesOperators = 'attributesOperators',
+            RuleLabelsOperators = 'labelsOperators',
+            RuleTypes = 'types'
+        }
+        type SupportedByObject = Record<RuleParameters, string[]>;
+
+        function getEnumValues(enumType: any) {
+            return Object.keys(enumType).map(enumValue => enumType[enumValue]);
+        }
```
Gelio, line 738: This is essentially the implementation of `Object.values` :slightly_smiling_face:

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -720,4 +723,44 @@ describe('Filters widget', () => {
...
+                [RuleParameters.RuleAttributes]: getEnumValues(FilterRuleAttributes),
+                [RuleParameters.RuleAttributesOperators]: getEnumValues(AttributesFilterRuleOperators),
+                [RuleParameters.RuleLabelsOperators]: getEnumValues(LabelsFilterRuleOperators),
+                [RuleParameters.RuleTypes]: getEnumValues(FilterRuleType)
+            };
+
+            getEnumValues(RuleParameters).forEach(parameter => {
+                expect(supportedByApi[parameter as RuleParameters]).to.include.all.members(
+                    supportedByUi[parameter as RuleParameters]
+                );
```
Gelio, line 762: I'm not sure that this is enough. Technically we could have an enum with some defined values but not use it in the form. IMO it would be safer to verify that the options are visible in the UI :slightly_smiling_face: This can be done in a separate ticket OFC

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -720,4 +723,44 @@ describe('Filters widget', () => {
...
+                [RuleParameters.RuleLabelsOperators]: operations.filter_rules_labels_operators,
+                [RuleParameters.RuleTypes]: operations.filter_rules_types
+            };
+
+            const supportedByUi: SupportedByObject = {
+                [RuleParameters.RuleAttributes]: getEnumValues(FilterRuleAttributes),
+                [RuleParameters.RuleAttributesOperators]: getEnumValues(AttributesFilterRuleOperators),
+                [RuleParameters.RuleLabelsOperators]: getEnumValues(LabelsFilterRuleOperators),
+                [RuleParameters.RuleTypes]: getEnumValues(FilterRuleType)
+            };
```
Gelio, line 757: Instead of defining 2 distinct objects and zipping them in the `forEach` below, how about the following implementation:

```ts
            function verifyValuesAreSupported(
                valuesSupportedByUi: Record<string, any>,
                valuesAvailableInAPI: string[]
            ) {
                expect(Object.values(valuesSupportedByUi)).to.include.all.members(valuesAvailableInAPI);
            }

            verifyValuesAreSupported(FilterRuleAttributes, operations.allowed_filter_rules_attrs);
            verifyValuesAreSupported(AttributesFilterRuleOperators, operations.filter_rules_attributes_operators);
            verifyValuesAreSupported(LabelsFilterRuleOperators, operations.filter_rules_labels_operators);
            verifyValuesAreSupported(FilterRuleType, operations.filter_rules_types);
```

This is shorter and easier to understand to me :slightly_smiling_face:

widgets/common/src/filters/types.ts
```diff
@@ -1,11 +1,17 @@
 /** @see https://docs.cloudify.co/api/v3.1/#the-filter-resource */
 export interface FilterRule {
-    key: string;
+    key: FilterRuleAttributes | string;
```
Gelio, line 3: Just FYI: this is just for documentation purposes, as having the union here is a no-op - the type will be `string` anyway :slightly_smiling_face: But I like it and I'm all for it :+1:

Gelio (commented):
The code looks good, but could be simplified :slightly_smiling_face:

test/cypress/integration/widgets/filters_spec.ts
qooban: Good idea. Changed.

test/cypress/integration/widgets/filters_spec.ts
qooban: It was done intentionally at the beginning as there are differences.
Anyways, I changed the implementation and added `without` calls to exclude values not supported by UI (no plans to support them AFAIK after discussing with Ofer).

test/cypress/integration/widgets/filters_spec.ts
qooban: Right, updated.

qooban:
@kubama 
> I noticed you added 'should handle all filter parameters'. However there is no such test case in the attached screen shot, there is 'should use only filter parameters supported by the API" though. So, is there a missing commit in the PR or it's just the screen shot not in line with the actual implementation?

Yeah, missing commit.

widgets/common/src/filters/types.ts
```diff
@@ -1,11 +1,17 @@
...
     values: string[];
     operator: FilterRuleOperator;
     type: FilterRuleType;
 }
 
+export enum FilterRuleAttributes {
```
kubama, line 9: I think a common convention would be to name this enum as a singular term, in this case `FilterRuleAttribute`

widgets/common/src/filters/types.ts
kubama: I think it would be even more descriptive if we created `LabelKey` type (which would be equal to a string) and used it here instead of `string`

qooban: @Gelio Yeah, I had that in mind.

@kubama Could be, added that type.

widgets/common/src/filters/types.ts
qooban: Right, my bad, changed.

widgets/common/src/filters/types.ts
```diff
@@ -1,16 +1,17 @@
 /** @see https://docs.cloudify.co/api/v3.1/#the-filter-resource */
 export interface FilterRule {
-    key: FilterRuleAttributes | string;
+    key: FilterRuleAttribute | LabelKey;
```
kubama, line 3: Optional: Looking at it further, maybe it's a good point to revisit my suggestion from https://github.com/cloudify-cosmo/cloudify-stage/pull/1296#discussion_r609795199?

kubama (approved):

test/cypress/integration/widgets/filters_spec.ts
Gelio: Hmm... I think there's a misunderstanding :slightly_smiling_face: My point was that just checking if frontend enums contain those values from the API may not be enough, since those enum values may not be used in the UI in the end :slightly_smiling_face: I suggested verifying the form displayed in the UI instead of checking what values are there in the enums, as IMO that would give higher confidence :slightly_smiling_face:

Gelio (approved):

test/cypress/integration/widgets/filters_spec.ts
qooban: Got, it, but do you think it is possible as we are populating dropdowns with `LabelsFilterRuleOperators` or `AttributesFilterRuleOperators` which are constructed from the enums?

```
export const LabelsFilterRuleOperators = { ...CommonRuleOperator, ...LabelsOnlyRuleOperator };
export const AttributesFilterRuleOperators = { ...CommonRuleOperator, ...AttributesOnlyRuleOperator };
```

widgets/common/src/filters/types.ts
qooban: I don't think it is straight-forward and I manage to do that before Tuesday. 
I started it, you can take a look at https://github.com/cloudify-cosmo/cloudify-stage/pull/1389

test/cypress/integration/widgets/filters_spec.ts
Gelio: Actually, I can see how that could be quite hard :worried: Especially that there is a mapping between the values sent to the API and the labels that UI displays (e.g. `ends_with` -> some text in the UI).

I don't have a good idea how to solve it, so I believe we have to settle with the existing solution :smile:

PR #1344, opened 2021-05-05
RD-2080 - Enhance support for Bulk Action - Run Workflow

qooban:
### Changes

* Start using new create deployment group API by passing filter_rules instead of deployment_ids
* Add search option to worklows selection dropdown
* Extracted common functionality for Deploy On and Run Workflow: `GoToExecutionsPageButton` component and `getBatchActionGroupId` function
* Add basic error handling in Run Worfklow modal
* Addressed some of the previous code review comments
* Added tests


test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -724,4 +733,49 @@ describe('Deployments View widget', () => {
...
+                        operator: FilterRuleOperators.AnyOf,
+                        values: [exampleSiteName]
+                    }
+                ]);
+        });
+
+        it('should allow to run workflow on filtered deployments', () => {
+            cy.interceptSp('POST', /^\/searches\/workflows/).as('searchWorkflows');
+            cy.interceptSp('PUT', /^\/deployment-groups\/BATCH_ACTION_.*/).as('createDeploymentGroup');
+            cy.interceptSp('POST', /^\/execution-groups/).as('startExecutionGroup');
```
kubama, line 758: Is there any reason for using regexps instead of plain strings?

widgets/common/src/SearchActions.ts
```diff
@@ -0,0 +1,31 @@
+import { FilterRule } from './filters/types';
+
+type ResourceName = 'blueprints' | 'deployments' | 'workflows';
+type Params = Record<string, any> | null;
+
+export default class SearchActions {
+    constructor(private toolbox: Stage.Types.Toolbox) {}
+
+    doList(resourceName: ResourceName, filterRules: FilterRule[], params: Params = null) {
```
kubama, line 9: Can it be marked as private?

widgets/common/src/SearchActions.ts
```diff
@@ -0,0 +1,31 @@
+import { FilterRule } from './filters/types';
+
+type ResourceName = 'blueprints' | 'deployments' | 'workflows';
+type Params = Record<string, any> | null;
```
kubama, line 4: I think if we used `undefined` here instead of `null` we wouldn't need default values for `params`.

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -73,13 +75,10 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
         return [...filterRules, ...additionalFilterRules, filteringByParentDeploymentResult.parentDeploymentRule];
     }, [filterRules, filteringByParentDeploymentResult.parentDeploymentRule, additionalFilterRules]);
 
-    const deploymentsUrl = '/searches/deployments';
     const deploymentsResult = useQuery(
-        [deploymentsUrl, gridParams, finalFilterRules],
+        ['deployments', gridParams, finalFilterRules],
```
kubama, line 79: Since queries are application wide I think it's worth to stick to a common convention when specifying query keys. I think query URL is the best candidate for key component if we want to maintain consistency.

widgets/common/src/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -46,59 +49,57 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterId,
...
-            return executionGroupsActions.doStart(workflowId!, groupId).then(_executionGroup => {
-                toolbox.getEventBus().trigger('deployments:refresh');
-                toolbox.getEventBus().trigger('executions:refresh');
-                resetLoadingMessage();
-                setExecutionGroupStarted();
-            });
-        });
-    }
+            toolbox.getEventBus().trigger('deployments:refresh');
+            toolbox.getEventBus().trigger('executions:refresh');
```
kubama, line 90: After my recent change it can be shortened to:
```suggestion
            toolbox.getEventBus().trigger('deployments:refresh').trigger('executions:refresh');
```

widgets/common/src/deploymentsView/header/common.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
+export function getGroupIdForBatchAction() {
```
kubama, line 2: Can we apply it in `DeployOnModal` at once? I'm also wondering if group name generation shouldn't be handled internally in `DeploymentGroupsActions`.

widgets/common/src/ExecutionGroupsActions.ts
```diff
@@ -2,7 +2,7 @@ export default class ExecutionGroupsActions {
     constructor(private toolbox: Stage.Types.Toolbox) {}
 
     doStart(workflowId: string, groupId: string, defaultParameters?: Record<string, any>) {
```
kubama, line 4: I think it would be slightly more intuitive if `groupId` was the first parameter and `workflowId` the second. We are acting on a group after all, and workflow ID is more an action parameter than an identifier in this case.

widgets/common/src/SearchActions.ts
qooban: Yes, changed.

widgets/common/src/SearchActions.ts
qooban: Hmm.. For me it didn't work. I had to specify default value in both cases (`undefined` / `null`).

widgets/common/src/deploymentsView/DeploymentsView.tsx
qooban: Make sense, reverted.

widgets/common/src/deploymentsView/header/RunWorkflowModal.tsx
qooban: Sure, done.

widgets/common/src/deploymentsView/header/common.ts
qooban: Updated `DeployOnModal`. For now (due to time pressure) I'd keep it external to `DeploymentGroupsActions`.

widgets/common/src/ExecutionGroupsActions.ts
qooban: Make sense. Changed.

widgets/common/src/SearchActions.ts
qooban: OK, I made it differently (`params` argument is now optional).

qooban:
System tests are now broken due to... Environmental issue again, so I started only local test for Deployments View widget:

image

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -753,9 +753,9 @@ describe('Deployments View widget', () => {
         });
 
         it('should allow to run workflow on filtered deployments', () => {
-            cy.interceptSp('POST', /^\/searches\/workflows/).as('searchWorkflows');
-            cy.interceptSp('PUT', /^\/deployment-groups\/BATCH_ACTION_.*/).as('createDeploymentGroup');
-            cy.interceptSp('POST', /^\/execution-groups/).as('startExecutionGroup');
+            cy.interceptSp('POST', 'searches/workflows').as('searchWorkflows');
+            cy.interceptSp('PUT', 'deployment-groups/BATCH_ACTION_').as('createDeploymentGroup');
+            cy.interceptSp('POST', 'execution-groups').as('startExecutionGroup');
```
kubama: I think it would be just a little bit more accurate if starting `/` was still included.

widgets/common/src/DeploymentGroupsActions.ts
```diff
@@ -1,20 +1,36 @@
...
     constructor(private toolbox: Stage.Types.Toolbox) {}
 
-    doCreate(id: string, data: CreateDeploymentGroupData) {
+    doCreateGroup(id: string, data: CreateDeploymentGroupData) {
+        return this.toolbox.getManager().doPut(`/deployment-groups/${id}`, null, data);
+    }
+
+    doCreateNewDeployments(id: string, data: CreateNewDeploymentsData) {
         return this.toolbox.getManager().doPut(`/deployment-groups/${id}`, null, data);
     }
```
kubama: We have two methods with same implementation but different arguments. So why not:

```suggestion
    doCreateGroup(id: string, data: CreateDeploymentGroupData | CreateNewDeploymentsData) {
        return this.toolbox.getManager().doPut(`/deployment-groups/${id}`, null, data);
    }
```

qooban: Could be, changed.

kubama (approved):

PR #1343, opened 2021-05-05
RD-2078 Added underlying logic to DeployOnModal

kubama:

PR #1349, opened 2021-05-07
RD-2078 Code review follow-ups

kubama:
Addresses code review comments from #1340 and #1343

Since there are some refactorings impacting broader context I scheduled system tests build: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/558/

### DIFF
--- a/widgets/common/src/UpdateDeploymentModal.jsx
+++ b/widgets/common/src/UpdateDeploymentModal.jsx
@@ -106,9 +106,6 @@ function UpdateDeploymentModal({ open, deploymentId, onHide, toolbox }) {
             validationErrors.blueprintName = 'Please select blueprint';
         }
 
-        const inputsWithoutValue = {};
-        InputsUtils.addErrors(inputsWithoutValue, validationErrors);
-
         if (!_.isEmpty(validationErrors)) {
             setErrors(validationErrors);
             unsetLoading();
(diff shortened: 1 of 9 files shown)


gregcfy (approved):

widgets/common/src/UpdateDeploymentModal.jsx
```diff
@@ -121,7 +118,7 @@ function UpdateDeploymentModal({ open, deploymentId, onHide, toolbox }) {
             .doUpdate(
                 deployment.id,
                 blueprint.id,
-                InputsUtils.getInputsToSend(inputsPlanForUpdate, deploymentInputs, inputsWithoutValue),
+                InputsUtils.getInputsMap(inputsPlanForUpdate, deploymentInputs),
```
qooban, line 121: I don't see `inputsWithoutVaue` after removing `InputsUtils.addErrors`, so it is not necessary here?

qooban (commented):
Just one comment-question from my side.

widgets/common/src/UpdateDeploymentModal.jsx
kubama: It's not necessary - in original implementation after calling `getInputsToSend` the value of `inputsWithoutValue` was not used.

qooban: Got it, approving then.

qooban (approved):

PR #1353, opened 2021-05-10
RD-2148 Add detailed E2E tests for Getting Started modal

gregcfy:
The main goal of this PR is to add tests for Getting Started Modal.

Ticket: https://cloudifysource.atlassian.net/browse/RD-2148

Build (all system tests passed with master): needed link here for last build result

image


test/cypress/support/commands.ts
```diff
@@ -146,8 +147,15 @@ const commands = {
             ...options
         });
     },
-    login: (username = 'admin', password = 'admin', expectSuccessfulLogin = true) => {
-        cy.disableGettingStarted();
+    login: (
+        username = 'admin',
+        password = 'admin',
+        expectSuccessfulLogin = true,
+        mockDisabledGettingStarted = true
```
kubama, line 154: Since all the parameters are optional I think it's a right moment to introduce a parameter object

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,445 @@
...
+import { mockAwsSecretsManager, mockEmptySecretsManager, mockSecretCreation, mockSecretUpdate } from './secretMocks';
+
+const gotoBackStep = () => cy.get('.modal button').contains('Back').click();
+const gotoNextStep = () => cy.get('.modal button').contains('Next').click();
+const gotoFinishStep = () => cy.get('.modal button').contains('Finish').click();
+const closeModal = () => cy.get('button:not([disabled])').contains('Close').click();
+
+describe('Getting started modal', () => {
+    before(() => {
+        cy.usePageMock().activate().login(undefined, undefined, true, false);
```
kubama, line 35: Looks awkward. See my other comment about introducing parameter object.
On the other hand, do we need to disable the disable of getting started modal here if in each test we are enabling it and calling `reload`?

test/cypress/support/commands.ts
```diff
@@ -325,7 +339,13 @@ const commands = {
...
+    mockEnabledGettingStarted: () => {
+        cy.interceptSp('GET', `/users/`, {
+            body: { show_getting_started: true }
+        });
+    },
+
+    mockDisabledGettingStarted: () => {
         cy.interceptSp('GET', `/users/`, {
             body: { show_getting_started: false }
         });
```
kubama, line 351: Minor code duplication

test/cypress/support/commands.ts
```diff
@@ -183,7 +191,9 @@ const commands = {
                     username
                 })
             );
-            cy.disableGettingStarted();
+            if (mockDisabledGettingStarted) {
+                cy.mockDisabledGettingStarted();
+            }
```
kubama, line 196: This condition is a common pattern, how about passing the flag to the `mockDisabledGettingStarted` function and handling the condition internally?

test/cypress/support/commands.ts
```diff
@@ -171,7 +179,7 @@ const commands = {
             cy.waitUntilLoaded().then(() => cy.saveLocalStorage());
         }
     },
-    mockLogin: (username = 'admin', password = 'admin') => {
+    mockLogin: (username = 'admin', password = 'admin', mockDisabledGettingStarted = true) => {
```
kubama, line 182: I'm not sure about `mockDisabledGettingStarted` variable name, the 'mock' term is a bit confusing when thinking about what this flag actually does. Why not just `disableGettingStarted`?

qooban:
@gregcfy Due to time pressure and since @kubama already reviewed this PR, I'm not going to review this one. I hope you'll be OK with it.

test/cypress/integration/gettingStarted/blueprintMocks.ts
```diff
@@ -0,0 +1,125 @@
+export const mockAwsBlueprintsManager = () =>
+    cy.interceptSp(
+        'GET',
+        '/blueprints?_include=id%2Cdescription%2Cmain_file_name%2Ctenant_name%2Ccreated_at%2Cupdated_at%2Ccreated_by%2Cprivate_resource%2Cvisibility',
+        req => {
+            req.reply({
+                metadata: { pagination: { total: 3, size: 50, offset: 0 }, filtered: null },
```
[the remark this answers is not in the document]
gregcfy: fixed

test/cypress/integration/gettingStarted/pluginMocks.ts
```diff
@@ -0,0 +1,279 @@
...
+        }
+    ],
+    icon: 'https://cloudify.co/wp-content/uploads/2020/07/ansible-icon.png',
+    name: 'cloudify-ansible-plugin'
+});
+
+export const mockEmptyPluginsCatalog = () =>
+    cy.intercept(
+        'GET',
+        '/console/external/content?url=http%3A%2F%2Frepository.cloudifysource.org%2Fcloudify%2Fwagons%2Fplugins.json',
```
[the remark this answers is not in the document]
gregcfy: done

test/cypress/integration/gettingStarted/blueprintMocks.ts
```diff
@@ -0,0 +1,125 @@
...
+    );
+
+export const mockKubernetesGcpGkeBlueprintUpload = () =>
+    cy.interceptSp(
+        'PUT',
+        '/blueprints/Kubernetes-GCP-GKE?visibility=tenant&async_upload=true&application_file_name=blueprint.yaml&blueprint_archive_url=https%3A%2F%2Fgithub.com%2Fcloudify-community%2Fblueprint-examples%2Freleases%2Fdownload%2Flatest%2Fkubernetes-gcp-gke.zip',
+        req => {
+            req.reply({ id: '', state: 'pending' });
+        }
+    );
```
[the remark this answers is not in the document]
gregcfy: fixed

test/cypress/support/commands.ts
gregcfy: fixed

test/cypress/support/commands.ts
gregcfy: fixed

test/cypress/support/commands.ts
gregcfy: fixed

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,445 @@
...
+    mockCloudifyGcpPluginUpload,
+    mockCloudifyKubernetesPluginUpload,
+    mockCloudifyTerraformPluginUpload,
+    mockCloudifyUtilitiesPluginUpload,
+    mockEmptyPluginsCatalog,
+    mockEmptyPluginsManager
+} from './pluginMocks';
+import { mockAwsSecretsManager, mockEmptySecretsManager, mockSecretCreation, mockSecretUpdate } from './secretMocks';
+
+const gotoBackStep = () => cy.get('.modal button').contains('Back').click();
```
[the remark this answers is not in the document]
gregcfy: done

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: common parts moved into before method

test/cypress/integration/gettingStarted/pluginMocks.ts
```diff
@@ -0,0 +1,279 @@
+const createCloudifyAwsPluginItem = () => ({
```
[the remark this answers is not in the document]
gregcfy: done

test/cypress/support/commands.ts
gregcfy: Ticket created: https://cloudifysource.atlassian.net/browse/RD-2314

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: disable / enable modal rest api mocked

kubama (commented):
I did only partial review - I would say the variable/function/fixture naming could be improved. I'd like it to be addressed before I continue with the review.

test/cypress/integration/gettingStarted/blueprintMocks.ts
```diff
@@ -0,0 +1,84 @@
+import awsBlueprintsManager from '../../fixtures/getting_started/aws_blueprints_manager.json';
+
+export const mockBlueprintsManager = (items: any[]) =>
+    cy.interceptSp(
+        'GET',
+        /^\/blueprints\?.*\b_include=(\bid\b|\bdescription\b|\bmain_file_name\b|\btenant_name\b|\bcreated_at\b|\bupdated_at\b|\bcreated_by\b|\bprivate_resource\b|\bvisibility\b|,)+/,
```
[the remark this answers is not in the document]
gregcfy: parameters order can be different

test/cypress/integration/gettingStarted/blueprintMocks.ts
```diff
@@ -0,0 +1,84 @@
...
+    applicationFileName: string,
+    blueprintArchiveUrl: string
+) => {
+    const pathname = `/blueprints/${blueprintName}`;
+    const query =
+        `visibility=tenant&` +
+        `async_upload=true&` +
+        `application_file_name=${encodeURIComponent(applicationFileName)}&` +
+        `blueprint_archive_url=${encodeURIComponent(blueprintArchiveUrl)}`;
+    return cy.interceptSp('PUT', `${pathname}?${query}`, { body: { id: blueprintName, state: 'pending' } });
```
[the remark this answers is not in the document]
gregcfy: it is split to multiple lines to make code readable

qooban:
Please remember about changing base to `6.0.0-build` branch.

test/cypress/integration/gettingStarted/blueprintMocks.ts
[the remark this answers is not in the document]
kubama: I'm not sure how this addresses my comment

test/cypress/integration/gettingStarted/blueprintMocks.ts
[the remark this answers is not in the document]
kubama: Ok, if so then it looks like we may have the same issue in L29. How about creating a generic function that would create such regexp given the query parameters map? It could be reused there as well.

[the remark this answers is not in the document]
kubama: On the other hand, we have RD-1277 - once it's implemented we will be able to work with regular `RouteMatcher` as the manager URL will be passed as part of the URL itself, not as a query parameter. If so, I'd suggest leavening it as it is.

test/cypress/integration/gettingStarted/blueprintMocks.ts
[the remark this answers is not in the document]
kubama: As I commented elsewhere we have RD-1277 that will allow us to work with RouteMatcher in such cases - let's leave it as it is then.

test/cypress/integration/gettingStarted/blueprintMocks.ts
```diff
@@ -0,0 +1,84 @@
+import awsBlueprintsManager from '../../fixtures/getting_started/aws_blueprints_manager.json';
+
+export const mockBlueprintsManager = (items: any[]) =>
```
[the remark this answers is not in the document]
gregcfy: we have resources from the catalog and from the manager, I wanted to distinguish the source. maybe `mockBlueprintsInManager` ?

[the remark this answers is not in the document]
gregcfy: `awsBlueprintsManager` refers only to blueprints in the manager that are associated with AWS technology.
maybe `awsBlueprintsInManager` ?

test/cypress/integration/gettingStarted/blueprintMocks.ts
```diff
@@ -0,0 +1,84 @@
+import awsBlueprintsManager from '../../fixtures/getting_started/aws_blueprints_manager.json';
```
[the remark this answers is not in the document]
gregcfy: the same line in the previous comment, maybe `awsBlueprintsInManager` ?

test/cypress/integration/gettingStarted/blueprintMocks.ts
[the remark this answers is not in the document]
gregcfy: the best way is to normalize `/^\/blueprints\?.*\...` with some function and later set as an argument.

test/cypress/integration/gettingStarted/blueprintMocks.ts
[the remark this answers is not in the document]
gregcfy: Lub:
```
function UrlBuilder(prefix) {
  	let cache = null;
  	const parameters = {};
  	this.getParameter = function(name, defaultValue) {
    	return parameters[name] ?? defaultValue;
    };
	this.setParameter = function(name, value) {
      	const parameter = parameters[name];
      	if (parameter == value) {
          	return;
        }
      	cache = null;
        if (value == null) {
          	delete parameters[name];
        } else {
          	parameters[name] = String(value);
        }
    };
  	this.toString = function() {
    	if (cache == null) {
        	let url = '';
          	const entries = Object.entries(parameters);
          	for (const entry of entries) {
                const name = encodeURIComponent(entry[0]);
                const value = encodeURIComponent(entry[1]);
                url += (url ? '&' : '?') + name + '=' + value;
            }
          	cache = prefix + url; 
        }
      	return cache;
    };
}
  
// Usage example:
  
const builder = new UrlBuilder('https://domain.com/page');
  
builder.setParameter('preview', true);
builder.setParameter('user', 'john');
builder.setParameter('age', 20);

console.log(builder.toString()); // https://domain.com/page?preview=true&user=john&age=20

builder.setParameter('preview', null);
builder.setParameter('user', null);
  
console.log(builder.toString()); // https://domain.com/page?age=20
```

test/cypress/integration/gettingStarted/blueprintMocks.ts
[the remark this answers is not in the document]
gregcfy: fixed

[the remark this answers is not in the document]
gregcfy: fixed

test/cypress/integration/gettingStarted/blueprintMocks.ts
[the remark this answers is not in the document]
gregcfy: fixed

test/cypress/integration/gettingStarted/blueprintMocks.ts
```diff
@@ -0,0 +1,84 @@
...
+        {
+            body: {
+                metadata: { pagination: { total: items.length, size: 1000, offset: 0 }, filtered: null },
+                items
+            }
+        }
+    );
+
+export const mockAwsBlueprintsManager = () => mockBlueprintsManager(awsBlueprintsManager);
+export const mockEmptyBlueprintsManager = () => mockBlueprintsManager([]);
```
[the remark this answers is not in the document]
gregcfy: done

test/cypress/integration/widgets/user_management_spec.js
```diff
@@ -35,6 +35,13 @@ describe('User management widget', () => {
                 cy.get('.checkbox:not(.checked)').click();
                 cy.get('.checkbox.checked');
             });
+
+            cy.log('Verifying Getting started checkbox is working');
+            cy.get('td:eq(4)').within(() => {
+                cy.get('.checkbox').click();
+                cy.get('.checkbox:not(.checked)').click();
+                cy.get('.checkbox.checked');
+            });
```
kubama, line 44: Shouldn't we also verify that proper backend request was done?

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,444 @@
...
+    mockAwsSecretsInManager,
+    mockEmptySecretsInManager,
+    mockSecretCreation,
+    mockSecretUpdate
+} from './secretMocks';
+
+const gotoBackStep = () => cy.contains('.modal button', 'Back').click();
+const gotoNextStep = () => cy.contains('.modal button', 'Next').click();
+const gotoFinishStep = () => cy.contains('.modal button', 'Finish').click();
+const closeModal = () => cy.contains('button:not([disabled])', 'Close').click();
```
kubama, line 36: I'm wondering if `:not([disabled])` is needed, won't Cypress wait for the button to be enabled if it's disabled when `click` command is called?

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,444 @@
...
+            .interceptSp('POST', `/users/`, res => {
+                showGettingStarted = res.body.show_getting_started;
+                res.reply({ show_getting_started: showGettingStarted });
+            })
+            .mockLogin();
+        cy.contains('label', "Don't show next time").click();
+        cy.contains('button', 'Close').click();
+        cy.reload();
+        cy.get('.modal').should('not.exist'); // the way to check if modal is not visible
+    });
```
kubama, line 55: Why are you mocking the /users/ endpoint if in the end the mocked implementation performs exactly the same logic as the real implementation? 🙂
I'm also not sure about the check in L54 - what would happen if you commented out L51? Would the test fail? The reason I'm asking is that I think the modal doesn't show up immediately after the `reload` command returns, so the test seems to be flaky.

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,444 @@
...
+});
+
+describe('Mocked getting started modal', () => {
```
kubama, line 58: Why do we have two sections? 'Getting started modal' and 'Mocked getting started modal'?

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,444 @@
...
+        cy.contains('.modal', /aws_access_key_id.*secret will be created/);
+        cy.contains('.modal', /aws_secret_access_key.*secret will be created/);
+        cy.contains('.modal', /AWS-Basics-VM-Setup.*blueprint will be uploaded/);
+        cy.contains('.modal', /AWS-VM-Setup-using-CloudFormation.*blueprint will be uploaded/);
+        cy.contains('.modal', /Kubernetes-AWS-EKS.*blueprint will be uploaded/);
+        gotoFinishStep();
+
+        cy.contains('.modal .progress .progress', '100%');
+        cy.contains('.modal .progress .label', 'Installation done!');
+        cy.get('.modal .ui.red.message', { timeout: 0 }).should('not.exist'); // there shouldn't be visible error messages
```
kubama, line 117: What's the purpose of `{ timeout: 0 }`? Is it possible that there is an error message that disappears automatically?

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,444 @@
...
+            }
+        );
+
+        cy.intercept(
+            'POST',
+            '/console/plugins/upload?visibility=tenant&title=Kubernetes&iconUrl=https%3A%2F%2Fcloudify.co%2Fwp-content%2Fuploads%2F2020%2F07%2Fkube-icon.png&yamlUrl=http%3A%2F%2Frepository.cloudifysource.org%2Fcloudify%2Fwagons%2Fcloudify-kubernetes-plugin%2F2.12.1%2Fplugin.yaml&wagonUrl=http%3A%2F%2Frepository.cloudifysource.org%2Fcloudify%2Fwagons%2Fcloudify-kubernetes-plugin%2F2.12.1%2Fcloudify_kubernetes_plugin-2.12.1-centos-Core-py36-none-linux_x86_64.wgn',
+            () => {
+                throw new Error('This case should not occur.');
+            }
+        );
```
kubama, line 153: Why not use RouteMatcher?

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,444 @@
...
+
+        cy.contains('.modal .header', 'Summary');
+        cy.contains('.modal', /cloudify-utilities-plugin.*plugin is already installed/);
+        cy.contains('.modal', /cloudify-kubernetes-plugin.*plugin is already installed/);
+        cy.contains('.modal', /cloudify-aws-plugin.*plugin is already installed/);
+        cy.contains('.modal', /aws_access_key_id.*secret will be updated/);
+        cy.contains('.modal', /aws_secret_access_key.*secret will be updated/);
+        cy.contains('.modal', /AWS-Basics-VM-Setup.*blueprint is already uploaded/);
+        cy.contains('.modal', /AWS-VM-Setup-using-CloudFormation.*blueprint is already uploaded/);
+        cy.contains('.modal', /Kubernetes-AWS-EKS.*blueprint is already uploaded/);
```
kubama, line 216: General comment to this file - looks like `cy.get('.modal').within` could be used to eliminate `.modal` selector repetition

test/cypress/integration/gettingStarted/modal_spec.ts
```diff
@@ -0,0 +1,444 @@
+import {
```
kubama, line 1: General comment to this spec - it would be good if it was verified if actual calls to the manager were performed. At the moment it's validating only the modal UI behaviour but not the underlaying logic, right?

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: it is reversed case, get is waiting for 30s checking if the message is not available all time. After 'Installation done!' message should be visible just. Removing it will increase test duration by about 30s.

test/cypress/integration/widgets/user_management_spec.js
gregcfy: ok, i will restore that test back

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: `/users/admin` path fix

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: yes, only

test/cypress/integration/gettingStarted/modal_spec.ts
kubama: I don't think this is the case. We have `not.exist` assertion - once the element doesn't exist the test continues. So if the message doesn't exist at the time the `should` is performed the test continues and there is no waiting.

test/cypress/integration/gettingStarted/modal_spec.ts
kubama: > `/users/admin` path fix

Not sure how this addresses my comment

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: done

test/cypress/integration/widgets/user_management_spec.js
gregcfy: all user_management_spec test should be changed ?

`cy.get('.checkbox.checked');` doesn't verify change?

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: `:not([disabled])` removed

test/cypress/integration/widgets/user_management_spec.js
kubama: > all user_management_spec test should be changed ?

I didn't review the whole spec, I was commenting only on the added fragment.

> cy.get('.checkbox.checked'); doesn't verify change?

It verifies the change in the UI. It doesn't verify if the change request was actually issued to the backend.

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: quick solution for waiting if the modal is not loaded after the web site is loaded (modal is loaded after additional request):
```
        // eslint-disable-next-line cypress/no-unnecessary-waiting
        cy.wait(7000).get('.modal', { timeout: 0 }).should('not.exist'); // the way to check if modal is not visible
```

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: removed unnecessary description

test/cypress/integration/gettingStarted/modal_spec.ts
gregcfy: done

kubama:
System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/620

@qooban Please review - I refactored it heavily, so it's almost like a new implementation.

app/components/GettingStartedModal/schema.json
```diff
@@ -67,37 +67,37 @@
        ],
        "secrets": [
           {
-             "name": "gpc_client_x509_cert_url",
+             "name": "gcp_client_x509_cert_url",
```
Gelio, line 70: I think I'm blind :smile: What's the change here? and in other `name` fields in below

test/cypress/support/getting_started.ts
```diff
@@ -0,0 +1,20 @@
+import { addCommands, GetCypressChainableFromCommands } from 'cloudify-ui-common/cypress/support';
+// import { waitUntilEmpty } from './resource_commons';
```
Gelio, line 2: Let's remove this line :slightly_smiling_face: 

```suggestion
```

test/cypress/support/getting_started.ts
```diff
@@ -0,0 +1,20 @@
...
+declare global {
+    namespace Cypress {
+        // NOTE: necessary for extending the Cypress API
+        // eslint-disable-next-line @typescript-eslint/no-empty-interface
+        export interface Chainable extends GetCypressChainableFromCommands<typeof commands> {}
+    }
+}
+
+const switchGettingStarted = (username = 'admin', modalEnabled: boolean) =>
+    cy.cfyRequest(`/users/${username}`, 'POST', undefined, { show_getting_started: modalEnabled });
```
Gelio, line 13: Minor:

If we did

```suggestion
const switchGettingStarted = (modalEnabled: boolean) => (username = 'admin') =>
    cy.cfyRequest(`/users/${username}`, 'POST', undefined, { show_getting_started: modalEnabled });
```

Then we could declare the commands like so:

```ts
const commands = {
    enableGettingStarted: switchGettingStarted(true),
    disableGettingStarted: switchGettingStarted(false)
};
```

test/cypress/support/commands.ts
```diff
@@ -147,8 +153,9 @@ const commands = {
             ...options
         });
     },
-    login: (username = 'admin', password = 'admin', expectSuccessfulLogin = true) => {
-        cy.disableGettingStarted();
+    // TODO(RD-2314): object instead of multiple optional parameters
+    login: (username = 'admin', password = 'admin', expectSuccessfulLogin = true, disableGettingStarted = true) => {
+        mockGettingStarted(!disableGettingStarted);
```
Gelio, line 158: What if we don't want to mock `gettingStarted` but instead get the real response from `/users`? I think we need to handle 3 cases

test/cypress/support/commands.ts
```diff
@@ -184,7 +191,7 @@ const commands = {
                     username
                 })
             );
-            cy.disableGettingStarted();
+            if (disableGettingStarted) mockGettingStarted(false);
```
Gelio, line 194: Similarly to the `login` function, we probably want to handle 3 cases:
1. Mock, set `true`
2. Mock, set `false`
3. Don't mock

test/cypress/integration/getting_started_spec.ts
```diff
@@ -0,0 +1,330 @@
...
+    'gcp_private_key',
+    'gcp_zone'
+];
+
+const goToBackStep = () => cy.contains('button', 'Back').click();
+const goToNextStep = () => cy.contains('button', 'Next').click();
+const goToFinishStep = () => cy.contains('button', 'Finish').click();
+const closeModal = () => cy.contains('button', 'Close').click();
+
+function verifyInstallationSucceed(blueprints: string[]) {
```
Gelio, line 23: I think it should be 

```suggestion
function verifyInstallationSucceeded(blueprints: string[]) {
```

test/cypress/integration/getting_started_spec.ts
```diff
@@ -0,0 +1,330 @@
...
+    cy.contains('.progress .progress', '100%', { timeout: blueprints.length * 2 * 60 * 1000 });
+    cy.contains('.progress .label', 'Installation done!');
+    cy.get('.ui.red.message').should('not.exist'); // there shouldn't be visible error messages
```
Gelio, line 26: Not sure if we need the comment :smile: The assertion is self-explanatory to me

```suggestion
    cy.get('.ui.red.message').should('not.exist');
```

Gelio (approved):
Mostly minor comments left :slightly_smiling_face: 

I really enjoy how the tests read, it's very easy to follow them, nice work :ok_hand:

app/components/GettingStartedModal/schema.json
kubama: `gpc` -> `gcp` 🙂

test/cypress/support/getting_started.ts
kubama: Looks like `disableGettingStarted` was not used, so I ended up removing it

test/cypress/integration/getting_started_spec.ts
kubama: Fixed

test/cypress/support/commands.ts
kubama: In fact we only have two use cases:
1) Totally disabled = mocked (all specs except getting_started_spec)
2) Handled by /users endpoint (only getting_started_spec)

test/cypress/support/commands.ts
kubama: Please see https://github.com/cloudify-cosmo/cloudify-stage/pull/1353#discussion_r635219453

qooban:
Thanks for reducing ~300 lines 🙂 
I'll take a look tomorrow morning.

test/cypress/support/commands.ts
```diff
@@ -172,7 +179,7 @@ const commands = {
             cy.waitUntilLoaded().then(() => cy.saveLocalStorage());
         }
     },
-    mockLogin: (username = 'admin', password = 'admin') => {
+    mockLogin: (username = 'admin', password = 'admin', disableGettingStarted = true) => {
```
qooban, line 182: Optional:

I'm not sure what is the scope of RD-2314, but maybe we can add the same TODO as in the `login` command?
```
// TODO(RD-2314): object instead of multiple optional parameters
```

test/cypress/support/commands.ts
```diff
@@ -36,6 +37,11 @@ const getCommonHeaders = () => ({
     tenant: 'default_tenant'
 });
 
+const mockGettingStarted = (modalEnabled: boolean) =>
+    cy.interceptSp('GET', `/users/`, {
+        body: { show_getting_started: modalEnabled }
+    });
```
qooban, line 43: What about the rest of the data provided in the original response? Is it added by cypress `intercept` command automatically? Or we are sure we don't need it?

test/cypress/integration/getting_started_spec.ts
```diff
@@ -0,0 +1,330 @@
...
+            }
+        }).as(toAlias(plugin));
+    });
+}
+
+function interceptBlueprintsUpload(blueprints: string[]) {
+    blueprints.forEach(blueprint => cy.interceptSp('PUT', `/blueprints/${blueprint}`).as(toAlias(blueprint)));
+}
+
+function mockPluginsCatalog(catalogData: any[]) {
```
qooban, line 101: Except `should display information about not available plugins` test case do we need mocking plugins catalog?

Do I get it right that full installation with real plugins list is only in `should omit uploaded plugins and blueprints updating existing secrets` test case?

qooban (commented):
Looks clear and easy to follow 👍🏻 

Added some comments/questions to get better understanding.

test/cypress/integration/getting_started_spec.ts
kubama: > Except `should display information about not available plugins` test case do we need mocking plugins catalog?

Well, we could test without catalog mock. The reason why I decided to use mock catalog is that I can assert detailed query parameters in `interceptPluginsUpload` function that needs to match data from the catalog. This could be still possible with real catalog, but in this case the test would need to fetch catalog data from external resource, which I wanted to avoid. It's also easier to work with a fixture. I also wanted to avoid accessing catalog URL located in application code base.

> Do I get it right that full installation with real plugins list is only in `should omit uploaded plugins and blueprints updating existing secrets` test case?

Full installation is performed in 'should install selected technology' and 'should group common plugins and secrets' test cases. Nothing except catalog data is mocked.

test/cypress/support/commands.ts
kubama: Looks like at the moment nothing more from this response is relevant as the current implementation doesn't break anything.
This is a pre-existing solution and to be honest I'm not sure why it was decided to mock the response instead of just updating the flag on the manager.

app/components/GettingStartedModal/schema.json
Gelio: Oh... right :+1: Thanks!

test/cypress/support/commands.ts
Gelio: Hmm ok, makes sense :+1:

Gelio (approved):
LGTM :slightly_smiling_face:

test/cypress/integration/getting_started_spec.ts
qooban: OK, thanks for the explanation. Let's keep it, then.

test/cypress/support/commands.ts
qooban: Hmm... Taking into account that you already spent long time on that, I think we can leave it.

I still think that mocking the whole response is not the right way to do it, especially that `mockGettingStarted` is used in many quite generic and widely used commands (like `login`). So could you create tech-debt issue to fix it?

qooban (approved):

test/cypress/support/commands.ts
kubama: Created RD-2390

PR #1354, opened 2021-05-10
RD-2252 - The getting started progress bar should be always visible

gregcfy:
The main goal of this PR is to make the installation/upload progress bar always visible - even the list of items is higher than modal height.

Ticket: https://cloudifysource.atlassian.net/browse/RD-2252

Build (system tests passed): https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/575/

Scrollable 'Task List' list:
image

### DIFF
--- a/app/components/GettingStartedModal/ModalContent.tsx
+++ b/app/components/GettingStartedModal/ModalContent.tsx
@@ -51,7 +51,7 @@ const ModalContent = ({
     const secretsStepData = secretsStepsData[secretsStepSchema?.name];
     const statusStepActive = stepName === StepName.Status;
     return (
-        <Modal.Content style={{ minHeight: 220 }}>
+        <Modal.Content style={{ minHeight: 220, display: 'flex', flexDirection: 'column' }}>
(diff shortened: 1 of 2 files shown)


Gelio (approved):
I did not test it in the UI, but the code and the solution look good to me :+1: 

I would not complicate it more than necessary by moving the progress bar into the footer - that would increase the diff and the complexity, but IMO the benefit would not justify it :smile: 

@gregcfy did you test it with multiple screen sizes? Could you share a video how it behaves when the viewport gets smaller?

gregcfy:
> I did not test it in the UI, but the code and the solution look good to me 
> 
> I would not complicate it more than necessary by moving the progress bar into the footer - that would increase the diff and the complexity, but IMO the benefit would not justify it 
> 
> @gregcfy did you test it with multiple screen sizes? Could you share a video how it behaves when the viewport gets smaller?

it works for different viewport

qooban:
@gregcfy Merged to `master`? Didn't we agree to merge to `6.0.0-build` branch only until 6.0 release?

PR #1361, opened 2021-05-12
RD-2045 & RD-2041 - Adjust allowed characters in labels input

qooban:
This PR changes validations on label key and value.

Current validations:
* The key only contains letters, digits and the characters `-`, `.` and `_`
* The key doesn't have to start with a letter
* The value doesn't contain any control characters
* The value doesn't contain a double quote character `"`
* They are shorter than 256 characters

There is a separate issue for adding validations in filter definition form for label key and value inputs - RD-2313 (planned for 6.1).

As our system tests are not working right now, I'm not scheduling a job right now.


kubama (approved):

app/translations/en.json
```diff
@@ -542,7 +542,8 @@
                 "newKey": "New key",
                 "newValue": "New value",
                 "valuePlaceholder": "Select or type in a value",
-                "validationError": "Only letters, digits and the characters \"-\", \".\" and \"_\" are allowed and it must begin with a letter.",
+                "keyValidationError": "Only letters, digits and the characters \"-\", \".\" and \"_\" are allowed.",
+                "valueValidationError": "All characters except \" are allowed.",
```
Gelio, line 546: Minor:

The error tells the user what's allowed, not what they did wrong. I would rather be told what is not allowed, than what's allowed:

```suggestion
                "valueValidationError": "The \" character is not allowed.",
```

widgets/common/src/hooks/useLabelInput.ts
```diff
@@ -1,31 +1,41 @@
...
+const allowedCharactersForLabelValue = /^[^\p{C}"]*$/u;
 const maxInputLength = 256;
 
-// TODO: Add proper typing for onChange function
-function useLabelInput(onChange: (value: any) => void, { allowAnyValue = false, initialValue = '' }) {
+function removeControlCharacters(value: string) {
+    return value.replace(/\p{C}/gu, '');
+}
+
+function formatNewValue(type: 'key' | 'value', newValue: string) {
```
Gelio, line 12: Minor:

```suggestion
function formatNewValue(type: LabelInputType, newValue: string) {
```

widgets/common/src/hooks/useLabelInput.ts
```diff
@@ -1,31 +1,41 @@
...
+    const truncatedNewValue = newValue.substr(0, maxInputLength);
+    return type === 'key' ? truncatedNewValue.toLowerCase() : removeControlCharacters(truncatedNewValue);
+}
+
+function useLabelInput(
+    onChange: (value: string) => void,
+    type: LabelInputType,
+    { initialValue } = { initialValue: '' }
```
Gelio, line 20: Minor:

Technically, doing the following allows for more:

```suggestion
    { initialValue = '' } = { }
```

and is shorter :smile: If there were more properties in the object, the existing solution would make them all required, whereas the one I suggested makes the `initialValue` property optional

widgets/common/src/hooks/useLabelInput.ts
```diff
@@ -1,31 +1,41 @@
...
-            }
-
+        submitChange: (_event: SyntheticEvent | null, data: { value: string; searchQuery?: string }) => {
             // supports both dropdown as well as regular input
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-            const lowercasedNewTypedValue = _.toLower(data.searchQuery! ?? data.value!).substr(0, maxInputLength);
-            if (lowercasedNewTypedValue === '' || allowedCharacters.test(lowercasedNewTypedValue)) {
-                setInputValue(lowercasedNewTypedValue);
-                onChange(lowercasedNewTypedValue);
+            const formattedNewValue = formatNewValue(type, data.searchQuery! ?? data.value!);
```
Gelio, line 34: Minor:

`value` is now required, so we don't need the `!` (unless it's required by mistake)

```suggestion
            const formattedNewValue = formatNewValue(type, data.searchQuery! ?? data.value);
```

widgets/common/src/labels/CommonDropdown.tsx
```diff
@@ -19,14 +20,47 @@ interface CommonDropdownProps
...
+
+const CommonDropdownWithAdditions: FunctionComponent<CommonDropdownProps> = ({
     innerRef = null,
     onChange,
     value = null,
+    type,
+    ...rest
+}) => {
+    const {
+        // @ts-ignore Property 'DynamicDropdown' does not exist on type 'typeof Common'
```
Gelio, line 40: Let's try to use `ts-expect-error` and give a more helpful error message, as in the previous PR :slightly_smiling_face:

Gelio (commented):
I did not fully read through the PR with caution, so I'm only commenting, but it looks good overall :slightly_smiling_face: I believe we can rely on @kubama's approval

app/translations/en.json
qooban: Make sense, I also added info about control characters, which are not allowed, too.

widgets/common/src/hooks/useLabelInput.ts
qooban: Good catch, forgot about that 👍🏻 Changed.

widgets/common/src/hooks/useLabelInput.ts
qooban: Right, changed 👍🏻

widgets/common/src/hooks/useLabelInput.ts
qooban: Good catch 👍🏻 Changed.

widgets/common/src/labels/CommonDropdown.tsx
qooban: Yep, copy-paste issue 🙂 Changed.

qooban:
System tests started: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/587/

app/translations/en.json
Gelio: Good idea :+1: Although it would probably be pretty rare that someone will use control characters :smile: Maybe that would just confuse the user? I'm not sure, but whatever, let's leave it :smile:

Gelio:
The latest commits look good :+1: I don't have more comments on those parts, but still I have not read the PR again, so I'll only leave my review as a comment, not an explicit approval :smile:

app/translations/en.json
qooban: Yeah, I was debating with myself about adding/not adding it having the same concern - "Maybe that would just confuse the user?"

But then I received such question from Sandy on Slack yesterday:
> Out of interest (not required information or anything, I'm just interested), if we're supporting all utf-8 for labels, what do you do if someone puts say a right-to-left special character (that switches text direction after that point) in a label?

so I think it make sense to be more informative 🙂

Gelio: Oh, I see :smile: Makes sense now

One way we could get the best of both worlds (be informative when there are invalid control characters and do not confuse the user otherwise) would be to use different error messages for those 2 distinct cases :smile: But that is a minor issue

kubama (dismissed):

qooban:
I found that after system test execution the following tests failed:
```
Test Result ("6 errors / +3")
  Labels widget should allow to edit labels.should allow to edit labels
  Labels widget should allow to remove labels.should allow to remove labels
  Deployments View widget bulk actions should allow to create child deployments on filtered deployments.should allow to create 
child deployments on filtered deployments
  Filters widget should provide basic functionality: list existing filters.list existing filters
  Deployments View widget drill-down functionality should support the drill-down workflow.should support the drill-down workflow
  Deployments View widget drill-down functionality should allow deleting a deployment without redirecting to the parent page.should allow deleting a deployment without redirecting to the parent page
```
I fixed the first two, which were caused by not updated `labels_spec`, the last 3 are coming from known issue with lack of system filters right now. Unknown for me is the `Deployments View widget bulk actions should allow to create child deployments on filtered deployments.should allow to create child deployments on filtered deployments` failure. It passes locally, so I think I'll leave it for now, maybe that's one time issue, maybe due to many retries taken, I don't know. Let's see in nightly.

kubama (approved):

PR #1363, opened 2021-05-12
RD-2142 - Add "is_not" filter operator

qooban:
image

### DIFF
--- a/widgets/common/src/deploymentsView/detailsPane/DrilldownButtons.tsx
+++ b/widgets/common/src/deploymentsView/detailsPane/DrilldownButtons.tsx
@@ -125,8 +125,7 @@ const deploymentTypeRule: Record<DrilldownButtonProps['type'], Stage.Common.Filt
     services: {
         type: FilterRuleType.Label,
         key: 'csys-obj-type',
-        // TODO(RD-2145): use FilterRuleOperators enum after adding the `is_not` member in it
-        operator: 'is_not' as any,
+        operator: FilterRuleOperators.IsNot,
         values: ['environment']
     }
 };
--- a/widgets/common/src/filters/common.ts
+++ b/widgets/common/src/filters/common.ts
@@ -6,7 +6,11 @@ const i18nPrefix = 'widgets.common.filters.form';
 export const getTranslation = (key: string) => i18n.t(`${i18nPrefix}.${key}`);
 export const getPlaceholderTranslation = (key: string) => getTranslation(`inputsPlaceholders.${key}`);
 
-const anyOperators: FilterRuleOperator[] = [FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf];
-export function isAnyOperator(operator: FilterRuleOperator) {
-    return anyOperators.includes(operator);
+const multipleValuesOperators: FilterRuleOperator[] = [
+    FilterRuleOperators.AnyOf,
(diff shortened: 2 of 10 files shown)


kubama:
Looking just at the dropdown content I need to say it's quite a bit confusing now. I managed to forget the exact semantics of the operators, and looking at it with a fresh eye now I would expect 'is not one of' to work exactly as the newly added operator. If so, the added 'is not one of (or no such key)' option looks confusing. Not sure if it is a subject to change, but I would consider the following:
- label the newly added operator as the existing one - 'is not one of'
- rename the existing 'is not one of' operator so it was clear that the key is expected but with a different value (I don't have a good proposal though)

qooban:
> Not sure if it is a subject to change

@kubama I'm open to change in that matter. There was already a discussion about it in RD-2145 comments. Please add your suggestion there and let Ofer knows about it, so he can take that into account.

kubama:
> Please add your suggestion there and let Ofer knows about it, so he can take that into account.

I added a comment to the ticket. I don't have any comments to the implementation so I'm approving 👍.

kubama (approved):

widgets/common/src/deploymentsView/detailsPane/DrilldownButtons.tsx
```diff
@@ -125,8 +125,7 @@ const deploymentTypeRule: Record<DrilldownButtonProps['type'], Stage.Common.Filt
     services: {
         type: FilterRuleType.Label,
         key: 'csys-obj-type',
-        // TODO(RD-2145): use FilterRuleOperators enum after adding the `is_not` member in it
```
Gelio, line 128: :tada:

widgets/common/src/filters/common.ts
```diff
@@ -6,7 +6,11 @@ const i18nPrefix = 'widgets.common.filters.form';
...
 
-const anyOperators: FilterRuleOperator[] = [FilterRuleOperators.AnyOf, FilterRuleOperators.NotAnyOf];
-export function isAnyOperator(operator: FilterRuleOperator) {
-    return anyOperators.includes(operator);
+const multipleValuesOperators: FilterRuleOperator[] = [
+    FilterRuleOperators.AnyOf,
+    FilterRuleOperators.NotAnyOf,
+    FilterRuleOperators.IsNot
+];
+export function isMultipleValueOperator(operator: FilterRuleOperator) {
```
Gelio, line 14: I like the new name :+1: Optionally, I would name it

```suggestion
export function isMultipleValuesOperator(operator: FilterRuleOperator) {
```

Gelio (approved):
Looks good :ok_hand:

widgets/common/src/deploymentsView/detailsPane/DrilldownButtons.tsx
qooban: 🙂

widgets/common/src/filters/common.ts
qooban: Make sense. Changed.

qooban:
Oh, added a commit and re-approval not necessary. That's probably GH configuration issue for `*-build` branches. I'll let Yuval know. 

I'll wait with merging for the final conclusions about the operators' naming in the UI (see: discussion in RD-2145).

PR #1368, opened 2021-05-13
RD-1654 - Add support for automatic UI layout pages resetting

qooban:
### Changes
* Moved `CURRENT_APP_DATA_VERSION` to common appllication consts file
* Converted `CURRENT_APP_DATA_VERSION` to be set dynamically based on the app version taken from the `package.json` file (as `appDataVersion` column in `UserApps` table is of integer type, I decided to convert version string into a number by simply removing all non-digit characters)
* Updated Cypress tests

System tests: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/597/


kubama (dismissed):

Gelio:
I see you're facing the same issues with the bulk actions tests in the system tests that I do in #1369

app/utils/consts.ts
```diff
@@ -1,4 +1,10 @@
+import { version } from '../../package.json';
+
+const convertVersionStringToNumber = (versionString: string) => Number(versionString.replace(/[^\d]/g, ''));
```
Gelio, line 3: Clever logic :smile: 

I'm thinking out loud:
1. Do we want `6.0.0-build` and `6.0.0` to be the same version? Probably yes
2. Do we want `6.12.3` and `6.1.23` to be the same `6123`? Probably not

While the second point is very rare, we could solve it by:
1. First, changing all dots to a valid digit, e.g. `0`
2. Then, removing all non-digit characters
This would leave us with `601203` and `601023` which are different :smile: 

I doubt this is a perfect solution, but it covers the edge case I just described :smile: We may as well improve it in the future, so I'm happy with having a ticket for it too if you're not convinced to change it now :slightly_smiling_face:

app/utils/consts.ts
```diff
@@ -1,4 +1,10 @@
...
+
 export default {
+    CURRENT_APP_DATA_VERSION: convertVersionStringToNumber(version),
```
Gelio, line 6: Minor:

I'm wondering about the name. Do we need the `DATA` part?

```suggestion
    CURRENT_APP_VERSION: convertVersionStringToNumber(version),
```

or even

```suggestion
    APP_VERSION: convertVersionStringToNumber(version),
```

both looks reasonable to me and are shorter

test/cypress/integration/login_spec.js
```diff
@@ -16,6 +18,30 @@ describe('Login', () => {
         cy.location('pathname').should('be.equal', redirectUrl);
     });
 
+    it('succeeds and resets user pages when application version is different than the one stored in the DB', () => {
+        const currentAppDataVersion = getCurrentAppVersion();
+        const fetchUserAppsTimeout = 15000;
+
+        cy.intercept('GET', '/console/ua', req => {
+            req.reply(res => {
+                res.body.appDataVersion = currentAppDataVersion - 1;
```
Gelio, line 27: Very clever, I like it :+1:

Gelio (dismissed):
IMO looks solid, I just left a few things to consider :slightly_smiling_face:

app/utils/consts.ts
qooban: Got your point. Right now (knowing that - at least since 4.0 - we have never reached two-digits values in any part of the version string) I would keep it as a future enhancement. Created RD-2346.

app/utils/consts.ts
qooban: Seems reasonable. I'll change it.

test/cypress/integration/login_spec.js
qooban: 🙂

app/utils/consts.ts
qooban: Done.

qooban:
Re-run only `login_spec` locally. All passed:
image

Not planning to re-run full system tests suite in Jenkins.

kubama (approved):

app/utils/consts.ts
Gelio: Sure thing :+1:

Gelio (approved):

PR #1369, opened 2021-05-14
RD-1527 Deployments Map marker clustering

Gelio:
This PR adds support for marker clustering in the Deployments Maps widget. This PR only contains the code for the functionality itself and does not contain e2e tests, since the PR is large enough as is. I will add tests in another PR (for RD-1527).

## Features

1. Marker color is based on deployment states
2. The number in the cluster marker represents the number of deployments grouped
3. The currently selected deployment is not clustered so it always appears on top of any markers.

    Useful to see the results immediately when selecting a deployment in the table.
4. Clicking a cluster marker will do the following, depending on the zoom level:
    * zoom in to expand the marker, if zooming in is possible
    * "spiderfy" (leaflet's term) the cluster when the zoom is already maxed out

## Missing functionalities / known bugs

1. Cluster marker tooltips RD-2305
2. Automatically panning/zooming the map when selecting a deployment RD-2135
3. The font used inside the map is different from the app's font RD-2321
4. Multiple deployments from the same site are always clustered RD-2107

## Video

https://user-images.githubusercontent.com/889383/118257784-4f70c880-b4af-11eb-9dd1-604a46076307.mp4

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/598/pipeline

Ran 2021-05-14 12:29 CEST

The deployments view spec passed as I was running it locally (the clustering functionality did not break existing tests, at least locally)


app/typings/stage-api.d.ts
```diff
@@ -21,7 +20,6 @@ declare global {
         React: typeof React;
         ReactDOM: typeof ReactDOM;
         _: typeof _;
-        L: typeof L;
```
Gelio, line 24: This import broke being able to reference `leaflet.markercluster` types

widgets/common/src/MarkerIcon.ts
```diff
@@ -4,7 +4,7 @@ type MarkerIconColor = 'blue' | 'green' | 'grey' | 'red' | 'yellow';
 function createMarkerIcon(color: MarkerIconColor) {
     const widgetId = 'common';
 
-    return new L.Icon({
+    return new window.L.Icon({
```
Gelio, line 7: Necessary since `L` is no longer a global variable available in modules due to https://github.com/cloudify-cosmo/cloudify-stage/pull/1369/files#r632438693

widgets/common/src/deploymentsView/map/marker/single-deployment-site.tsx
```diff
@@ -54,16 +62,36 @@ const DeploymentSiteMarker: FunctionComponent<DeploymentSiteMarkerProps> = ({
...
+
 const BareDeploymentSiteMarker: FunctionComponent<{
     status: DeploymentStatus;
     position: [number, number];
     onClick?: () => void;
 }> = ({ status, position, children, onClick }) => {
     const { Marker } = Stage.Basic.Leaflet;
     const icon = Stage.Common.createMarkerIcon(deploymentStatusToIconColorMapping[status]);
+    const markerRef = useRef<
+        import('react-leaflet').Marker<import('react-leaflet').MarkerProps, DeploymentMarkerWithStatus>
```
Gelio, line 75: Both `leaflet` and `react-leaflet` use the `Marker` name, so I decided to use this import syntax to get the type results to avoid importing one or the other library in the whole file (and cause naming collisions)

widgets/common/src/deploymentsView/map/marker/cluster.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+type DeploymentStatusCounts = Partial<Record<DeploymentStatus, number>>;
+const worstToBestDeploymentStatuses = [
+    DeploymentStatus.RequiresAttention,
+    DeploymentStatus.InProgress,
+    DeploymentStatus.Good
+];
+const getAggregatedClusterStatus = (statusCounts: DeploymentStatusCounts): DeploymentStatus => {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const aggregatedStatus = worstToBestDeploymentStatuses.find(status => statusCounts[status]! > 0);
```
kubama, line 96: Is this 'non null assertion' used to narrow the type and strip the possible null from it?
I've already seen it in other places, if it's usable then why not disable the eslint rule globally?

kubama (approved):
Looks very thoughtful. I only have one comment/question.

widgets/common/src/deploymentsView/map/marker/cluster.tsx
Gelio: > Is this 'non null assertion' used to narrow the type and strip the possible null from it?

Yes, exactly :slightly_smiling_face: 



> I've already seen it in other places, if it's usable then why not disable the eslint rule globally?

IMO the rule generally makes sense. We want to be handling `null` and `undefined` values explicitly in most cases. I use `!` only when I am sure that it's impossible that the value is null, or it does not matter (e.g. as in this case, it does not matter because `null > 0` is `false`)

Maybe I could have used `(statusCounts[status] ?? 0) > 0` here which would be legal from ESLint perspective :thinking: 

Anyway, I see the value in keeping the `no-non-null-assertion` rule but I am also open to a wider discussion whether to keep it, e.g. during the weekly :slightly_smiling_face:

Gelio:
Aside from the known failures in system tests, the bulk actions test in the `deployments_view_spec` failed because the blueprint could not have been picked:
https://jenkins.cloudify.co/job/Stage-UI-System-Test/598/artifact/cloudify-stage/test/cypress/screenshots/widgets/deployments_view_spec.ts/Deployments%20View%20widget%20--%20bulk%20actions%20--%20should%20allow%20to%20create%20child%20deployments%20on%20filtered%20deployments%20(failed)%20(attempt%202).png

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/598/artifacts

Looks unrelated to this PR IMO

qooban:
I won't be able to review this PR today, I'll review it on Monday morning.

If you'd like to merge it before the weekend, I'm fine as Kuba M. already approved.
And knowing that there'll be another PR with system tests update, any comments I would have here can be applied in the next PR.

Gelio:
> I won't be able to review this PR today, I'll review it on Monday morning.
> 
> If you'd like to merge it before the weekend, I'm fine as Kuba M. already approved.
> And knowing that there'll be another PR with system tests update, any comments I would have here can be applied in the next PR.

No problem, I still need to put some effort into the tests so we can wait with this PR until Monday :slightly_smiling_face: Thanks for notifying me

widgets/common/src/deploymentsView/map/DeploymentsMap.tsx
```diff
@@ -50,6 +56,19 @@ const DeploymentsMap: FunctionComponent<DeploymentsMapProps> = ({
     const { Map: MapComponent } = Stage.Basic.Leaflet;
     const { DefaultTileLayer } = Stage.Common.Map;
 
+    function renderDeploymentSiteMarker({ deployment, site }: DeploymentSitePair) {
+        return (
+            <DeploymentSiteMarker
+                deployment={deployment}
+                site={site}
+                key={`${deployment.id}\n${site.name}`}
```
qooban, line 64: Clever, I see multi-line React key for the first time 🙂 I assume it was intentional to clearly separate deployment ID from the site name?

qooban (approved):
Looks good 👍🏻

widgets/common/src/deploymentsView/map/DeploymentsMap.tsx
Gelio: Yes, exactly :smile: I was using a dash as a separator, but @kubama rightly noted that it could lead to problems :smile: React does not seem to care about the contents of the string, so we may as well use such constructs :slightly_smiling_face:

Gelio:
The only tests that failed in CI are those that also fail in other PRs and seem to be unrelated to this PR so I'll merge it :slightly_smiling_face: Thanks for the reviews :eyes:

PR #1374, opened 2021-05-17
WDR-21 Add polling to drill-down buttons

Gelio:
The PR adds polling (`refreshInterval`) to drill-down buttons in the Deployments View widget. Fixes https://cloudifysource.atlassian.net/browse/WDR-21

I have also taken the chance to improve the margins in the details pane's header.

## Video

https://user-images.githubusercontent.com/889383/118489841-96fc8c00-b71d-11eb-9089-b2444aa1fa7d.mp4

Notice that there's a loading overlay for the drill-down buttons now. Also, notice that there's a gap between the drill-down buttons and the deployment action buttons (they were right next to each other previously, without any vertical space).

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/605/pipeline

Queued 2021-05-17 14:37 CEST

### DIFF
--- a/app/utils/StageAPI.ts
+++ b/app/utils/StageAPI.ts
@@ -23,7 +23,7 @@ type StageHooks = typeof StageHooks;
 /** @see https://docs.cloudify.co/developer/writing_widgets/widget-apis/#toolbox-object */
 interface StageToolbox {
     drillDown(
-        widget: StageWidget,
+        widget: StageWidget<unknown>,
         defaultTemplate: string,
         drilldownContext: Record<string, any>,
         drilldownPageName?: any
--- a/.size-limit
+++ b/.size-limit
@@ -20,7 +20,7 @@
     },
     {
         "name": "Application package",
-        "limit": "13.39 MB",
+        "limit": "13.45 MB",
         "gzip": false,
         "running": false,
         "path": ["dist", "!dist/**/*.gz"]
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -529,6 +529,25 @@ describe('Deployments View widget', () => {
             getDeploymentsViewTable().within(() => cy.contains(tempDeploymentId).should('not.exist'));
             getBreadcrumbs().contains(parentDeploymentId);
         });
+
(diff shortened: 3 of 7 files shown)


kubama (dismissed):

Gelio:
The only system tests that failed are the known failures. I will only need to increase the app bundle size to get the CI on this PR to pass :slightly_smiling_face:

app/utils/StageAPI.ts
```diff
@@ -23,7 +23,7 @@ type StageHooks = typeof StageHooks;
 /** @see https://docs.cloudify.co/developer/writing_widgets/widget-apis/#toolbox-object */
 interface StageToolbox {
     drillDown(
-        widget: StageWidget,
+        widget: StageWidget<unknown>,
```
qooban, line 26: This is because it differs between widgets and it is not possible to have "common ground"?

qooban (approved):
LGTM 👍🏻

app/utils/StageAPI.ts
Gelio: Oh, sorry, I did not leave a comment about it :sweat_smile: 

The problem here has to do with covariance and contravariance of types in TS :smile: I hope I can explain in simple terms

1. When creating the `StageWidget` type I thought that the configuration can be any object that has string keys (`Record<string, unknown>`), so I set it as a default value: https://github.com/cloudify-cosmo/cloudify-stage/blob/6968d562a9198645ebb66723810370536b7877ba/app/utils/StageAPI.ts#L48
2. This works well in the `Stage` API in the functions that accept a `widget`, because they can do `widget.configuration.anything` and the type of this expression will be `unknown`
3. However, when you are calling this function (e.g. `toolbox.drillDown`), the argument is `StageWidget<Record<string, unknown>>` which means that you need to pass a widget whose `configuration` accepts any property.
    The problem manifests itself exactly in this case, because the `widget` that we have here has a concrete type for its configuration, that only has known properties (`SharedDeploymentsViewWidgetConfiguration`):
    
    image

I'm leaning towards changing the default type for `StageWidget`'s configuration to be `unknown` instead of `Record<string, unknown>` which could help, but I did not think about it that much :slightly_smiling_face:

Gelio:
Thanks for the reviews :+1:

PR #1376, opened 2021-05-17
WDR-31 Prevent using invalid site latitudes and longitudes

Gelio:
Do not allow `NaN` or `Infinity` (which are of type `number`).

This PR aims to be a hotfix, because the root cause is still unknown to me. As I have followed the source code, we do the following operations:
1. Filter out sites that have `latitude` and `longitude`
2. Match deployments with those sites

I don't see any other operation that would make latitude or longitude be a `NaN`.

If, after merging this PR, Ofer and Gabriel will not have satisfactory results in the UI, I will have to investigate using their managers, as I could not reproduce the problem locally.

## System tests

https://jenkins.cloudify.co/job/Stage-UI-System-Test/606/

Queued 2021-05-17 20:31 CEST

### DIFF
--- a/test/jest/widgets/common/map/site.test.ts
+++ b/test/jest/widgets/common/map/site.test.ts
@@ -1,0 +1,43 @@
+import StageUtils from 'utils/stageUtils';
+
+// @ts-expect-error Necessary global API
+Stage.Utils = StageUtils;
+
+// NOTE: use `require` to lazily load the module after `Stage.Utils` is set
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { isSiteWithPosition } = require('../../../../../widgets/common/src/map/site');
+
+describe('(Widgets common) isSiteWithPosition', () => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const declareValidityCheck = (expectedResult: boolean, position: [any, any]) => {
+        it(`should return ${expectedResult} for ${position}`, () => {
+            expect(
+                isSiteWithPosition({
+                    name: 'abc',
+                    latitude: position[0],
+                    longitude: position[1]
+                })
+            ).toBe(expectedResult);
+        });
+    };
+
+    const validPositions: [number, number][] = [
+        [1, 2],
+        [1.543241, -12.123123],
+        [51.5224160825326, -0.1318359375],
+        [42.322001080603, -71.12548828125],
+        [32.7734193549752, -117.04833984375]
+    ];
+    validPositions.forEach(position => declareValidityCheck(true, position));
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const invalidPositions: [any, any][] = [
+        [null, null],
+        [NaN, NaN],
+        [Infinity, Infinity],
+        [-Infinity, -Infinity],
+        ['12.33', '1.8'],
+        [true, false]
+    ];
+    invalidPositions.forEach(position => declareValidityCheck(false, position));
+});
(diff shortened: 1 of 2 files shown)


test/jest/widgets/common/map/site.test.ts
```diff
@@ -0,0 +1,43 @@
...
+            ).toBe(expectedResult);
+        });
+    };
+
+    const validPositions: [number, number][] = [
+        [1, 2],
+        [1.543241, -12.123123],
+        [51.5224160825326, -0.1318359375],
+        [42.322001080603, -71.12548828125],
+        [32.7734193549752, -117.04833984375]
```
Gelio, line 29: These are locations that are present in the email that Ofer sent

qooban (approved):
Change looks good. But I'm not sure it'll fix it.

I did a quick search and found that there are few items containing "invalid LatLng object" phrase in [Leaflet project](https://github.com/Leaflet/Leaflet/issues?q=is%3Aissue+is%3Aopen+%22invalid+latlng+object%22). Maybe there's something in the library itself...

I'm wondering maybe you could get a snapshot from Ofer and try to upload it locally to your environment and then you'll manage to reproduce the problem?

Gelio:
@qooban I am also not certain that this is the fix for the problem :smile: I have scheduled a debugging session with Ofer for today at 10 AM. I will try to get the responses from `/sites` and `/searches/deployments` and then see if there's anything suspicious.

I am not familiar with the snapshots functionality, so I will rely on returning a hardcoded response from the API calls, I hope that will be enough :slightly_smiling_face: 

I'll wait with merging the PR before I get some more investigation info

---

There are only known failures in the system tests for this PR, so it is safe to merge from my perspective

qooban:
@Gelio Sure 👍🏻 Live debugging session can help indeed.

Reg. snapshots - it's relatively easy. Cloudify Manager snapshot is an archive (eg. ZIP file). You just need to upload it to a system and then restore. All these steps can be accomplished in the UI, but as snapshot files might be big it is recommended to do that in the CLI.

Once you have an archive, on the destination CM you need to run:
```
cfy snapshots upload [OPTIONS] SNAPSHOT_PATH
cfy snapshots restore [OPTIONS] SNAPSHOT_ID
```
More about it at: https://docs.cloudify.co/latest/cli/maint_cli/snapshots

test/jest/widgets/common/map/site.test.ts
```diff
@@ -0,0 +1,43 @@
+import StageUtils from 'utils/stageUtils';
+
+// @ts-expect-error Necessary global API
+Stage.Utils = StageUtils;
+
+// NOTE: use `require` to lazily load the module after `Stage.Utils` is set
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { isSiteWithPosition } = require('../../../../../widgets/common/src/map/site');
```
kubama, line 8: We have `widgets` set as module directory (see `jest.config.js`) so you should be able to use path relative to it.

kubama (approved):

test/jest/widgets/common/map/site.test.ts
Gelio: Good point, I'll try :+1:

Gelio: I am getting an ESLint error when trying to use `require` with a bare path
image

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md

I tried enabling the `commonjs` option, but it did not help. I would rather have a long path than `eslint-disable`, so I'll leave it as is :slightly_smiling_face:

PR #1375, opened 2021-05-17
RD-2210 - Fix Sites Map widget to work properly with large number of deployments

qooban:
### Changes
* Changed way of fetching data in Sites Map widget
* Refactors:
  * Full TS conversion of Sites Map widget
  * `Stage.Common` improvements - added `SummaryActions` class and improved `DeploymentActions` class
  * Few widgets updates to use new API added to `Stage.Common`

PR is not as small as I would like it to be, but it was really hard to make it smaller without wasting time on having artificial solution. There is for sure more to be done with Sites Map widget. But on the other hand I think I have already done some parts of RD-1780 or at least prepared a base for it.

Sites Map widget system tests locally are passing. I have scheduled full run: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/612/
(first try no. 610 failed due to environmental issues)

### Screenshot
image

### To Do
- [X] add tooltip statuses description (I'll ask Ofer) => UPDATE: I got answer from him, so I updated the PR.
- [x] documentation update as there was an update of the statuses provided in the tooltip (there will be 2 more PRs - one in docs.cloudify.co repo and one in cloudify-stage repo to update README.md file)  => UPDATE: PR in docs repo merged, so I updated also `README.md` file in this PR.

PR #1378, opened 2021-05-18
WDR-28 Propagate map open state in drill-down pages

Gelio:
This PR fixes [WDR-28](https://cloudifysource.atlassian.net/browse/WDR-28).

1. The map _open_ state is maintained when drilling down.

    This is done by setting the `mapOpen` property in the drilled-down page's context.
2. When going back to the parent page, the map _open_ state is restored to what it was before drilling down.

    This means that when the user
    1. Opens the map in the parent page
    2. Drills down (the map will stay open)
    3. Closes the map in the drilled-down page
    4. Goes to the parent page
    the map will be opened because the `mapOpen` was `true` in the parent page's context. I did not find a better solution that would address this minor issue.

3. The drilled-down version of the Deployments View widget does not have "Show map by default" configuration. It gets this information from the parent page.

## Known issues

1. When going to the parent page, the map open state is restored to the moment before the drill-down (described in point 2. above)
2. "Show map by default" configuration is not applied immediately after changing the setting. The user has to refresh the page or move to another page and back to the Deployments View widget to see the map by default.
    I did not find a way to fix it.

## Video

https://user-images.githubusercontent.com/889383/118657142-7f8bd480-b7eb-11eb-9f38-58b2aa97ef92.mp4

## System tests

They passed locally (I created a `csys-environment-filter` substitute myself to get the drill-down tests to pass):
image

https://jenkins.cloudify.co/job/Stage-UI-System-Test/613/

Queued 2021-05-18 15:11 CEST


Gelio:
Only known failures in the system tests + the new test failed because it relies on the `csys-environment-filter` system filter. It passed for me locally (see the PR description). Aside from that, no regressions

Gelio:
The PR pipeline failed because the readme for `sitesMap` is out of sync. To be fixed by https://github.com/cloudify-cosmo/cloudify-stage/pull/1375

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -143,6 +148,10 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
         context.setValue('deploymentId', deployments[0].id);
     }
 
+    // eslint-disable-next-line react/destructuring-assignment
+    const mapOpen = (context.getValue(mapOpenContextKey) as boolean | undefined) ?? false;
```
kubama, line 152: Wouldn't `!!` work here?

kubama (approved):

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -143,6 +148,10 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
         context.setValue('deploymentId', deployments[0].id);
     }
 
+    // eslint-disable-next-line react/destructuring-assignment
```
kubama, line 151: I suspect the reason behind this rule might be that it confuses the `context` variable with react context. Would it still complain if we renamed `context` to `pageContext`?

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: It would :slightly_smiling_face: 

The line would be 

```suggestion
    const mapOpen = !!(context.getValue(mapOpenContextKey) as boolean | undefined);
```

I don't see a major difference and I don't mind so I'll use `!!~

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: Good point :+1: I did not read the rule docs well enough, thanks :smile: I'll rename it to `toolboxContext`, I hope you don't mind

widgets/common/src/deploymentsView/DeploymentsView.tsx
kubama: I just thought `as boolean | undefined` would be unnecessary in this case. Is it still needed?

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -130,19 +135,22 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
...
     }
 
-    const context = toolbox.getContext();
+    const toolboxContext = toolbox.getContext();
     const deployments = deploymentsResult.data.items;
     const selectedDeployment = find(deployments, {
         // NOTE: type assertion since lodash has problems receiving string[] in the object
         // eslint-disable-next-line react/destructuring-assignment
-        id: context.getValue('deploymentId') as string | undefined
+        id: toolboxContext.getValue('deploymentId') as string | undefined
```
qooban, line 143: Digression:

I know this code was before, so I'm just wondering how actually ` as string | undefined` works. 
What if we have an array of ids. It is then converted to `undefined`?

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -130,19 +135,22 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
         );
     }
 
-    const context = toolbox.getContext();
+    const toolboxContext = toolbox.getContext();
```
qooban, line 138: Just wondering was there any reason for changing the name?

qooban (approved):
LGTM 👍🏻 

Just added some question - more for me to understand.

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: IMO it serves as additional documentation and tells you what values you can expect from the context :slightly_smiling_face: I don't think it hurts

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: See https://github.com/cloudify-cosmo/cloudify-stage/pull/1378#discussion_r634482945 :slightly_smiling_face:

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: `as` in TS is a TS-only construct that is erased when the code is compiled. It does not generate any type conversions - it's just to tell TS that we are sure that the type is `string | undefined`, even if the return type of `context.getValue('deploymentId')` can also include `string[]`. It's basically us saying:
> yeah, we know that the types say it can be different, but trust me, I know what I'm doing :slightly_smiling_face: 

More info: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions

The code does handle `string[]` since we're using lodash's `find` (when `deploymentId` will be an array, `selectedDeployment` won't be found), but TS reports some obscure error when trying to pass `string[]` as an argument here. See the comment on line 141 :slightly_smiling_face: This is why I added the assertion some time ago

widgets/common/src/deploymentsView/DeploymentsView.tsx
qooban: I see, so is the comment
```
// eslint-disable-next-line react/destructuring-assignment
```
in L142 still needed?

widgets/common/src/deploymentsView/DeploymentsView.tsx
qooban: Got it, thx for the explanation 👍🏻

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: Good catch, it's redundant :smile: I've removed it

widgets/common/src/deploymentsView/DeploymentsView.tsx
kubama: Since we are using `!` operator on the value I don't think this type documentation is of any use. Also, there is nothing that would ensure validity of this type - you could specify anything as the type here and it would still work. And if the actual type eventually changed this type specification would be misleading. I would still remove it, but I leave it up to you.

qooban (approved):

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: Valid points :+1: I will merge it as-is to not ask for reviews again and prolong this PR, but I generally agree :slightly_smiling_face:

PR #1380, opened 2021-05-19
RD-2371 Fix event row ID calculation in Events/Logs widget

qooban:
See RD-2371 for details about the problem.

### Before

Not only there was React error, but also the rows were not filtered properly (notice that with `hello-world` blueprint selected I see also events from `nodecellar` blueprint):

image

### After

No React error and accurate filtering:

image


Gelio (approved):
Wow :smile: That's a huge simplification :ok_hand: I did not test in the UI, but the code looks good :+1:

widgets/events/src/widget.jsx
```diff
@@ -142,20 +142,8 @@ Stage.defineWidget({
...
-                        item.level +
-                        item.node_name +
-                        item.workflow_id +
-                        item.reported_timestamp +
-                        item.deployment_id +
-                        item.type +
-                        item.execution_id
-                );
+                // eslint-disable-next-line no-underscore-dangle
+                const id = item._storage_id;
```
kubama, line 146: Another undocumented feature?🙂 I'm just curious how safe it is to rely on it - isn't it something that can disappear from the API at any time without any notice? If it's something usable maybe we should first try to have it moved as an 'official' API feature?

tehasdf: as noted on today's standup, let's probably use this for 6.0 indeed, and consider for the next version whether events/logs should have an actual (uu)id, or maybe if we would promote the storage id to something official

kubama (approved):

qooban:
Scheduled system tests run for this PR: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/622/

qooban:
System tests did not complete due to... `java.io.EOFException`.
Hmm... Anybody has already complained about the system tests environment stability? 🙂 

I'll wait with merge anyway (when I get more information in the e-mail thread about `_storage_id`).

qooban:
I'll not rerun the system tests again. I ran just `events_spec` locally and it passed without any issues, so I'm going to merge this PR once CI passes.

PR #1382, opened 2021-05-19
RD-2368 Updated deployment drill down page to display display_name in place of id

kubama:
System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/622/

I submitted full system tests execution as this change has a relatively big impact. I expect some fixes to be required.


qooban (approved):
Looks good at first review. Let's see what comes from the system tests 🙂

kubama:
I fixed deployment_button_spec and verified it locally. I'm not planning another full system tests execution as it was the only spec requiring an update.

qooban (approved):

PR #1386, opened 2021-05-21
RD-2078 Deploy On test fix

kubama:
The reason of the failures was a timing issue when selecting values from dynamically loaded dropdown (Blueprint field in this case).
As a fix and also to prevent such issues in the future I created ` setDropdownValue` command that does the thing in a safe way.

It is possible to reproduce the issue manually if you type fast. However, this is such an edge case that I don't consider it worth addressing as a DynamicDropdown component fix.

System tests all green on the branch: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/627/


test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -865,7 +865,7 @@ describe('Deployments View widget', () => {
             const labelKey = 'label_key';
             const labelValue = 'label_value';
             cy.get('.modal').within(() => {
-                cy.contains('.field', 'Blueprint').find('input').type(`${blueprintName}{enter}`);
+                cy.setDropdownValue('Blueprint', blueprintName);
```
qooban, line 868: Could you also apply the same solution for workflow selection in "should allow to run workflow on filtered deployments" test in this spec?

qooban (commented):
Added one comment.

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: Good point, applied

qooban (approved):

kubama:
Since this is a simple change I'm merging it with just one approval

Gelio (commented):
Nice :ok_hand:

PR #1388, opened 2021-05-21
RD-1532 Update Deployments View widgets READMEs

Gelio:
This PR updates the READMEs and images of both variants of the Deployments View widgets using the `updateReadmes.js` script.

Because https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1840 is not yet merged, I have changed the branch name in https://github.com/cloudify-cosmo/cloudify-stage/blob/RD-1532-update-Deployments-View-widget-readme/scripts/readmesConfig.json#L2 to prepare the PR ahead of time.

I have based on the https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1840/commits/74d8f2f7dfbe01892b6d7358d35467059198b5e2 commit of the readmes when copying them.

## Noticed issue

I have noticed an issue with relative links in the readmes and I have reported RD-2417 to eventually fix it.

## Screenshots

Main variant:
image

The images are broken as we are using the 5.1 prefix in https://github.com/cloudify-cosmo/cloudify-stage/blob/1f278f95115d16da8e7fca98938b41ef9293bcb6/app/translations/en.json#L609

btw. maybe it would make 

Drilled-down:
image

## System tests

The only logic change is enabling `hasReadme` which is not tested anywhere :smile: The README contents are verified by another job. I am not planning to run system tests for this PR.


widgets/deploymentsViewDrilledDown/README.md
```diff
@@ -1,5 +1,12 @@
-# Deployments View (drilled-down)
+# Deployments View
```
qooban, line 1: So the header is the same for both widgets?

qooban (commented):
Looking at the size increase, let's compress changed PNG files.

Reg.:
> The images are broken as we are using the 5.1 prefix in
It'll be fixed within RD-2380.

Gelio:
@qooban 

> Looking at the size increase, let's compress changed PNG files.

They are compressed :slightly_smiling_face: It's probably the map that increases the file size, as it was not present in the previous screenshots

widgets/deploymentsViewDrilledDown/README.md
Gelio: That's what I got from `updateReadmes.js`, I did not dig into why. The title in https://github.com/cloudify-cosmo/docs.getcloudify.org/blob/74d8f2f7dfbe01892b6d7358d35467059198b5e2/content/working_with/console/widgets/deploymentsViewDrilledDown.md is different

Looks like the regexp in https://github.com/cloudify-cosmo/cloudify-stage/blob/1f278f95115d16da8e7fca98938b41ef9293bcb6/scripts/updateReadmes.js#L43 does not handle parentheses. I will try to update it in this PR

Gelio: Done, I've used a slightly different regexp that handles non-word characters

kubama (dismissed):

Gelio:
I have rebased the PR on top of `6.0.0-build` and force pushed. No other code changes were done in the meantime.

I have run `npm run docWidgets` again but looks like the output is wrong:

image

I have a hunch that the `updateReadmes` script does not handle multiple `note`s. I'll try to come up with a fix in this PR, as otherwise the readme for the Deployments View widget will be broken

widgets/deploymentsViewDrilledDown/README.md
qooban: Great, thanks!

qooban:
@Gelio 


> They are compressed slightly_smiling_face It's probably the map that increases the file size, as it was not present in the previous screenshots

Oh, got it. Map has a lot of colors. OK, we have to live with it right now.

> I have a hunch that the updateReadmes script does not handle multiple notes. I'll try to come up with a fix in this PR, as otherwise the readme for the Deployments View widget will be broken

Yeah, probably there's an issue with regex again. This time with:
```
const noteRegex = /{{%\s*note.*%}}([^]*){{%\s*\/note\s*%}}/gm;
```

I think I'll also need to update styling for tables: 
image

I'll try to do that within my PR: https://github.com/cloudify-cosmo/cloudify-stage/pull/1391

Gelio:
@qooban  
> I think I'll also need to update styling for tables:

Yup, looks like it :smile: 

---

I have fixed the transformation for multiple note/warning/fix tags in widget readmes. There was a problem with the greediness of the regexp used.

It works correctly now:

image


With that, the PR should be ready for final reviews

qooban (approved):
Looking at the screenshot, I'm wondering would it also be possible to fix links in `note`'s? :

image

PR #1393, opened 2021-05-24
RD-2440 Fix page redirection on clicking site marker tooltip

qooban:
As we removed/changed recently some of the pages from the layout Sites Map redirection (on clicking the deployment status in the site marker tooltip) stopped working. This PR fixes that.

Within that PR I also did two small alignments to Deployments View widget:
* Changed "Good" site marker color (from green to blue)
* Changed "Failed" deployment status name (from “Failed” to “Requires attention”)

As some of the pages were removed I also updated `usePageMock` function.

Full system test run: https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/651/

### DIFF
--- a/test/cypress/integration/cluster_status_spec.js
+++ b/test/cypress/integration/cluster_status_spec.js
@@ -52,11 +52,7 @@ describe('Cluster Status', () => {
         cy.get('table.servicesData').within(() => {
             cy.get('button.refreshButton').should('not.have.class', 'loading');
             checkServicesStatus('Degraded', 'OK', 'OK');
-
-            cy.get('tbody tr:nth-child(1)').click();
         });
-
-        cy.location('pathname').should('be.equal', '/console/page/admin_operations');
     });
 
     it('shows correct data', () => {
--- a/test/cypress/integration/widgets/sites_map_spec.js
+++ b/test/cypress/integration/widgets/sites_map_spec.js
@@ -30,16 +30,8 @@ describe('Sites Map', () => {
         cy.get('.leaflet-marker-icon').should('have.length', 1).click();
         cy.get('.leaflet-popup .leaflet-popup-content')
             .should('be.visible')
-            .within(() => {
-                cy.get('h5.header').should('have.text', testSite.name).click();
-                cy.get('.deploymentState').first().click();
-            });
-        cy.location('pathname').should('be.equal', '/console/page/deployments');
-        cy.location('search').then(queryString =>
-            expect(JSON.parse(new URLSearchParams(queryString).get('c'))).to.deep.equal([
-                { context: { siteName: testSite.name } }
-            ])
-        );
+            .find('h5.header')
+            .should('have.text', testSite.name);
 
         cy.log('Add second site');
         const secondSite = { name: 'Bergen', location: '60.389433, 5.332489', visibility: 'private' };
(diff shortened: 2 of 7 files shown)


test/cypress/integration/cluster_status_spec.js
```diff
@@ -52,11 +52,7 @@ describe('Cluster Status', () => {
         cy.get('table.servicesData').within(() => {
             cy.get('button.refreshButton').should('not.have.class', 'loading');
             checkServicesStatus('Degraded', 'OK', 'OK');
-
-            cy.get('tbody tr:nth-child(1)').click();
         });
-
-        cy.location('pathname').should('be.equal', '/console/page/admin_operations');
```
qooban, line 59: Fix may come within RD-2442.

test/cypress/integration/widgets/sites_map_spec.js
```diff
@@ -49,4 +41,19 @@ describe('Sites Map', () => {
         cy.log('Verify second site is present on the map');
         cy.get('.leaflet-marker-icon').should('have.length', 2);
     });
+
+    it('opens a page showing list of deployments per selected site', () => {
+        // NOTE: Do not mock login to load all currently available pages and test drill-down to site
+        cy.login();
```
kubama, line 47: So far for testing drill down functionality we were using mocked empty pages defined in `commands.ts` (removed in this PR) and this way we could still use `mockLogin` in such cases. Was there any issue with this approach that it was changed?

Gelio: If I understand it right, I have already solved a similar problem in Deployments View by using `additionalPageTemplates`:

https://github.com/cloudify-cosmo/cloudify-stage/blob/576b9805ccd748179363c5163c47a57c24166187/test/cypress/integration/widgets/deployments_view_spec.ts#L70-L74

Gelio (dismissed):
Looks great :+1:

test/cypress/integration/widgets/sites_map_spec.js
qooban: > So far for testing drill down functionality we were using mocked empty pages defined in commands.ts (removed in this PR) and this way we could still use `mockLogin` in such cases. Was there any issue with this approach that it was changed?

In my opinion it is error-prone solution as it didn't brake when those pages were removed from the template (ref. Alex's PR). We can add there anything and we will not get any error in the test. That's why we didn't spot RD-2440 in the tests.

> If I understand it right, I have already solved a similar problem in Deployments View by using `additionalPageTemplates`

This solution has the same flaw as described before.

I don't think I will have time to implement anything more advanced. I think what we can do in the future is to have automatically added all pages available in `/templates/pages` catalog.

test/cypress/support/commands.ts
```diff
@@ -284,9 +284,7 @@ const commands = {
                                 ]
                             }
                         ]
-                    },
-                    { id: 'admin_operations' },
```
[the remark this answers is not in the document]
qooban: Yes, it should do that.

kubama (approved):

qooban:
Fixed failing `login_spec`. There's something mysterious with `sites_map_spec`.

Re-run with just those two tests:
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/653/

Gelio (approved):

PR #1395, opened 2021-05-25
Updated browserslist

kubama:


qooban (commented):
Two things:

1. Don't we need to have also `cloudify-ui-common`, `cloudify-ui-components` and `cloudify-blueprint-topology` to be released and updated here as well?
2. I know it should not have big impact, but still I'm not sure we should push to `6.0.0-build` branch right now.

kubama:
> 1. Don't we need to have also `cloudify-ui-common`, `cloudify-ui-components` and `cloudify-blueprint-topology` to be released and updated here as well?

It's not necessary.

> 2. I know it should not have big impact, but still I'm not sure we should push to `6.0.0-build` branch right now.

The goal was to fix the nightly build. But I understand the concern. If we are OK with failing nightly build we can switch this PR to master.

qooban (approved):
`master` branch is ready.

PR #1397, opened 2021-05-27
CY-1275 Workflows menu - made whole group name clickable

kubama:
The fix involves removing wrapping element padding and adding the same padding to the group name element. The change is visually transparent, thus no screenshot.

This is a styling update only, I'm not planning system tests execution.


widgets/common/src/WorkflowsMenu.jsx
```diff
@@ -74,14 +74,18 @@ class AccordionWorkflowsMenu extends React.Component {
...
                             active={activeGroup === group.name}
                             index={group.name}
                             content={<StyledTitle bold={group.name === defaultGroupName} name={group.name} />}
                             onClick={this.onGroupClick}
+                            style={{ padding: '13px 16px' }}
                         />
-                        <Accordion.Content active={activeGroup === group.name}>
+                        <Accordion.Content
+                            active={activeGroup === group.name}
+                            style={{ paddingTop: 0, paddingLeft: 16 }}
```
qooban, line 87: Looking at the style in `According.Title` and `According.Content` I think that `16` can be extracted to a const.
Also, maybe a reference from where those numbers come from would be useful here.

qooban (commented):
Just one comment.

widgets/common/src/WorkflowsMenu.jsx
kubama: > Looking at the style in `According.Title` and `According.Content` I think that `16` can be extracted to a const.

The two places where `16` is used by no means are required to use the same value. So I'd rather not extract a single constant, as it would indicate that the values need to be in line, which is not the case.

> Also, maybe a reference from where those numbers come from would be useful here.

I chose `16` so the change was visually transparent, but it could be any reasonable value and it would still look fine. So there is no place where this number comes from and with which it has to be in line.

qooban (approved):

PR #1398, opened 2021-05-27
CY-943 Handle ESOCKETTIMEDOUT when proxying requests to the Manager

Gelio:
This PR aims to improve the error message visible in [CY-943](https://cloudifysource.atlassian.net/browse/CY-943) when there is an `ESOCKETTIMEDOUT` error.

I have decided to install the https://github.com/nock/nock library to make testing making external requests easier.

Aside from the change to the proxy's `errorHandler` I have added 3 new tests:
* 2 tests verifying existing error handling options
* 1 test for verifying `ESOCKETTIMEDOUT`

## Addressing CY-943

The root cause for CY-943 is the fact that the Manager takes too long when sending a response for the Deployment Update request. I have reported RD-2497 to make the API asynchronous and return immediately, without possibilities for waiting indefinitely for a response (which triggers timeouts).

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/699/pipeline

Queued 2021-06-01 14:55 CEST


kubama (approved):

qooban (approved):
LGTM 👍🏻

Gelio:
Thanks for quick reviews :+1: I'll wait for the results of the system tests before merging :slightly_smiling_face:

Gelio:
Looks like failures in the system tests again :worried: I can't run them myself because Rackspace is down :worried: I will wait with merging this PR, as I don't see similar failures on another System tests run https://jenkins.cloudify.co/job/Stage-UI-System-Test/672/

qooban:
@Gelio

I did a quick check and it seems that in both runs, in the maintenance mode widget test, some of the executions were still in progress:
image

image
which I suspect caused maintenance mode to not be deactivated.

I think we need to improve our test to allow switching (actually asking to switch) to maintenance mode only if there are no running executions. That way we'll always be able to deactivate maintenance mode. Or we should always (by REST call) deactivate maintenance mode and wait for it to be deactivated before we start any test. 

I have described my observations and proposal for the solution here: RD-2501 .
I think you can try re-running test again or implement RD-2501 (potential fix should not be very difficult to apply).

Gelio:
@qooban Thanks for doing the work for me :smile: I appreciate it. I was about to begin investigating.

I think waiting for all executions to finish inside of the test makes sense. I'll try that approach first and see if it fixes the problem.

Gelio:
With #1401 merged I can finally rerun the system tests:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/699/pipeline

If they pass, I'll be merging the PR :sligh

Gelio:
The tests passed, so I'm finally merging the PR :smile: :tada:

PR #1400, opened 2021-05-28
RD-1755 Fix Cluster Status widget test to work properly without retries

qooban:
### Experiments

1# Test: https://jenkins.cloudify.co/job/Stage-UI-System-Test/675/
* default timeout: 30s (instead of 90s)
* no. of retries: 1 (instead of 2)
* results: 
  * ✔  All specs passed!
  * `cluster_status_spec` - 1 retry

2# Test: https://jenkins.cloudify.co/job/Stage-UI-System-Test/720/
* default timeout: 10s
* no. of retries: 1
* results: 4 unexpected failures (I'm not counting 2 known failures - RD-2564)
  * User flow installs deployment from scratch.installs deployment from scratch
  * Snapshots list widget should allow to upload and restore a snapshot.should allow to upload and restore a snapshot
  * Snapshots list widget should allow to create and delete a snapshot.should allow to create and delete a snapshot
  * Deployments View widget drill-down functionality should allow deleting a deployment without redirecting to the parent page.should allow deleting a deployment without redirecting to the parent page

3# Test: https://jenkins.cloudify.co/job/Stage-UI-System-Test/721
* default timeout: 30s
* no. of retries: 0
* Improved Cluster Status Widget test which always required 1 retry (I added additional checks which removed the problem with triggering events on removed DOM elements, ref.: https://docs.cypress.io/guides/references/error-messages#cy-failed-because-the-element-you-are-chaining-off-of-has-become-detached-or-removed-from-the-dom)
* results: No unexpected failures (only 2 known failures - RD-2564)

I did much more tests, but sometimes I had to retry it due to CI env issues or the fact that job history is quite short and sometimes get deleted to free disk space.

### Conclusions

* change default timeout from 90s to 30s (it could probably be lower, but I think such decrease is enough for now)
* remove test retries (it can be turned on if necessary only on specific tests if we see that there is real need)

### To do

- [ ] Update cloudify-ui-common after merging https://github.com/cloudify-cosmo/cloudify-ui-common/pull/61 and releasing new version


test/cypress/integration/widgets/cluster_status_spec.js
```diff
@@ -30,9 +30,21 @@ describe('Cluster Status widget', () => {
             manager: 2,
             broker: 3
         };
+        const nodeName = 'cloudify';
+        const privateIp = '10.110.0.39';
+        const version = '5.2.0';
 
         const checkService = (service, nodeStatus, nodeServices) => {
-            const nodeStatusIcon = `tbody tr:nth-child(${rowNumberAIO[service]}) > td:nth-child(3) > i.icon`;
```
Gelio, line 35: Just by reading the code I did not understand what was wrong :worried: Can you shed some light on the problem?

As I see it, the new test code tests more than the previous version and the new assertions are a superset of the old ones. How is it more reliable? Is it because the rows change order during the test?

test/cypress/integration/widgets/cluster_status_spec.js
```diff
@@ -30,9 +30,21 @@ describe('Cluster Status widget', () => {
             manager: 2,
             broker: 3
         };
+        const nodeName = 'cloudify';
+        const privateIp = '10.110.0.39';
+        const version = '5.2.0';
```
Gelio, line 35: Just to confirm - we expect all of them to be the same for all rows?

Gelio (approved):
I'm wondering... how about lowering the timeout to 10s and increasing the timeouts for just specific tests that fail? If we run e.g. 3 Stage test runs and they pass, maybe we could have the confidence to leave such a low timeout.

This is just a suggestion. IMO 30s is good enough for this PR, but maybe we could get it even lower and reap more benefits :slightly_smiling_face:

test/cypress/integration/widgets/cluster_status_spec.js
qooban: Yes, they are the same for all rows. All services have the same node name, private IP and version. See: https://github.com/cloudify-cosmo/cloudify-stage/blob/master/test/cypress/fixtures/cluster_status/fail_with_services.json.

test/cypress/integration/widgets/cluster_status_spec.js
qooban: Yeah, I should have added some description about that :-) 

The main reason for the failure was that some of the Cypress commands were executed in between cluster status table re-rendering (which was triggered by the cluster status data fetch). It happened, for example, that from that chain:
```
cy.get(nodeStatusIcon).trigger('mouseover');
```
`get` worked fine, but the `trigger` not and the error described in:
https://docs.cypress.io/guides/references/error-messages#cy-failed-because-the-element-you-are-chaining-off-of-has-become-detached-or-removed-from-the-dom was thrown.

To make it less error prone I added more checks (node name, private IP, version) before interacting with the icon. That way we are sure that the table got fully re-rendered (as the new data is visible) as added texts are quite specific and unlikely to be present in the real environment.

qooban:
> I'm wondering... how about lowering the timeout to 10s and increasing the timeouts for just specific tests that fail? If we run e.g. 3 Stage test runs and they pass, maybe we could have the confidence to leave such a low timeout.
> This is just a suggestion. IMO 30s is good enough for this PR, but maybe we could get it even lower and reap more benefits 

That's good suggestion. I'll try it tomorrow within another PR and we'll see. 
In that case I'll merge this PR and will update `cloudify-ui-common` in the next one.

General problem with those experiments is that for the test to be reliable I would need to run many tests (but as we now with the current environment it's quite hard to get reliable results). The second thing is that I'm a bit afraid that if we get too low with the timeout, we would observe other tests failing in the future (due to e.g. different network conditions).

test/cypress/integration/widgets/cluster_status_spec.js
Gelio: Oh, cool, thanks :+1:

test/cypress/integration/widgets/cluster_status_spec.js
Gelio: Makes sense :+1: Thanks for explaining

Gelio:
I see, sure, my suggestion to get it lower is optional :slightly_smiling_face: I appreciate effort, but not if you think it's wasted time :slightly_smiling_face:

PR #1401, opened 2021-05-28
RD-2501 Wait for executions to finish before activating maintenance mode

Gelio:
Modify the system tests for maintenance mode to wait for executions to finish before activating maintenance mode. Otherwise, the test fails because activating the maintenance mode takes time (the Manager waits for executions to finish first) which causes cascading failures in other tests since the maintenance mode is activated.

I have verified locally that the test actually waits for the execution to finish (I have run a restart workflow for some deployment when the maintenance mode test was running).

image

Moreover, I have implemented the changes mentioned in https://github.com/cloudify-cosmo/cloudify-stage/pull/1401#issuecomment-850403788:

image

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/690/pipeline

Queued 2021-05-31 12:04 CEST


test/cypress/integration/widgets/maintenance_mode_button_spec.ts
```diff
@@ -1,7 +1,12 @@
+import { waitUntilEmpty } from '../../support/resource_commons';
+
 describe('Maintenance mode button widget', () => {
     before(() => cy.activate('valid_trial_license').usePageMock('maintenanceModeButton').mockLogin());
 
     it('should enter maintenance mode on click', () => {
+        waitUntilEmpty(
+            'executions?status=scheduled&status=queued&status=pending&status=started&status=cancelling&status=force_cancelling&status=kill_cancelling'
```
Gelio, line 8: The statuses are combined in a very clumsy way in the UI:

https://github.com/cloudify-cosmo/cloudify-stage/blob/d7ce26eaf83d26555a5a8de46bc6b0742ad02629/app/actions/managers.js#L176-L179

and 

https://github.com/cloudify-cosmo/cloudify-stage/blob/d7ce26eaf83d26555a5a8de46bc6b0742ad02629/app/utils/shared/ExecutionUtils.js#L73-L80

So I just copied the exact parameters that the maintenance mode button widget uses when sending the request

qooban: Optional:

If you'd like to you can also experiment, you can check `ended_at` field from `executions` response. 
I'd expect that this field is only set for finished executions, so maybe something like:
`executions?ended_at='` will work as well?

qooban (approved):

kubama:
Don't you think that this issue should be rather fixed at the button implementation level? Looking at the failing test recording it looks like there is a modal showing executions. It shows one pending execution, but the action is still available, there is also no any kind of warning. Confirming the activation in the modal takes the user back to the application where activation is still available as an action, there is no information that it was scheduled, the impression is rather that it didn't actually work.

qooban:
@kubama 
The action should be still available IMHO (as the API allows it and it is desired behavior = activate maintenance mode once all executions are finished). 
Additional message in the modal (when there are running executions) - that's probably a good idea.

kubama:
> Additional message in the modal (when there are running executions) - that's probably a good idea.

What about the state of the maintenance button? It allows for the activation to be submitted again, suggesting that the action didn't actually work, as nothing actually changes after returning to the UI.

qooban:
> What about the state of the maintenance button? It allows for the activation to be submitted again, suggesting that the action didn't actually work, as nothing actually changes after returning to the UI.

Good point. Probably we can check `GET /maintenance` result and set `activate` prop on  `MaintenanceModeButton` component properly (instead of setting it always to true).

Gelio:
So, to sum up, I will:

1. Add a message in the UI saying that the Manager will go into maintenance mode as soon as the executions are done, in case there are some active executions.
2. Change the  `MaintenanceModeButton` based on the state of `/maintenance`. Looking at the available information 
image I'm planning to:
    * set the `activate` prop based on `status`
    * disable the button if the activation was requested, but the maintenance mode is not activated yet (with a tooltip saying that activating maintenance mode is in progress)

Does that suit you both :slightly_smiling_face: ? @qooban @kubama

qooban:
@Gelio Yes, sound good to me 👍🏻

kubama:
Sounds great 🙂

test/cypress/integration/widgets/maintenance_mode_button_spec.ts
Gelio: Unfortunately, that got me a 500 error :slightly_frowning_face: 

image

Gelio:
I have implemented just as I said above :slightly_smiling_face: I'm sorry for a single large commit, I wanted to do it today and I don't have much time left to divide it.

System tests:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/681/pipeline

They pass locally:
image

app/components/shared/MaintenanceModeModal.jsx
```diff
@@ -156,121 +161,92 @@ function MaintenanceModeModal({
...
-                            />
-                            <DataTable.Column
-                                label={i18n.t('maintenanceMode.confirmModal.executions.status', 'Status')}
-                                width="15%"
-                            />
-                            <DataTable.Column
-                                label={i18n.t('maintenanceMode.confirmModal.executions.action', 'Action')}
-                            />
+                        <>
+                            <Message content={tConfirmModal('activeExecutionsDelay')} info />
```
Gelio, line 174: This should be the only functional change in this file

The rest are just extracting translation path prefixes

app/utils/StageAPI.ts
```diff
@@ -119,7 +119,7 @@ interface CommonWidgetDefinition<Params, Data, Configuration> {
     permission: Stage.Types.ObjectKeys<typeof GenericConfigType['CUSTOM_WIDGET_PERMISSIONS']> | string;
     showBorder: boolean;
     showHeader: boolean;
-    supportedEditions: string[];
+    supportedEditions?: string[];
```
Gelio, line 122: I have made these optional as it turns out some widgets (e.g. the maintenance mode button widget) do not define them and it works just fine. I did not want to define those props in that widget, so I made the properties optional. I may as well do it the other way - keep them required and define them on the widget. Let me know if you'd prefer the other option :slightly_smiling_face:

test/cypress/integration/widgets/maintenance_mode_button_spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    it('should enter maintenance mode on click', () => {
+        waitUntilEmpty(
+            'executions?status=scheduled&status=queued&status=pending&status=started&status=cancelling&status=force_cancelling&status=kill_cancelling'
+        );
+        getActivateMaintenanceModeButton().click();
+        cy.contains('Yes').click();
+        getDeactivateMaintenanceModeButton().click();
+        cy.contains('Yes').click();
+
+        cy.location('pathname').should('be.equal', '/console/');
```
Gelio, line 19: This part remained unchanged from the initial test implementation. The only change is extracting the button getter functions.

All other code in this file is new :slightly_smiling_face:

Gelio:
Looks like there is a single failure in the filters spec. I doubt it is related to this PR, IMO it's a flake - to be verified on Monday :slightly_smiling_face: I'm not planning to rerun the system tests, as all other tests passed, so I'm waiting for reviews and merging the PR should there be no comment :slightly_smiling_face:

app/components/shared/MaintenanceModeModal.jsx
```diff
@@ -28,6 +30,9 @@ import ExecutionStatus from './ExecutionStatus';
 
 const POLLING_INTERVAL = 2000;
 
+const tConfirmModal = (suffix, options) => i18n.t(`maintenanceMode.confirmModal.${suffix}`, options);
```
kubama, line 33: You could use `Stage.Utils.getT` here

app/components/shared/MaintenanceModeModal.jsx
```diff
@@ -28,6 +30,9 @@ import ExecutionStatus from './ExecutionStatus';
...
+const tExecutions = (suffix, options) => tConfirmModal(`executions.${suffix}`, options);
```
kubama, line 34: Optional: I would consider extending the `Stage.Utils.getT` function so it optionally accepted an existing 't' function, so it could be used here like this: `Stage.Utils.getT(tConfirmModal, 'executions')`

app/components/shared/MaintenanceModeModal.jsx
```diff
@@ -156,121 +161,92 @@ function MaintenanceModeModal({
...
-                    ? i18n.t(
-                          'maintenanceMode.confirmModal.header.activate',
-                          'Are you sure you want to enter maintenance mode?'
-                      )
-                    : i18n.t(
-                          'maintenanceMode.confirmModal.header.deactivate',
-                          'Are you sure you want to exit maintenance mode?'
-                      )}
+                    ? tConfirmModal('header.activate', 'Are you sure you want to enter maintenance mode?')
+                    : tConfirmModal('header.deactivate', 'Are you sure you want to exit maintenance mode?')}
```
kubama, line 165: I guess default strings (2nd argument) could be safely removed

app/components/shared/MaintenanceModeModal.jsx
Gelio: Good idea :+1: I'll use it

app/components/shared/MaintenanceModeModal.jsx
Gelio: Definitely, I'll try to make it work

app/components/shared/MaintenanceModeModal.jsx
Gelio: Done :slightly_smiling_face:

app/components/shared/MaintenanceModeModal.jsx
Gelio: I ended up adding a new `composeT` function instead of overriding `getT` to keep the complexity low :slightly_smiling_face:

app/components/shared/MaintenanceModeModal.jsx
Gelio: Done :slightly_smiling_face:

test/cypress/integration/widgets/maintenance_mode_button_spec.ts
```diff
@@ -0,0 +1,90 @@
...
+                        status_display: 'completed',
+                        tenant_name: 'default_tenant',
+                        created_by: 'admin',
+                        resource_availability: 'tenant',
+                        private_resource: false
+                    }
+                ],
+                metadata: {
+                    pagination: {
+                        total: 13,
```
kubama, line 62: Shouldn't it be 1?

Gelio: Yup, it should be :smile: I forgot to change it. Thanks for catching

test/cypress/integration/widgets/maintenance_mode_button_spec.ts
```diff
@@ -0,0 +1,90 @@
+import { waitUntilEmpty } from '../../support/resource_commons';
+
+describe('Maintenance mode button widget', () => {
+    before(() => cy.activate('valid_trial_license'));
+    beforeEach(() => cy.usePageMock('maintenanceModeButton', { pollingTime: 2 }).mockLogin());
+
+    const getActivateMaintenanceModeButton = () => cy.contains('Activate Maintenance Mode');
+    const getDeactivateMaintenanceModeButton = () => cy.contains('Deactivate Maintenance Mode');
```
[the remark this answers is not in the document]
Gelio: Good idea, done :+1:

test/cypress/integration/widgets/maintenance_mode_button_spec.ts
Gelio: Fixed

kubama (approved):

app/utils/StageAPI.ts
qooban: Let's keep it optional. I assume that `normalizeWidgetDefinition` function should do the work (add default values for not defined properties). 

So does `StageAPI.ts` is now aligned with `normalizeWidgetDefinition`? (I didn't go over the props one by one)

test/cypress/integration/widgets/maintenance_mode_button_spec.ts
qooban: I see. `ended_at` support is probably missing in the REST API. Let's leave it then. 

Optionally you can create a ticket for that or just ask sb from the core team. It's just a suggestion.

qooban (dismissed):

app/utils/StageAPI.ts
Gelio: I checked the properties and looks like there are 8 properties that are required in the interface but have default values in `normalizeWidgetDefinition` (https://github.com/cloudify-cosmo/cloudify-stage/blob/93e565b5857bfafcc23b7dd580e47ef3e9b3ca37/app/utils/normalizeWidgetDefinition.ts#L19-L33)

It will be faster for me to make them optional right away than to create a ticket for it, so I'll append a single commit with this change to this PR :slightly_smiling_face: Good idea to check it, btw

test/cypress/integration/widgets/maintenance_mode_button_spec.ts
Gelio: Having support for filtering by `ended_at` (which is a time-related value) would probably require a new way of filtering. Just filtering `ended_at=some timestamp` does not sound that useful to me :smile: To provide some real benefits for the user, we would probably need an API to allow `ended_at > X` and all other logical operators (`<`, `>=`, etc), and I believe that cannot be easily done by just slapping `ended_at` support in the URL params as we do for other attributes

I agree it would be nice to have such support, but if there are bigger fish to fry when it comes to improvements, I'd rather tackle those and not even add for work the core team :slightly_smiling_face: 

Not having full logical operators support for the `ended_at` and just supporting the equals operator would only help in 2 cases as I see (when we would query for `ended_at=null`), so IMO, the improvement it rather small :slightly_smiling_face: 

As it's optional, I hope we can move on :slightly_smiling_face:

app/utils/StageAPI.ts
Gelio: Nevermind again, the ticket I wanted to create is already implemented. I forgot about that other type I created. Anyway, I need to make some small changes, so there'll be likely a commit :slightly_smiling_face:

Gelio: Done, I've reverted the changes to `StageAPI` and used the correct type for defining the widget configuration (one that already knows which fields are optional as they are later normalized) :slightly_smiling_face:

Gelio:
1 small TS-only commit to revert unnecessary changes done earlier in this PR :slightly_smiling_face: 

I'm not rerunning the system tests, since the changes are TS only and are validated by this PR's CI :slightly_smiling_face: The previous test run passed, so the PR should be good to go

kubama (approved):

qooban (approved):

PR #1402, opened 2021-05-28
RD-1075 Refactored External class interface

kubama:
Although this PR is huge in terms of number of changed files it should be relatively easy and fast to review. The core is the changed interface of the `External` class. Changes in 56 files are just alignments to the new interface.

An additional change is the migration of `External` and `Internal` classes to TS.

System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/689


app/utils/External.ts
```diff
@@ -257,21 +279,18 @@ export default class External {
     }
 
     // eslint-disable-next-line class-methods-use-this
-    buildActualUrl(url, data) {
+    buildActualUrl(url: string, data?: Record<string, any>) {
+        // TODO: RD-258
+        // @ts-ignore Cannot find $
```
Gelio, line 284: Optional:

We could install the types for jquery (`npm i @types/jquery -D`) to fix the problem here :slightly_smiling_face:

app/utils/External.ts
```diff
@@ -212,16 +234,16 @@ export default class External {
     }
 
     // eslint-disable-next-line class-methods-use-this
-    isUnauthorized() {
+    isUnauthorized(_response: Response) {
```
Gelio, line 237: Minor:

I would appreciate having a comment explaining that the parameter is used in overrides and that is why it's declared here :smile: I would prefer clearly stating that than having to wonder why it's here

The same applies to `isLicenseError`

app/utils/External.ts
```diff
@@ -212,16 +234,16 @@ export default class External {
...
         return false;
     }
 
     // eslint-disable-next-line class-methods-use-this
-    isLicenseError() {
+    isLicenseError(_response: Response, _body: any) {
```
Gelio, line 242: Using the  `unknown` type for the body would probably be safer :slightly_smiling_face:

app/utils/Internal.ts
```diff
@@ -9,33 +9,34 @@ import Consts from './consts';
 import StageUtils from './stageUtils';
 
 export default class Internal extends External {
+    constructor(managerData: any) {
+        super(managerData);
+    }
```
Gelio, line 14: Is this required? If I'm not mistaken, this is the default constructor

app/utils/External.ts
```diff
@@ -74,7 +94,7 @@ export default class External {
                     if (response.message) {
                         reject({ message: StageUtils.resolveMessage(response.message) });
                     } else {
-                        reject({ message: e.message });
+                        reject({ message: undefined });
```
Gelio, line 97: What's the reason for this change?

app/utils/External.ts
```diff
@@ -62,9 +83,8 @@ export default class External {
             // Call upload method
             const xhr = new XMLHttpRequest();
             (xhr.upload || xhr).addEventListener('progress', e => {
-                const done = e.position || e.loaded;
-                const total = e.totalSize || e.total;
```
Gelio, line 66: We are no longer taking into account `position` and `totalSize`. What's the reason?

app/utils/External.ts
```diff
@@ -212,16 +234,16 @@ export default class External {
...
     }
 
     // eslint-disable-next-line class-methods-use-this
-    isLicenseError() {
+    isLicenseError(_response: Response, _body: any) {
         return false;
     }
 
-    checkStatus(response) {
+    checkStatus(response: Response) {
```
Gelio, line 246: Probably this method can be marked as `protected` or `private` to limit the visibility. What do you think?

Similar modifiers can be added to other methods too (`isUnauthorized`, `isLicenseError`, `checkStatus`,  maybe `ajaxCall`) :smile:

app/utils/Internal.ts
```diff
@@ -9,33 +9,34 @@ import Consts from './consts';
...
+        const headers = {
+            tenant: _.get(this.managerData, 'tenants.selected', Consts.DEFAULT_TENANT),
+            'Authentication-Token': Cookies.get(Consts.TOKEN_COOKIE_NAME) || undefined
+        };
 
         return headers;
     }
 
-    buildActualUrl(path, data) {
+    buildActualUrl(path: string, data: Record<string, any>) {
```
Gelio, line 29: Too bad we don't have TS 4.3 yet :crying_cat_face: If we did, we could mark this as an override :smile: https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/#override

app/utils/External.ts
```diff
@@ -20,40 +20,61 @@ Text form of class hierarchy diagram to be used at: https://yuml.me/diagram/nofu
...
 
 export default class External {
-    constructor(data) {
-        this.data = data;
-    }
+    constructor(protected managerData: any) {}
 
-    doGet(url, params, parseResponse, headers) {
-        return this.ajaxCall(url, 'get', params, null, parseResponse, headers);
+    doGet(url: string, requestOptions?: RequestOptions) {
```
Gelio, line 38: We should probably omit the `data` property from `RequestOptions`. Not sure about the `fileName`

app/utils/External.ts
```diff
@@ -20,40 +20,61 @@ Text form of class hierarchy diagram to be used at: https://yuml.me/diagram/nofu
 
 */
 
-function getContentType(type) {
+interface RequestOptions {
+    params?: Record<string, any>;
+    data?: any;
+    parseResponse?: boolean;
+    fileName?: string;
```
Gelio, line 27: Is this property used only for `doDownload`? If yes, can we remove it from here and only add it (e.g. using `&`) to the params for `doDownload`?

app/utils/External.ts
```diff
@@ -20,40 +20,61 @@ Text form of class hierarchy diagram to be used at: https://yuml.me/diagram/nofu
 
 */
 
-function getContentType(type) {
+interface RequestOptions {
+    params?: Record<string, any>;
+    data?: any;
```
Gelio, line 25: How about using `body` as the name? `body` is a bit more specific than `data` IMO :slightly_smiling_face: 

```suggestion
    body?: any;
```

Gelio (commented):
This new API looks much better :smile: I like that we don't have to be as verbose

This is a partial review. I did not review the changes inside the `widgets` directory

For the future, IMO it would be a bit easier to review if we separated the refactoring into 2 commits:
1. Change the API
2. Refactor the usage
I am aware that then, the first commit leaves the UI in a non-functional state, but I would prefer that tradeoff :slightly_smiling_face:

.size-limit
```diff
@@ -1,7 +1,7 @@
 [
     {
         "name": "Main JS bundle",
-        "limit": "332.05 KB",
+        "limit": "332.4 KB",
```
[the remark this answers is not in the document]
kubama: The size increased by 0.52KB

app/utils/External.ts
kubama: Done

app/utils/External.ts
kubama: Done

app/utils/External.ts
kubama: Done

app/utils/Internal.ts
kubama: It's not required - removed

app/utils/External.ts
kubama: There is no `message` field on the corresponding type, so it was giving TS error. If so my understanding is that `e.message` must have been always resolved to `undefined`, thus the change.

app/utils/External.ts
kubama: They are not in the type. Looks like a legacy API.

app/utils/External.ts
kubama: Done

app/utils/External.ts
kubama: Done

app/utils/External.ts
kubama: Done

app/utils/External.ts
kubama: I named it after the pre-existing parameter name, but you are right - updated

app/utils/External.ts
kubama: Unfortunately it doesn't seem to be working - I reverted the change.
I didn't want to investigate why it doesn't work as jQuery is intended to be removed anyway.

Gelio:
@kubama Is the PR ready for a re-review :slightly_smiling_face: ?

kubama:
@Gelio I still need to provide updates to address failing system tests. All your comments are addressed however.

qooban:
@kubama Looking at the last comment, if you don't mind, I'll defer the review until you push the final changes to the system tests.

app/utils/External.ts
Gelio: Sure thing, no worries :+1: Thanks for taking the time to try

app/utils/External.ts
```diff
@@ -35,7 +34,7 @@ function getContentType(type?: string) {
 export default class External {
     constructor(protected managerData: any) {}
 
-    doGet(url: string, requestOptions?: RequestOptions) {
+    doGet(url: string, requestOptions?: Omit<RequestOptions, 'body' | 'fileName'>) {
```
Gelio, line 37: After the latest changes we don't have `fileName` in `RequestOptions`, so we could remove the union :slightly_smiling_face:

app/utils/External.ts
Gelio: I'm not sure we are on the same page, so I'll try to provide some details on TS that hopefully clear things up. If you already knew it - my bad :slightly_smiling_face: 

Just because TypeScript does not know about some property does not mean that the property is not there at runtime. TypeScript is mostly just type annotations and does very little in terms of runtime.

For example, in the following case:
https://www.typescriptlang.org/play?#code/MYewdgzgLgBCBGArGBeGBvAUDHMC2AphBAIYDmBAXDAOQnzA0A02uA1mCAO5gDSBAT2oBGAEwBmTAF8A3JkwAzAK5hgUAJbgYJYMAIAHKAHkkAdXVQAFr049+AgBQJE5qze59B1dDA4f71GBKePAEAE4wUgCUWLgwoJAgADYEAHRJIGROZhbWtp4CUXKsOAkQyWkZWdkuue52gtoQ2mCFqYTE5ARF0vI6eoYmtW759jVFQA

If we run the snippet (Ctrl + Enter), we'll see the following:
image

Meaning that even though TypeScript does not know about the `message` property at compile-type, the property may still be there at runtime :slightly_smiling_face: 

It should be clear if we remove the `as any` annotation (meaning there will be a TS error, just as in this case in the PR) and take a look at the compiled JS code:
```js
"use strict";
const obj = {
    message: 'abc',
    knownKey: 123
};
function acceptObjWithKnownKey(objWithKnownKey) {
    console.log(objWithKnownKey);
    console.log(objWithKnownKey.message);
}
acceptObjWithKnownKey(obj);
```

As you can see, even though TS emits an error as it does not know about the `message` property, the resulting code still tries to access such property :slightly_smiling_face: 

Thus, with this change in the PR, as well as https://github.com/cloudify-cosmo/cloudify-stage/pull/1402#discussion_r642411760, unless we have tested every possibility of running this code, there may be a scenario when there is a `message` property in runtime, it's just that we did not tell TypeScript that it is possible :slightly_smiling_face: So, I would be careful when making such changes. This change is the same whether done in TypeScript or JavaScript, meaning that we are changing the behavior :slightly_smiling_face: 

With that out of the way, I'll leave it up to you as you know best - whether we are sure that this will always be `undefined` since there is no case in which this property could exist in JS (same for the other comment https://github.com/cloudify-cosmo/cloudify-stage/pull/1402#discussion_r642411760), or we are not sure and we need to tell TS to ignore this problem (with `@ts-expect-error`, probably)

app/utils/External.ts
Gelio: If I'm not mistaken, this is a very similar case to https://github.com/cloudify-cosmo/cloudify-stage/pull/1402#discussion_r643091563, so the resolution for this discussion should be the same, whichever option you decide in https://github.com/cloudify-cosmo/cloudify-stage/pull/1402#discussion_r643091563

Gelio (commented):
I have skimmed through the rest of the files (the widgets directory) and did not find anything concerning :slightly_smiling_face: I'll hold on with approving until the tests are fixed

app/utils/External.ts
kubama: I'm almost sure it is legacy API and it's safe to be changed - this function is well covered by our tests. However, revisiting the context this is a listener that logs the upload progress on debug level. As for me, this is not useful at all and I'm tempted to remove this listener at all. This would be 100% safe solution 🙂. What do you think?

app/utils/External.ts
kubama: Done

app/utils/External.ts
kubama: Looks like this line is not covered with our tests. I decided to revert back to the original solution as I'm not 100% sure the change is safe and I don't want to invest time in checking.

kubama:
I believe I addressed system tests failures. Submitted new execution: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/701/

I'm not planning any more changes at this point - please review.

qooban (approved):
Change itself looks good. Good work!

I've got few more general comments:
1. As this is backward incompatible change we should be verbose on what we are doing, so I suggest to communicate that (probably by e-mail) to Alex, Ofer and Michael. Proper information in release notes should be provided. As a user installing new version of Cloudify I would also expect a migration path for my custom widgets, so I think that you can also bring that up in that e-mail. It will probably be very simple, just an example on how to change function calls. 
2. Do you plan to also change `backend/handler/services/ManagerService.js` and `backend/handler/services/RequestService.js`? Those services are exposed in widget backend feature. They were based on External class when they were introduced and I think that it would not be wise to keep them unchanged. Of course it can (probably should) be done within separate PR.
3. Just as a reminder - user documentation should be updated accordingly. Scanning the docs quickly, I see that at least [Widget APIs](https://docs.cloudify.co/latest/developer/writing_widgets/widget-apis) page must be updated. Probably there are more places which should be updated (probably only in that [Writing Widgets](https://docs.cloudify.co/latest/developer/writing_widgets/) section). Documentation changes will be required also when p. 2 will be implemented ([Widget Backend](https://docs.cloudify.co/latest/developer/writing_widgets/widget-backend/) page).

kubama:
Thanks @qooban, I'll address your points.
As for now I'm merging the PR as system tests are all green.

kubama:
@Gelio Looks like I'm missing your approval though 🙂

Gelio:
> @Gelio Looks like I'm missing your approval though

Taking a look right now :eye:

app/utils/External.ts
```diff
@@ -93,7 +93,9 @@ export default class External {
                     if (response.message) {
                         reject({ message: StageUtils.resolveMessage(response.message) });
                     } else {
-                        reject({ message: undefined });
+                        // NOTE: legacy solution carried from JS
+                        // @ts-ignore property message does not exist
```
Gelio, line 97: I don't expect any changes here, but IMO this could be compressed to be more concise and elegant, and also use `@ts-expect-error` to make sure the comment is revisited in case `e.message` becomes valid:

```suggestion
                        // @ts-expect-error legacy solution carried from JS
```

app/utils/External.ts
Gelio: Yeah, I agree, let's remove it :smile: Less code to maintain is always nice

Gelio (dismissed):
I didn't review the zips :smile: Having zips in the repo that include code is IMO not the most maintainable way, but well, what can we do now :man_shrugging:

Gelio (approved):

app/utils/External.ts
kubama: Applied

app/utils/External.ts
kubama: Done

PR #1404, opened 2021-06-01
RD-1295 Add TS project references

Gelio:
This PR adds support for [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) in our TS projects.

The projects structure is as follows:

image

and then there is the main `tsconfig.json` solution project that includes all of the projects. Arrow `A -> B` means that project `A` uses `B` as a reference.

TypeScript compilation results (only declaration files and declaration maps, as well as a `tsbuildinfo` file) are cached in the `tsc-dist` directory. They are not used in any user-facing solution.

## Benefits

1. Errors from unrelated projects can be shown at the same time in webpack/`npm run check-types`

    For example, if there were errors in both `test/jest` and `test/cypress` and `backend`, all those projects do not depend on each other, so they can be type-checked in parallel. Previously, due to using `&&` in the `npm run check-types` command, only errors from a single project were shown.

2. TypeScript detects references across projects.

    For example, when searching for references to some `Stage` API declared in `app`, IDEs should now correctly detect the use of that API in other projects, e.g. `widgets`.

    With project references: 
    image

    Without project references: 
    image

3. Webpack reports up-to-date errors when API in `app` changes.

    If there is an API change in `app`, webpack recompiles the types in `app` and then recompiles `widgets`. Previously, webpack relied on the initial version of `app`, meaning that if you changed some `app` API, errors in `widgets` due do that API change were not reported.

    For example, I replaced `StageUtils.isEmptyWidgetData` with `null`.
    With project references: 
    image

    Without project references, webpack is happy:
    image

4. Less resource usage in IDEs

    IDEs can now use a single TS language server for the main `tsconfig.json` solution file instead of starting up a new TS language server for each subproject. This means that when working in multiple directories (and thus, multiple projects), you will have just a single TS language server and will compile each file only once, instead of N language servers and N compilations.

    With project references (a single TS language server): 
    image

    Without project references (multiple TS servers):
    image

5. Better performance when type-checking (see [Performance](#performance))

## Code changes

Any code changes done in this PR are due to the `declarationMap` setting being enabled in TS projects. Because of it there can be no private types used in public APIs. I had to modify some files that had unusual code structure, e.g. did not export the main component, or used some private variable to extend some global API.

Feel free to comment on specific changes and I will try to provide explanations, but perhaps the fastest way would be to try to revert the change locally and inspect the error from `npm run check-types` :slightly_smiling_face: 

## Not addressed in this PR

I have not addressed the following items in this PR:
1. Configure path aliases to shorten import paths. This would introduce even more changes that are not strictly related to project references. Project references support turned out to be hard enough IMO that configuring import paths can be done in a separate ticket.
    I have modified RD-1724 (an existing ticket for import paths) to capture this requirement
2. Upgrade to TypeScript 4.3. It introduces some improvements for TS project references performance and files size. I will do that in a separate PR, still as a part of RD-1295

## Performance

Performance comparison of type-checking is available here:
https://docs.google.com/spreadsheets/d/1rMQYPDzRhN44GER2TXPcvnBBszUUodZNK-w11IS4jAE/edit#gid=0

Conclusion:
1. Rechecking in webpack may be now slower, since webpack potentially has to recompile 2 projects, not just 1 (see benefit number 3).
2. `npm run check-types` is 50% slower when doing a cold start (without the `tsc-dist` cache) (30s vs 19s).
3. `npm run check-types` is 98% or 50% faster with a warm cache (0.45s when no changes, 8.18s when there are changes, compared to 19s without project references)

I hope all of the above are improved with the TS 4.3 upgrade based on these premises:
1. https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/#tsbuildinfo-is-smol
2. https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/#lazier-incremental

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/714/pipeline

Queued 2021-06-07 9:00 CEST


README.md
```diff
@@ -55,6 +55,27 @@ Changes in the source code shall be loaded to the development version of the app
...
+
+Compilation results are stored in the `tsc-dist` directory. Keep in mind those are only used by the TypeScript compiler
+and are not used in the UI application. Keeping those files serves as a cache to speed up subsequent compilations.
+
+When developing, make sure your IDE TypeScript plugin supports project references and is using the main `tsconfig.json`
+file as the configuration file. If your IDE uses project-specific `tsconfig.json` files
+(e.g. [`app/tsconfig.json`](./app/tsconfig.json) for files in the `app` directory), your IDE will be doing unnecessary
+work compiling subprojects multiple times.
+
+IDEs known to work with project references:
```
Gelio, line 72: Let me know if you see references in WebStorm in all projects when looking for references for some Stage API, e.g. `StageUtils` (if you see their usage in `widgets` too)

If yes, I'll also mention WebStorm here

app/utils/External.ts
```diff
@@ -1,7 +1,5 @@
-/**
- * Created by pposel on 09/02/2017.
- */
-
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
Gelio, line 1: Whoops, looks like I'll have conflicts with #1402 :smile:

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -5,6 +5,7 @@ import type { SystemLabel } from '../../support/deployments';
 
 import { exampleBlueprintUrl } from '../../support/resource_urls';
 import { FilterRuleOperators, FilterRuleType } from '../../../../widgets/common/src/filters/types';
+import '../../../../widgets/common/src/deploymentsView';
```
Gelio, line 8: Unfortunately, TS did not detect those `namespace` extensions, so I had to import manually :worried: 

Same for filters

Gelio:
There are errors in the system tests setup now. I'll check it out shortly

app/utils/External.ts
kubama: Could we leave this file as js to make the merge as easy as possible?

widgets/common/src/filters/AddRuleButton.tsx
```diff
@@ -4,7 +4,7 @@ import { getTranslation } from './common';
 const { Button } = Stage.Basic;
```
kubama, line 4: I guess this line could be now moved to `AddRuleButton`

widgets/common/src/filters/RuleRemoveButton.tsx
```diff
@@ -4,7 +4,7 @@ import { getTranslation } from './common';
 const { Button } = Stage.Basic;
```
kubama, line 4: Could be moved to where it is used?

widgets/common/src/labels/AddButton.tsx
```diff
@@ -3,8 +3,8 @@ import type { ComponentProps, FunctionComponent } from 'react';
 const { Button } = Stage.Basic;
```
kubama, line 3: Move to `LabelAddButton`?

widgets/common/src/labels/LabelErrorPopup.tsx
```diff
@@ -3,7 +3,7 @@ import type { ComponentProps, FunctionComponent } from 'react';
 const { Popup } = Stage.Basic;
```
kubama, line 3: Once again, how about moving it as a result of the change in L6?

app/utils/External.ts
Gelio: Nope, unfortunately, we cannot because TS did not know which function parameters were optional in some cases :worried: 

I believe we should merge #1402 first and then I'll just use your file

widgets/common/src/filters/AddRuleButton.tsx
Gelio: Done :+1:

widgets/common/src/filters/RuleRemoveButton.tsx
Gelio: Done :+1:

widgets/common/src/labels/AddButton.tsx
Gelio: Done :+1:

widgets/common/src/labels/LabelErrorPopup.tsx
Gelio: Done :+1:

Gelio:
I believe I have fixed the system tests - we needed to use `import type` in 2 places that I added in this PR.

Here's a rerun:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/695/pipeline

app/utils/External.ts
kubama: Looking at the changes made I think it could be also done the other way around - I guess changes form #1402 could safely overwrite changes made here.

Gelio: I see, then looks like it does not matter which PR goes in first :slightly_smiling_face: Thanks

kubama (approved):
I don't have knowledge to review the approach or its implementation. I'm assuming it just does its job as there are visible benefits, so I'm happy to approve.

qooban:
I'm sure I won't be able to review it today. I'll try tomorrow morning, but looking at my time-plan I may also leave it to be approved just by @kubama . Hope that's fine with you.

Gelio:
@qooban Sure, no worries, I believe a single approval could be enough, as most of the changes are related to the configuration and then addressing any TS errors the came up

@kubama @qooban Can either of you check if the project still works well in WebStorm, if you can go to references, etc? I don't want to break the development experience for you :smile:

Gelio:
The only failing test seems to be flaky. I have raised RD-2540 to fix it and I'm working on it ATM

README.md
qooban: Yes, it works fine in WebStorm:
image

README.md
```diff
@@ -55,6 +55,27 @@ Changes in the source code shall be loaded to the development version of the app
 - for changes in [app](./app) and [widgets](./widgets) directory you need to reload page to see your updates,
 - for changes in [backend](./backend) directory you don't need to reload page as backend server will automatically be restarted.
 
+## TypeScript support in IDEs
```
qooban, line 58: Digression:

I'm not sure if main `README.md` file is the best place for such description, but right now I don't have better suggestion. Probably we would need to create something new, or add another `md` file in `doc` folder for IDE/TS related descriptions.

package.json
```diff
@@ -22,20 +22,15 @@
     "build:analyse:widgets": "npm run build -- --env.analyse=widgets",
     "build:analyse:widgets-common": "npm run build -- --env.analyse=widgets-common",
     "build:coverage": "cross-env COVERAGE_CHECK=1 npm run build",
-    "check-types": "npm run check-types:backend && npm run check-types:frontend && npm run check-types:widgets && npm run check-types:cypress",
-    "//": "The backend and app directories are type-checked when compiling their respective test projects, so there's no need to type-check them individually",
-    "check-types:backend": "tsc -p backend/test --checkJs false",
-    "check-types:frontend": "tsc -p test/jest --checkJs false",
-    "check-types:widgets": "tsc -p widgets --checkJs false",
-    "check-types:cypress": "tsc -p test/cypress --checkJs false",
+    "check-types": "tsc --build",
```
qooban, line 25: Nice! I like this change 👍🏻

widgets/common/src/filters/RulesForm.tsx
```diff
@@ -46,7 +46,7 @@ function getFilterRule(filterRuleRow: FilterRuleRow): FilterRule {
     return filterRuleRow.rule;
 }
 
-interface RulesFormProps {
+export interface RulesFormProps {
```
qooban, line 49: I checked removing that and type-checking called from webpack devserver did not complain about it, but when I run `npm run check-types` I got errors (the same goes to `LabelsInputProps` and `ValidationErrorPopupProps`).

Is it expected that we will get different output from webpack with type-checking and check-types script?

widgets/common/src/DataTypesButton.jsx
```diff
@@ -1,6 +1,4 @@
-/**
- * Created by jakubniezgoda on 15/04/2019.
- */
+export {};
```
qooban, line 1: Hmmm... Wouldn't it be better to export `DataTypesButton` if the export is necessary?

qooban (commented):
Similarly to Kuba M. I don't have relevant knowledge about the feature to provide the real review. Tested it a bit in WebStorm and seems it works fine. Few comments added.

README.md
Gelio: I can move it to `doc/README.md` if you'd like

widgets/common/src/DataTypesButton.jsx
Gelio: I suppose we could. I added the `export` because the file was not an ES module and TS complained :smile: I'll export `DataTypesButton` :+1:

widgets/common/src/filters/RulesForm.tsx
Gelio: It turns out it was not the case. This error occurred in `npm run check-types` despite having `declarationMap: false`. My other hunch is that declaration errors are not emitted by default due to `declaration: false` in `diagnosticOptions` in https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/5b7c333d91f3eed9fb134ba9b9ed10f52115e7f6/README.md#typescript-options.

I tried setting it to `true` in `webpack.config.json`, but the errors still were not reported by webpack.

It looks like only declaration errors are not reported, as other errors were reported normally, as I have seen. We can always remove `ForkTsCheckerWebpackPlugin` and run `tsc --build --watch` ourselves. It will always report all the errors, and it may be faster (just a thought) :smile: 

I think the issue is rather minor and I don't mind not seeing rare errors in webpack, so I'd rather move on. Let me know if you have other expectations :slightly_smiling_face: 

---

When experimenting, I created a shared `tsconfig.base.json` for the stage repo, which IMO is a good idea, as it reduces the number of lines of code in the configs. I will commit the changes

widgets/common/src/DataTypesButton.jsx
Gelio: Done

README.md
Gelio: Done

README.md
qooban: Let's leave it here for now.

qooban (dismissed):

widgets/common/src/filters/RulesForm.tsx
qooban: Thanks for checking. OK, let's leave it.

Adding shared config - good idea 👍🏻

Gelio:
This PR had conflicts anticipated in https://github.com/cloudify-cosmo/cloudify-stage/pull/1404#discussion_r642871222 since #1402 was merged.

You can see the merge conflicts resolution by running:

```sh
git show e9f1d1bd2382d7faae0f859db4a077227b36499a
```


`npm run check-types` still works after resolving those. However, I'll run a full system test suite just in case :smile: 

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/706/pipeline

qooban (approved):

Gelio:
Looks like this can finally be merged :smile: 

I just got 2 failures in the e2e tests, both in the Getting Started modal, both coming from some incompatibility with some AWS blueprint:

image

:worried:

I don't think it's a blocker for this PR, since it does not seem to be related, but anyway, that should probably be fixed soon. I'll create a bug for that and post it on Slack, and I'll merge the PR in the meantime

PR #1409, opened 2021-06-01
RD-2540 Fix race condition in filter_spec

Gelio:
There was a race condition between refetching the list of available
blueprints in the Resource Filter widget and the blueprint being
removed.

Removing the blueprint could have happened after the Resource Filter
widget's dropdown is opened, meaning that the test will fail as the
blueprint will supposedly still be in the system.

## System tests

Only a single test file is changed, so I'm running the tests just for that file:

https://jenkins.cloudify.co/job/Stage-UI-System-Test/698/

Queued 2021-06-01 14:53 CEST

### DIFF
--- a/test/cypress/integration/widgets/filter_spec.js
+++ b/test/cypress/integration/widgets/filter_spec.js
@@ -81,15 +81,15 @@ describe('Filter', () => {
         });
 
         it('blueprint upload and removal', () => {
-            cy.get('.blueprintsWidget').within(() => {
-                cy.getSearchInput().scrollIntoView().clear().type(blueprintName);
-                cy.get(`.${blueprintName}`).parent().find('.trash').click();
-            });
+            cy.get('.blueprintsWidget')
+                .as('blueprintsWidget')
+                .within(() => {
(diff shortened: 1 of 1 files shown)


Gelio:
The tests for `filter_spec` passed, so it should be safe to merge. I hope they will be more stable now :slightly_smiling_face:

kubama (commented):
Please see https://cloudifysource.atlassian.net/browse/RD-2453.

As I can see this PR only relaxes the test so the blueprint dropdown is not verified. But actually this is at the core of the test case. I don't think this is the way to go. I believe RD-2453 should be implemented and the test actually enhanced, as proposed in the ticket.

Gelio:
@kubama You are right, I agree. I did not read the test description carefully enough. I'll close this PR and close this ticket as a duplicate. I'm not planning on completing RD-2453 right now, in this case, as it looks to be larger than I anticipated.

Btw. should it not be visible on our known CI failures epic, as it causes tests to be flaky? Maybe then it would receive more attention during sprint planning :slightly_smiling_face:

PR #1415, opened 2021-06-08
RD-2027 Rename app files to TypeScript

Gelio:
This PR renames js/jsx files to ts/tsx and adds [`@ts-nocheck` comments](https://devblogs.microsoft.com/typescript/announcing-typescript-3-7-beta/#ts-nocheck-in-typescript-files) in files that had TS errors (some renamed files were already correct TS - :tada: ). After that was done, there were some ESLint/TS errors that remained, so I had to fix them manually.

I plan on migrating other directories (`test`, `widgets`) in the meantime, but I wanted to run this PR by you to see if the general approach is understood and agreed on before I make the changes across the repo :slightly_smiling_face: 

I believe this PR is best reviewed commit-by-commit, as I tried to separate changes done by tooling from changes done by hand.

## Migration steps

I have used `ts-migrate` to rename the files in the `app` directory:

```sh
npm i ts-migrate -D # need to install as a dev dependency, using npx directly does not work
git reset --hard  # reset package.json
npx ts-migrate rename app
```

Then, to add `// @ts-nocheck` comments, I have used the following [zx](https://github.com/google/zx) script:

```ts
$.verbose = false;
const result = await nothrow($`npx typescript --build`);

const filesWithErrors = [
  ...new Set(
    result.stdout
      .split("\n")
      .filter((line) => !line.startsWith(" ") && line.trim().length > 0)
      .map((line) => line.slice(0, line.indexOf("(")))
  ),
];

const textToPrepend = `// @ts-nocheck File not migrated fully to TS`;

await Promise.all(
  filesWithErrors.map((filePath) =>
    fs
      .readFile(filePath, { encoding: 'utf8' })
      .then((fileContents) =>
        fileContents.startsWith(textToPrepend)
          ? Promise.resolve()
          : fs.writeFile(filePath, `${textToPrepend}\n${fileContents}`)
      )
  )
);
```

I saved that file as `ts-ignore-files.mjs` in a directory outside of the cloudify-stage repo and ran it as follows:

```sh
nvm use lts/* # make sure to use node v14, since it's required by zx
npx zx ../ts-ignore-files.mjs
nvm use # go back to using the node that Stage uses in `.nvmrc`
```

So, the full instructions are:

```sh
# Done only once:
npm i ts-migrate -D
git reset --hard
# Done for each directory to be migrated:
npx ts-migrate rename app
git commit -m "Rename files"
nvm use lts/*
npx zx ../ts-ignore-files.mjs
git commit -m "Add @ts-nocheck comment to TS files"
nvm use
```

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/726/pipeline

Queued 2021-06-08 14:03 CEST


.eslintrc.temp-overrides.json
```diff
@@ -3,7 +3,14 @@
     "rules": {
         // NOTE: disabled during the migration to TypeScript
         "@typescript-eslint/no-explicit-any": "off",
-        "@typescript-eslint/explicit-module-boundary-types": "off"
+        "@typescript-eslint/explicit-module-boundary-types": "off",
+        "@typescript-eslint/ban-ts-comment": [
+            "error",
+            {
+                "ts-ignore": "allow-with-description",
+                "ts-nocheck": "allow-with-description"
```
Gelio, line 11: `ts-nocheck` comments were forbidden before

Gelio:
Strange issues in System tests:
1. Snapshots list widget should allow to create and delete a snapshot
2. Snapshots list widget should allow to upload and restore a snapshot
3. Maintenance mode button widget should enter maintenance mode on click
4. Number of nodes widget "before all" hook for "should display nodes count"

aside from the _getting started_ ones. Let's wait for the results of https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/729/pipeline (a run against the `master` branch) as I suspect those may be due to some backend change

app/components/layout/GridItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+interface GridItemProps {
+    id: string;
+    className?: string;
+    onItemAdded?: (id: string) => void;
+    onItemRemoved?: (id: string) => void;
+    // NOTE: These props are only used outside, in Grid component
+    x?: number;
+    y?: number;
+    width?: number;
+    height?: number;
```
qooban, line 13: I'm wondering would it be possible to remove it completely as they are not used here?

app/utils/scriptLoader.ts
```diff
@@ -1,14 +1,14 @@
 // @ts-nocheck File not migrated fully to TS
-/**
- * Created by kinneretzin on 11/09/2016.
- */
```
qooban, line 3: General:

Your script seems like a good base to do a global removal of those comments 🙂 
I don't expect any changes in this PR. Just sharing a thought.

qooban (approved):
Looks good! I agree with the approach, so feel free to continue.

Nice script. I learned today about `zx` - looks like a cool tool for JS-minded people wanting to do some shell scripting 🙂

app/components/layout/GridItem.tsx
Gelio: Not really, unfortunately. TypeScript prevents passing excessive props to a component, so if those props would not be mentioned in the interface, you could not pass them using the regular `<GridItem x={123} height={5} />` :slightly_frowning_face:

app/utils/scriptLoader.ts
Gelio: Yeah, definitely :smile: Although, it would have to be modified to not rely on the output of `typescript --build`, but instead maybe on `grep` :thinking: Anyway, good idea, we could do it sometime soon, but IMO we may as well do it when modifying files one by one :smile: I don't have a preference

Gelio:
@qooban Thanks for the review :+1: 
> Nice script. I learned today about zx - looks like a cool tool for JS-minded people wanting to do some shell scripting slightly_smiling_face

Yeah, I saw it a month ago somewhere and thought that it would be perfect for medium-complexity scripts :smile: I am not a bash expert, so doing more advanced stuff like parse the file paths from the list of `tsc` errors and then modify them would definitely be out of my reach and I would have to learn, whereas I was able to be quite productive immediately with `zx` :smile: It has some quirks (e.g. it always wants to escape arguments to shell commands, so I had to use the `fs` module instead of e.g. `cat "content" > file`, but overall, it's been cool :+1: 

---

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/730/pipeline/

Failed only with known getting started errors, so looks like the errors reported in https://github.com/cloudify-cosmo/cloudify-stage/pull/1415#issuecomment-856790726 were indeed false-positives :thinking: I'll leave it be, as I am not too keen on investigating it further and would rather focus on known problems :smile: 

With that, I'll be working on merging this PR soon :slightly_smiling_face:

PR #1417, opened 2021-06-09
RD-2582 Exclude `tsc-dist` directory from collecting coverage

Gelio:
Simplify the array of globs to collect coverage from to only include `app` and `widgets` directories.

I believe the previous solution was overly complex and ended up doing the same thing as one I propose, but in a roundabout manner :smile: 

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/737/pipeline

Queued 2021-06-09 9:51 CEST


qooban (approved):
Seems like a good simplification 👍🏻 

Optional: I don't suspect any issues, but if you find some time you can compare the coverage results with the results from before merging #1404. Probably it can be done by simply running `npm run test:frontend:coverage` on two different commits.

Gelio:
> Optional: I don't suspect any issues, but if you find some time you can compare the coverage results with the results from before merging #1404. Probably it can be done by simply running `npm run test:frontend:coverage` on two different commits.

I ran `npm run test:frontend:coverage` but the coverage was not collected at all for some reason :smile: I'll investigate after the system tests pass and report coverage

Gelio:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/737/pipeline passed and I did not see any coverage-related problems there.

Now I will only see that `npm run test:frontend:coverage` is reporting similar values and then merge the PR

Gelio:
`coverage-jest/lcov-report/index.html` content from:

* `master` including `tsc-dist`, but without this PR:
  image

* `master` with `tsc-dist` directory manually removed and without this PR: 
  image
 

* this PR: 
  image

My conclusions are:
* the aim of this PR is met - `tsc-dist` is not taken into account when collecting coverage now
* the coverage for lines and statements was slightly lower when `tsc-dist` was taken into account. This is according to my expectations - there was code that was not covered in that directory :smile: 

With that, the PR is safe to be merged, which I'll do right now

PR #1420, opened 2021-06-10
RD-2453 Fix options list in DynamicDropdown

qooban:
This PR fixes invalid options listing in `DynamicDropdown`. The reason for the problem was that `overrideOptionsAfterFetch` state variable was not working properly. `options` for `DynamicDropdown` were not reset for the fetch triggered by typing in a text (auto-completion). I fixed also a small issue with `VisibilitySensor` not working properly for all cases.

System tests: [Stage-UI-System-Test #745](https://jenkins.cloudify.co/job/Stage-UI-System-Test/745/)


widgets/common/src/DynamicDropdown.tsx
```diff
@@ -149,10 +141,9 @@ export default function DynamicDropdown({
                     }
 
                     setOptions(latestOptions => [
-                        ...(overrideOptionsAfterFetch ? [] : latestOptions),
```
qooban, line 152: Options should be reset only when we are fetching the first page.
`overrideOptionsAfterFetch` is not necessary to detect that case.

widgets/common/src/DynamicDropdown.tsx
```diff
@@ -11,25 +11,18 @@ let instanceCount = 0;
...
     useUpdateEffect(() => {
-        delayedFetchTrigger({ shouldReset: false });
-    }, withoutResetFetchDeps);
-
-    useUpdateEffect(() => {
-        delayedFetchTrigger({ shouldReset: true });
-    }, withResetFetchDeps);
+        delayedFetchTrigger();
+    }, fetchDeps);
 }
```
qooban, line 26: Options reset not necessary anymore, so I simplified `useFetchTrigger` hook.

widgets/common/src/DynamicDropdown.tsx
```diff
@@ -267,7 +253,8 @@ export default function DynamicDropdown({
                                     <Loading message="" />
                                 </VisibilitySensor>
                             ),
-                            key: 'loader'
+                            key: 'loader',
+                            text: searchQuery
```
qooban, line 257: That addition helped detect change when user was typing in a value and allowed `VisibilitySensor` to re-render.

It was removed originally due to unnecessary re-renders, but when we introduced `fetchState` with `useReducer` hook, it is not re-rendering unexpectedly.

test/cypress/integration/widgets/filter_spec.js
```diff
@@ -48,9 +48,10 @@ describe('Filter', () => {
                 .deleteBlueprints(blueprintName)
```
qooban, line 48: All changes in that file were introduced by @kubama on `filter-spec-enhancement` branch (used as a reproduction branch for RD-2453). 

I think they are worth to include.

test/cypress/integration/widgets/filter_spec.ts
```diff
@@ -49,9 +49,10 @@ describe('Filter', () => {
                 .deleteBlueprints(blueprintName)
```
qooban, line 49: Changes in this file were introduced by @kubama on reproduction branch (see RD-2453). I think it is worth to include it.

Gelio:
I'm afraid `widgets/filters_spec.ts` failed in https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/745/pipeline :worried:

widgets/common/src/DynamicDropdown.tsx
```diff
@@ -11,25 +11,18 @@ let instanceCount = 0;
...
+ * @param {function(): void} fetchTrigger function to be called to trigger data fetching,
  * accepts single boolean argument
- * @param {React.DependencyList} withoutResetFetchDeps list of dependencies for delayed `fetchTrigger` call with
-`shouldReset` argument set to false
- * @param {React.DependencyList} withResetFetchDeps list of dependencies for delayed `fetchTrigger` call with
- * `shouldReset` argument set to true
+ * @param {React.DependencyList} fetchDeps list of dependencies for delayed `fetchTrigger` call
  */
-function useFetchTrigger(fetchTrigger, withoutResetFetchDeps, withResetFetchDeps) {
+function useFetchTrigger(fetchTrigger, fetchDeps) {
```
Gelio, line 18: Optional:

Since the file is TSX now, we could move the types to TS type annotations instead of JSDoc, only leaving the descriptions in JSDoc

Gelio (commented):
The changes look sensible, but I did not read deep enough to verify that those do not break anything. I can do that, but that would increase the time of review.

Generally, `DynamicDropdown` seems like one of the most complex components in the app and each time I review it, I have to do a deep-dive to remind myself the implementation :smile: Maybe we could refactor it to use react-query and it's https://react-query.tanstack.com/reference/useInfiniteQuery :thinking: I could create a ticket if you agree to at least try and lower the complexity

Besides, I see that `filters_spec` is not failing :worried: I'll hold on with the approval for now just because of that fact, but otherwise, I would approve :slightly_smiling_face:

widgets/common/src/DynamicDropdown.tsx
qooban: Good idea, I'll do that.

qooban:
> Generally, DynamicDropdown seems like one of the most complex components in the app and each time I review it, I have to do a deep-dive to remind myself the implementation smile Maybe we could refactor it to use react-query and it's https://react-query.tanstack.com/reference/useInfiniteQuery thinking I could create a ticket if you agree to at least try and lower the complexity

I agree totally. It's very hard to work with it. It was complex before I started doing changes there required for Labels and Filters. And now I consider this component as our technical debt. We need to refactor it, so I would appreciate having a ticket for that. I'll probably have some ideas to add there as well - e.g. we need to first clearly list different use cases for that component (it is used in different combinations depending on the options: `multiple`, `allowAdditions`, `search`-related options, etc. and adds on top of the `Dropdown` component functionality). Maybe that will lead us to split it into smaller components.

> Besides, I see that filters_spec is not failing worried I'll hold on with the approval for now just because of that fact, but otherwise, I would approve slightly_smiling_face

Yes, some more rework needs to be done (dropdowns with multiple values allowed are not handled properly after my change). I'm working on the fix.

Gelio:
I'm glad we are on the same page as to the complexity of that component :+1: 

> I would appreciate having a ticket for that

Here it goes: RD-2591

Feel free to add your suggestions, as I only listed using `useInfiniteQuery` as one of the refactoring ideas

widgets/common/src/DynamicDropdown.tsx
qooban: Done.

qooban:
> Here it goes: RD-2591

Thanks. Added point to my TODO list to update it with my ideas.

Regarding the PR itself, I added a quick fix for multiple values selection. And another system test execution started: [Stage-UI-System-Test #748](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/748/pipeline).

qooban:
Only known agent test failure (RD-2589), so at this point PR is ready for the next round.

widgets/common/src/DynamicDropdown.tsx
Gelio: Does it mean that there will always be a new disabled item at the bottom of the dropdown whose content is the same as what the user typed? Just asking

widgets/common/src/DynamicDropdown.tsx
```diff
@@ -95,7 +87,24 @@ export default function DynamicDropdown({
...
+     * elements from `value` prop and elements from `newOptions` argument.
+     *
+     * @param newOptions list of options used for comparison
+     */
+    function getImplicitOptions(newOptions: { [valueProp: string]: string }[]) {
+        const implicitOptions = [];
+        if (!_.isEmpty(value)) {
+            _.castArray(value).forEach(singleValue => {
+                if (!_.find(newOptions, { [valueProp]: singleValue })) {
+                    implicitOptions.push({ [valueProp]: singleValue, implicit: true });
```
Gelio, line 102: Can you tell me what does `implicit` mean in this context? I do not see it being used in the file, so I assume it's used by some underlying component, but I did not find it in Semantic's dropdown nor in `cloudify-ui-components`

Does it mean that this is an option that is known in the API (e.g. a label that is already used in the system)?

Gelio (approved):
If I understand it right, there were 3 problems:

1. `overrideOptionsAfterFetch` was not a great solution
2. Logic now captured by `getImplicitOptions` was previously only executed in 1 case, but not the other one
3. `VisibilitySensor` sometimes did not appear

Do I understand it right?

Again, I did not dive deep into the code to fully understand the fix. The changes look sensible, but I don't have full confidence that this indeed is the fix, so I'll just place trust in you :smile:

qooban:
Thanks for the review. I'll answer all the questions tomorrow.

widgets/common/src/DynamicDropdown.tsx
qooban: It was there in case `fetchState.hasMore` was truthy. It was disabled always.

This sensor - at least how I understand it - is to:
1. Detect that user is scrolling down to get more options
2. Show loader at the bottom of the list of options when next page fetching starts

The problem was that without `text: searchQuery` the sensor did not get re-rendered properly.

I'll add a point about this sensor - to take closer look at it in RD-2591.

widgets/common/src/DynamicDropdown.tsx
qooban: As far as I understand it is used to:
1. indicate that specific option was added implicitly
2. remove such implicitly added options when the value gets cleared, see [L181](https://github.com/cloudify-cosmo/cloudify-stage/blob/4d64e6780264fae9e7bc52cacba8264a85a563b5/widgets/common/src/DynamicDropdown.tsx#L181)

And it is only for `DynamicDropdown` component use, I think there is no support for such field in Semantic's dropdown.

qooban:
> If I understand it right, there were 3 problems:
> 
> 1. `overrideOptionsAfterFetch` was not a great solution
> 2. Logic now captured by `getImplicitOptions` was previously only executed in 1 case, but not the other one
> 3. `VisibilitySensor` sometimes did not appear
> 
> Do I understand it right?

Yes, exactly, those were the problems I detected.

> Again, I did not dive deep into the code to fully understand the fix. The changes look sensible, but I don't have full confidence that this indeed is the fix, so I'll just place trust in you

I must say that I'm not 100% confident as well 🙁 but I believe looking at my manual tests and system test results that `DynamicDropdown` is a bit better now, and with RD-2591 it will get new shape and will be much clearer for all of us.

qooban:
I'm going to merge the PR and keep an eye on uses of `DynamicDropdown` in general and hope we'll get RD-2591 prioritized soon.

widgets/common/src/DynamicDropdown.tsx
Gelio: Ok, thanks!

widgets/common/src/DynamicDropdown.tsx
Gelio: I see. Those `implicit` options are then also a thing that we could consider simplifying :smile:

PR #1426, opened 2021-06-11
RD-2027 Finalize renaming files to TS

Gelio:
This PR disables importing JS files in TS files, which is some automation to prevent us from going back to JS :smile: 

Besides, I have added a script for calculating how many JS/TS/fully migrated TS files are there in the repo.

I have created https://docs.google.com/spreadsheets/d/1pDXaJHCUZWTYbhoHl9pc07VdRwre85Yx7OoRH9xtLzg/edit#gid=0 to track it over time.

## System tests

The only changes are TS config, so if `npm run check-types` passes in Multibranch CI, the PR is good to go. There is no need for full tests.

Just to be sure, I'll run them on Monday, as the logs will probably be erased anyway over the weekend.

### DIFF
--- a/scripts/getTsMigrationProgress.sh
+++ b/scripts/getTsMigrationProgress.sh
@@ -13,7 +13,7 @@ echo "Found $JS_FILES_COUNT JS/JSX files"
 TS_FILES=`find $DIRECTORIES_TO_CHECK -type f \( -name '*.ts' -o -name '*.tsx' \) ! -path '*node_modules*'`
 TS_FILES_COUNT=`echo "$TS_FILES" | wc -l`
 echo "Found $TS_FILES_COUNT TS/TSX files"
-TS_NOCHECK_COUNT=`grep '@ts-nocheck' $TS_FILES | wc -l`
+TS_NOCHECK_COUNT=`grep '@ts-nocheck' $TS_FILES | wc -l || true`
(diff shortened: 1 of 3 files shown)


qooban:
Won't be able to review it today, I'll review it on Monday.

qooban (approved):
Code looks good!

The chart in the spreadsheet also looks good. Seems like the migration pace is ~25%. Nice. Of course we need to be aware that we added a lot of new code in which TS was used from the beginning.

If I understand correctly it takes into account `backend` folder, but we don't have TS support there yet. Right? I'm just wondering what is the migration status in the FE part?

Gelio:
> If I understand correctly it takes into account `backend` folder, but we don't have TS support there yet. Right? I'm just wondering what is the migration status in the FE part?

Yup, that's correct. The `backend` folder only contains JS files, so the 98 files come from there. All other files come from the frontend folders :slightly_smiling_face: I believe this is sufficient to track the migration, but let me know if you'd rather have a strict split between frontend and backend

qooban:
> I believe this is sufficient to track the migration, but let me know if you'd rather have a strict split between frontend and backend

It's perfectly fine. No need to change it.

Gelio:
Great, I will move onto merging this PR in this case :slightly_smiling_face:

PR #1429, opened 2021-06-14
RD-674 Fixed executions widget crash when graph displaying widget exits maximized state

kubama:
In order to fix the issue properly I had to switch the executions table layout from auto to fixed. This is because the graph component requires its width and height to be set with fixed values. These should be set according to its parent size, but for auto layout its the content that drives columns width, so it didn't work properly.

After switching the layout a number of issues occurred as non-breaking table content no longer caused the table to stretch. In order to mitigate the issues I did the following:
- set fixed width for constant width columns (id, actions)
- set word breaking rules for all cells with variable length content (blueprint id, deployment, workflow, creator)
- set minimum table width
- set all percentage width columns to always sum to at least 100% regardless of which columns are actually displayed - this is required as if the columns widths sum up to less than 100% then the fixed width columns (id, actions) are getting stretched, which is not desired

Original look of executions widget as seen on the dashboard page:
image

After the change:
image

In both cases 4 rows are visible, but after the change there is no scrollbar and all columns are visible. Having non-breakable words broken is not ideal, but I consider it a good trade-off in terms if usability.

Original look of executions widget as seen in the deployments view widget:
image

After the change:
image

In this case the Status column went out of the view. On the other hand the row is lower (more rows will fit on the screen) and date columns are easier to read. In this case I consider the change neutral - IMHO pros and cons are balanced.


widgets/executions/src/ExecutionsTable.tsx
```diff
@@ -181,7 +181,7 @@ export default class ExecutionsTable extends React.Component {
                     className="executionsTable"
                     noDataMessage={NO_DATA_MESSAGE}
                 >
-                    <DataTable.Column label="" width="1%" />
+                    <DataTable.Column label="" width="43px" />
```
Gelio, line 184: Are these width changes relevant to the fix? I'm asking out of curiosity

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -41,9 +41,12 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
 
     const timer = useRef(null);
     const cancelablePromise = useRef(null);
-    const wrapper = useRef();
     const modal = useRef();
 
+    const wrapper = useRef();
+    const [wrapperWidth, setWrapperWidth] = useState();
```
Gelio, line 47: Minor:

Since this is a TS file, consider writing valid TS to make the migration easier :slightly_smiling_face: 

```suggestion
    const [wrapperWidth, setWrapperWidth] = useState<number>(null);
```

or something else :smile: 

Same for the `wrapper` above (probably we want to use `HTMLElement`)

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -41,9 +41,12 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
...
+    const wrapperResizeObserver = useRef(new ResizeObserver(() => setWrapperWidth(wrapper.current?.offsetWidth)));
```
Gelio, line 48: This looks like a perfect use-case for `useMemo` with an empty dependency array to cache the value indefinitely :slightly_smiling_face:

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -96,6 +99,15 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
...
     }, [selectedExecution.id]);
 
+    useEffect(() => {
+        const wrapperRef = wrapper.current;
+        if (wrapperRef) {
+            wrapperResizeObserver.current.observe(wrapperRef);
+            return () => wrapperResizeObserver.current.unobserve(wrapperRef);
+        }
+        return undefined;
+    }, [wrapper.current]);
```
Gelio, line 109: Optional:

How about extracting this code along with declaring `wrapperResizeObserver` to a separate hook to encapsulate the logic, name it, and make the component shorter?

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -96,6 +99,15 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
         return stopPolling;
     }, [selectedExecution.id]);
 
+    useEffect(() => {
+        const wrapperRef = wrapper.current;
+        if (wrapperRef) {
```
Gelio, line 104: Minor, nitpick:

Doing an early return saves us indentation on the main logic:

```ts
if (!wrapperRef) {
  return;
}

// main logic is indented less
```

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
Gelio, line 47: I assume that we don't use `wrapperWidth` deliberately. If that's the case, I suggest:

1. Prefixing it with `_` (`_wrapperWidth`
2. Adding an explanation comment why it's done that way (I imagine it is to cause the component to rerender)

Gelio (commented):
I've only reviewed the code in terms of quality. It would be nice to have some video/screenshot of the new UI since the widths changed, and also an explanation what the problem was :slightly_smiling_face: That would make the review much faster for me as I would not have to dive deep to understand the problem myself

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -234,7 +246,7 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
 
             {graphData && (
                 <div ref={wrapper} style={{ position: 'relative' }}>
-                    {renderGraph(Math.max(0, getContainerWidth() - 1), containerHeight, position, setPosition)}
+                    {renderGraph(Math.max(0, getContainerWidth()), containerHeight, position, setPosition)}
```
Gelio, line 249: This `Math.max` is probably not necessary now, since the offset is always >= 0 :slightly_smiling_face:

widgets/executions/src/ExecutionsTable.tsx
kubama: Yes, as the table layout changed from auto to fixed.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: Would there be any benefit over `useRef`?

qooban:
I see @Gelio did a review, so I think I'll leave that for you both to finalize this PR and do not review it by myself (that's due to lack of time).

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
Gelio: We would not create unnecessary `ResizeObserver`s on each render :slightly_smiling_face:

kubama: With `useMemo` we would create new function instance on each render instead, right? I don't think it would be considerably cheaper, if at all.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: Done

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: Done

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: Done

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: Right, removed

kubama:
@Gelio I applied most of your suggestions. However, looks like I need to rethink changing the table layout to fixed, as it's causing layout issues when the widget is narrow, for example when nested in deployments view widget.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
Gelio: Yup, we are creating new functions very often and that seems to be cheap. I am not sure about the cost of instantiating new observers on each render. I don't have hard data, my hunch is that there could be a difference, as it's not as common as functions, but w/e.

The other aspect that I believe is worth here is readability - IMO this is the exact use-case for `useMemo`, as we don't need to change the `.current` anywhere from the component, we just want a value that does not change for the lifetime of the component, kind of like a property initialized in the constructor. Personally, `useMemo` sounds like a more fitting function in this case that conveys more meaning about the intent. `useRef` would be fine to me if we mutated `current`

Since we don't have an agreement, I'll ask a third person to say what they prefer :smile: @qooban, would you see `useRef` or `useMemo` as more fitting in this case?

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -122,17 +112,13 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
...
-    }
-
     function getModalWidth() {
         return _.get(modal.current, 'offsetWidth', 0);
     }
 
     function fitToView() {
         const { width: graphWidth, height: graphHeight } = graphData;
-        const width = isMaximized ? getModalWidth() : Math.max(0, getContainerWidth() - 1);
+        const width = isMaximized ? getModalWidth() : Math.max(0, getWrapperWidth() - 1);
```
Gelio, line 121: Hmm... should we be using `Math.max` and subtracting 0 here too? Just wondering, as it looks similar to the code you removed earlier

Gelio (commented):
The code looks good to me ATM, but looks like you want to do some more experimenting, so I'll leave it without approval yet.

Also, I did not see any changes to the description and I would appreciate if you did fill in the description as I suggested in my previous review summary :slightly_smiling_face: Or at least not leave that main comment unanswered

kubama:
@Gelio Please see the description provided

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: Actually I'm convinced by your points - changed.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: I tested it without Math.max - works perfectly fine, I committed the change.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
qooban: Yeah, using `useMemo` in this case seems reasonable.
I see you have an agreement, so great 👍🏻

Gelio: Yup, I'm glad :+1:

widgets/executions/src/ExecutionsTable.tsx
```diff
@@ -181,58 +181,58 @@ export default class ExecutionsTable extends React.Component {
...
                     <DataTable.Column
                         label="Blueprint"
                         name="blueprint_id"
-                        width="15%"
+                        width="150%"
```
Gelio, line 188: I'm curious, as this does not look normal with columns having > 100% width. I see it still works fine on the screenshots. I have not worked much with styling tables before, so I gotta say - is this a common practice? How does it work that it still renders fine?

widgets/executions/widget.css
```diff
@@ -11,6 +11,11 @@
     left: unset !important;
 }
 
+.executionsTable {
+    table-layout: fixed;
+    min-width: 860px;
```
Gelio, line 16: Minor:

I removed the `min-width` and could not reproduce the bug. This leads me to believe that maybe we do not need it. Did you try without this property?

This way we can get away without a horizontal scrollbar in Deployments View:
image

However, I see that in this case the table headers are overlapping :worried:

Gelio (approved):
Thanks for a very thorough description :+1: It clarifies things that I would not think about :smile: 

Looks good, I have just added 2 comments to get clarification :smile:

widgets/executions/src/ExecutionsTable.tsx
kubama: I'm not sure if it's a common practice, but it's not something out of spec. For sure it would be nicer to have the widths to sum up to 100%, but since individual columns can be shown or hidden it would require some complex calculations depending on which columns are displayed and which are not. This approach eliminates the need for such logic.

Gelio: I see. So, this works similarly to the `flex-grow` property, meaning that all the widths are summed up by the browser and the actual width of this column will be a weighted average (`150% / sum of all %`)?

kubama: Looks like this is the case.

Gelio: Thanks for the answers :+1:

widgets/executions/widget.css
kubama: Yes, I set the min-width only to ensure headers are not overlapping. However, if you think it's worth trying to get rid of the scrollbars we could set it to a lower value and try one of the following:
- set the Attributes column to be wider (not perfect as it would be wider also when the whole table is wider)
- set the Attributes column a fixed width (looking at the code it looks like it can contain only at most two icons, so maybe it could be aligned to that)
- add overflow hidden and text-overflow ellipsis - it would then look like this: 
image

Gelio: Good ideas :+1: The third option speaks best to me :smile: Although I would also suggest adding tooltips/`title`s to those columns so it's clear on hover :slightly_smiling_face: 

This can be done in a separate PR to not prolong this one, though.

Aside from that, I don't really mind the table having the scrollbar, there's little we can do to fit it with limited width. 

However, we could also document why the `min-width` is there - it looks surprising to me and I probably won't remember it after some time :slightly_smiling_face:

kubama: OK, I added comment on `min-width`. I'm not planning any more changes in this PR - please review.

Gelio: Sounds good :+1:

Gelio (approved):

PR #1437, opened 2021-06-16
WDR-44 Added word breaking rule to filter id column

kubama:
This addresses filters table issue reported in WDR-44.

Regarding the reported dropdown issue there is RD-1575 to address it application wide, so I'm not going to fix this as part of this PR.


Gelio (approved):

Gelio:
I did not run it in the UI myself. I assume this actually fixes the problem, the code looks alright :slightly_smiling_face: Seeing a screenshot would provide a better proof, though :smile: Not required, though

PR #1440, opened 2021-06-17
RD-1829 Improve handling of incompatibilities between Resource Filter and Deployments View widgets

Gelio:
This PR implements the new way of interoperability between the Resource Filter and the Deployments View widgets.

The behavior is described in points listed in the following Jira comment: https://cloudifysource.atlassian.net/browse/RD-1829?focusedCommentId=72281. The logic is as follows:

1. If there is no deployment selected (not in Resource Filter or Deployments View), select the first deployment from the table in Deployments View

    1.1. If there are no deployments (“No Results Found”), nothing is selected and there is only that information “No Results Found”. There is no warning now (the warning shown on https://user-images.githubusercontent.com/889383/122370093-1240a000-cf5f-11eb-98bb-e219ea25075d.png was unnecessary)

2. If the user selected a deployment in the Resource Filter widget (no matter whether the user has the “Allow multiple selection” enabled or not)

    2.1. If that deployment is on the current table page, select it

    2.2. If that deployment is on some other table page, select the first deployment in the table and show a warning in the developer console

3. If the user selected multiple deployments in the Resource Filter widget, try to unwrap the selected deployment from the array. If it's not possible, select the first deployment in the table and show a warning in the developer console

To sum up, the Deployments View will override any `deploymentId` picked by the Resource Filter in cases where otherwise there would be an "unknown deployments selected'. There is no user-facing warning, only warnings in the console in those rare scenarios.

Changes to the documentation of the Resource Filter widget are done in https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1880.

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/794/pipeline

Queued 2021-06-17 11:24 CEST

### DIFF
--- a/widgets/common/src/deploymentsView/getSelectedDeployment.ts
+++ b/widgets/common/src/deploymentsView/getSelectedDeployment.ts
@@ -33,12 +33,6 @@ const getSelectedDeployment = (
         id: deploymentIdInContext as string | undefined
     });
 
-    if (!selectedDeployment) {
-        log.warn(
-            'The selected deployment is not visible in the Deployments View table. It will be overridden to the first one in the table'
-        );
-    }
-
     return {
         selectedDeployment,
         fallbackDeployment
--- a/widgets/common/src/deploymentsView/getSelectedDeployment.ts
+++ b/widgets/common/src/deploymentsView/getSelectedDeployment.ts
@@ -1,0 +1,48 @@
+import { find } from 'lodash';
+import type { Deployment } from './types';
+
+interface SelectedDeploymentResult {
+    selectedDeployment: Deployment | undefined;
+    fallbackDeployment: Deployment | undefined;
+}
+
+const getSelectedDeployment = (
+    deploymentIdInContext: Stage.ContextEntries['deploymentId'] | undefined,
+    deployments: Deployment[]
+): SelectedDeploymentResult => {
+    const fallbackDeployment: Deployment | undefined = deployments[0];
+
+    if (Array.isArray(deploymentIdInContext)) {
+        if (deploymentIdInContext.length === 1) {
+            // NOTE: try to work nicely if there is a singleton array
+            return getSelectedDeployment(deploymentIdInContext[0], deployments);
+        }
+
+        log.warn(
+            'The Deployments View widget does not support selecting multiple deployments. ' +
+                'Most likely you are using a Resource Filter widget alongside it and the combination is not supported'
+        );
+
+        return {
+            selectedDeployment: undefined,
+            fallbackDeployment
+        };
+    }
+
+    const selectedDeployment = find(deployments, {
+        id: deploymentIdInContext as string | undefined
+    });
+
+    if (!selectedDeployment) {
+        log.warn(
+            'The selected deployment is not visible in the Deployments View table. It will be overridden to the first one in the table'
+        );
+    }
+
+    return {
+        selectedDeployment,
+        fallbackDeployment
+    };
+};
+
+export default getSelectedDeployment;
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -202,6 +202,12 @@ describe('Deployments View widget', () => {
         });
     });
 
+    it('should select the first visible deployment by default', () => {
+        useDeploymentsViewWidget();
+
+        getDeploymentsViewTable().find('tbody tr:first-of-type.active').should('exist');
+    });
+
     describe('with filters', () => {
         const deploymentNameThatMatchesFilter = `${specPrefix}precious_deployment`;
         const filterId = 'only-precious';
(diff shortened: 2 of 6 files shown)


kubama (dismissed):

widgets/common/src/deploymentsView/getSelectedDeployment.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    const selectedDeployment = find(deployments, {
+        id: deploymentIdInContext as string | undefined
+    });
+
+    if (!selectedDeployment) {
+        log.warn(
+            'The selected deployment is not visible in the Deployments View table. It will be overridden to the first one in the table'
+        );
+    }
```
Gelio, line 40: Maybe it makes sense to move this `if` to the place where `toolbox.setContext('deploymentId', fallbackDeployment)` is called in `DeploymentsView.tsx`. This way it's closer to where the actual overriding happens :thinking:

Gelio:
I have encountered an error in `filters_spec` https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/794/tests:

image

Looks like yet another flaky test :thinking: It comes from the following line of code:
https://github.com/cloudify-cosmo/cloudify-stage/blob/56fc44a5d35497aee62115376b24edf3103e2da2/test/cypress/integration/widgets/filters_spec.ts#L458

I will report a ticket for that. I doubt it's related to this PR, so I assume this PR is mergeable

widgets/common/src/deploymentsView/getSelectedDeployment.ts
```diff
@@ -0,0 +1,48 @@
...
+    fallbackDeployment: Deployment | undefined;
+}
+
+const getSelectedDeployment = (
+    deploymentIdInContext: Stage.ContextEntries['deploymentId'] | undefined,
+    deployments: Deployment[]
+): SelectedDeploymentResult => {
+    const fallbackDeployment: Deployment | undefined = deployments[0];
+
+    if (Array.isArray(deploymentIdInContext)) {
```
qooban, line 15: Optional:

I didn't analyse deeply the code, but maybe we can avoid that if and simplify the code by using `castArray` from `lodash`? Just wondering, I'm not sure.

widgets/common/src/deploymentsView/getSelectedDeployment.ts
qooban: Seems reasonable.

qooban (approved):
Looks good. 

Minor optional comment added.

qooban:
> EDIT: I have reported RD-2632 for that failure. Let's see if it fails in CI too

I'm working on it.

widgets/common/src/deploymentsView/getSelectedDeployment.ts
Gelio: Hmm I don't see how we would then detect if there are multiple items in the array. By doing `castArray(arr)[0]` we are not handling separately the case when there are 2 or more elements.

This existing logic handles that and sets `deploymentId` in the context to be a single, first item, rather than ignore this case.

I like this kind of precision :slightly_smiling_face: At the moment, it does not have any user-facing consequences, since the Resource Filter widget does not read `deploymentId` from the context when it shows what is selected in the _Deployment_ filter :worried: It only reads it initially https://github.com/cloudify-cosmo/cloudify-stage/blob/56fc44a5d35497aee62115376b24edf3103e2da2/widgets/filter/src/Filter.tsx#L20-L23

I suppose if we fixed that widget to not have its own copy of the context, then with the logic I provided, the user would not be allowed to pick more than a single item, which IMO makes sense. I am not creating a ticket for it, since we are doing less and less with that widget, so I don't think it's worth the effort :smile: 

To sum up, not using `castArray` gives us more precision that I like, which is why I would leave the code as-is :slightly_smiling_face:

widgets/common/src/deploymentsView/getSelectedDeployment.ts
Gelio: Thanks for confirming :+1: Done :slightly_smiling_face:

kubama (approved):

widgets/common/src/deploymentsView/getSelectedDeployment.ts
qooban: OK, I'm not arguing, I did not analyse the code deeply 👍🏻

qooban (approved):

widgets/common/src/deploymentsView/getSelectedDeployment.ts
Gelio: No worries :smile: I wanted to provide all the details about why I made this decision so you hopefully have full information :slightly_smiling_face: I can provide less next time if my tone was too argumentative :sweat_smile:

Gelio:
Thanks for re-reviews :+1: I'll merge the PR as soon as CI passes.

I will raise another PR with README updates from https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1880 and merge it after https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1880 is merged :slightly_smiling_face:

widgets/common/src/deploymentsView/getSelectedDeployment.ts
qooban: No, no, all fine 🙂 I appreciate detailed answer 👍🏻

PR #1441, opened 2021-06-17
RD-2626 Improve Filters spec to be more robust by more precise dropdown option selection

qooban:
### Changes

Instead of waiting for the part which is generic, we are now waiting for the exact value in the dropdown option.

### System Tests

Filters spec passed locally two times in a row.

[Stage-UI-System-Test/797](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/797/pipeline) (only Filters spec)
```
[2021-06-17T11:40:19.231Z]        Spec                                              Tests  Passing  Failing  Pending  Skipped  
[2021-06-17T11:40:19.231Z]   ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
[2021-06-17T11:40:19.231Z]   │ ✔  widgets/filters_spec.ts                  01:35       19       19        -        -        - │
[2021-06-17T11:40:19.231Z]   └────────────────────────────────────────────────────────────────────────────────────────────────┘
[2021-06-17T11:40:19.231Z]     ✔  All specs passed!                        01:35       19       19        -        -        -  
```

### DIFF
--- a/test/cypress/integration/widgets/filters_spec.ts
+++ b/test/cypress/integration/widgets/filters_spec.ts
@@ -439,7 +439,7 @@ describe('Filters widget', () => {
                             cy.get('input').type(`${value}`);
                             if (withAutocomplete) cy.wait(`@valueSearch_${value}`);
 
-                            if (isNew) cy.contains('[role="option"]', 'Add ').click();
+                            if (isNew) cy.contains('[role="option"]', `Add ${value}`).click();
                             else cy.get(`div[option-value="${value}"]`).click();
 
                             cy.get(`.label[value="${value}"]`).should('exist');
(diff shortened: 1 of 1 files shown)


kubama (approved):

Gelio (approved):
The change looks good, but do you think this is the root cause? I thought that the problem was with `.within` capturing some DOM node that then is removed, and all assertions fail inside, since that parent node is no longer mounted.

I don't know whether this is the case, I did not check myself, but that's how I interpreted the error :smile: I recently also had a problem with Cypress retrying some command on the wrong parent node instead of retrying multiple commands, that's why I am raising this topic :slightly_smiling_face: 

Anyway, this is a good improvement, so we may as well merge it and not spend more time on it :slightly_smiling_face:

qooban:
> The change looks good, but do you think this is the root cause? I thought that the problem was with .within capturing some DOM node that then is removed, and all assertions fail inside, since that parent node is no longer mounted.

I don't think the problem was with `.within` capturing. Actually looking few times at the video from the failing test I think right now that the problem comes from re-render happened too late after fetch was finished which caused small blink ("New key edition" option disappeared for some short time). So my fix can help (as the text to find is a bit longer), but not necessarily.

> Anyway, this is a good improvement, so we may as well merge it and not spend more time on it slightly_smiling_face

Yeah, that's also my assumption. Let's not spend more time on it and see how it works.

PR #1444, opened 2021-06-18
RD-1769 Added support for new or modified filter

kubama:
image

### DIFF
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -470,9 +470,17 @@ describe('Deployments View widget', () => {
             });
 
             getDeploymentsViewDetailsPane().within(() => {
-                getSubservicesButton().contains('1');
+                getSubservicesButton().within(() => {
+                    cy.containsNumber(1);
+                    cy.get('i[aria-label="Requires attention"]').should('exist');
+                });
                 cy.log('Drill down to subenvironments of app-env');
-                getSubenvironmentsButton().contains('1').click();
+                getSubenvironmentsButton()
+                    .within(() => {
+                        cy.containsNumber(1);
+                        cy.get('i[aria-label="Requires attention"]').should('exist');
+                    })
+                    .click();
             });
 
             getBreadcrumbs().contains('app-env [Environments]');
@@ -489,9 +497,9 @@ describe('Deployments View widget', () => {
             verifySubdeploymentsOfAppEnv();
 
             getDeploymentsViewDetailsPane().within(() => {
-                getSubenvironmentsButton().contains('0').should('be.disabled');
+                getSubenvironmentsButton().containsNumber(0).should('be.disabled');
                 cy.log('Drill down to subservices of db-env');
-                getSubservicesButton().contains('2').click();
+                getSubservicesButton().containsNumber(2).click();
             });
 
             getBreadcrumbs().contains('db-env [Services]');
@@ -503,9 +511,23 @@ describe('Deployments View widget', () => {
                 cy.contains('db-env').should('not.exist');
             });
 
+            const expectOnlySubdeploymentTypeIcon = () => {
+                cy.get('i').should('have.length', 1).not('.object.group.icon, .cube.icon').should('have.length', 0);
+            };
+
             getDeploymentsViewDetailsPane().within(() => {
-                getSubenvironmentsButton().contains('0').should('be.disabled');
-                getSubservicesButton().contains('0').should('be.disabled');
+                getSubenvironmentsButton()
+                    .should('be.disabled')
+                    .within(() => {
+                        cy.containsNumber(0);
+                        expectOnlySubdeploymentTypeIcon();
+                    });
+                getSubservicesButton()
+                    .should('be.disabled')
+                    .within(() => {
+                        cy.containsNumber(0);
+                        expectOnlySubdeploymentTypeIcon();
+                    });
             });
 
             cy.log('Go back to the parent environment');
@@ -516,10 +538,10 @@ describe('Deployments View widget', () => {
             getBreadcrumbs().contains('Test Page').click();
             getDeploymentsViewDetailsPane().within(() => {
                 cy.log('Drill down to subservices of app-env');
-                getSubservicesButton().contains('1').click();
+                getSubservicesButton().containsNumber(1).click();
             });
             getDeploymentsViewTable().within(() => {
-                cy.log('Subservices of app-end should be visible (web-app)');
+                cy.log('Subservices of app-env should be visible (web-app)');
                 cy.contains('web-app');
                 cy.contains('db-env').should('not.exist');
             });
@@ -575,7 +597,7 @@ describe('Deployments View widget', () => {
             cy.getSearchInput().type(deploymentName);
 
             getDeploymentsViewDetailsPane().within(() => {
-                getSubservicesButton().contains('0');
+                getSubservicesButton().containsNumber(0);
 
                 cy.interceptSp('GET', `${deploymentName}?all_sub_deployments=false`, req =>
                     req.reply(res => {
@@ -584,7 +606,7 @@ describe('Deployments View widget', () => {
                 ).as('deploymentDetails');
                 cy.wait('@deploymentDetails');
 
-                getSubservicesButton().contains('50');
+                getSubservicesButton().containsNumber(50);
             });
         });
     });
@@ -711,9 +733,9 @@ describe('Deployments View widget', () => {
                         cy.contains('Cloudify-Hello-World');
                         cy.contains(siteNames.london);
                         cy.contains('In progress').parent().find('i.orange.spinner');
-                        getTooltipSubenvironments().contains('5');
+                        getTooltipSubenvironments().containsNumber(5);
                         getTooltipSubservices().within(() => {
-                            cy.contains('80');
+                            cy.containsNumber(80);
                             cy.get('[aria-label="In progress"]');
                         });
                     }
@@ -727,11 +749,11 @@ describe('Deployments View widget', () => {
                         cy.contains(siteNames.olsztyn);
                         cy.contains('Requires attention').parent().find('i.red.exclamation');
                         getTooltipSubenvironments().within(() => {
-                            cy.contains('1');
+                            cy.containsNumber(1);
                             cy.get('[aria-label="In progress"]');
                         });
                         getTooltipSubservices().within(() => {
-                            cy.contains('3');
+                            cy.containsNumber(3);
                             cy.get('[aria-label="Requires attention"]');
                         });
                     }
@@ -744,8 +766,8 @@ describe('Deployments View widget', () => {
                         cy.contains('Cloudify-Hello-World');
                         cy.contains(siteNames.warsaw);
                         cy.contains('Good');
-                        getTooltipSubenvironments().contains('10');
-                        getTooltipSubservices().contains('8');
+                        getTooltipSubenvironments().containsNumber(10);
+                        getTooltipSubservices().containsNumber(8);
                     }
                 );
             });
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -202,6 +202,12 @@ describe('Deployments View widget', () => {
         });
     });
 
+    it('should select the first visible deployment by default', () => {
+        useDeploymentsViewWidget();
+
+        getDeploymentsViewTable().find('tbody tr:first-of-type.active').should('exist');
+    });
+
     describe('with filters', () => {
         const deploymentNameThatMatchesFilter = `${specPrefix}precious_deployment`;
         const filterId = 'only-precious';
@@ -270,6 +276,18 @@ describe('Deployments View widget', () => {
 
             cy.contains(deploymentName);
             cy.contains(deploymentNameThatMatchesFilter);
+        });
+
+        // NOTE: this test does not really belong to the filters functionality, but it needs at least two deployments to work
+        it('should allow selecting a deployment through the Resource Filter widget', () => {
+            useDeploymentsViewWidget();
+
+            const getSelectedDeployment = () => getDeploymentsViewTable().find('tbody tr.active');
+
+            cy.setDeploymentContext(deploymentName);
+            getSelectedDeployment().contains(deploymentName);
+            cy.setDeploymentContext(deploymentNameThatMatchesFilter);
+            getSelectedDeployment().contains(deploymentNameThatMatchesFilter);
         });
     });
 
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -97,7 +97,8 @@ describe('Deployments View widget', () => {
             cy.contains('button', 'Filter').click();
 
             cy.get('.modal').within(() => {
-                cy.get('input').type(`${filterId}{enter}`);
+                cy.setSearchableDropdownValue('Filter ID', filterId);
+                cy.get('.loader').should('not.exist');
                 cy.contains('OK').click();
             });
 
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -241,7 +241,7 @@ describe('Deployments View widget', () => {
             cy.contains(/with ID .* was not found/);
         });
 
-        it('should take the selected existing filter into account when displaying deployments', () => {
+        it.only('should take the selected existing filter into account when displaying deployments', () => {
             useDeploymentsViewWidget();
 
             cy.contains(deploymentNameThatMatchesFilter);
@@ -252,7 +252,7 @@ describe('Deployments View widget', () => {
 
             cy.get('.modal').within(() => {
                 cy.setSearchableDropdownValue('Filter ID', filterId);
-                cy.get('.loader').should('not.exist');
+                cy.contains('Label').should('be.visible');
                 cy.contains('OK').click();
             });
 
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -208,6 +208,13 @@ describe('Deployments View widget', () => {
             }
         ];
 
+        function verifyFilterRulesForm() {
+            cy.contains('Label').should('be.visible');
+            cy.contains('is one of').should('be.visible');
+            cy.contains('precious');
+            cy.contains('yes');
+        }
+
         before(() => {
             cy.deleteDeploymentsFilter(filterId, { ignoreFailure: true })
                 .createDeploymentsFilter(filterId, filterRules)
@@ -251,7 +258,7 @@ describe('Deployments View widget', () => {
 
             cy.get('.modal').within(() => {
                 cy.setSearchableDropdownValue('Filter ID', filterId);
-                cy.contains('Label').should('be.visible');
+                verifyFilterRulesForm();
                 cy.interceptSp('POST', '/searches/deployments').as('deploymentsSearchRequest');
                 cy.contains('OK').click();
             });
@@ -286,10 +293,7 @@ describe('Deployments View widget', () => {
 
             cy.get('.modal').within(() => {
                 cy.setSearchableDropdownValue('Filter ID', filterId);
-                cy.contains('Label').should('be.visible');
-                cy.contains('is one of').should('be.visible');
-                cy.contains('precious');
-                cy.contains('yes');
+                verifyFilterRulesForm();
 
                 cy.contains('Add new rule').click();
                 cy.contains('.selection', 'Type in values')
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -4,7 +4,7 @@ import { without } from 'lodash';
 import type { SystemLabel } from '../../support/deployments';
 
 import { exampleBlueprintUrl } from '../../support/resource_urls';
-import { FilterRuleOperators, FilterRuleType } from '../../../../widgets/common/src/filters/types';
+import { FilterRuleAttribute, FilterRuleOperators, FilterRuleType } from '../../../../widgets/common/src/filters/types';
 import type {} from '../../../../widgets/common/src/deploymentsView';
 
 describe('Deployments View widget', () => {
@@ -93,19 +93,6 @@ describe('Deployments View widget', () => {
     };
 
     const widgetHeader = {
-        setFilter: (filterId: string) => {
-            cy.contains('button', 'Filter').click();
-
-            cy.get('.modal').within(() => {
-                cy.setSearchableDropdownValue('Filter ID', filterId);
-                cy.get('.loader').should('not.exist');
-                cy.contains('OK').click();
-            });
-
-            cy.get('.modal').should('not.exist');
-            cy.contains('button', 'Filter').should('not.exist');
-            cy.contains('button', filterId);
-        },
         openRunWorkflowModal: () => {
             cy.contains('Bulk Actions').click();
             cy.contains('Run Workflow').click();
@@ -209,7 +196,7 @@ describe('Deployments View widget', () => {
         getDeploymentsViewTable().find('tbody tr:first-of-type.active').should('exist');
     });
 
-    describe('with filters', () => {
+    describe.only('with filters', () => {
         const deploymentNameThatMatchesFilter = `${specPrefix}precious_deployment`;
         const filterId = 'only-precious';
         const filterRules: Stage.Common.Filters.Rule[] = [
@@ -254,14 +241,24 @@ describe('Deployments View widget', () => {
             cy.contains(/with ID .* was not found/);
         });
 
-        it('should take the selected filter into account when displaying deployments', () => {
+        it('should take the selected existing filter into account when displaying deployments', () => {
             useDeploymentsViewWidget();
 
             cy.contains(deploymentNameThatMatchesFilter);
             cy.contains(deploymentName);
             cy.interceptSp('POST', '/searches/deployments').as('deploymentsSearchRequest');
 
-            widgetHeader.setFilter(filterId);
+            cy.contains('button', 'Filter').click();
+
+            cy.get('.modal').within(() => {
+                cy.setSearchableDropdownValue('Filter ID', filterId);
+                cy.get('.loader').should('not.exist');
+                cy.contains('OK').click();
+            });
+
+            cy.get('.modal').should('not.exist');
+            cy.contains('button', 'Filter').should('not.exist');
+            cy.contains('button', filterId);
 
             cy.wait('@deploymentsSearchRequest').then(({ request }) => {
                 const requestRules = request.body.filter_rules;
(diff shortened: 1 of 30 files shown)


widgets/filters/src/FilterModal.tsx
```diff
@@ -24,13 +24,16 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
+
     const { useInput, useErrors } = Stage.Hooks;
     const [filterId, setFilterId] = useInput((initialFilter?.id ?? '') + initialFilterIdSuffix);
-    const [filterRules, setFilterRules] = useInput(
-        initialFilter ? [...initialFilter.attrs_filter_rules, ...initialFilter.labels_filter_rules] : []
-    );
+    const [filterRules, setFilterRules] = useInput(initialFilterRules);
     const { errors, setErrors, clearErrors, setMessageAsError } = useErrors();
-    const [filterRulesInvalid, setFilterRulesInvalid] = useState(false);
+    const [filterRulesInvalid, setFilterRulesInvalid] = useState(!initialFilter);
```
Gelio, line 36: In `RulesForm` we have a `useEffect` that changes the rows when `initialFilters` change. I believe we should also change the `filterRulesInvalid` when the `initialFilter` changes.

Besides, would it not make more sense to have the `invalid` state be derived based on `filterRules` via `useMemo`? Then we would not need any manual synchronization

widgets/common/src/deploymentsView/useFilterQuery.ts
```diff
@@ -0,0 +1,28 @@
+import { useQuery } from 'react-query';
+
+export default function useFilterQuery(toolbox: Stage.Types.Toolbox, filterId?: string) {
```
Gelio, line 3: Good idea to extract it :+1:

widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -28,15 +28,17 @@ const DeploymentsViewHeader: FunctionComponent<DeploymentsViewHeaderProps> = ({
     const [filterModalOpen, openFilterModal, closeFilterModal] = useBoolean();
     const [deployOnModalOpen, openDeployOnModal, closeDeployOnModal] = useBoolean();
     const [runWorkflowModalOpen, openRunWorkflowModal, closeRunWorkflowModal] = useBoolean();
-    const [filterId, setFilterId] = useState<string>();
+    const [userFilterRules, setUserFilterRules] = useState<FilterRule[]>();
```
Gelio, line 31: We have `userFilterRules` both in `DeploymentsView` and in here, as well as in `FilterModal`. Having duplicated state that needs to be synchronized could lead to problems and increased maintenance. Can we only have them in 2 places?
1. Rules that the widget uses to fetch the deployments
2. Local form state - the rules that the user sees

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -52,32 +52,11 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
...
-             * Such a case should not happen naturally, unless an `enabled` option is added to `useQuery`. If it is added,
-             * it should be here.
-             */
-            throw new Error('Idle state for fetching filter rules is not implemented.');
-        }
+    const [userFilterRules, setUserFilterRules] = useState<FilterRule[] | undefined>(undefined);
 
-        return result;
-    })();
+    const defaultFilterRulesResult = useFilterQuery(toolbox, userFilterRules ? undefined : defaultFilterId);
```
Gelio, line 57: `FilterModal` is already responsible for resolving the rules. Maybe we could also move the logic of resolving the initial filter rules there, so it will be the single point of resolving the filter rules? Then, it could propagate the state (`loading`, `error`, the actual rules) to the parent `DeploymentsView` component.

IMO that would make it easier to reason about, but I don't have strong arguments as I did not investigate the code deeply

widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -50,20 +52,16 @@ const DeploymentsViewHeader: FunctionComponent<DeploymentsViewHeaderProps> = ({
...
+                        content={userFilterId ?? headerT('filter.unsavedFilter')}
                         onClick={openFilterModal}
                         style={{ whiteSpace: 'nowrap', maxWidth: 200, textOverflow: 'ellipsis', overflow: 'hidden' }}
                     />
-                    <Button
-                        icon="delete"
-                        onClick={() => handleFilterChange(undefined)}
-                        title={headerT('filter.clearButton')}
-                    />
+                    <Button icon="delete" onClick={() => handleFilterChange()} title={headerT('filter.clearButton')} />
```
Gelio, line 64: Minor:

I prefer seeing `undefined` passed explicitly rather than use default arguments  `handleFilterChange()`. I would make that function declaration look like:

```ts
function handleFilterChange(newFilterRules: FilterRule[] | undefined, newFilterId: string | undefined) {
```

which requires being explicit

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -56,15 +107,26 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
-                            onChange={handleChange}
+                            onChange={handleFilterIdChange}
                             fetchUrl="/filters/deployments?_include=id"
                             prefetch
                             value={filterId}
                         />
                     </UnsafelyTypedFormField>
+                    <UnsafelyTypedFormField label={tModal('filterRules')}>
+                        {filterRulesResult.isFetching && <Stage.Basic.LoadingOverlay />}
+                        {filterRulesResult.isSuccess && (
```
Gelio, line 121: What about `filterRulesResult.isError` :smile: ?

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -21,33 +26,79 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
-    const [filterId, setFilterId, resetFilterId] = useResettableState<string | undefined>(filterIdProp);
+    const [submitedFilterId, setSubmitedFilterId] = useState<string>();
+    const [filterId, setFilterId] = useState<string>();
+    const [initialFilterRules, setInitialFilterRules] = useState<FilterRule[]>([]);
+    const [filterRules, setFilterRules] = useState<FilterRule[]>();
+    const [rulesErrorsPresent, setRuleErrorsPresent] = useState<boolean>();
+
+    // Controlls filter 'dirty' state
+    const [filterEdited, setFilterEdited, unsetFilterEdited] = useBoolean();
+    const [submitedFilterEdited, setSubmitedFilterEdited] = useState<boolean>();
```
Gelio, line 40: Typo: submited -> submitted

Gelio (commented):
The structure looks cool :+1: I did not test it in the UI yet.

Do you plan on running a full run of system tests for this PR?

I see you're targeting the `6.1.0-build` branch with this PR. Do you know if we already switched the CI to build off this branch? I am not aware of any such changes that have happened so far

widgets/common/src/deploymentsView/DeploymentsView.tsx
kubama: My understanding is that there is no direct correlation between the default (configured) filter and filter selection modal, so I would actually consider such solution artificial. The extracted `useFilterQuery` is a common hook, there is almost no extra logic needed to get the filter rules, so I consider the current solution sufficient.

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: Done

widgets/common/src/deploymentsView/header/index.tsx
kubama: Done

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: I added `useEffect` that displays an error message when `filterRulesResult.isError` flag gets set

widgets/common/src/deploymentsView/header/index.tsx
kubama: This state value is present in 3 places but it is used in each of them:
1. Top level - `header/FilterModal` - local form state, value is read and written
2. Middle level - `header/index` - used to set proper button label, but the value is also unset from here when the user filter is removed, so again the value is read and written
3. Bottom level - `DeploymentsView` - rules that the widget uses to fetch the deployments, read only

Summing up, I don't see an easy way to remove it from the second level without introducing one or more other state variables in place.

widgets/filters/src/FilterModal.tsx
kubama: > In `RulesForm` we have a `useEffect` that changes the rows when `initialFilters` change. I believe we should also change the `filterRulesInvalid` when the `initialFilter` changes.

It's a different component, so I'm not sure what's the reasoning. Besides, by design the `!initialFilter` value never changes - it's constant for each of three modal types (add/edit/clone).

> Besides, would it not make more sense to have the `invalid` state be derived based on `filterRules` via `useMemo`? Then we would not need any manual synchronization

The `filterRulesInvalid` is not just a derived value, it is an actual state variable that gets mutated.

kubama:
> Do you plan on running a full run of system tests for this PR?

I added related system tests to this PR, scheduled full build: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/816/

Gelio:
Minor:

I have noticed that when there is a filter set, there is no gap between the blue filter button and "Bulk Actions":

image

kubama:
Another system tests run after correcting flaky test: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/821/

kubama:
> I have noticed that when there is a filter set, there is no gap between the blue filter button and "Bulk Actions"

Thanks for noticing, fixed.

kubama:
Looks like the build is failing due to https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1880 being merged and related README not updated.

widgets/common/src/deploymentsView/header/index.tsx
Gelio: I ended up drawing as I would do that myself and I ended up with a very similar structure to yours:

image

So, looks like I was wrong initially and this is the way :slightly_smiling_face:

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: Yup, after drawing it myself, looks ok :+1:

Gelio: I can't open a new conversation for some reason, so I will use this one.

Minor:

Do we need this ternary operator for the 2nd argument? IMO this could always point to `defaultFilterId`. It's just the `defaultFilterRulesResult` won't be used later on when there are used defined filter rules.

Otherwise, it may happen that the user will see a loading/error state even though they just filled in the form and are using user-defined rules

Gelio (commented):
As I am reading the code, I find the logic complex :worried: There's a lot of similar state, unfortunately, and that makes it more confusing. I don't have any idea how to simplify it, as I would have to probably try to implement something similar myself.

One idea may be to use `useReducer` to simplify state management for similar pieces of state. Maybe that would centralize the state management instead of using individual multiple `set...` calls one after another (in `header/index` and `header/FilterModal`)

I did not read the tests and I do not fully comprehend the `FilterModal`. I think I will only review the tests next (later today) and trust you that `FilterModal` is fine, because frankly, the code is complex to me and reading it takes too much of my time (I have spent ~2.5h for this PR already)

widgets/filters/src/FilterModal.tsx
kubama: > Right, I did not notice this is a separate component. I was confused because this was changed in the PR that I believe was scoped only for Deployments View.
> 
> Anyway, a minor risk I see now is that we have this `initialFilter`/`initialFilters` props that are sometimes reactive (a change will update the component's internal state) and sometimes they don't.
> 
> As this is not related to this PR (as is pretty much this whole change, if I'm not mistaken 😄 ), we can tackle it later, in some separate ticket. We can keep this discussion going to clarify it, but it does not block the PR 🙂

The changes here are related to the PR. Please note the change in `RulesForm` - I added an effect to update form content on `initialFilters` change. I needed this to implement form reloading when changing the filter selected in the dropdown. All changes in this `FilterModal` component are a result of this change and they are intended to preserve the pre-existing behaviour.

> Right, I can see that 🙂 However, from the logical perspective, the fact that the form is valid or not, to my mind, depends on the form's content, meaning the rules. Thus, if we make it a separate state variable, we run the risk of desynchronizing the state in some case in which we only change the rules and forget to update the validity. Looks like a risk to me and higher maintenance cost 🙂
> 
> Is there a case when we need to mutate the validity so that it is inconsistent with the form rules?

The `RulesForm` component reports form validity as part of the `onChange` callback. However it is called only as a result of user interaction with the form. When the form is in its initial state the `onChange` is not yet called and the form is valid if it's prepopulated with existing filter rules and invalid when it is empty. Hence the state variable with the initial value. The variable is always in sync with the form state and there is no way it could be inconsistent, as it is only updated as part of the `onChange` callback.

widgets/filters/src/FilterModal.tsx
```diff
@@ -24,13 +24,16 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
     onSubmit,
     toolbox
 }) => {
+    const initialFilterRules = useMemo(
+        () => (initialFilter ? [...initialFilter.attrs_filter_rules, ...initialFilter.labels_filter_rules] : []),
+        []
```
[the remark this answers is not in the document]
kubama: `initialFilterRules` will never become stale or outdated. By design this modal is always opened in the context of a single filter. `initialFilter` is always passed as a static data, there is no async loading involved. In other words there is no scenario in which `initialFilter` would ever change.

widgets/filters/src/FilterModal.tsx
```diff
@@ -24,13 +24,16 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
     onSubmit,
     toolbox
 }) => {
+    const initialFilterRules = useMemo(
```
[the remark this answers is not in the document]
kubama: Please see https://github.com/cloudify-cosmo/cloudify-stage/pull/1444#discussion_r656048177

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -21,33 +27,83 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
     const { errors, setErrors, clearErrors } = Stage.Hooks.useErrors();
-    const [filterId, setFilterId, resetFilterId] = useResettableState<string | undefined>(filterIdProp);
+    const [submittedFilterId, setSubmittedFilterId] = useState<string>();
+    const [filterId, setFilterId] = useState<string>();
+    const [initialFilterRules, setInitialFilterRules] = useState<FilterRule[]>([]);
+    const [filterRules, setFilterRules] = useState<FilterRule[]>();
+    const [rulesErrorsPresent, setRuleErrorsPresent] = useState<boolean>();
+
+    // Controlls filter 'dirty' state
+    const [filterEdited, setFilterEdited, unsetFilterEdited] = useBoolean();
```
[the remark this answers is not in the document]
kubama: Right, renamed

widgets/filters/src/FilterModal.tsx
Gelio: > I needed this to implement form reloading when changing the filter selected in the dropdown. All changes in this `FilterModal` component are a result of this change and they are intended to preserve the pre-existing behaviour.

Makes sense now :+1: 

For the other part - maybe there could be some pattern to make it clearer and avoid so many interactions between components, but I did not come up with one just by reviewing. Let's leave it be

widgets/filters/src/FilterModal.tsx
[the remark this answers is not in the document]
Gelio: > `initialFilter` is always passed as a static data, there is no async loading involved. In other words there is no scenario in which `initialFilter` would ever change.

We're making this assumption implicitly :slightly_smiling_face: Maybe you know it after reading the interaction between a few components, but just by looking at this code, it's unexpected - you have a variable that is used in the callback, but not used in the dependency array. This is a basic violation of rules of hooks and would be reported as such by the ESLint plugin for rules of hooks :slightly_smiling_face: I agree that it will likely not change, but there is no harm in adding the `initialFilter` to the dependency array. Then, we are more compliant with how `useMemo` is used.

If you are not too keen on modifying the dependency array, a comment stating what you just said (that `initialFilter` is static and will never change) makes this clear enough.

I would advise going with the way of modifying the dependency array anyway

widgets/common/src/filters/RulesForm.tsx
```diff
@@ -56,23 +56,33 @@ export interface RulesFormProps {
...
         );
+        setRows(newRows);
+        handleChange(newRows);
     }
 
     useEffect(() => {
-        onChange(rows.map(getFilterRule), rows.filter(row => row.hasError).length > 0);
-    }, [rows]);
+        setRows(getFilterRuleRows(initialFilters));
+    }, [initialFilters]);
```
Gelio, line 85: `RulesForm` is a controlled and an uncontrolled element at the same time, which is probably why I find the solution confusing. It has its own state, but also takes state from above, and communicated the changes to parent components. I am ok with having it here for now, but I would definitely try to find some better approach in the future - maybe centralized `useReducer` in some parent component or some advanced React patterns

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -21,33 +27,83 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
     const { ApproveButton, CancelButton, Icon, Modal, UnsafelyTypedForm, UnsafelyTypedFormField } = Stage.Basic;
     // @ts-expect-error DynamicDropdown is not converted to TS yet
     const { DynamicDropdown } = Stage.Common;
-    const { useResettableState } = Stage.Hooks;
+    const { useBoolean } = Stage.Hooks;
 
     const { errors, setErrors, clearErrors } = Stage.Hooks.useErrors();
-    const [filterId, setFilterId, resetFilterId] = useResettableState<string | undefined>(filterIdProp);
+    const [submittedFilterId, setSubmittedFilterId] = useState<string>();
+    const [filterId, setFilterId] = useState<string>();
```
Gelio, line 34: Did you consider my previous suggestion about using `useReducer`?

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -253,14 +241,24 @@ describe('Deployments View widget', () => {
...
             cy.contains(deploymentNameThatMatchesFilter);
             cy.contains(deploymentName);
-            cy.interceptSp('POST', '/searches/deployments').as('deploymentsSearchRequest');
 
-            widgetHeader.setFilter(filterId);
+            cy.contains('button', 'Filter').click();
+
+            cy.get('.modal').within(() => {
+                cy.setSearchableDropdownValue('Filter ID', filterId);
+                cy.contains('Label').should('be.visible');
```
Gelio, line 254: This is unexpected to me. Which "Label" are we looking for? Could you add some comment?

Gelio (approved):
Looks ok to me, but it still looks complex :worried: I did not fully comprehend all the changes, but I don't want to spend more time digging into the details, as I have already spent quite some time on this PR

Since this is quite a substantial change, I believe @stepek should also take a look to have 2 approvals

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: I extracted a function, I think it should be obvious now.

kubama:
> As I am reading the code, I find the logic complex 😟 There's a lot of similar state, unfortunately, and that makes it more confusing. I don't have any idea how to simplify it, as I would have to probably try to implement something similar myself.
> 
> One idea may be to use `useReducer` to simplify state management for similar pieces of state. Maybe that would centralize the state management instead of using individual multiple `set...` calls one after another (in `header/index` and `header/FilterModal`)

One reason it looks complex is that except 'current' modal state we also have 'submitted' state, which is saved to be able to revert to it when closing the modal with cancel. I took advantage of this pattern and created `useRevertableState` hook which I reused for handling `filterId`, `filterRules` and `filterDirty` state values. I think it's much cleaner now and also easier to understand. The only confusing thing now is the `initialFilterRules` state variable which is needed only because, as you noted, `RulesForm` is not a fully controlled component at the moment. I'm not willing to convert it at this point as its internal state is derived from the initial state provided through the prop, so the conversion wouldn't be straightforward - looks like for each change there would need to be a conversion from internal state to external data representation and vice versa.

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: Yes, I considered it but decided to take a different path - please see https://github.com/cloudify-cosmo/cloudify-stage/pull/1444#issuecomment-866082811

widgets/filters/src/FilterModal.tsx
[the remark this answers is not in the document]
kubama: OK, updated

kubama:
I'm not planning any more changes at this point. So I'm waiting for a final review.

@stepek Please let me know if you are willing to submit your review.

widgets/filters/src/FilterModal.tsx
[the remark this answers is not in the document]
Gelio: Thanks! :+1:

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Yup, much better :smile:

widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -28,15 +28,15 @@ const DeploymentsViewHeader: FunctionComponent<DeploymentsViewHeaderProps> = ({
     const [filterModalOpen, openFilterModal, closeFilterModal] = useBoolean();
     const [deployOnModalOpen, openDeployOnModal, closeDeployOnModal] = useBoolean();
     const [runWorkflowModalOpen, openRunWorkflowModal, closeRunWorkflowModal] = useBoolean();
-    const [userFilterRules, setUserFilterRules] = useState<FilterRule[]>();
+    const [userFilterSelected, setUserFilterSelected] = useState<boolean>(false);
```
Gelio, line 31: Good simplification :+1:

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -16,8 +16,37 @@ interface FilterModalProps {
...
+    }
+
+    const result: [T, typeof setValue, typeof saveValue, typeof revertValue, typeof resetValues] = [
+        value,
+        setValue,
+        saveValue,
+        revertValue,
+        resetValues
+    ];
+    return result;
```
Gelio, line 45: Minor:

This looks like a bit too many values in "tuple" to me :slightly_smiling_face: If we returned an object with properties `value`, `set`, `save, `revert`, `reset`, we could use it like:

```ts
const submittedFilterIdState = useRevertableState(...);

submittedFilterIdState.value;
submittedFilterIdState.set('abc');
```

IMO less typing for us and fewer places to mistakes when using the hook (with an object it's harder to use the wrong function, as they are explicitly named when exported)

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -16,8 +16,37 @@ interface FilterModalProps {
 const tModal = Stage.Utils.getT(`${i18nPrefix}.header.filter.modal`);
 const tMessage = Stage.Utils.getT(i18nMessagesPrefix);
 
+function useRevertableState<T>(initialValue: T) {
```
Gelio, line 19: I like the idea :+1: Also, good work keeping this hook local to the component and not extracting it preemptively - I like it

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -27,27 +56,33 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
+    // The values are 'saved' on modal submit and 'reverted' on modal cancel
+    const [filterId, setFilterId, saveFilterId, revertFilterId, resetFilterId] = useRevertableState<string | undefined>(
+        undefined
+    );
+    const [filterRules, setFilterRules, saveFilterRules, revertFilterRules, resetFilterRules] = useRevertableState<
+        FilterRule[]
+    >([]);
+    const [filterDirty, setFilterDirty, saveFilterDirty, revertFilterDirty, resetFilterDirty] = useRevertableState<
+        boolean
+    >(false);
```
Gelio, line 72: Minor:

Maybe we could extract these 3 to a custom hook and expose functions to work with the form in an easier way? I see they are often called together, so maybe we could group this "call together" logic into functions inside of a custom hook

Gelio (dismissed):
Looks good :+1: The `useRevertableState` hook cleared some stuff up, nice idea :smile: 

I have suggested some possible improvements, but they are not a must-have and could be postponed if you think they are worth it :slightly_smiling_face:

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: Great idea, implemented.

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: I implemented the idea, probably not in the most optimal way though - please review.

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -22,27 +22,24 @@ function useRevertableState<T>(initialValue: T) {
...
+        save: () => setSavedValue(value),
+        revert: () => setValue(savedValue),
+        reset: () => {
+            resetValue();
+            resetSavedValue();
+        }
+    };
+}
+
+function useRevertableStates(...revertableStates: Omit<ReturnType<typeof useRevertableState>, 'set'>[]) {
```
Gelio, line 37: Nice TS magic :smile:

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -61,31 +58,22 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
-    const [filterRules, setFilterRules, saveFilterRules, revertFilterRules, resetFilterRules] = useRevertableState<
-        FilterRule[]
-    >([]);
-    const [filterDirty, setFilterDirty, saveFilterDirty, revertFilterDirty, resetFilterDirty] = useRevertableState<
-        boolean
-    >(false);
+    const filterId = useRevertableState<string | undefined>(undefined);
+    const filterRules = useRevertableState<FilterRule[]>([]);
+    const filterDirty = useRevertableState<boolean>(false);
+    const modalState = useRevertableStates(filterId, filterRules, filterDirty);
```
Gelio, line 64: Nice :+1:

Gelio (approved):
LGTM :+1: A considerable improvement in terms of readability :smile:

widgets/common/src/deploymentsView/header/FilterModal.tsx
Gelio: Looks good, I did not think of `useRevertableStates`, but IMO that's a nice pattern :+1: I meant more of a hardcoded solution, but yours looks to scale better and is alright in terms of readability

kubama:
Since I'm not receiving any comments from @stepek and this PR is open for a while I'm merging this PR with one approval.

@Gelio Thanks for spending your time on the review. Even though it was lengthy I think we managed to improve the quality drastically, so I think it was worth it.

stepek (commented):
LGTM

PR #1445, opened 2021-06-20
RD-1766 redesign ci workflow stage

yuvlar:
Suggested changes to Cloudify-stage in order to reduce the build runtime

Tweaked the changes a bit,
In order to decrease the build time even more, we'll need to do caching (next task)


jenkins/Jenkinsfile
```diff
@@ -22,37 +22,101 @@ pipeline {
...
+        stage('Prepare'){
+          steps {
+            container('node'){
+              dir("${STAGE_DIR}/conf") {
+                withVault([configuration: configuration, vaultSecrets: secrets]) {
+                  sh """
+                    jq '.maps.accessToken = "${env.MAPS_ACCESS_TOKEN}"' app.json > app.json.tmp && mv app.json.tmp app.json
+                  """
+                }
+                dir("${STAGE_DIR}") {
```
Gelio, line 58: Technically this is not 1-1 with https://cawemo.com/share/06df993a-57bc-4926-8741-d6b81025068f as building the RPM can start before `npm run beforebuild` is done :slightly_smiling_face: 

Is it what you agreed on with @qooban?

jenkins/Jenkinsfile
```diff
@@ -93,37 +157,15 @@ pipeline {
...
-            sh 'sudo rm -rf /var/lib/apt/lists/* && sudo apt update && sudo apt install postgresql-client'
-            sh 'echo \'{"app":{"ssl":{"ca":null},"logLevelConf":"","logLevel":"debug"}}\' > conf/me.json'
-            sh 'npm run test:backend'
-          }
-        }
-      }
-    }
     stage('Run Tests') {
       parallel {
         stage('Size') {
```
Gelio, line 171: Shouldn't this be removed? We already have a size check in the `Run Builds & Build RPM` stage :slightly_smiling_face:

jenkins/Jenkinsfile
```diff
@@ -157,34 +199,19 @@ pipeline {
             }
```
Gelio, line 199: I can't comment in the `Docs` stage directly, but if I'm not mistaken, it can also be parallelized, so it can be moved to the `parallel` block in the `Run Builds & Build RPM` stage :slightly_smiling_face:

Gelio (changes requested):
This looks great :+1: 

I know you said that this improves the time by 5 minutes. Does it mean the job now runs in ~10 mins instead of ~15? That would be awesome :smile: 

I believe `.DS_Store` was committed by mistake, which is why I'm marking this as _request changes_ :slightly_smiling_face:

jenkins/Jenkinsfile
yuvlar: npm run size does run in Build stage already, let me know if it can be removed

jenkins/Jenkinsfile
yuvlar: Basically yes, but it has a command there to delete a folder that the tests use, so it crashes.
If it doesn't need to delete the folder that'll be great, let me know what you think

jenkins/Jenkinsfile
yuvlar: right, it wasn't possible to do it exactly like the diagram.
I could put the RPM stage before, but that would increase the build time

jenkins/Jenkinsfile
yuvlar: nvm, my bad, I put it twice.
will be fixed

jenkins/Jenkinsfile
Gelio: I see, no worries :slightly_smiling_face: We can leave it as-is if that's faster

I'm curious, though, as it would seem to me that running an independent step (building the RPM) earlier and in parallel would make the overall process finish sooner. Is it because there's not enough parallelization possible in the build system that it is overall slower? :thinking: 

I mean, just looking at the `Jenkinsfile` code I believe we could add more parallelism and try to do it just like the diagram, but I am not an expert in Jenkins :smile: I simply do not see the limitations that make the diagram impossible/hard to implement :thinking:

jenkins/Jenkinsfile
```diff
@@ -22,37 +22,54 @@ pipeline {
...
-          dir("${stage_dir}/conf") {
-            withVault([configuration: configuration, vaultSecrets: secrets]) {
-              sh """
-                jq '.maps.accessToken = "${env.MAPS_ACCESS_TOKEN}"' app.json > app.json.tmp && mv app.json.tmp app.json
-              """
+    stage('Prepare & Audit master'){
+      parallel{
+        stage('Audit'){
+          when {
+            branch "master"
```
Gelio, line 36: Just FYI, we will have conflicts with #1446 :smile:

jenkins/Jenkinsfile
```diff
@@ -22,37 +22,54 @@ pipeline {
...
+                }
+              }
+              echo 'prepare test folder'
+              sh "mkdir -p ${WORKSPACE}/test && cp -rf ${WORKSPACE}/${STAGE_DIR}/. ${WORKSPACE}/test"
+              dir("${WORKSPACE}/test"){
+                echo 'Run static code analysis and code style check'
+                sh 'npm run lint'
+
+                echo 'Check TypeScript types'
+                sh 'npm run check-types'
```
Gelio, line 72: Theoretically, these two can run in parallel :slightly_smiling_face: Not sure if that helps

jenkins/Jenkinsfile
```diff
@@ -93,33 +110,37 @@ pipeline {
...
-            sh 'npm run check-types'
-
-            echo 'Run frontend tests'
-            sh 'npm run test:frontend'
-
-            echo 'Run backend tests'
-            sh 'sudo rm -rf /var/lib/apt/lists/* && sudo apt update && sudo apt install postgresql-client'
-            sh 'echo \'{"app":{"ssl":{"ca":null},"logLevelConf":"","logLevel":"debug"}}\' > conf/me.json'
-            sh 'npm run test:backend'
+    stage('Run Builds & Build RPM') {
```
Gelio, line 122: This should probably say

```suggestion
    stage('Run tests') {
```

as the build step happens way earlier :slightly_smiling_face:

jenkins/Jenkinsfile
Gelio: Interesting :smile: I believe it does not need to delete that `test` directory. Generally, if I'm not mistaken, it looks like some stages copy the whole project to a separate directory (e.g. `${workspace}/test`) and then run some commands there. Thus, the removal of the test directory is something that is only required by the build system.

We could even clean it up a bit and make each stage clean up after itself. The structure would then be like this, for example for docs:

```jenkinsfile
        stage('Docs'){
          steps {
            container('node'){
              sh "mkdir -p ${WORKSPACE}/docs && cp -rf ${WORKSPACE}/${stage_dir}/. ${WORKSPACE}/docs"
              dir("docs") {
                echo 'Generate widgets documentation'
                sh 'npm run docWidgets'

                echo 'Check if widgets documentation is up to date'
                sh 'git diff --exit-code ./widgets'

                echo 'Check if cloudify-ui-components version is the same as in official documentation'
                sh 'npm run docCheck'
              }
              sh 'rm -rf ${WORKSPACE}/docs'
            }
          }
        }
```

I believe this makes the stage:
1. Create a copy of the project to run the commands in isolation
2. Run the commands
3. Clean up

It would solve the problem of some stages trying to clean other stage's copies (e.g. `docs` trying to clean up `tests`) :slightly_smiling_face: 

Then, we could parallelize almost all the stages, just as in https://cawemo.com/share/06df993a-57bc-4926-8741-d6b81025068f

Gelio (changes requested):
Looks definitely better, but I am wondering if we could squeeze even more parallelism :smile: 

Also, take a look at my previous review summary, because:

> I believe `.DS_Store` was committed by mistake, which is why I'm marking this as _request changes_

:slightly_smiling_face: 

We could also have a quick call to discuss the parallelism issues that I am not seeing :slightly_smiling_face:

Gelio (approved):
LGTM! :smile: 

Just FYI, we will most likely need the same changes cherry-picked to 6.1.0-build :slightly_smiling_face:

kubama:
Seeing an in-depth review done by @Gelio I'll opt out of reviewing myself.

jenkins/Jenkinsfile
```diff
@@ -22,38 +22,55 @@ pipeline {
...
+              }
             }
           }
         }
-      }
-    }
-    stage('Run Builds') {
-      parallel {
-        stage('Build') {
+        stage('Prepare & run lint&check-types tests'){
```
warolv, line 50: Can you put more human-readable text?:-)

jenkins/build-pod.yaml
```diff
@@ -8,32 +8,32 @@ spec:
...
+  - name: node
+    image: circleci/node:12.22.1-browsers
     resources:
       requests:
         cpu: 1.5
-        memory: 3Gi
+        memory: 8Gi
+        ephemeral-storage: "15Gi"
       limits:
-        cpu: 1.5
```
warolv, line 18: not QOS anymore?

warolv (changes requested):

jenkins/Jenkinsfile
Gelio: An example could be `Install dependencies and execute basic checks`

Gelio (changes requested):
The Jenkinsfile looks good. I have no clue about `build-pod` though. Also, please revert `.DS_Store` :smile: Aside from that, cool!

jenkins/build-pod.yaml
yuvlar: It will be Burstable QoS, because in one parallel part, only the node container is used, and it should utilize the instance resources.
According to k8s, a higher cpu limit than the instance has won't evict the pod, but just might delay some processes.
I'll keep an eye on it anyway and if it'll cause issues I'll revert the limits to be like the requests

jenkins/Jenkinsfile
yuvlar: Will do and request a new review

Gelio (changes requested):
I believe you may be missing [all](https://github.com/cloudify-cosmo/cloudify-stage/pull/1445#pullrequestreview-688113983) [my](https://github.com/cloudify-cosmo/cloudify-stage/pull/1445#pullrequestreview-689093770) [comments](https://github.com/cloudify-cosmo/cloudify-stage/pull/1445#pullrequestreview-690540991) about removing `.DS_Store` :slightly_smiling_face: Please remove it, we don't want it in the repo :smile:

warolv (approved):
Looks good!

Gelio (approved):
LGTM

PR #1446, opened 2021-06-21
RD-2647 Switch UI development to 6.1.0 build

Gelio:
This branch switches the main branch for the UI CI to `6.1.0-build`. I have also refactored slightly the Jenkinsfile to avoid having to change the name in multiple places in the future :slightly_smiling_face: 

I will raise a cherry-pick for the `6.1.0-build` branch after this PR is merged.

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/819/pipeline

Queued 2021-06-21 20:02 CEST

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -26,7 +26,7 @@ pipeline {
     stage_dir = "cloudify-stage"
     cfy_manager_url = 'https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager'
     cfy_node_rpm = 'http://repository.cloudifysource.org/cloudify/components/nodejs-12.22.1-1nodesource.x86_64.rpm'
-    MAIN_BRANCH = "master"
+    MAIN_BRANCH = "6.1.0-build"
   }
 
   stages {
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -26,6 +26,7 @@ pipeline {
     stage_dir = "cloudify-stage"
     cfy_manager_url = 'https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager'
     cfy_node_rpm = 'http://repository.cloudifysource.org/cloudify/components/nodejs-12.22.1-1nodesource.x86_64.rpm'
+    MAIN_BRANCH = "master"
   }
 
   stages {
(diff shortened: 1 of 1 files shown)


jenkins/Jenkinsfile
```diff
@@ -26,6 +26,7 @@ pipeline {
     stage_dir = "cloudify-stage"
     cfy_manager_url = 'https://raw.githubusercontent.com/cloudify-cosmo/cloudify-manager'
     cfy_node_rpm = 'http://repository.cloudifysource.org/cloudify/components/nodejs-12.22.1-1nodesource.x86_64.rpm'
+    MAIN_BRANCH = "6.1.0-build"
```
yuvlar, line 29: @Gelio How static is this var? is it changed only when we bump versions or you might need to change it sometimes regardless? If it's the latter, maybe it should be a parameter instead of an env var.
If it's not, then see this change as approved

Gelio: I believe we only changed it when we bumped versions, so it's changed once per release, and then reverted to `master` after the release is over :slightly_smiling_face: IMO we don't need this to be parametrized per-build for now

yuvlar: Sounds good, Approved

yuvlar (approved):

kubama (approved):

jenkins/Jenkinsfile
```diff
@@ -176,13 +177,13 @@ pipeline {
...
-        branch "master"
+        branch "${MAIN_BRANCH}"
       }
       steps {
         echo 'Trigger Stage-UI System-tests'
         build(job: 'Stage-UI-System-Test', parameters: [
-          string(name: 'BRANCH', value: 'master'),
-          string(name: 'STAGE_BRANCH', value: 'master')
+          string(name: 'BRANCH', value: '${MAIN_BRANCH}'),
+          string(name: 'STAGE_BRANCH', value: '${MAIN_BRANCH}')
```
Gelio, line 186: I believe I should use double quotes here. I cannot push to this branch anymore since it's protected. I will raise a separate PR with the updated code

Gelio:
@yuvlar Could you remove the `RD-2647-switch-UI-development-to-6.1.0-build` branch? I cannot delete it since it's protected

jenkins/Jenkinsfile
yuvlar: Sounds fair enough.
I'll make sure to delete this one

Gelio: Thanks!

PR #1451, opened 2021-06-23
RD-2633 Allow setting up intercepts right before login

Gelio:
See https://cloudifysource.atlassian.net/browse/RD-2633?focusedCommentId=72312 for more information about the problem.

Long story short, when the `useDeploymentsView` function was executing for some test, the UI from the previous test is still displayed in the browser. When cypress sets up an intercept for the `/searches/deployments` before the page is refreshed (via `cy.visit`) in the new test, it may happen that due to polling, the UI from the previous test sends a request that is matched by Cypress. Because the browser unloads the UI and aborts the requests, the request never gets any response, which is what I believe caused the problems mentioned in RD-2633.

Now if there is a `/searches/deployments` request that happens before `/visit`, it will not be intercepted. I could not reproduce this exact situation locally, but a "close enough" case is below:
image

I inspected all uses of `mockLogin` to see if there are `intercept(...).as(...)` before `mockLogin` calls. I only found one aside from `deployments_view_spec`. Other `intercept`s before `mockLogin` did not have `as`, meaning they only mocked the response, which is irrelevant to this problem. I left those as they are, as IMO those are harmless

I will cherry-pick the PR to `master` after it's merged.

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/828/pipeline

Queued 2021-06-23 15:02 CEST

### DIFF
--- a/test/cypress/integration/change_password_spec.ts
+++ b/test/cypress/integration/change_password_spec.ts
@@ -80,9 +80,9 @@ describe('Change Password modal', () => {
     });
 
     it('should not be available when LDAP is enabled', () => {
-        cy.interceptSp('GET', `/ldap`, 'enabled').as('ldap');
-
-        cy.usePageMock().mockLogin();
+        cy.usePageMock().mockLogin(undefined, undefined, undefined, () => {
+            cy.interceptSp('GET', `/ldap`, 'enabled').as('ldap');
+        });
 
         cy.get('.usersMenu').click();
         cy.get('#changePasswordMenuItem').should('have.class', 'disabled');
--- a/test/cypress/integration/change_password_spec.ts
+++ b/test/cypress/integration/change_password_spec.ts
@@ -80,9 +80,9 @@ describe('Change Password modal', () => {
     });
 
     it('should not be available when LDAP is enabled', () => {
-        cy.usePageMock().mockLogin(undefined, undefined, undefined, () => {
-            cy.interceptSp('GET', `/ldap`, 'enabled').as('ldap');
-        });
+        cy.interceptSp('GET', `/ldap`, 'enabled').as('ldap');
+
+        cy.usePageMock().mockLogin();
 
         cy.get('.usersMenu').click();
         cy.get('#changePasswordMenuItem').should('have.class', 'disabled');
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -67,12 +67,13 @@ describe('Deployments View widget', () => {
         routeHandler,
         configurationOverrides = {}
     }: { routeHandler?: RouteHandler; configurationOverrides?: Partial<typeof widgetConfiguration> } = {}) => {
-        cy.interceptSp('POST', /^\/searches\/deployments/, routeHandler).as('deployments');
         cy.usePageMock(
             [widgetId],
             { ...widgetConfiguration, ...configurationOverrides },
             { additionalWidgetIdsToLoad, widgetsWidth: 12, additionalPageTemplates: ['drilldownDeployments'] }
-        ).mockLogin();
+        ).mockLogin(undefined, undefined, undefined, () => {
+            cy.interceptSp('POST', /^\/searches\/deployments/, routeHandler).as('deployments');
+        });
         cy.wait('@deployments');
     };
 
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -71,9 +71,8 @@ describe('Deployments View widget', () => {
             [widgetId],
             { ...widgetConfiguration, ...configurationOverrides },
             { additionalWidgetIdsToLoad, widgetsWidth: 12, additionalPageTemplates: ['drilldownDeployments'] }
-        ).mockLogin(undefined, undefined, undefined, () => {
-            cy.interceptSp('POST', /^\/searches\/deployments/, routeHandler).as('deployments');
-        });
+        ).mockLoginWithoutWaiting();
+        cy.interceptSp('POST', /^\/searches\/deployments/, routeHandler).as('deployments');
         cy.wait('@deployments');
     };
 
--- a/test/cypress/support/commands.ts
+++ b/test/cypress/support/commands.ts
@@ -183,18 +183,20 @@ const commands = {
             cy.waitUntilLoaded().then(() => cy.saveLocalStorage());
         }
     },
-    /**
-     * @param setupIntercepts Allows setting up network intercepts at a moment that ensures that the page from
-     * the previous test is already unloaded. Prevents race conditions (e.g. the intercept catching a polling request
-     * from the previous test)
-     */
     // TODO(RD-2314): object instead of multiple optional parameters
-    mockLogin: (
+    mockLogin: (username?: string, password?: string, disableGettingStarted?: boolean) => {
+        cy.mockLoginWithoutWaiting({ username, password, disableGettingStarted });
+        cy.waitUntilLoaded();
+    },
+    mockLoginWithoutWaiting: ({
         username = 'admin',
         password = 'admin',
-        disableGettingStarted = true,
-        setupIntercepts: () => void = _.noop
-    ) => {
+        disableGettingStarted = true
+    }: {
+        username?: string;
+        password?: string;
+        disableGettingStarted?: boolean;
+    } = {}) => {
         cy.stageRequest('/console/auth/login', 'POST', undefined, {
             Authorization: `Basic ${btoa(`${username}:${password}`)}`
         }).then(response => {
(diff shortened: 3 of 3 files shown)


test/cypress/support/commands.ts
```diff
@@ -183,8 +183,13 @@ const commands = {
             cy.waitUntilLoaded().then(() => cy.saveLocalStorage());
         }
     },
+    /**
+     * @param setupIntercepts Allows setting up network intercepts at a moment that ensures that the page from
+     * the previous test is already unloaded. Prevents race conditions (e.g. the intercept catching a polling request
+     * from the previous test)
+     */
     // TODO(RD-2314): object instead of multiple optional parameters
```
Gelio, line 191: I have decided not to complete this TODO in this PR, as it would make the diff a bit messier, and also I have been working on this task for way too long already.

I plan on completing RD-2314 if there is still some time in the sprint (basically after I am done with the current sprint's scope)

I don't think it's as pressing now as delivering planned features for 6.1

test/cypress/support/commands.ts
```diff
@@ -198,7 +203,10 @@ const commands = {
             );
             if (disableGettingStarted) mockGettingStarted(false);
         });
-        cy.visit('/console').waitUntilLoaded();
+        cy.visit('/console');
+        // eslint-disable-next-line chai-friendly/no-unused-expressions
+        setupIntercepts?.();
```
kubama, line 208: How about a default value for `setupIntercepts`? I guess rule disable wouldn't be necessary then.

test/cypress/integration/change_password_spec.ts
```diff
@@ -80,9 +80,9 @@ describe('Change Password modal', () => {
     });
 
     it('should not be available when LDAP is enabled', () => {
-        cy.interceptSp('GET', `/ldap`, 'enabled').as('ldap');
-
-        cy.usePageMock().mockLogin();
+        cy.usePageMock().mockLogin(undefined, undefined, undefined, () => {
+            cy.interceptSp('GET', `/ldap`, 'enabled').as('ldap');
+        });
```
kubama, line 85: Just to confirm I understand - this solution is needed here as the call to `/ldap` happens before page is fully loaded (before `cy.waitUntilLoaded();` completes), right?

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -67,12 +67,13 @@ describe('Deployments View widget', () => {
...
     }: { routeHandler?: RouteHandler; configurationOverrides?: Partial<typeof widgetConfiguration> } = {}) => {
-        cy.interceptSp('POST', /^\/searches\/deployments/, routeHandler).as('deployments');
         cy.usePageMock(
             [widgetId],
             { ...widgetConfiguration, ...configurationOverrides },
             { additionalWidgetIdsToLoad, widgetsWidth: 12, additionalPageTemplates: ['drilldownDeployments'] }
-        ).mockLogin();
+        ).mockLogin(undefined, undefined, undefined, () => {
+            cy.interceptSp('POST', /^\/searches\/deployments/, routeHandler).as('deployments');
+        });
```
kubama, line 76: Why not just call `cy.interceptSp` after `mockLogin` in this case?

test/cypress/support/commands.ts
```diff
@@ -183,8 +183,13 @@ const commands = {
             cy.waitUntilLoaded().then(() => cy.saveLocalStorage());
         }
     },
+    /**
+     * @param setupIntercepts Allows setting up network intercepts at a moment that ensures that the page from
+     * the previous test is already unloaded. Prevents race conditions (e.g. the intercept catching a polling request
+     * from the previous test)
```
kubama, line 189: I'm not sure if this description is accurate. I would add a sentence explaining that interceptors are installed before page is actually loaded, which I believe is the point why it is actually used in `change_password_spec.ts`.

test/cypress/integration/change_password_spec.ts
Gelio: Initially I thought that the behavior here is similar to the one in `deployments_view_spec`. Deployments View was the main point in this PR, I only changed `change_password_spec` as I thought that it could have a similar flaw. Looks like it does not, because there is no `cy.wait('@ldap')` in this test. Thus, I will revert this change

test/cypress/support/commands.ts
Gelio: What you described is not the problem this parameter is trying to solve :slightly_smiling_face: It is just as it says - prevents intercepting requests from the UI loaded in a previous tset

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: That would work, although that would extend the time of the tests as `cy.wait` below would wait for the next polling interval, as the initial request would likely not be caught.

I could remove `cy.wait` and see what happens, but that could introduce other flakiness in the tests and I'm not a fan of doing that ATM, especially having had this bug on me for almost a week. I would appreciate having such feedback earlier, when I asked if the approach outlined here is right (see https://cloudifysource.atlassian.net/browse/RD-2633?focusedCommentId=72312)

test/cypress/support/commands.ts
Gelio: Good idea :+1: done

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: I'm not saying the approach is bad 🙂. Having a test that executes faster is always a plus.
I'm just thinking about consistency with other tests. Looks like other tests install interceptors after `mockLogin` and aside from waiting for the next polling interval they are working just fine. I would take one of two paths:
1. Drop this solution and just install the interceptor after `mockLogin` just as other tests do, or
2. Accept this solution and review other specs to check if this solution could be used in other places to benefit from faster test execution there as well.

test/cypress/integration/change_password_spec.ts
Gelio: Done

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: I see, thanks for a detailed explanation of your viewpoint :+1: 

> Accept this solution and review other specs to check if this solution could be used in other places to benefit from faster test execution there as well.

This is what I have done :slightly_smiling_face: I have inspected all uses of `mockLogin` to see if there is a test that does:
1. `intercept(...).as(X)`
2. `mockLogin`
3. `wait(X)`

which could be problematic. I only found `change_password_spec`, but it turned out to be a mistake on my part and I reverted that after seeing https://github.com/cloudify-cosmo/cloudify-stage/pull/1451#discussion_r656872961  :slightly_smiling_face: 

Thus, I believe there are no tests that have this pattern. There are tests that do

1. `intercept`
2. `mockLogin`

but the intercept for point 1 is not `wait`ed anywhere, so IMO it is fine and there is no benefit in moving it to the `setupIntercepts` callback :slightly_smiling_face:

kubama: What about tests that do `mockLogin` and then `intercept` in the same function? They could use this solution as well and benefit from faster execution by not waiting for the next polling, right?

test/cypress/support/commands.ts
kubama: What this description states is exactly what can be achieved by calling `intercept` after `mockLogin`. So it's not clear what are the benefits of using this feature. I would mention here about catching the first data poll as you mentioned in https://github.com/cloudify-cosmo/cloudify-stage/pull/1451#discussion_r656917345. Additionally, even though it's not the issue that is being solved in this PR I would still mention about intercepting calls taking place before the page is fully loaded as it may be relevant in some cases.

On the other hand, did you consider adding another command instead of introducing a callback? Like `mockLoginNoWait` that would drop call to `waitUntilLoaded`? It could be then used like that:
```
cy.mockLoginNoWait();
cy.intercept();
cy.waitUntilLoaded();
```
Would be easier to read and understand even without the documentation IMHO.

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: I have looked through all the uses of `mockLogin` and I have not noticed this case :slightly_smiling_face: You can take a look yourself and suggest cases.

Generally, this is IMO quite a rare case that we do an intercept for the request for which we are polling and then wait for that request before proceeding with the test. I don't remember why it's done this way for deployments view :slightly_smiling_face:

test/cypress/support/commands.ts
Gelio: I like your suggestion, let me refactor :smile: We could probably even drop the `waitUntilLoaded` since we're waiting for the data to be populated in the table anyway

Gelio: Done :slightly_smiling_face:

kubama (approved):

stepek (approved):
LGTM!

Gelio:
Thanks for reviews :+1: I'll merge after we're done with the system tests: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/828/pipeline

Gelio:
The tests passed, so I'll merge the PR and raise a cherry-pick for master

PR #1459, opened 2021-06-24
RD-1771 Added support for creating new filter

kubama:
The existing "Save" button was converted to a combination of the button and a dropdown:
image

Dropdown clicked:
image

Once "Save As..." is clicked the component is replaced with:
image

System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/839/

### DIFF
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -713,9 +713,13 @@
                         "header": "Select a filter",
                         "submit": "Apply",
                         "save": "Save",
+                        "saveAs": "Save as...",
+                        "saveNew": "Save new filter",
+                        "filterIdPlaceholder": "Enter new filter ID...",
                         "noFilterError": "Please select a filter or specify filter rules",
                         "noFilterRulesError": "Please specify filter rules",
                         "filterRulesInvalidError": "Please provide all the values in filter rules section",
+                        "filterIdMissingError": "Please specify new filter ID",
                         "filterId": "Filter ID",
                         "filterRules": "Filter rules"
                     }
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -715,6 +715,7 @@
                         "save": "Save",
                         "saveAs": "Save as...",
                         "saveNew": "Save new filter",
+                        "cancel": "Cancel",
                         "filterIdPlaceholder": "Enter new filter ID...",
                         "noFilterError": "Please select a filter or specify filter rules",
                         "noFilterRulesError": "Please specify filter rules",
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -719,7 +719,7 @@
                         "noFilterError": "Please select a filter or specify filter rules",
                         "noFilterRulesError": "Please specify filter rules",
                         "filterRulesInvalidError": "Please provide all the values in filter rules section",
-                        "filterIdMissingError": "Please specify new filter ID",
+                        "filterIdMissingError": "Please specify a new filter ID",
                         "filterId": "Filter ID",
                         "filterRules": "Filter rules"
                     }
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -265,6 +265,41 @@ describe('Deployments View widget', () => {
             });
         });
 
+        it('should allow to save new filter', () => {
+            const newFilterId = `${filterId}-new`;
+            cy.deleteDeploymentsFilter(newFilterId, { ignoreFailure: true });
+
+            useDeploymentsViewWidget();
+
+            cy.interceptSp('PUT', `/filters/deployments/${newFilterId}`).as('filterCreate');
+
+            cy.contains('button', 'Filter').click();
+            cy.get('.modal').within(() => {
+                cy.contains('.selection', 'Type in values')
+                    .find('input')
+                    .type(`${blueprintName}{enter}`, { force: true });
+
+                cy.contains('.buttons', 'Save').find('.ui.dropdown').click();
+                cy.contains('Save as').click();
+                cy.get('input[placeholder="Enter new filter ID..."]').type(newFilterId);
+                cy.contains('Save new filter').click();
+
+                cy.contains('Save new filter').should('not.exist');
+                cy.contains('.field', 'Filter ID').contains(newFilterId);
+            });
+
+            cy.wait('@filterCreate').then(({ request }) => {
+                const requestRules = request.body.filter_rules;
+                expect(requestRules).to.have.length(1);
+                expect(requestRules[0]).to.deep.equal({
+                    type: FilterRuleType.Attribute,
+                    key: FilterRuleAttribute.Blueprint,
+                    values: [blueprintName],
+                    operator: FilterRuleOperators.Contains
+                });
+            });
+        });
+
         it('should prevent modified system filter from being saved', () => {
             useDeploymentsViewWidget();
 
--- a/test/cypress/integration/widgets/deployments_view_spec.ts
+++ b/test/cypress/integration/widgets/deployments_view_spec.ts
@@ -265,9 +265,15 @@ describe('Deployments View widget', () => {
             });
         });
 
-        it('should allow to save new filter', () => {
+        it('should allow to save new filter and apply it', () => {
             const newFilterId = `${filterId}-new`;
             cy.deleteDeploymentsFilter(newFilterId, { ignoreFailure: true });
+            const newFilterRule = {
+                type: FilterRuleType.Attribute,
+                key: FilterRuleAttribute.Blueprint,
+                values: [blueprintName],
+                operator: FilterRuleOperators.Contains
+            };
 
             useDeploymentsViewWidget();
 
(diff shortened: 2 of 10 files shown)


Gelio:
I'll take a look at this PR tomorrow :slightly_smiling_face:

app/translations/en.json
```diff
@@ -713,9 +713,13 @@
                         "header": "Select a filter",
                         "submit": "Apply",
                         "save": "Save",
+                        "saveAs": "Save as...",
```
Gelio, line 716: To me, the following would look more natural:

```suggestion
                        "saveAs": "Save as",
```

or with a space

```suggestion
                        "saveAs": "Save as ...",
```

The same comment applies to `filterIdPlaceholder`

app/translations/en.json
```diff
@@ -713,9 +713,13 @@
...
+                        "saveNew": "Save new filter",
+                        "filterIdPlaceholder": "Enter new filter ID...",
                         "noFilterError": "Please select a filter or specify filter rules",
                         "noFilterRulesError": "Please specify filter rules",
                         "filterRulesInvalidError": "Please provide all the values in filter rules section",
+                        "filterIdMissingError": "Please specify new filter ID",
```
Gelio, line 722:
```suggestion
                        "filterIdMissingError": "Please specify a new filter ID",
```

to be more grammatically correct :smile:

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -265,6 +265,41 @@ describe('Deployments View widget', () => {
...
+                cy.contains('.selection', 'Type in values')
+                    .find('input')
+                    .type(`${blueprintName}{enter}`, { force: true });
+
+                cy.contains('.buttons', 'Save').find('.ui.dropdown').click();
+                cy.contains('Save as').click();
+                cy.get('input[placeholder="Enter new filter ID..."]').type(newFilterId);
+                cy.contains('Save new filter').click();
+
+                cy.contains('Save new filter').should('not.exist');
```
Gelio, line 287: Minor:

I believe we could combine these

```suggestion
                cy.contains('Save new filter').click().should('not.exist');
```

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -265,6 +265,41 @@ describe('Deployments View widget', () => {
...
+            cy.wait('@filterCreate').then(({ request }) => {
+                const requestRules = request.body.filter_rules;
+                expect(requestRules).to.have.length(1);
+                expect(requestRules[0]).to.deep.equal({
+                    type: FilterRuleType.Attribute,
+                    key: FilterRuleAttribute.Blueprint,
+                    values: [blueprintName],
+                    operator: FilterRuleOperators.Contains
+                });
+            });
```
Gelio, line 300: Maybe we could also apply the filter and make sure that it actually works as intended? This way we would not have to assert that the body is correct, but do more of an end-to-end test :slightly_smiling_face: We could keep both parts too, of course :slightly_smiling_face:

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -265,6 +265,41 @@ describe('Deployments View widget', () => {
...
+            const newFilterId = `${filterId}-new`;
+            cy.deleteDeploymentsFilter(newFilterId, { ignoreFailure: true });
+
+            useDeploymentsViewWidget();
+
+            cy.interceptSp('PUT', `/filters/deployments/${newFilterId}`).as('filterCreate');
+
+            cy.contains('button', 'Filter').click();
+            cy.get('.modal').within(() => {
+                cy.contains('.selection', 'Type in values')
```
Gelio, line 278: Could we use some more descriptive selector, like:

```suggestion
                cy.contains('.fields', 'Blueprint').contains('.selection', 'Type in values')
```

This will express in the code that we're looking for the Blueprint filter (which could be inferred by the fact that we're putting `blueprintName` there, but I hope the added context does not hurt)

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -189,12 +235,32 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
+                            onClick={handleSave}
+                        />
+                        <Dropdown
+                            className="button icon"
+                            options={[{ text: tModal('saveAs'), onClick: activateSaveAs, key: '' }]}
+                            trigger={<></>}
+                            style={{ position: 'unset' }}
+                        />
+                    </Button.Group>
+                )}
```
Gelio, line 263: Optional:

Could we extract a component out of these, so that the `FilterModal` gets a bit shorter and maybe less complex :slightly_smiling_face: ?

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -56,19 +56,23 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
     // @ts-expect-error DynamicDropdown is not converted to TS yet
     const { DynamicDropdown } = Stage.Common;
-    const { useBoolean, useErrors } = Stage.Hooks;
+    const { useBoolean, useErrors, useInput } = Stage.Hooks;
 
     const { errors, setErrors, clearErrors, setMessageAsError } = useErrors();
     const [rulesErrorsPresent, setRuleErrorsPresent] = useState<boolean>();
 
     const [filterSaving, setFilterSaving, unsetFilterSaving] = useBoolean();
+    const [saveAsActive, activateSaveAs, deactivateSaveAs] = useBoolean();
```
Gelio, line 74: Minor:

```suggestion
    const [saveAsModeActive, activateSaveAsMode, deactivateSaveAsMode] = useBoolean();
```

The same applies to other variable/function names, to generally rename `saveAs` to `saveAsMode`

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -189,12 +235,32 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
-                />
+                {saveAsActive ? (
+                    <div style={{ float: 'left' }}>
+                        <Input
+                            placeholder={tModal('filterIdPlaceholder')}
+                            style={{ marginRight: '0.25em' }}
+                            value={newFilterId}
+                            onChange={setNewFilterId}
+                        />
+                        <Button icon="cancel" onClick={handleSaveAsCancel} />
```
Gelio, line 246: Minor:

Can we add an `aria-label` or a `title` here?

Gelio: Minor:

How about testing that the "cancel" button works in the tests?

widgets/common/src/deploymentsView/header/FilterModal.tsx
```diff
@@ -189,12 +235,32 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
...
+                    <Button.Group style={{ float: 'left', position: 'relative' }}>
+                        <Button
+                            content={tModal('save')}
+                            disabled={!filterId.value || !filterDirty.value || filterRulesResult.data?.is_system_filter}
+                            onClick={handleSave}
+                        />
+                        <Dropdown
+                            className="button icon"
+                            options={[{ text: tModal('saveAs'), onClick: activateSaveAs, key: '' }]}
+                            trigger={<></>}
```
Gelio, line 259: Clever :smile:

Gelio (commented):
The logic looks clean :+1: Nice work

Hmm looks like the system tests failed - worth investigating :slightly_smiling_face: It's also failing for me locally

image

app/translations/en.json
kubama: I think an ellipsis as a common way to express that certain action needs some more user input. It's also usually used without the extra space. For example, please see the following WebStorm menu:
image

test/cypress/integration/widgets/deployments_view_spec.ts
kubama: Yes, extended.

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: Applied both suggestions

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: Applied

app/translations/en.json
Gelio: Yup, you're right, thanks for the pushback :+1:

test/cypress/integration/widgets/deployments_view_spec.ts
Gelio: Thanks :+1:

Gelio (commented):
I've taken a look again, no further comments aside from the one already opened :slightly_smiling_face:

widgets/common/src/deploymentsView/header/FilterModal.tsx
kubama: Done

widgets/common/src/deploymentsView/header/FilterModalSaveButton.tsx
```diff
@@ -0,0 +1,53 @@
+import { FunctionComponent, useEffect } from 'react';
+import { tModal } from './FilterModal';
```
Gelio, line 2: I'm afraid this is a cyclic dependency, as `FilterModal` imports `FilterModalSaveButton`, and we do an import in the reverse direction. AFAIK, webpack handles these, but generally, they should be avoided as they can be confusing to debug.

Can we extract this `tModal` to some `common.ts`/`utils.ts`?

widgets/common/src/deploymentsView/header/FilterModalSaveButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface SaveButtonProps {
+    saveDisabled: boolean;
+    onSave: () => void;
+    onSaveAsCancel: () => void;
+    onSaveAsSubmit: (newFilterId: string) => Promise<unknown>;
```
Gelio, line 8: Minor:

`unknown` means that the caller (`SaveButton` in this case) does something with the return value, but it can handle any value. I would say `void` would be better in this case, as `void` signifies that the caller does not use the return value (but the `Promise` can still resolve with any value)

I have not tested myself whether this will won't cause TS errors, but I would advise trying to change `Promise<unknown>` to `Promise<void>`

Gelio (commented):
Looks generally okay :+1: The only significant thing from this review IMO is the cyclic dependency to be broken apart :slightly_smiling_face:

widgets/common/src/deploymentsView/header/FilterModalSaveButton.tsx
kubama: Done

widgets/common/src/deploymentsView/header/FilterModalSaveButton.tsx
kubama: Done

Gelio (approved):
LGTM :+1:

PR #1460, opened 2021-06-24
RD-1848 Add vertical slider in Deployments View

Gelio:
This PR add a vertical slider to allow resizing the panes in the Deployments View, as well a Cypress test for it.

The library that I used (https://github.com/tomkp/react-split-pane) is [what we are using in Composer](https://github.com/cloudify-cosmo/cloudify-blueprint-composer/blob/f71b49cb1afed547d8c7af80d4b7b512e5ef71eb/package.json#L60).

Also, I had to change the way the `react-grid-layout` gets the `width` prop. The default `WidthProvider` only reacts to window `resize` events and not to other ways the grid can be resized. Thus, I have used [`react-sizeme`](https://github.com/ctrlplusb/react-sizeme) (recommended at the bottom of https://github.com/react-grid-layout/react-grid-layout#providing-grid-width)

## Pending items

I have asked Ofer in [the comments of RD-1848](https://cloudifysource.atlassian.net/browse/RD-1848?focusedCommentId=72343) whether the layout (pane widths) should be persisted and restored on component mount. I have yet to receive an answer.

If that will be the requirement, I will add persisting the percentage width to `localStorage`. The change should be minimal, so IMO the PR is reviewable ATM.

## Video

https://user-images.githubusercontent.com/889383/123272893-77b60300-d502-11eb-850d-8381c20350c1.mp4

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/847/pipeline

Queued 2021-06-25 16:28 CEST


app/components/layout/Grid.tsx
```diff
@@ -1,18 +1,20 @@
-// @ts-nocheck File not migrated fully to TS
```
Gelio, line 1: I have decided to migrate the component to TS. It was rather easy

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { FunctionComponent, useMemo, useState } from 'react';
 import { useQuery } from 'react-query';
+import SplitPane from 'react-split-pane';
```
Gelio, line 3: Keep in mind that `react-split-pane` will be imported into the widget's bundle, not the `vendor` bundle. IMO that is not a problem, as it's a _local_ widget dependency, not something that's reused across widgets. It can be pulled into the `Stage.Shared` later

kubama (commented):
The code looks just fine 👍.

Regarding the introduced `react-sizeme`, isn't it something that serves the same purpose as the recently introduced `widgets/executions/src/tasksGraph/useWidthObserver.ts`? If yes, maybe it would be good to make things consistent by using the same solution in both places?

Gelio:
Good point, I'll take a look at `useWidthObserver` and try to make it consistent :slightly_smiling_face: I'll also fix the failing unit test

Both of these - tomorrow

Gelio:
There are some very weird errors in https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/840/artifacts that seem unrelated to each other. Here's a rerun to make sure they are actually valid ones:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/842/pipeline

Gelio:
Looks like there are valid edit mode failures. I'll investigate them

Gelio:
## `useWidthObserver` vs `react-sizeme`

I ended up uninstalling `react-sizeme` as `useWidthObserver` seems to be doing the job. I moved it to `Stage.Hooks` to make it more reusable, since we need it in widgets and in the core app.

I tried using `react-sizeme` instead of `useWidthObserver` in the `ExecutionWorkflowGraph`, but due to `SizeMe` accepting a render prop, there would be more changes than I was comfortable with doing.

## Failing unit tests

There were some problems with using `react-sizeme` in the tests. I fixed them initially, but since we're doing away with that library, I ended up only improving the quality of the tests.

## Failing Cypress tests

It turns out that due to using `React.Children.map`, the keys of `GridItem`s were modified. I ended up going back to `_.map` to preserve the keys.

---

Here's a rerun of system tests:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/845/pipeline

kubama (approved):

Gelio:
Urgh... I hit RD-2598 in [the system tests](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/845/artifacts). Here's a rerun:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/847/pipeline

---

Besides, looks like I need to mock `ResizeObserver` in the unit tests and fix the size limits. I'll do that shortly, but the review can without any issue wait till Monday

Gelio:
Some changes later, I believe I have fixed the unit tests, so the multibranch CI should pass. Since I have modified the application's code, here's a new rerun of the system tests:

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/847/pipeline

I have cancelled the previous one, but the link is the same as for the previous one :slightly_smiling_face:

test/jest/setup.ts
```diff
@@ -45,6 +45,9 @@ global.log = log;
 global.Stage = { defineCommon: noop };
 global.PropTypes = PropTypes;
 global.React = React;
+global.ResizeObserver = function ResizeObserver() {
```
kubama, line 48: Cosmetic: why not anonymous function?

kubama (dismissed):

test/jest/setup.ts
Gelio: image

An arrow function is not a valid constructor, so `new` would not work, and with an anonymous function, the `name` property is missing, so it could lead to worse debugging experience (the class may not have a name) :slightly_smiling_face:

kubama: Right, thanks for explanation

Gelio:
[The system tests](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/847/pipeline/) passed. A screenshot, in case the job gets erased over the weekend:
image

Thus, it's just a matter of resolving the merge conflicts between we merge the PR. I'll update the PR on Monday

Gelio:
I've resolved the conflicts (only `.size-limit`). I'm looking for reapproval (@kubama ) and review (@stepek )

stepek (approved):
LGTM!

kubama (dismissed):

Gelio:
Thanks for the reviews :smile: 

Looks like I need to change the size limit once again :slightly_frowning_face: A single approval should be sufficient

kubama (approved):

PR #1461, opened 2021-06-25
 RD-266 Remove jQuery from CreateTemplateModal.

stepek:


kubama:
Two general comments after briefly going through the changes:
1. The `CreateTemplateModal.tsx` seems to be moved, but in this PR it's deleted and created in different dir. This makes reviewing changes in this file extremely difficult, as it's not known what was actually changed. Could you please correct this? I think it should be fairly simple to do by creating a patch, reverting the change, moving the file and applying the patch on it.
2. I still encourage to invest time in Cypress tests instead of enzyme tests 🙂.

app/components/templates/CreateTemplateModal/useFormState.ts
```diff
@@ -0,0 +1,51 @@
...
+                        break;
+                    case 'pages':
+                        setPages(val as string[]);
+                        break;
+                    default:
+                }
+            });
+        },
+        [setTemplateName, setRoles, setTenants, setPages]
+    );
```
Gelio, line 39: I believe the following would be a bit shorter and provide equal functionality:

```ts
    const [formState, setFormState] = useState(initialState);
    const setState = (value: Partial<FormState>) => setFormState(prevState => ({ ...prevState, ...value }));
    const clearState = () => setFormState({ templateName: '', roles: [], tenants: [], pages: [] });
```

Of course, that does not include the default values for `initialState` or `useCallback` :slightly_smiling_face:

app/components/templates/CreateTemplateModal/useFormState.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const clearState = useCallback(() => {
+        setTemplateName('');
+        setRoles([]);
+        setTenants([]);
+        setPages([]);
+    }, [setTemplateName, setRoles, setTenants, setPages]);
```
Gelio, line 46: General:

Do we want to clear values or revert to the default ones? I don't know, this is a general question, as I see that `initialState` is available :slightly_smiling_face:

stepek:
> 1. The `CreateTemplateModal.tsx` seems to be moved, but in this PR it's deleted and created in different dir. This makes reviewing changes in this file extremely difficult, as it's not known what was actually changed. Could you please correct this? I think it should be fairly simple to do by creating a patch, reverting the change, moving the file and applying the patch on it.

Will add one commit more before splitting to separate files

test/jest/components/templates/CreateTemplateModal.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        }
+    ];
+    const pages: string[] = ['Hoth'];
+
+    it('should render component with only required props', () => {
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        const wrapper = mount(<CreateTemplateModal {...defaultProps} />);
+
+        expect(wrapper).toHaveLength(1);
+    });
```
Gelio, line 45: What's the value of this test? Shouldn't TypeScript validate it?

test/jest/components/templates/CreateTemplateModal.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    it('should render component with 2x Messages', () => {
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        const wrapper = shallow(<CreateTemplateModal {...defaultProps} />);
+        const messages = wrapper.find(Segment).find(List).find(Message);
+
+        expect(messages).toHaveLength(2);
+
+        expect(messages.at(0).prop('content')).toEqual('No pages available');
+        expect(messages.at(1).prop('content')).toEqual('No pages selected');
```
Gelio, line 55: I'm not a fan of such white box testing. When refactoring the structure there is a high risk that the test will fail even though the messages will still be shown, just in a different way.

The same comment applies to other tests that are querying by React components, not DOM nodes, and are using props instead of user-visible values.

Can we instead check if the actual messages are shown somewhere in the DOM, no matter where? Refactoring the tests to Cypress would also be nice :smile:

app/components/templates/CreateTemplateModal/SortablePageItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Icon, List, Ref } from '../../basic';
+
+interface SortablePageItemProps {
+    name: string;
+    onRemove: (value: string) => void;
+}
+const SortablePageItem: FunctionComponent<SortablePageItemProps> = ({ name, onRemove }) => {
+    const { attributes, listeners, setNodeRef, transform, transition } = useSortable({ id: name });
+
+    const style = {
```
Gelio, line 16: Minor:

Can we have a type annotation for `style`? I believe it should be `CSSProperties`

app/components/templates/CreateTemplateModal/SortablePageItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+import i18n from 'i18next';
+
+import { useSortable } from '@dnd-kit/sortable';
+import { CSS } from '@dnd-kit/utilities';
+
+import { Icon, List, Ref } from '../../basic';
+
+interface SortablePageItemProps {
+    name: string;
+    onRemove: (value: string) => void;
```
Gelio, line 11: Minor, general:

This technically allows one item to tell that another item was removed :slightly_smiling_face: Probably not what we want. For more correctness, I would go for

```suggestion
    onRemove: () => void;
```

and the consumer using partial application/higher order functions to supply the `name`  to the callback

```tsx
name={name}
onRemove={() => onRemove(name)}
```

app/components/templates/CreateTemplateModal/SortablePageItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+        transform: CSS.Transform.toString(transform),
+        transition
+    };
+
+    const handleRemove = useCallback(() => {
+        onRemove(name);
+    }, [name]);
+
+    return (
+        <Ref innerRef={setNodeRef}>
```
Gelio, line 26: I believe this is the first time I see this component being used. What's the use case :smile: ? Is it a semantic replacement for `<div ref={setNodeRef}>`?

Gelio (commented):
This is only a partial review. I have not reviewed the `CreateTemplateModal.tsx` and did not check for the correctness of the component transformation :slightly_smiling_face: 

Unfortunately, this is a very big PR to review that mixes up functional changes and a refactor, which adds to the difficulty. We will probably spend this week getting this finalized :slightly_smiling_face: For the future, please try to prioritize multiple smaller PRs over a single large one

When making a function change + a large refactor, I encourage splitting that into 2 commits:
1. Use dnd-kit instead of jquery
2. Refactor the component to a function one

This way I could better assess the changes :slightly_smiling_face:

stepek:
CI https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/853/

app/components/templates/CreateTemplateModal/SortablePageItem.tsx
stepek: nop its more complex component https://github.com/microsoft/fluentui/blob/master/packages/fluentui/react-component-ref/src/Ref.tsx with finding correct dom element in child.

app/reducers/tenantsReducer.ts
```diff
@@ -6,7 +6,19 @@
...
+
+// TODO check TenantState what fields should be required
+export interface TenantState {
+    isFetching: boolean;
+    items: TenantDefinition[];
+    selected: string;
+    lastUpdated;
+}
+
+const tenants = (state: Record<string, any> = {}, action) => {
```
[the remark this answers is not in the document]
stepek: probably yes, but i don't want to do bigger refactoring, will check errors after defining `TenantState` as type of `state`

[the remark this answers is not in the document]
Gelio: I see :slightly_smiling_face: In that case, not to leave the code in this limbo state, how about adding a comment explaining the type mismatch?

app/components/templates/CreateTemplateModal/SortablePageItem.tsx
Gelio: I see, thanks! Today I learned :smile:

app/reducers/tenantsReducer.ts
```diff
@@ -6,7 +6,19 @@
 import _ from 'lodash';
 import * as types from '../actions/types';
 
-const tenants = (state = {}, action) => {
+export interface TenantDefinition {
+    name: string;
+}
+
+// TODO check TenantState what fields should be required
```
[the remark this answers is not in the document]
stepek: Fixed

app/components/templates/CreateTemplateModal/useFormState.ts
stepek: right there should be used `initialState`

app/components/templates/CreateTemplateModal/useFormState.ts
```diff
@@ -0,0 +1,51 @@
+import { useCallback, useState } from 'react';
+
+interface FormState {
+    templateName: string;
+    roles: string[];
+    tenants: string[];
+    pages: string[];
+}
+
+const useFormState = (
```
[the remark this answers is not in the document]
stepek: 👍

test/jest/components/templates/CreateTemplateModal.test.tsx
stepek: This test only checking if component render only.

test/jest/components/templates/CreateTemplateModal.test.tsx
stepek: I'm not a fan of e2e test over unit tests, will add cypress test for that feature

Gelio: We can have unit tests, no problem :slightly_smiling_face: My suggestion would be not to rely on anything react-specific in the tests, as IMO that's an anti-pattern. React Testing Library does not allow such operations to avoid shooting ourselves in the foot

Some post that I believe describes the problem with Enzyme's white-box testing approach :slightly_smiling_face: 
https://www.element7.io/2020/01/react-white-box-vs-black-box-testing/

test/jest/components/templates/CreateTemplateModal.test.tsx
Gelio: Isn't it covered by other tests, though? Will the test below (`should render component with 2x Messages`) fail if the component does not render with the default props? I believe it will fail, so IMO this test is redundant :smile: Just my opinion

app/components/templates/CreateTemplateModal/useFormState.ts
```diff
@@ -6,46 +6,27 @@ interface FormState {
...
     );
 
     const clearState = useCallback(() => {
-        setTemplateName('');
-        setRoles([]);
-        setTenants([]);
-        setPages([]);
-    }, [setTemplateName, setRoles, setTenants, setPages]);
+        setFormState({ ...defaultState, ...initialState });
+    }, [formState, setFormState]);
```
Gelio, line 27: Shouldn't it be

```suggestion
    }, [initialState, setFormState]);
```

?

package.json
```diff
@@ -172,6 +172,7 @@
     "imagemin-webpack-plugin": "^2.4.2",
     "ip": "^1.1.3",
     "jest": "^26.4.2",
+    "jest-react-hooks-shallow": "^1.5.1",
```
Gelio, line 175: As far as I have read, the library helps when doing shallow rendering in Enzyme. What if we used regular `mount`? Will it cause problems?

I'm asking as in React Testing Library (RTL), all renders are `mount`s, and that's the recommended way to do more integration testing.

Gelio (commented):
I tried to spend 15 minutes reviewing the `CreateTemplateModal` changes but I can't find a good way to do it with the changes mixed up. The diff from commit https://github.com/cloudify-cosmo/cloudify-stage/pull/1461/commits/7f3b31056286b6ad0b1fa5a6f242d3c560a91fe4 is still pretty hard to parse as
1. Adding TS
2. Changing the component to a function component
3. Swapping out jQuery for dnd-kit

is done in a single step.

To review this PR from me would mean to go through all the already-working code for `CreateTemplateModal` and check if it is right again after the refactor. I am convinced that it would be miles easier if it was 3 separete commits, as then maybe there would not be a need to do a full scan of the code.

Unless @kubama is eager to do a full-review of this PR, I would recommend to start from scratch and make the change as simple as possible and not refactor to a function component, as it was that change that makes the diff so messy :slightly_smiling_face:

package.json
stepek: This lib is not needed

stepek:
Sorry guys for wasting your time, there is one commit with removing only jquery UI from this component. Refactoring to functional component will be later, after merge this one.

app/components/templates/CreateTemplateModal.tsx
```diff
@@ -176,6 +205,12 @@ export default class CreateTemplateModal extends Component {
                 className="createTemplateButton"
             />
         );
+        const sensors: SensorDescriptor<any>[] = [
+            {
+                sensor: PointerSensor,
```
Gelio, line 210: Minor:

We should probably restrict the drag direction to vertical only, as it is misleading ATM - it looks as if the user is able drag the pages between the right and left lists, but in fact it's impossible :smile: 


https://user-images.githubusercontent.com/889383/123923389-394c9800-d989-11eb-9f3b-cd7b1af63182.mp4

Gelio (approved):
Much easier to review :+1: Thanks for reducing the scope of the changes :smile: I appreciate it

I've tested the PR in the UI and it seems to work fine. Let's just run the tests and it should be good to go :slightly_smiling_face:

kubama (dismissed):

stepek:
CI https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/865/

Gelio (approved):
I did not test the latest version in the UI. I hope it works well

stepek:
Thanks!

PR #1463, opened 2021-06-28
RD-2500 Remove jQuery dependency from External.ts.

stepek:


app/utils/External.ts
```diff
@@ -5,6 +5,8 @@ import log from 'loglevel';
 import 'isomorphic-fetch';
 import { saveAs } from 'file-saver';
 import JSZip from 'jszip';
+import { stringify as queryQtringify } from 'query-string';
```
kubama, line 8: `qtringify`? Is it a typo 🙂?

app/utils/External.ts
```diff
@@ -282,9 +284,7 @@ export default class External {
 
     // eslint-disable-next-line class-methods-use-this
     protected buildActualUrl(url: string, data?: Record<string, any>) {
-        // TODO: RD-258
-        // @ts-ignore Cannot find $
-        const queryString = data ? (url.indexOf('?') > 0 ? '&' : '?') + $.param(data, true) : '';
+        const queryString = data ? (url.indexOf('?') > 0 ? '&' : '?') + queryStringify(data) : '';
```
Gelio, line 287: Did you consider using https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams to build the query string?

This way we won't be pulling in third-party code :slightly_smiling_face:

test/jest/utils/External.test.ts
```diff
@@ -0,0 +1,18 @@
+import External from 'utils/External';
+
+describe('(Utils) External', () => {
+    describe('buildActualUrl (protected)', () => {
+        const external = new External({});
+
+        it(`should return correct url with params`, () => {
```
Gelio, line 7: Minor:

We could also add a test when the URL already contains a query string :slightly_smiling_face:

app/utils/External.ts
Gelio: Actually, nevermind, as it won't handle arrays out of the box :smile: The test uses arrays, not sure if we use them in production, but w/e :smile: It's ok :slightly_smiling_face:

app/utils/External.ts
```diff
@@ -5,6 +5,8 @@ import log from 'loglevel';
 import 'isomorphic-fetch';
 import { saveAs } from 'file-saver';
 import JSZip from 'jszip';
+import { stringify as queryStringify } from 'query-string';
```
Gelio, line 8: Minor:

I would call it

```suggestion
import { stringify as stringifyQueryString } from 'query-string';
```

as it sounds to me more like a function :slightly_smiling_face:

Gelio (dismissed):

kubama (approved):

app/utils/External.ts
stepek: 👍

test/jest/utils/External.test.ts
stepek: as you wish :)

Gelio (approved):

PR #1465, opened 2021-06-28
RD-2593 Migrate BlueprintActionButtons widget to TS

Gelio:
This PR migrates the Blueprint Action Buttons widget to TypeScript. There shouldn't be any functional changes in this PR.

I have decided to separate these 2 commits to another PR in order to make the reverting process easier in 6.2, when the button should be shown. By having the PRs for RD-2593 separated, we could only revert the 2nd PR (the one that hides the "Edit a copy in Composer" button) and not revert this TS migration.

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/856/pipeline

Queued 2021-06-28 14:55 CEST


widgets/blueprintActionButtons/src/BlueprintActionButtons.tsx
```diff
@@ -1,15 +1,34 @@
...
+
+export default class BlueprintActionButtons extends React.Component<
+    BlueprintActionButtonsProps,
+    BlueprintActionButtonsState
+> {
+    // eslint-disable-next-line react/static-property-placement
+    static propTypes = {
+        blueprintId: PropTypes.string,
+        toolbox: Stage.PropTypes.Toolbox.isRequired
+    };
```
kubama, line 24: Do we still need this? If yes, why it was moved from outside of the class to here? Also, why do we need eslint disable?

kubama (commented):
Looks good!
Did you consider converting `BlueprintActionButtons` to functional component at the same time?

Gelio:
> Did you consider converting BlueprintActionButtons to functional component at the same time?

Nope :smile: These conversions are usually not diff-friendly, so I decided not to do it this time and finish the scope earlier :slightly_smiling_face:

widgets/blueprintActionButtons/src/BlueprintActionButtons.tsx
Gelio: Done, I've removed the `propTypes` :slightly_smiling_face:

kubama (approved):

stepek (approved):
Lgtm!

PR #1469, opened 2021-06-29
RD-2331 Added deployment display name to workflow execution modal header

kubama:
System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/862/


widgets/common/src/ExecuteDeploymentModal.tsx
```diff
@@ -111,6 +112,11 @@ export default function ExecuteDeploymentModal({
         clearErrors();
     }, [dryRun, force, schedule, scheduledTime]);
 
+    let deploymentsList = deployments;
+    if (_.isEmpty(deployments)) {
+        deploymentsList = _.compact([deploymentId]);
+    }
```
Gelio, line 118: Minor:

```suggestion
    const deploymentsList: string[] = _.isEmpty(deployments) ? _.compact([deploymentId]) : deployments;
```

has the benefit of not making `deploymentsList` a mutable variable :slightly_smiling_face: 

Also, adding a `string[]` type adds a bit of documentation to the code

widgets/common/src/ExecuteDeploymentModal.tsx
```diff
@@ -30,6 +30,7 @@ function renderActionField(name, checked, onChange) {
 
 export default function ExecuteDeploymentModal({
     deploymentId,
+    deploymentName,
```
Gelio, line 33: This seems not to be included in the `propTypes` and we also don't have TS types for the props. Can you add it to `propTypes` and optionally add TS types for props? :smile:

widgets/common/src/ExecuteDeploymentModal.tsx
```diff
@@ -226,9 +227,11 @@ export default function ExecuteDeploymentModal({
...
-    if (!_.isEmpty(deployments)) {
-        if (_.size(deployments) > 1) {
+    if (!_.isEmpty(deploymentsList)) {
+        if (_.size(deploymentsList) > 1) {
             headerKey += 'multipleDeployments';
+        } else if (deploymentName) {
+            headerKey += 'singleNamedDeployemnt';
         } else {
             headerKey += 'singleDeployemnt';
         }
```
Gelio, line 237: General:

Is there ever a case when there is a deployment but it does not have a name?

widgets/deployments/src/ActionsMenus.tsx
```diff
@@ -1,20 +1,20 @@
 // @ts-nocheck File not migrated fully to TS
-export default function ActionsMenus({ deploymentId, onDeploymentAction, onWorkflowAction, toolbox, workflows }) {
+export default function ActionsMenus({ deployment, onDeploymentAction, onWorkflowAction, toolbox, workflows }) {
```
Gelio, line 2: Minor:

Can we add TS types for the props? This way we could show which properties for `deployment` we expect to be there.

Gelio (commented):
The approach looks good and I don't have anything major, but I left some open questions that may be worth discussing before we conclude the PR :slightly_smiling_face: 

A screenshot in the PR description would be great :slightly_smiling_face:

widgets/common/src/ExecuteDeploymentModal.tsx
kubama: No, but there is a case when the name is not passed to the modal, for example in the context of spire manager widget. This widget is not in scope of display name introduction, so to retain existing functionality it doesn't pass the display name, and the modal renders as before.

Gelio: I see :smile: I assumed it's only used in a few places that we refactor. Thanks for explaining :+1:

test/cypress/integration/widgets/deployment_actions_spec.ts
```diff
@@ -30,6 +31,7 @@ describe('Deployment Action Buttons widget', () => {
             cy.get('button.executeWorkflowButton').click();
             cy.get('.popupMenu > .menu').contains('Start').click();
             cy.get('.executeWorkflowModal').should('be.visible');
+            cy.contains(`Execute workflow start on ${deploymentName} (${deploymentId})`);
             cy.get('.executeWorkflowModal button.ok').click();
```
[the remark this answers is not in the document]
kubama: Yes, refactored.

test/cypress/integration/widgets/deployments_spec.ts
```diff
@@ -3,47 +3,48 @@ import { exampleBlueprintUrl } from '../../support/resource_urls';
...
+        selectDeploymentActionFromMenu(id, '.deploymentActionsMenu', action);
     };
-    const executeWorkflow = (deploymentId, workflow) => {
-        selectActionFromMenu(deploymentId, '.workflowsMenu', workflow);
+    const executeDeploymentWorkflow = (id, workflow) => {
+        selectDeploymentActionFromMenu(id, '.workflowsMenu', workflow);
     };
     const verifyWorkflowIsStarted = () => {
-        cy.contains('div.row', deploymentName).find('.spinner.loading.icon').should('be.visible');
+        cy.contains('div.row', deploymentId).find('.spinner.loading.icon').should('be.visible');
```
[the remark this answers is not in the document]
kubama: At the moment this widget still uses deployment ID in the view. I just renamed `deploymentName` to `deploymentId` as it was used as an ID in fact so now it better reflects what it actually is.

widgets/deployments/src/ActionsMenus.tsx
kubama: In this case the `deployment` is blindly passed to callbacks, so this component does not expect it to be of any particular type.

widgets/common/src/ExecuteDeploymentModal.tsx
kubama: Sure, just added a prop type.

kubama:
> A screenshot in the PR description would be great 🙂

This PR only changes the modal header text, so I didn't consider it a 'visual' change.

stepek (approved):
LGTM!

widgets/deployments/src/ActionsMenus.tsx
Gelio: Ok, maybe it was not the best place to add the comment, but I mean generally - it would be nice to have the types :smile: 

Anyway, it's minor, I'll resolve the comment

Gelio (approved):
LGTM. I did not run it in the UI, though :slightly_smiling_face:

PR #1470, opened 2021-06-29
RD-2593 Migrate Blueprints widget to TS

Gelio:
The PR migrates the _Blueprints_ widget to TypeScript. It is very similar to https://github.com/cloudify-cosmo/cloudify-stage/pull/1465 - the aim is to not introduce any functional changes, but instead do a pure TS migration.

Changes to the configuration to include a flag for toggling whether _Edit in Composer_ button should be visible (which is the main point of RD-2593) will be done in a separate PR, so it's easier to revert it.

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/863/pipeline

Queued 2021-06-29 20:51 CEST


widgets/blueprints/src/props/BlueprintsViewDefaultProps.ts
```diff
@@ -1,8 +0,0 @@
-export default {
-    fetchData: () => {},
```
Gelio, line 2: There is no point in having default values if the props are always provided in the component hierarchy

kubama (approved):

widgets/blueprints/src/BlueprintsList.tsx
```diff
@@ -1,15 +1,27 @@
...
+    widget: Stage.Types.Widget<BlueprintsWidgetConfiguration>;
+    data: BlueprintDataResponse;
+}
+
+interface BlueprintListState {
+    showDeploymentModal: boolean;
+    showUploadModal: boolean;
+    blueprintId: string;
+    confirmDelete: boolean;
+    error: any;
```
stepek, line 18: Can we define error type? I see this ( `error: any;`) 2nd time. Maybe in `ErrorMessage` component ?

stepek (dismissed):
One minor comment.
LGTM!

widgets/blueprints/src/BlueprintsList.tsx
Gelio: Probably we could, but I don't know what's the type of error and it was easily inspectable in the UI. I will skip the comment this time, but I will investigate the solutions as I'll be adding an `error: any` prop again :slightly_smiling_face:

Gelio:
I have rebased the code on top of 6.1.0 since #1465 was merged. I did not change any code

Gelio:
I need a re-approval here :slightly_smiling_face:

kubama (approved):

PR #1473, opened 2021-06-30
Remove jQuery-UI

stepek:
Removes jQuery UI from stage.


package.json
```diff
@@ -57,8 +57,6 @@
     "history": "^4.7.2",
     "i18next": "^19.8.3",
     "isomorphic-fetch": "^3.0.0",
-    "jquery": "^3.5.0",
```
stepek, line 60: Should I remove `jquery` ot it need to stay ?

Gelio: If we don't use it anywhere, do we need it?

Gelio (commented):
What did you test in this PR? The change seems small and easy, but the scope is large, so it should definitely be validated thoroughly :slightly_smiling_face: 

Besides, I do see:
```
12:47 $ rg '\$\.'
app/utils/Manager.ts
60:            queryString += $.param(params, true);
63:        const queryString = data ? (url.indexOf('?') > 0 ? '&' : '?') + $.param(data, true) : '';
```

Shouldn't it be removed already?

Also, I ran `rg '\$[^{/]'` and there are some results such as:

```
app/components/WidgetDynamicContent.tsx
232:                        $(container).find(event.selector).off(event.event);
233:                        $(container)
247:            widget.definition.postRender($(container), widget, data.data, this.getToolbox());

app/components/layout/ScrollToTop.ts
16:            $('div.main').scrollTop(0);

test/jest/setup.ts
8:import $ from 'jquery';
36:global.$ = $;
37:global.jQuery = $;
38:global.window.$ = $;
39:global.window.jQuery = $;

app/utils/SplashLoadingScreen.ts
6:const SPLASH_DIV = $('div.splashPage');

app/app.tsx
48:window.$ = $;
```

So looks like there is some more work to do

package.json
stepek: We use them

stepek:
> What did you test in this PR? The change seems small and easy, but the scope is large, so it should definitely be validated thoroughly 🙂
> 
> Besides, I do see:
> 
> ```
> 12:47 $ rg '\$\.'
> app/utils/Manager.ts
> 60:            queryString += $.param(params, true);
> 63:        const queryString = data ? (url.indexOf('?') > 0 ? '&' : '?') + $.param(data, true) : '';
> ```
> 
> Shouldn't it be removed already?
> 
> Also, I ran `rg '\$[^{/]'` and there are some results such as:
> 
> ```
> app/components/WidgetDynamicContent.tsx
> 232:                        $(container).find(event.selector).off(event.event);
> 233:                        $(container)
> 247:            widget.definition.postRender($(container), widget, data.data, this.getToolbox());
> 
> app/components/layout/ScrollToTop.ts
> 16:            $('div.main').scrollTop(0);
> 
> test/jest/setup.ts
> 8:import $ from 'jquery';
> 36:global.$ = $;
> 37:global.jQuery = $;
> 38:global.window.$ = $;
> 39:global.window.jQuery = $;
> 
> app/utils/SplashLoadingScreen.ts
> 6:const SPLASH_DIV = $('div.splashPage');
> 
> app/app.tsx
> 48:window.$ = $;
> ```
> 
> So looks like there is some more work to do

Let me create more subtasks regarding this places in code

stepek:
CI https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/869/

package.json
Gelio: If we use them, then I don't see why we should remove those from `package.json` :slightly_smiling_face:

Gelio:
Hmm... I don't get it now - if we're still using jQuery, then what does this PR do? Is it supposed to only remove jQuery UI, which we don't use now? A good description summarizing the current state of the PR could help :slightly_smiling_face: 

Besides, I still see uses of jQuery when executing `rg '\$\.'` and `rg '\$[^{/]' app` (btw. `rg` is https://github.com/BurntSushi/ripgrep) as I said in https://github.com/cloudify-cosmo/cloudify-stage/pull/1473#pullrequestreview-695970391 :slightly_smiling_face:

stepek:
@Gelio I know this can be some messy, but in ticket is described only removing jquery UI, probably there is some missing tasks regarding other places jquery places in the code. So let my create missing tasks for that and remove jQuery itself  in separate PR.

app/app.tsx
```diff
@@ -50,8 +45,6 @@ import Routes from './containers/Routes';
 import translation from './translations/en.json';
 import LoaderUtils from './utils/LoaderUtils';
 
-window.$ = $;
```
Gelio, line 53: If we're only removing jQuery UI, should we remove this line?

package.json
Gelio: Makes sense to only remove `jquery-ui`, so it seems clear to me now

package-lock.json
```diff
@@ -21913,14 +21913,9 @@
...
     "jquery": {
-      "version": "3.5.0",
-      "resolved": "https://registry.npmjs.org/jquery/-/jquery-3.5.0.tgz",
-      "integrity": "sha512-Xb7SVYMvygPxbFMpTFQiHh1J7HClEaThguL15N/Gg37Lri/qKyhRGZYzHRyLH8Stq3Aow0LsHO2O2ci86fCrNQ=="
-    },
-    "jquery-ui": {
-      "version": "1.12.1",
-      "resolved": "https://registry.npmjs.org/jquery-ui/-/jquery-ui-1.12.1.tgz",
-      "integrity": "sha1-vLQEXI3QU5wTS8FIjN0+dop6nlE="
+      "version": "3.6.0",
```
Gelio, line 21916: I see that jQuery version is upgraded to 3.6.0 now. Do you think it's safe?

I don't see anything concerning, but maybe you could also take a look

https://blog.jquery.com/2021/03/02/jquery-3-6-0-released/

Gelio (commented):
> but in ticket is described only removing jquery UI

Makes sense now :slightly_smiling_face: I did not check very thoroughly if that's all the places that use jQuery UI, but after doing some quick searches, looks good to me :slightly_smiling_face: I am not confident about removing `window.$ = $`, though, so that's my only concern right now

package-lock.json
stepek: lets stay with 3.5.0

Gelio (approved):
If the system tests pass, we should be good :smile: I did not do any manual testing

app/app.tsx
```diff
@@ -50,12 +45,12 @@ import Routes from './containers/Routes';
...
-
 const browserHistory = createBrowserHistory({
     basename: Consts.CONTEXT_PATH
 });
 
+window.$ = $;
+
```
kubama, line 53: Not sure why this line was moved - I think leaving it at its original position would make the diff smaller, which would be minor but still a benefit.

Gelio (approved):

kubama (approved):

PR #1479, opened 2021-06-30
RD-2335 Added deployment name column to nodes list widget

kubama:
The pre-existing Deployment column was renamed to Deployment ID and is hidden by default. As a result nothing has changed in widget appearance except that the Deployment column now shows deployment name instead of deployment ID.

System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/871

### DIFF
--- a/test/cypress/integration/widgets/nodes_spec.ts
+++ b/test/cypress/integration/widgets/nodes_spec.ts
@@ -1,0 +1,45 @@
+describe('Nodes list widget', () => {
+    const blueprintName = 'nodes_list_test';
+    const deployment1Id = `${blueprintName}_dep1_id`;
+    const deployment1Name = `${blueprintName}_dep1_name`;
+    const deployment2Id = `${blueprintName}_dep2_id`;
+    const deployment2Name = `${blueprintName}_dep2_name`;
+
+    before(() =>
+        cy
+            .activate('valid_trial_license')
+            .usePageMock('nodes', { fieldsToShow: ['Deployment', 'Deployment ID'] })
+            .mockLogin()
+            .deleteDeployments(blueprintName, true)
+            .deleteBlueprints(blueprintName, true)
+            .uploadBlueprint('blueprints/simple.zip', blueprintName)
+            .deployBlueprint(
+                blueprintName,
+                deployment1Id,
+                { server_ip: '127.0.0.1' },
+                { display_name: deployment1Name }
+            )
+            .deployBlueprint(
+                blueprintName,
+                deployment2Id,
+                { server_ip: '127.0.0.1' },
+                { display_name: deployment2Name }
+            )
+    );
+
+    it('should display nodes list', () => {
+        cy.setBlueprintContext(blueprintName);
+        cy.get('tbody tr').should('have.length', 2);
+        cy.get('table')
+            .getTable()
+            .should(tableData => {
+                expect(tableData[0].Deployment).to.eq(deployment1Name);
+                expect(tableData[0]['Deployment ID']).to.eq(deployment1Id);
+                expect(tableData[1].Deployment).to.eq(deployment2Name);
+                expect(tableData[1]['Deployment ID']).to.eq(deployment2Id);
+            });
+
+        cy.setDeploymentContext(deployment1Id);
+        cy.contains('th', 'Deployment').should('not.exist');
+    });
+});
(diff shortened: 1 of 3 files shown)


stepek (approved):
lgtm!

test/cypress/integration/widgets/nodes_spec.ts
```diff
@@ -0,0 +1,45 @@
+describe('Nodes list widget', () => {
```
Gelio, line 1: :medal_military: for adding new tests for the widget :tada:

test/cypress/integration/widgets/nodes_spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    const deployment1Id = `${blueprintName}_dep1_id`;
+    const deployment1Name = `${blueprintName}_dep1_name`;
+    const deployment2Id = `${blueprintName}_dep2_id`;
+    const deployment2Name = `${blueprintName}_dep2_name`;
+
+    before(() =>
+        cy
+            .activate('valid_trial_license')
+            .usePageMock('nodes', { fieldsToShow: ['Deployment', 'Deployment ID'] })
+            .mockLogin()
```
Gelio, line 12: Minor:

Would it not be safer/faster to `mockLogin` after all the blueprints are deployed?

test/cypress/integration/widgets/nodes_spec.ts
```diff
@@ -0,0 +1,45 @@
...
+            .deleteDeployments(blueprintName, true)
```
Gelio, line 13: Minor:

This looks to be optional, as `deleteBlueprints` will delete respective deployments too:

https://github.com/cloudify-cosmo/cloudify-stage/blob/0de621cbc844559d430ec97197e39be486ba49cd/test/cypress/support/blueprints.ts#L33-L35

We can keep it, though, for readability :smile:

test/cypress/integration/widgets/nodes_spec.ts
```diff
@@ -0,0 +1,45 @@
...
+                { server_ip: '127.0.0.1' },
+                { display_name: deployment2Name }
+            )
+    );
+
+    it('should display nodes list', () => {
+        cy.setBlueprintContext(blueprintName);
+        cy.get('tbody tr').should('have.length', 2);
+        cy.get('table')
+            .getTable()
```
Gelio, line 34: Nice, makes the test much more readable :+1:

test/cypress/integration/widgets/nodes_spec.ts
```diff
@@ -0,0 +1,45 @@
...
+            .should(tableData => {
+                expect(tableData[0].Deployment).to.eq(deployment1Name);
+                expect(tableData[0]['Deployment ID']).to.eq(deployment1Id);
+                expect(tableData[1].Deployment).to.eq(deployment2Name);
+                expect(tableData[1]['Deployment ID']).to.eq(deployment2Id);
+            });
+
+        cy.setDeploymentContext(deployment1Id);
+        cy.contains('th', 'Deployment').should('not.exist');
```
Gelio, line 43: Hmm... what does this test exactly? Should we not list that one single deployment with ID `deployment1Id` here?

Gelio: Minor:

If we expect there are no rows, maybe let's assert that `No Results Found` or some other `noDataMessage` is displayed?

Gelio (approved):
Looks generally ok, I do have a few minor clarification questions :smile: I'll approve anyway, as those seem minor

test/cypress/integration/widgets/nodes_spec.ts
kubama: This assertion tests if 'Deployment' and 'Deployment ID' columns get hidden when deployment is set in the context - please note the `th` selector.

test/cypress/integration/widgets/nodes_spec.ts
kubama: Ah, yes, thanks for reminding though.

test/cypress/integration/widgets/nodes_spec.ts
kubama: I think it doesn't matter - the test body starts with setting blueprint context, which causes widget content to be reloaded, so it's perfectly safe. There would be also no speed benefit in switching the order.

Gelio: Ok, makes sense

test/cypress/integration/widgets/nodes_spec.ts
Gelio: Right, makes sense :smile: So it tests the `data.deploymentSelected` part. And, conveniently, it also tests both columns. Clever, almost too clever :smile: 

Thanks for explaining

PR #1480, opened 2021-06-30
RD-2330 Added deployment display name to deployment actions modal headers

kubama:
There are no visual changes - only modals header text changes.

System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/872/

### DIFF
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -467,11 +467,11 @@
                     }
                 },
                 "removeModal": {
-                    "deleteDescription": "Are you sure you want to remove deployment '{{ deploymentId }}'?",
-                    "forceDeleteDescription": "Force delete will ignore any existing live nodes, or existing deployments which depend on this deployment.\nIt's recommended to first run uninstall to stop the live nodes, and make sure there are no running installations which depend on this deployment - and then run delete.\nAre you sure you want to ignore the live nodes and delete the deployment '{{ deploymentId }}'?"
+                    "deleteDescription": "Are you sure you want to remove deployment {{deploymentName}} ({{ deploymentId }})?",
+                    "forceDeleteDescription": "Force delete will ignore any existing live nodes, or existing deployments which depend on this deployment.\nIt's recommended to first run uninstall to stop the live nodes, and make sure there are no running installations which depend on this deployment - and then run delete.\nAre you sure you want to ignore the live nodes and delete the deployment {{deploymentName}} ({{ deploymentId }})?"
                 },
                 "setSiteModal": {
-                    "header": "Set the site of deployment '{{ deploymentId }}'",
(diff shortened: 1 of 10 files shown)


stepek (dismissed):
LGTM!

test/cypress/integration/widgets/deployments_spec.ts
```diff
@@ -125,22 +125,24 @@ describe('Deployments widget', () => {
             cy.interceptSp('POST', `/executions`).as('executeDeploymentWorkflow');
         });
 
-        const startAndVerifyWorkflowExecution = () => {
-            cy.get('.executeWorkflowModal button.ok').click();
+        const startAndVerifyWorkflowExecution = workflow => {
```
Gelio, line 128: Minor:

```suggestion
        const startAndVerifyWorkflowExecution = (workflow: string) => {
```

app/translations/en.json
```diff
@@ -467,11 +467,11 @@
...
+                    "deleteDescription": "Are you sure you want to remove deployment {{deploymentName}} ({{ deploymentId }})?",
+                    "forceDeleteDescription": "Force delete will ignore any existing live nodes, or existing deployments which depend on this deployment.\nIt's recommended to first run uninstall to stop the live nodes, and make sure there are no running installations which depend on this deployment - and then run delete.\nAre you sure you want to ignore the live nodes and delete the deployment {{deploymentName}} ({{ deploymentId }})?"
                 },
                 "setSiteModal": {
-                    "header": "Set the site of deployment '{{ deploymentId }}'",
+                    "header": "Set the site of deployment {{deploymentName}} ({{ deploymentId }})",
```
Gelio, line 474: General comment about always displaying the display name and ID:

IMO it looks a bit redundant and weird when the ID === name and it's repeated twice:
image

As a user, I could even consider it to be a bug :smile: 

Maybe we could have a ticket to make it smarter in the future (post 6.1) and not repeat the ID if it's the same as the name? :thinking: What do you think?

Gelio (dismissed):
Straightforward and easy to review - good job :+1: I've tested in the UI and it looks good :+1: 

Btw. I believe the title of the PR should say RD-2330 instead of RD-1330

kubama:
> Btw. I believe the title of the PR should say [RD-2330](https://cloudifysource.atlassian.net/browse/RD-2330) instead of [RD-1330](https://cloudifysource.atlassian.net/browse/RD-1330)

Yes, of course. Updated.

app/translations/en.json
kubama: I think it's a good suggestion since deployment ID is the default value for display name when it's not specified (at least when created via REST). I created RD-2693.

Gelio: Cool, thanks :+1:

Gelio (approved):

PR #1489, opened 2021-07-02
RD-2329 Add deployment display name to Node filter.

stepek:
image
Changed some fetch data code to use display_name from response, if is not name will have only resource `id`


test/cypress/integration/widgets/agents_spec.ts
```diff
@@ -99,6 +107,7 @@ describe('Agents widget', () => {
             cy.contains('Validate').click();
             cy.get('.modal').within(() => {
                 fillNodesFilter();
+                cy.contains(`${deploymentDisplayName} (${deploymentName})`).should('be.visible');
```
kubama, line 110: How about renaming `deploymentName` to `deploymentId`? I think it would be more accurate name for the variable after introducing this change.

stepek:
## system tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/884/

widgets/common/src/NodeFilter.tsx
```diff
@@ -1,4 +1,6 @@
 // @ts-nocheck File not migrated fully to TS
+import { DropdownItemProps } from 'semantic-ui-react/dist/commonjs/modules/Dropdown/DropdownItem';
```
kubama, line 2: Can't we obtain `DropdownItemProps` having access to `Stage.Basic.Dropdown.Item`?

widgets/common/src/NodeFilter.tsx
```diff
@@ -133,15 +137,26 @@ export default class NodeFilter extends React.Component {
...
             .then(data => {
-                let ids = _.chain(data.items || [])
-                    .map(item => item.id)
-                    .uniqWith(_.isEqual)
-                    .value();
+                let additionalOptions = [];
+
                 if (this.isFilteringSetFor(optionsField)) {
-                    ids = _.intersection(ids, this.getAllowedOptionsFor(optionsField));
+                    additionalOptions = _.intersection(additionalOptions, this.getAllowedOptionsFor(optionsField));
```
kubama, line 143: `additionalOptions` is an empty array at this point, so why the intersection?

widgets/common/src/NodeFilter.tsx
stepek: oh thanks, will change

widgets/common/src/NodeFilter.tsx
stepek: right

widgets/common/src/NodeFilter.tsx
```diff
@@ -140,10 +138,10 @@ export default class NodeFilter extends React.Component {
                 let additionalOptions = [];
 
                 if (this.isFilteringSetFor(optionsField)) {
-                    additionalOptions = _.intersection(additionalOptions, this.getAllowedOptionsFor(optionsField));
+                    additionalOptions = this.getAllowedOptionsFor(optionsField);
```
kubama: Just to make sure: intersection with an empty array is an empty array, regardless of other arguments, so here you actually changed the logic. Are you sure this is the right implementation? Can we assure the logic implemented by this line is verified by any test?

stepek: oh i misunderstood this part of code, this should be done by other way.

stepek: How can i set `allowedDeployments` props in e2e tests?

kubama: Going through the code quickly it looks like it should be always set for a given list of agents.

widgets/common/src/NodeFilter.tsx
```diff
@@ -133,15 +135,28 @@ export default class NodeFilter extends React.Component {
             .getManager()
             .doGet(fetchUrl, { params })
             .then(data => {
-                let ids = _.chain(data.items || [])
-                    .map(item => item.id)
-                    .uniqWith(_.isEqual)
-                    .value();
+                let allowedOptions = [];
```
Gelio, line 138: This file can freely use TypeScript that can make understanding the code easier :slightly_smiling_face: Thus, I suggest we sprinkle a bit of TS here:

```suggestion
                let allowedOptions: string[] = [];
```

also, we could add types in many other places - the function signature:

```ts
    fetchData(fetchUrl: string, params: any, optionsField: string) {
```

widgets/common/src/NodeFilter.tsx
```diff
@@ -133,15 +135,28 @@ export default class NodeFilter extends React.Component {
...
+
                 if (this.isFilteringSetFor(optionsField)) {
-                    ids = _.intersection(ids, this.getAllowedOptionsFor(optionsField));
+                    allowedOptions = this.getAllowedOptionsFor(optionsField);
                 }
 
-                const options = _.map(ids, id => ({ text: id, value: id, key: id }));
+                const options: Stage.Basic.Dropdown.Item.DropdownItemProps[] = Object.entries(
```
Gelio, line 144: I turned on TypeScript (removed `@ts-nocheck`) and it turns out this type is invalid :smile: `Dropdown` does not have a registered `Item` property that TypeScript knows about. Thus, we need to make it a bit different.

Also, it would be easier to understand for me if we moved the type annotation to the return type of the last `map` callback. Thus, the final solution that I suggest would be:

```ts
                ).map(([id, displayName]): NonNullable<
                    ComponentProps<typeof Stage.Basic.Dropdown>['options']
                >[number] => ({
```

for the final `map` call (line 154), and then, removing the type annotation from this line:

```suggestion
                const options = Object.entries(
```

widgets/common/src/NodeFilter.tsx
```diff
@@ -133,15 +135,28 @@ export default class NodeFilter extends React.Component {
...
+                    (data.items || []).reduce((result: Record<string, string>, item) => {
```
Gelio, line 145: Minor:

Can we have a type annotation for `item`?

widgets/common/src/NodeFilter.tsx
```diff
@@ -133,15 +135,28 @@ export default class NodeFilter extends React.Component {
...
+                        if (allowedOptions.length !== 0 && !allowedOptions.includes(item)) {
+                            return result;
+                        }
+
+                        result[item.id] = item.display_name || item.id;
+
+                        return result;
+                    }, {})
```
Gelio, line 153: Can we extract this code to some smaller function that would be named? Having such low-lever operations mixed with high-level operations reduces my comprehension :slightly_smiling_face: I assume this is some kind of a custom intersection function

widgets/common/src/NodeFilter.tsx
```diff
@@ -133,15 +135,28 @@ export default class NodeFilter extends React.Component {
...
+                ).map(([id, displayName]) => ({
+                    value: id,
+                    text: id === displayName ? displayName : `${displayName} (${id})`,
```
Gelio, line 156: Nice :+1:

Gelio (commented):
Looks reasonable :+1: I did not test it in the UI myself :slightly_smiling_face: There are some TS errors in the newly written code, so I would appreciate if we fixed them :smile: They do not show up, because the file has TS checking disabled, but IMO we should not be writing new invalid TypeScript, only fix it :slightly_smiling_face:

widgets/common/src/NodeFilter.tsx
stepek: Sure we could add it in whole of ts file, but the changes will be very big and unneeded regarding this feature. Will add  ts in new code.

widgets/common/src/NodeFilter.tsx
stepek: nice thanks!

widgets/common/src/NodeFilter.tsx
stepek: sure

widgets/common/src/NodeFilter.tsx
Gelio: I agree adding it in the whole file will result in a large diff :slightly_smiling_face: One of the benefits behind having a TS file and a `@ts-nocheck` flag instead of just a JS file is that we can add types gradually, to some parts of the file, resulting in less work in the future and better comprehension for the PR reviewers :slightly_smiling_face: For example, I try to add type annotations for the code that I modify in TS files (even if they don't have types themselves yet) just to have that much work already done. It also helps me write code at the same time :smile: 

Anyway, it's minor, we can let it be for now. I'll resolve the comment

widgets/common/src/NodeFilter.tsx
```diff
@@ -120,6 +132,22 @@ export default class NodeFilter extends React.Component {
...
     };
 
+    optionsReducer(result: Record<string, string>, item: Partial<ResponseItem>): Record<string, string> {
+        let allowedOptions: string[] = [];
+
+        if (this.isFilteringSetFor(optionsField)) {
+            allowedOptions = this.getAllowedOptionsFor(optionsField);
+        }
+
+        if (allowedOptions.length !== 0 && !allowedOptions.includes(item)) {
```
Gelio, line 142: Shouldn't this be

```suggestion
        if (allowedOptions.length !== 0 && !allowedOptions.includes(item.id)) {
```

since `string[]` will never contain an object (`Partial<ResponseItem>`) :thinking: In the previous logic there was an intersection of `ids` (`string[]`) with the options, so IMO we should change it :slightly_smiling_face:

Gelio (commented):
Just one small comment that affects correctness, I believe :slightly_smiling_face: Looks good aside from that overall

widgets/common/src/NodeFilter.tsx
stepek: Right, fixed

widgets/common/src/NodeFilter.tsx
```diff
@@ -120,6 +132,22 @@ export default class NodeFilter extends React.Component {
...
+    optionsReducer(result: Record<string, string>, item: Partial<ResponseItem>): Record<string, string> {
+        let allowedOptions: string[] = [];
+
+        if (this.isFilteringSetFor(optionsField)) {
+            allowedOptions = this.getAllowedOptionsFor(optionsField);
+        }
+
+        if (allowedOptions.length !== 0 && !allowedOptions.includes(item.id)) {
+            return result;
+        }
```
Gelio, line 144: I've compared this logic to the previous logic and I am afraid this is not 1-1, but I don't know whether it matters :smile: 

With the new logic, if `this.getAllowedOptionsFor` returns an empty array, all items will be accepted, and with the previous logic, due to `_.intersection` with an empty array, no items would be accepted.

Thus, I believe it should be more along the lines of:

```suggestion
        if (this.isFilteringSetFor(optionsField) && !this.getAllowedOptionsFor(optionsField).includes(item.id)) {
            return result;
        }
```

meaning that we don't handle the `allowedOptions.length === 0` case in a special way

Gelio (approved):
The code looks good to me, I just have a single doubt about one of the refactors that I'd like to clarify before we merge the PR (assuming the system tests will pass, of course :smile: )

widgets/common/src/NodeFilter.tsx
stepek: oh... right, i missmashed the logic :(

stepek:
## system tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/906/

Gelio (approved):

stepek:
I changed base to 6.1 and rebased this to 6.1 because on master some of tests failing after removing jQuery url parsing. 

### system test 
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/910

Gelio (dismissed):

stepek:
## system tests 
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/915/

widgets/common/src/NodeFilter.tsx
```diff
@@ -132,18 +132,6 @@ export default class NodeFilter extends React.Component {
         this.handleInputChange({}, event, field);
     };
 
-    optionsReducer(result: Record<string, string>, item: Partial<ResponseItem>): Record<string, string> {
```
Gelio: Out of curiosity, why did you choose to inline the function rather than bind it / use an arrow function? The latter would have resulted in a smaller diff and the code and the method would still be extracted :slightly_smiling_face:

Gelio (approved):

widgets/common/src/NodeFilter.tsx
stepek: When function was arrow function as class property, was some weird error, function was not called properly, and not throwing any error in console.

PR #1490, opened 2021-07-02
RD-2694 Remove jQuery usage from Manager.ts file.

stepek:
changed `$.param` to `stringify` from `queryString` lib and adjust tests.

### DIFF
--- a/app/utils/Manager.ts
+++ b/app/utils/Manager.ts
@@ -2,6 +2,8 @@
 /* eslint no-underscore-dangle: ["error", { "allow": ["_size", "_offset"] }] */
 
 import _ from 'lodash';
+import { stringify as stringifyQueryString } from 'query-string';
+
 import Internal from './Internal';
 import StageUtils from './stageUtils';
 import Consts from './consts';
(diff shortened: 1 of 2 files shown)


stepek:
## system tests 
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/883/

app/utils/Manager.ts
```diff
@@ -2,6 +2,8 @@
 /* eslint no-underscore-dangle: ["error", { "allow": ["_size", "_offset"] }] */
 
 import _ from 'lodash';
+import { stringify as stringifyQueryString } from 'query-string';
```
kubama, line 5: As for me the name `stringifyQueryString` implies a function that... stringifies a query string 🙂. How about naming it `stringifyToQueryString` or just `toQueryString`?

stepek: the usage of this import comes from previous commit changes, so IMO should be consistently

kubama (approved):

app/utils/Manager.ts
stepek: maybe with ts migration can be done, Thanks!

PR #1491, opened 2021-07-02
RD-2695 Remove jQuery usage from WidgetDynamicContent component.

stepek:
Remove jQuery usage from Widget Dynamic Context by rewriting to 'native' javascript code


app/components/WidgetDynamicContent.tsx
```diff
@@ -229,12 +231,32 @@ export default class WidgetDynamicContent extends Component {
...
-                            .on(event.event, e => {
-                                event.fn(e, widget, this.getToolbox());
-                            });
+
+                        if (!container) {
+                            log.warn('Cannot attach event, missing container element. Event data is ', event);
+                            return;
+                        }
+
+                        if (this.attachedEvents.current === null) {
```
Gelio, line 240: Will it ever be `null`? We initialize the ref with `{}`

app/components/WidgetDynamicContent.tsx
```diff
@@ -220,6 +221,7 @@ export default class WidgetDynamicContent extends Component {
 
     attachEvents(container) {
         const { data, widget } = this.props;
+
         if (widget.definition && widget.definition.events) {
```
Gelio, line 225: Wow, this seems like a bizarre, outdated API that is not even public :smile: As I have grepped through the widgets code, I didn't find a single widget that uses `events` in its definition. I'm wondering if it's worth maintaining such an API.

Let's wait for @kubama's and @qooban's opinions on whether to keep it.

Personally, I would remove this `definition.events` behavior :smile:

app/components/WidgetDynamicContent.tsx
```diff
@@ -220,6 +221,7 @@ export default class WidgetDynamicContent extends Component {
 
     attachEvents(container) {
```
Gelio, line 222: Minor:

Can we add a type annotation here?

```suggestion
    attachEvents(container: HTMLDivElement | null) {
```

app/components/WidgetDynamicContent.tsx
```diff
@@ -229,12 +231,32 @@ export default class WidgetDynamicContent extends Component {
...
+                        if (!container) {
+                            log.warn('Cannot attach event, missing container element. Event data is ', event);
+                            return;
+                        }
+
+                        if (this.attachedEvents.current === null) {
+                            this.attachedEvents.current = {};
+                        }
+
+                        if (this.attachedEvents.current[`${event.selector}_${event}`]) {
```
Gelio, line 244: Can we be a bit more explicit with the stringification of `event`?

```suggestion
                        if (this.attachedEvents.current[`${event.selector}_${JSON.strinfigy(event)}`]) {
```

app/components/WidgetDynamicContent.tsx
```diff
@@ -20,6 +20,7 @@ export default class WidgetDynamicContent extends Component {
         this.state = {
             loading: false
         };
+        this.attachedEvents = createRef<Record<string, any>>({});
```
Gelio, line 23: Minor:

Do we need a ref? I believe we could just have a regular property here :slightly_smiling_face:

app/components/WidgetDynamicContent.tsx
Gelio: We could avoid the stringification if instead of a single `Record<string, (event: Event) => void>` we had: ` Record<string, Record<object, (event: Event) => void>>` (first map for the selectors, second map for the event type -> listener mapping)

app/components/WidgetDynamicContent.tsx
```diff
@@ -229,12 +231,32 @@ export default class WidgetDynamicContent extends Component {
...
+                            return;
+                        }
+
+                        if (this.attachedEvents.current === null) {
+                            this.attachedEvents.current = {};
+                        }
+
+                        if (this.attachedEvents.current[`${event.selector}_${event}`]) {
+                            container
+                                .querySelector(event.selector)
```
Gelio, line 246: Minor:

Can we extract this part to a separate variable so it can be reused (it's used in 2 places in this logic)? Like `const eventTarget`

app/components/WidgetDynamicContent.tsx
Gelio: Minor:

Can we extract the condition to a separate variable to reuse it later (e.g. `const previousEventListener`)?

app/components/WidgetDynamicContent.tsx
```diff
@@ -229,12 +231,32 @@ export default class WidgetDynamicContent extends Component {
...
+                                .removeEventListener(
```
Gelio, line 247: Minor:

Can we add a bit more precise error handling? If `querySelector` won't find any elements, the error will be a very general `cannot invoke removeEventListener on undefined` and I believe we could make the error message a bit more developer-friendly :slightly_smiling_face: 

The same goes for `addEventListener` below :slightly_smiling_face: If we extract the result of `querySelector` to a variable (as I suggested in another comment), that check would only have to happen once :slightly_smiling_face:

Gelio (commented):
Nice jQuery simulation :+1: 

I added some comments about possible improvements :slightly_smiling_face: I'm not sure whether we want to maintain this `definition.events` feature :smile: Let's wait for the opinions from the rest of the team before we merge it :slightly_smiling_face: You can also postpone the code improvements until then, as maybe we will decide to remove this code altogether

app/components/WidgetDynamicContent.tsx
stepek: This probably is old widget style related API

app/components/WidgetDynamicContent.tsx
stepek: `createRef` don't have possibility to set initialized state
```
function createRef<T>(): RefObject<T>;
``` 

react has some small mishmash with class and hooks ref handling

app/components/WidgetDynamicContent.tsx
Gelio: Another minor, if you decide to stick with `createRef` - from your comment in https://github.com/cloudify-cosmo/cloudify-stage/pull/1491#discussion_r669406841 it looks like `createRef` does not take any arguments :slightly_smiling_face: Thus, having `{}` is misleading

```suggestion
        this.attachedEvents = createRef<Record<string, any>>();
```

app/components/WidgetDynamicContent.tsx
Gelio: You are right :smile: Good point :+1: 

Did you consider initializing the ref with an empty object to avoid this `null` check? Or, we could just follow https://github.com/cloudify-cosmo/cloudify-stage/pull/1491/files/251be192fb3bedd45b2b66c7485b3c6f76e34af4#r667787378

app/components/WidgetDynamicContent.tsx
qooban: Yes, this is very old API and there is very little chance that somebody is using it.
Searching through JIRA I found that we don't have it documented since v4.3 (ref.: https://cloudifysource.atlassian.net/browse/CYBL-195), so I think we can safely remove it from the code.

qooban (commented):
Added one comment. I think we can safely remove `definition.events` from the codebase.

app/components/WidgetDynamicContent.tsx
```diff
@@ -218,36 +218,6 @@ export default class WidgetDynamicContent extends Component {
...
-                    this
-                );
-            } catch (e) {
-                log.error('Error attaching events to widget', e);
-            }
-        }
-
-        if (widget.definition.postRender) {
-            widget.definition.postRender($(container), widget, data.data, this.getToolbox());
-        }
```
qooban, line 248: I'm wondering if it is possible to keep it somehow without having jQuery?

As this method (`postRender`) is mentioned in our documentation (https://docs.cloudify.co/latest/developer/writing_widgets/widget-apis/), but similarly to `events` it is very unlikely that anybody uses it.

qooban (commented):
I added one more sub-task to RD-258 to update documentation after all changes related to jQuery removal - RD-2775.

app/components/WidgetDynamicContent.tsx
Gelio: I would also opt for keeping this API (it's also mentioned in our typings https://github.com/cloudify-cosmo/cloudify-stage/blob/efd8ecea5c7a9509d71f9d7ab1fb234af596312a/app/utils/StageAPI.ts#L163-L168)

Gelio:
Let's update the branch with `master` to get rid of the multibranch CI failure :smile:

app/components/WidgetDynamicContent.tsx
stepek: `componentDidUpdate` in this case can be better :) imho

Gelio: I just remembered - we will need it in `componentDidMount` too, if I'm not mistaken :smile:

qooban: @stepek Please mention that during standup that it is going to be - after all - a breaking change.

app/components/WidgetDynamicContent.tsx
```diff
@@ -17,6 +17,9 @@ import combineClassNames from '../utils/shared/combineClassNames';
 export default class WidgetDynamicContent extends Component {
     constructor(props) {
         super(props);
+
+        this.containerRef = createRef<HTMLElement>();
```
Gelio, line 21: Can you also declare this property in the class, so TypeScript sees it? I mean

```ts
private readonly containerRef: RefObject<HTMLElement>;
```

(not sure about the type, though, I did not check it in the editor)

Gelio, line 21: Can you also declare this property in the class, so TypeScript sees it? I mean

```ts
private readonly containerRef: RefObject<HTMLElement>;
```

(not sure about the type, though, I did not check it in the editor)

app/components/WidgetDynamicContent.tsx
```diff
@@ -78,6 +81,10 @@ export default class WidgetDynamicContent extends Component {
         if (requiresFetch) {
             this.fetchData();
         }
+
+        if (widget.definition.postRender && this.containerRef) {
+            widget.definition.postRender(this.containerRef, widget, data.data, this.getToolbox());
```
qooban, line 86: Hmm... Shouldn't we pass `this.containerRef.current`?

app/components/WidgetDynamicContent.tsx
```diff
@@ -78,6 +81,10 @@ export default class WidgetDynamicContent extends Component {
         if (requiresFetch) {
             this.fetchData();
         }
+
+        if (widget.definition.postRender && this.containerRef) {
```
Gelio, line 85: We should probably check if `this.containerRef.current` is defined, instead of `this.containerRef` (the ref object is always created in the constructor).

```suggestion
        if (widget.definition.postRender && this.containerRef.current) {
```

Also, can we have some console error if `this.containerRef.current` is not defined? This would mean there is a bug in the component :smile:

app/components/WidgetDynamicContent.tsx
Gelio, line 86: We should probably be doing

```suggestion
            widget.definition.postRender(this.containerRef.current, widget, data.data, this.getToolbox());
```

app/components/WidgetDynamicContent.tsx
Gelio: Do we not need this logic in `componentDidMount`?

Gelio (changes requested):
To me, there is now a bug with how refs are used :smile: Thus, I'm blocking the review, just to be safe :slightly_smiling_face:

qooban (commented):
Looks better now. Added one comment.

app/components/WidgetDynamicContent.tsx
stepek: sorry, right should be

stepek:
### System tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/945/

app/components/WidgetDynamicContent.tsx
stepek: sure, but i want to keep the same convention as other fields like `state`

Gelio (commented):
Now it's just a matter of also calling the function in `componentDidMount` and this PR should be good to go :slightly_smiling_face:

qooban (commented):
Agree with @Gelio last comment.

stepek:
@Gelio @qooban move to separate function and call it on `componentDidMount`

Gelio (approved):
LGTM :smile:

stepek:
###System test

https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/957/

qooban (approved):
Looks fine 👍🏻  Let's just wait for the confirmation from Mr Jenkins :-)

Gelio:
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/957/ is green, so we're finally ready to merge :smile:

stepek:
@Gelio @qooban Thanks !

PR #1492, opened 2021-07-02
Revert "Remove jQuery-UI"

stepek:
Reverts cloudify-cosmo/cloudify-stage#1473

Adding this revert because `cloudify-blueprint-topology` is not fully support jquery ui itself.

### DIFF
--- a/app/app.tsx
+++ b/app/app.tsx
@@ -1,4 +1,9 @@
 // @ts-nocheck File not migrated fully to TS
+import 'jquery-ui/ui/core';
+import 'jquery-ui/ui/widget';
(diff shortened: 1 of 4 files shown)


stepek:
## System tests 
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/886/

kubama (approved):

kubama:
For the future - I think it's a good practice to wait for the system tests to complete before merging.

Gelio:
Do we still need the branch? Maybe we should remove it :slightly_smiling_face:

PR #1493, opened 2021-07-02
RD-267  Remove jQuery-UI (#1473)

stepek:
* RD-267 Remove jQuery and jQuery-UI


Gelio (commented):
What's the purpose of this PR? I see that there is #1492 and #1473 and I can't figure out what's happening

stepek:
This should be merged after removing jQuery from repos.

qooban:
This is to be merged only after https://github.com/cloudify-cosmo/cloudify-blueprint-topology/pull/162 is merged and new version of cloudify-blueprint-topology is released, right?

stepek:
@qooban Yes, before that we should merger #162

qooban:
I assume you had https://github.com/cloudify-cosmo/cloudify-blueprint-topology/pull/162 in mind :-) 
OK, so let's finalize PR in cloudify-blueprint-topology repo first, then release new version and update the version here. Then we can review it finally.

stepek:
@qooban `cloudify-blueprint-topology` updated

package.json
```diff
@@ -58,7 +58,6 @@
     "i18next": "^19.8.3",
     "isomorphic-fetch": "^3.0.0",
     "jquery": "^3.5.0",
```
Gelio, line 60: Can we also remove `jquery`, similarly to https://github.com/cloudify-cosmo/cloudify-blueprint-composer/pull/905#discussion_r674550763?

webpack.config.js
```diff
@@ -160,7 +160,6 @@ module.exports = (env, argv) => {
             resolve: {
                 extensions: resolveExtensions,
                 alias: {
-                    'jquery-ui': 'jquery-ui/ui',
                     jquery: `${__dirname}/node_modules/jquery`, // Always make sure we take jquery from the same place
```
Gelio, line 163: If you will remove jquery, let's also remember to remove this line

Gelio: There are also some uses of `jquery` down below in `webpack.config.js`

Gelio (commented):
btw. I see that jquery is also used in `test/jest/setup.ts`. Maybe we could remove it from there too?

package.json
stepek: No, i'll do this after removing last jquery usage. https://github.com/cloudify-cosmo/cloudify-stage/pull/1491

Gelio: So what's the order of merging the PRs :smile: ? There are 3 open PRs for stage regarding jQuery and I've lost track

qooban: I've lost track, too. I think we should discuss that on UI weekly (I'll add a point). Looks like there are (or at least there were) many open PRs related to jQuery removal at the same time which are not independent and all of us are engaged in it. That's not optimal situation I think.

qooban (commented):
I see that there will be more changes in this PR as it looks like the CI is failing due to size of the package, so I'll postpone my review until it is finalized.

Can you check the size? That's probably because the Topology widget bundle is now bigger (contains jQuery) and we still have jQuery in the deps? 

That's, IMHO, another reason for having merging order of all of jQuery removal PRs settled and clarified.

stepek:
### System Tests 
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1001/

package-lock.json
```diff
@@ -21809,11 +21816,6 @@
         "logalot": "^2.0.0"
       }
     },
-    "jquery": {
-      "version": "3.5.0",
-      "resolved": "https://registry.npmjs.org/jquery/-/jquery-3.5.0.tgz",
-      "integrity": "sha512-Xb7SVYMvygPxbFMpTFQiHh1J7HClEaThguL15N/Gg37Lri/qKyhRGZYzHRyLH8Stq3Aow0LsHO2O2ci86fCrNQ=="
-    },
     "jquery-ui": {
```
Gelio, line 21819: Minor:

Interesting... `jquery-ui` is still there, and `jquery` is not (it's moved to be a direct dependency of `cloudify-blueprint-topology`). Can you check why there is such a difference between the 2 packages?

I thought we could remove it entirely and keep it only included in `cloudify-blueprint-topology` :thinking:

Gelio (approved):
Looks clean :+1: Nice work getting rid of jquery :smile: 

I see that `jquery-ui` is still at the top-level of `package-lock.json`, which worries me :thinking: Let's check why the way it's stored is different from how `jquery` is stored (it's not on the top level, but it only is present as a dependency of `cloudify-blueprint-topology`)

qooban (approved):
Looks good, just double check `package-lock.json` file as mentioned by @Gelio .

package-lock.json
stepek: Fixed,

Gelio: What did you to do fix it? What was the problem :smile: ? Did you move it manually?

stepek:
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1008/ pass

qooban (approved):

PR #1495, opened 2021-07-06
RD-2483 Migrate Plugins Catalog widget to TypeScript

Gelio:
This PR migrates the Plugins Catalog widget fully to TypeScript. My aim was only to make the code simpler to understand and keep functional changes to a minimum. I have also added system tests for the Plugins Catalog widget to make sure that the basic workflow (upload a plugin and check that it is visible in the Plugins widget) works.

There are no user-facing changes, only internal ones, so I'm not posting any screenshots.

I have verified that the error message is shown correctly after uploading a malformed plugin (some data was missing in the request body) and that the plugins are actually uploaded (which is also verified by the system test).

There will be a separate PR for RD-2483 that actually adds a new "Uploaded version" column to the Plugins Catalog widget. I am raising these changes as a separate PR to make it easier to review since those are 2 distinct changes.

I will cherry-pick this PR to `master` once it is merged.

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/892/pipeline

Queued 2021-07-06 9:48 CEST

### DIFF
--- a/widgets/pluginsCatalog/src/PluginsCatalogList.tsx
+++ b/widgets/pluginsCatalog/src/PluginsCatalogList.tsx
@@ -1,22 +1,31 @@
-// @ts-nocheck File not migrated fully to TS
-/**
- * Created by Tamer on 30/07/2017.
- */
-import PluginsCatalogModal from './PluginsCatalogModal';
+import PluginsCatalogModal, { PluginsCatalogModalProps } from './PluginsCatalogModal';
 import Actions from './Actions';
+import type { PluginDescription, PluginsCatalogWidgetConfiguration } from './types';
 
-export default class PluginsCatalogList extends React.Component {
-    constructor(props, context) {
-        super(props, context);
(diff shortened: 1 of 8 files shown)


widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -89,8 +102,9 @@ export default class PluginsCatalogList extends React.Component {
                                         onClick={event => {
                                             event.preventDefault();
                                             this.onUpload({
-                                                ...item.wagon,
```
Gelio, line 92: I've decided to be more precise here and only use the necessary properties explicitly. Otherwise, there could be potentially some additional unexpected data sent to the API, or the underlying JavaScript objects will have more properties than TS knows about, which can be unexpected

alexmolev (approved):
LGTM

PR #1500, opened 2021-07-12
RD-2750 Remove jQuery code deps

stepek:
Removed last jQuery code usage.


app/app.tsx
```diff
@@ -50,7 +50,7 @@ import Routes from './containers/Routes';
 import translation from './translations/en.json';
 import LoaderUtils from './utils/LoaderUtils';
 
-window.$ = $;
+window.$ = $; // Needed by topology widget.
```
Gelio, line 53: Minor:

If that's the case, maybe we could move this line to the topology widget itself? This way it will be added closer to where it's used

test/jest/setup.ts
```diff
@@ -33,10 +32,6 @@ Object.keys(document.defaultView).forEach(property => {
     }
 });
 
-global.$ = $;
```
Gelio, line 36: Yay :smile:

app/utils/SplashLoadingScreen.ts
```diff
@@ -1,24 +1,30 @@
...
-
-const SPLASH_DIV = $('div.splashPage');
+const SPLASH_DIV: HTMLElement | null = document.querySelector('div.splashPage');
 const ACTIVE_CLASS = 'active';
-const FADE_OUT_SPEED = 'slow';
+const FADE_OUT_SPEED = 300;
 
 export default class SplashLoadingScreen {
+    static fadeOut(element: HTMLElement) {
+        element.style.setProperty('transition', `all ${FADE_OUT_SPEED}ms linear`);
```
Gelio, line 7: If I'm not mistaken, we can make the code simpler by adding the `transition` in CSS (https://github.com/cloudify-cosmo/cloudify-stage/blob/stepek/RD-2750/backend/templates/style.ejs). Then, we would not have to remove it manually.

Also, how about being more specific and only transitioning the `opacity`, instead of `all`? :slightly_smiling_face:

app/utils/SplashLoadingScreen.ts
```diff
@@ -1,24 +1,30 @@
...
+            element.style.removeProperty('opacity');
+            element.style.setProperty('display', 'none');
+        }, FADE_OUT_SPEED);
+    }
+
     static turnOn() {
-        if (!SPLASH_DIV.hasClass(ACTIVE_CLASS)) {
-            SPLASH_DIV.show();
-            SPLASH_DIV.addClass(ACTIVE_CLASS);
+        if (SPLASH_DIV && !SPLASH_DIV.classList.contains(ACTIVE_CLASS)) {
```
Gelio, line 17: You could add 

```ts
if (!SPLASH_DIV) {
    throw new Error('Splash page element not found');
}
```

before the class declaration which would make checking if `SPLASH_DIV` is defined unnecessary :slightly_smiling_face:

app/utils/SplashLoadingScreen.ts
```diff
@@ -1,24 +1,30 @@
-// @ts-nocheck File not migrated fully to TS
-/**
- * Created by jakubniezgoda on 06/13/2017.
- */
-
-const SPLASH_DIV = $('div.splashPage');
+const SPLASH_DIV: HTMLElement | null = document.querySelector('div.splashPage');
```
Gelio, line 1: Minor:

The type annotation seems redundant to me, as it's the same as the default one for `querySelector`

```suggestion
const SPLASH_DIV = document.querySelector('div.splashPage');
```

app/utils/SplashLoadingScreen.ts
```diff
@@ -1,24 +1,30 @@
...
 const ACTIVE_CLASS = 'active';
-const FADE_OUT_SPEED = 'slow';
+const FADE_OUT_SPEED = 300;
 
 export default class SplashLoadingScreen {
+    static fadeOut(element: HTMLElement) {
```
Gelio, line 6: Can we make it a `private` method? Also, this always will refer to `SPLASH_DIV`, so we could ditch the parameter now :slightly_smiling_face: 

```suggestion
    private static fadeOut() {
```

app/utils/SplashLoadingScreen.ts
```diff
@@ -1,24 +1,30 @@
...
+        element.style.setProperty('transition', `all ${FADE_OUT_SPEED}ms linear`);
+        element.style.setProperty('opacity', '0');
```
Gelio, line 8: How about adding the `opacity` manipulation to the `.active` class in CSS? Then, we would not need to take care of it manually, and all this function would do it add `display: none` after a timeout :smile:

Gelio (commented):
Nice TS migrations :+1: I appreciate those

app/components/layout/ScrollToTop.ts
```diff
@@ -1,19 +1,15 @@
...
-    componentDidUpdate(prevProps) {
+class ScrollToTop extends Component<RouteComponentProps> {
+    componentDidUpdate(prevProps: RouteComponentProps) {
         const { location } = this.props;
+
         if (location !== prevProps.location) {
-            $('div.main').scrollTop(0);
+            window.scrollTo(0, 0);
         }
     }
```
qooban, line 14: We don't have a test for that scrolling functionality IIRC, so I assume you verified it and works fine, right?

app/utils/SplashLoadingScreen.ts
```diff
@@ -1,24 +1,30 @@
-// @ts-nocheck File not migrated fully to TS
-/**
- * Created by jakubniezgoda on 06/13/2017.
- */
-
-const SPLASH_DIV = $('div.splashPage');
+const SPLASH_DIV: HTMLElement | null = document.querySelector('div.splashPage');
 const ACTIVE_CLASS = 'active';
-const FADE_OUT_SPEED = 'slow';
+const FADE_OUT_SPEED = 300;
```
qooban, line 3: As none of these consts are exported, probably all of them can be lowercased to be aligned with https://github.com/airbnb/javascript#naming--uppercase .

(we started following AirBnB long after this file was created)

app/app.tsx
qooban: Good idea. Let's move it as close to the usage as possible to avoid potential confusion in the future.
As leaving it here suggests that the whole application is using jQuery.

qooban (commented):
Added some comments.

app/utils/SplashLoadingScreen.ts
stepek: No, because we have only animation of fade out.

app/components/layout/ScrollToTop.ts
stepek: it's works but i'll change it to use only `div.main` container.

app/utils/SplashLoadingScreen.ts
```diff
@@ -1,24 +1,39 @@
...
         }
     }
 
     static turnOff() {
-        if (SPLASH_DIV.hasClass(ACTIVE_CLASS)) {
-            SPLASH_DIV.fadeOut(FADE_OUT_SPEED);
-            SPLASH_DIV.removeClass(ACTIVE_CLASS);
+        if (!splashDiv) {
+            throw new Error('Splash page element not found');
+        }
```
qooban, line 32: I see it is common for all methods, maybe we can extract it to avoid duplication?

Optional:

Actually it is very unlikely it will happen as `div.splashPage` is part of the `index.html` from the very early stage of the application running time, so maybe that check should be done only before `SplashLoadingScreen` class definition?

widgets/topology/src/Topology.tsx
```diff
@@ -6,6 +6,8 @@ import { createExpandedTopology } from './DataProcessor';
 import ScrollerGlassHandler from './ScrollerGlassHandler';
 import TerraformDetailsModal from './TerraformDetailsModal';
 
+window.$ = $;
+
```
qooban, line 10: This will work only if we have https://github.com/cloudify-cosmo/cloudify-blueprint-topology/pull/162 implemented and cloudify-blueprint-topology version updated in stage, right?

qooban (commented):
Commented. 

I assume that this PR needs to wait for https://github.com/cloudify-cosmo/cloudify-blueprint-topology/pull/162 and possibly https://github.com/cloudify-cosmo/cloudify-stage/pull/1493 (depending on in which PR you'd like to update `package.json`).

widgets/topology/src/Topology.tsx
```diff
@@ -6,6 +6,8 @@ import { createExpandedTopology } from './DataProcessor';
 import ScrollerGlassHandler from './ScrollerGlassHandler';
 import TerraformDetailsModal from './TerraformDetailsModal';
 
+window.$ = $;
```
Gelio, line 9: Wait a second :smile: This looks like a magic import to me at the moment. Where do we get the `$` from? Should we not `import $ from 'jquery';` or something? If not, then maybe jQuery is already a global and this assignment is no necessary (which I would not be a fan of :smile:)?

Gelio: Actually, since `cloudify-blueprint-topology` will have its own instance of jQuery thanks to https://github.com/cloudify-cosmo/cloudify-blueprint-topology/pull/162, maybe we can skip this line completely and remove jQuery altogether from stage?

`cloudify-blueprint-topology` could import jQuery itself and expose it globally if it wants to

app/utils/SplashLoadingScreen.ts
Gelio: I see that does not work :worried: 

image

I believe a helper method/function with such an `if` should do the job here :+1:

app/utils/SplashLoadingScreen.ts
Gelio: Ok, makes sense, thanks :+1:

Gelio (commented):
The code looks mostly good to me. I want to only clarify the location and necessity of `window.$ = $` before approving :slightly_smiling_face:

app/utils/SplashLoadingScreen.ts
stepek: nop, the same errors about possibility of null 
image

Gelio: You need to use an assertion function:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

stepek:
`window.$ = $` can be removed after https://github.com/cloudify-cosmo/cloudify-stage/pull/1493

app/utils/SplashLoadingScreen.ts
```diff
@@ -1,12 +1,16 @@
...
+
 const splashDiv = document.querySelector<HTMLElement>('div.splashPage');
 const activeClass = 'active';
 const fadeOutSpeed = 300;
 
 export default class SplashLoadingScreen {
     private static fadeOut() {
         if (!splashDiv) {
-            throw new Error('Splash page element not found');
+            log.error('Splash page element not found');
```
Gelio, line 10: Why is this changed to an `log.error` instead of a `throw`? What's the reason we want to fail silently :smile: ? IMO if such an error happens in the app, we should not even continue as something is really broken and should be fixed by a developer

Gelio (commented):
https://github.com/cloudify-cosmo/cloudify-stage/pull/1500#discussion_r673741492 is still unresolved :slightly_smiling_face:

app/utils/SplashLoadingScreen.ts
stepek: Failing tests unit tests in many places

Gelio: Sooo... maybe we should get our unit tests setup to more closely resemble the actual UI, instead of changing the runtime logic to accommodate the tests? :slightly_smiling_face: I'm ok with creating a task to revert this change and add a `splashDiv` in the jsdom in unit tests :slightly_smiling_face:

qooban (commented):
I updated the title to have JIRA ID reference. 
I have no more specific comments, but I'm assuming that there are some discussions not resolved, yet, so I'll not approve yet.

From my side: 
1. Please verify if all tests are passing.
2. I assume that the last PR for jQuery full removal will be https://github.com/cloudify-cosmo/cloudify-stage/pull/1493 right?

stepek:
> 1. Please verify if all tests are passing.

System tests: 
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/979/



> 2. I assume that the last PR for jQuery full removal will be [RD-267  Remove jQuery-UI (#1473) #1493](https://github.com/cloudify-cosmo/cloudify-stage/pull/1493) right?

Yes, right

app/utils/SplashLoadingScreen.ts
stepek: Ok, ticket [RD-2841](https://cloudifysource.atlassian.net/browse/RD-2841).

widgets/topology/src/Topology.tsx
stepek: No, upgrade will be in https://github.com/cloudify-cosmo/cloudify-stage/pull/1493

Gelio: Alright :+1:

Gelio:
I am irritated by https://github.com/cloudify-cosmo/cloudify-stage/pull/1500#discussion_r673741492 not being addressed for 6 days despite other communication happening in this PR :frowning_face: Other than that comment, it looks good to me

app/utils/SplashLoadingScreen.ts
stepek: Fixed but added try catches to prevent failing tests

widgets/topology/src/Topology.tsx
stepek: no, Changes regarding cloudify-cosmo/cloudify-blueprint-topology added in separate RP

app/utils/SplashLoadingScreen.ts
Gelio: Thanks :+1:

widgets/topology/src/Topology.tsx
Gelio: What's the reason to update `cloudify-blueprint-topology` in a separate PR instead of here? Doing it here would allow us to remove this unnecessary change, if I'm not mistaken

Gelio (dismissed):
Generally looks good to me :slightly_smiling_face:

widgets/topology/src/Topology.tsx
stepek: in next PR i'll remove jquery jquery ui and upgrade `cloudify-blueprint-topology`

qooban (approved):

widgets/topology/src/widget.tsx
```diff
@@ -3,11 +3,20 @@
...
+import 'jquery-ui/ui/widgets/mouse';
+import 'jquery-ui/ui/widgets/draggable';
+import 'jquery-ui/ui/widgets/droppable';
+
 import { castArray } from 'lodash';
 import Topology from './Topology';
 import DataFetcher from './DataFetcher';
 import { createBaseTopology } from './DataProcessor';
 
+window.$ = $;
```
stepek, line 18: Mover from `Topology.tsx` to be initialized before importing `cloudify-blueprint-topology`

widgets/topology/src/widget.tsx
```diff
@@ -3,11 +3,20 @@
  * Created by kinneretzin on 07/09/2016.
  */
 
+import $ from 'jquery';
+import 'jquery-ui/ui/core';
+import 'jquery-ui/ui/widget';
```
stepek, line 8: Moved closure to topology widget. will be removed in next pr with upgrade of `cloudify-blueprint-topology` and removement  jQuery

stepek:
### system tests 
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/995/

widgets/topology/src/widget.tsx
qooban: I assume you are talking about this PR https://github.com/cloudify-cosmo/cloudify-stage/pull/1493 ?

🙂 Why not remove it at once within this PR?

qooban (approved):
I think it would be easier to remove all deps within one PR (especially that #1493 seems very small), but I'm also fine with having two as this is hopefully last phase of the removal, I suspect 🙂

widgets/topology/src/widget.tsx
stepek: because I have one open :)

PR #1501, opened 2021-07-13
RD-2747 Fix expanding Terraform nodes in Topology widget

Gelio:
This PR fixes the error that occurred in RD-2747 when expanding Terraform topology nodes.

The main fix is in the first commit of this PR. Reviewing commit-by-commit should help distinguish the functional change from the refactor. Other commits are only TS-related changes - an effort to migrate more code to TypeScript.

`Topology.tsx` contains a lot of code and was hard to add type annotations all at once, so I added some of them. My aim was to be defensive and cautious and only add type annotations when I was sure that was the type. I added `unknown` in places I was not sure about.

There are existing system tests that verify the "expand" functionality, so I did not add any new system tests. The system tests are not really end-to-end, as they use their own blueprint and not the Terraform one from the catalog. RD-2747 was caused not by any code change in the UI, but by the Terraform blueprint/some plugin changing, which I believe is out of scope for the UI system tests. Having a system test to guard against RD-2747 regressions requires having AWS secrets and waiting around 3 minutes for the blueprint to deploy (provided that there are enough resources on AWS to do so), which IMO could be flaky very fast. This is why I decided not to add any system tests in this PR.

## Media

Expanded `cloud_resources` node:

image

https://user-images.githubusercontent.com/889383/125415070-f0dc32a7-fdfb-4cd3-beb0-6f6a1a387667.mp4

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/908/pipeline

Queued 2021-07-13 10:53 CEST


widgets/topology/src/Topology.tsx
```diff
@@ -22,13 +24,44 @@ function isNodesChanged(topologyNodes, newNodes) {
...
 
-function findExpandedNode(currentTopology, deploymentId) {
+function findExpandedNode(currentTopology: TopologyProps['data']['blueprintDeploymentData'], deploymentId: string) {
     return _.find(currentTopology.nodes, node => node.templateData.deploymentId === deploymentId);
 }
 
-export default class Topology extends React.Component {
-    constructor(props, context) {
-        super(props, context);
+interface TopologyProps {
```
stepek, line 31: IMHO this should be defined before usage, up to 27 line

widgets/topology/src/Topology.tsx
```diff
@@ -22,13 +24,44 @@ function isNodesChanged(topologyNodes, newNodes) {
...
+        blueprintDeploymentData: unknown;
+        icons: Record<string, string>;
+        layout: unknown;
+    };
+    deploymentId: string;
+}
+
+interface TopologyState {
+    expandedDeployments: string[];
+    // eslint-disable-next-line camelcase
```
stepek, line 45: General:
what sens is have `camelcase` in eslint config? in this case, always when we defining backend returned values, we need to disable this rule.

stepek (approved):

widgets/topology/src/Topology.tsx
Gelio: Right, good point :+1: I'll move it

widgets/topology/src/Topology.tsx
Gelio: Good point :+1: I'm also not sure :smile: We can talk about it during the weekly (perhaps not the upcoming one, as then it will just be the 2 of us, but the one next week). Can you add it to the agenda?

widgets/topology/src/Topology.tsx
Gelio: Done

widgets/topology/src/Topology.tsx
stepek: Done

stepek (approved):
LGTM

widgets/topology/src/Topology.tsx
Gelio: Thanks :+1:

PR #1502, opened 2021-07-13
Fix order of params in External and Manager classes.

stepek:
Fix build failing after removing jquery params function call. In some cases query string sorts parsed string, to have consistently with current tests I decided to add sorting the same as jQuery params.


stepek:
### system tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/910/

test/jest/utils/External.test.ts
```diff
@@ -1,26 +0,0 @@
-import External from 'utils/External';
-
-describe('(Utils) External', () => {
```
Gelio, line 3: Do you believe the tests are no longer needed?

app/utils/getQueryParamSorter.ts
```diff
@@ -0,0 +1,9 @@
+const getQueryParamSorter = (params: Record<string, any> = {}) => {
```
Gelio, line 1: I'm not sure this is the best way to go. In each place we're building URLs, we're now expected to use this utility function. This not only adds unnecessary operations in runtime (sorting the params), but also adds something to remember.

I believe we should go another way and cater the tests to handle different order of URL params, rather than go the other way around :smile: We could use the `URLSearchParams` API or use some other way of converting the params into an object and then inspect individual properties

Gelio (commented):
Aside from the inline comments, we are missing a ticket reference in this PR :slightly_smiling_face:

test/jest/utils/External.test.ts
stepek: ups, i committed this accidentally ;/

app/utils/getQueryParamSorter.ts
stepek: > I'm not sure this is the best way to go. In each place we're building URLs, we're now expected to use this utility function. This not only adds unnecessary operations in runtime (sorting the params), but also adds something to remember.
No, you don't need remember this, because this function is called only in External and Manager classes. 
Right it's additional operation added to parsing URL's,

> I believe we should go another way and cater the tests to handle different order of URL params, rather than go the other way around 😄 We could use the `URLSearchParams` API or use some other way of converting the params into an object and then inspect individual properties
IMHO too but i want to stay with current way url handling.

Gelio: Hmm so if we don't agree on which way to go, let's wait for a third person to share their opinion :slightly_smiling_face: 

@qooban would you rather
1. Make the query params order consistent by adding more operations in runtime
2. Make the tests more complex to handle any param order
?

qooban: Wait... Before we used `$.param` which seems to me that is not sorting keys automatically (according to some tests I did and https://api.jquery.com/jquery.param/). Now we started using `query-string` which is sorting keys automatically (according to https://github.com/sindresorhus/query-string#sort-1).

Why just not turn off automatic keys sorting in all `stringify` calls to keep backward compatibility?

Example taken from https://github.com/sindresorhus/query-string#sort-1
```
const queryString = require('query-string');

queryString.stringify({b: 1, c: 2, a: 3}, {sort: false});
//=> 'b=1&c=2&a=3'
```

qooban (commented):
Added one comment in the discussion: https://github.com/cloudify-cosmo/cloudify-stage/pull/1502/files#r668699807

stepek:
@qooban thanks:) fixed

qooban (commented):
After setting `{ sort: false }` option to `stringify`, I think you need to revert changes in unit tests, am I right?
Let me know once it is done, then I'll review again.

stepek:
## System Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/936/

qooban:
> > After setting `{ sort: false }` option to `stringify`, I think you need to revert changes in unit tests, am I right?
> > Let me know once it is done, then I'll review again.
> 
> No, the tests should be the same as jQuery, and it is.
> It's done

Hm... I'm not sure I understand what you meant, but it is strange to me that the solution with and without sorting have the same asserts in the tests 😮 It means to me that either tests are not working fine or the sorting solution you implemented originally did not work 🙂 Can you double-check?

qooban (approved):
OK, clear. Sorting implemented originally in this PR was meant to do what actually `{ sort: false }` is doing (preserve the order from the object).

Once failing previously system tests related query-string change pass, I'm OK with merging this one.

app/utils/getQueryParamSorter.ts
Gelio: I see we've diverged from this question a bit. Wouldn't it make the tests more robust? If the tests would not depend on the order of params, they would require fewer modifications when we make changes to the requests' params :smile: 

I'm not intending to block the PR, if @qooban thinks that it's ok for now, but IMO I would rather have tests be agnostic of the order

Gelio:
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/936/console passed, so I guess we can merge it, but I would still go back to https://github.com/cloudify-cosmo/cloudify-stage/pull/1502#discussion_r673015623 even to create a tech debt item for the future and avoid order-related problems in the future :smile:

Gelio:
Also, https://github.com/cloudify-cosmo/cloudify-stage/pull/1502#pullrequestreview-705096488 still applies - there is no ticket reference in this commit/PR

Gelio:
@stepek I feel ignored - I did not get a response to 2 of my comments (one of which I left on the previous review) :worried:

app/utils/getQueryParamSorter.ts
qooban: You're right, we diverged from the original question a bit.

I agree, our tests should be agnostic of the order. The question is how difficult would it be to change it. We have at least two areas to update:
* unit tests (for External, Manager classes)
* system tests (plenty of them I suspect)
* any other place?

I think we should be able to change unit tests quite fast, I'm not sure about system tests.

Anyways, I agree with @gelio we should aim for making the tests not depend on the order of params, so @stepek could you create technical debt issue for that?

PR #1506, opened 2021-07-14
RD-1827 Remove double loader in DeploymentActionButtons widget

Gelio:
This PR removes the extra loading spinner that was shown inside of the `DeploymentActionButtons` widget. Now there will only be a single loader when the deployment data is being fetched from the Manager (the loader comes from our widget framework).

Additionally, this PR:
1. Migrates some code to TypeScript
2. Makes the data fetching states more explicit (explicit error, loading, success states)
3. Fixes some minor problems that I encountered along the way (static properties missing the `static` keyword, `Error` not being handled as widget's `data`)

I tried to add a test to verify that there is only a single loading spinner active, but Cypress was too slow (or the data fetching was too fast) and did not catch the moment that there was a spinner. It always saw 0 spinners when deployment was being picked. I could add some throttling logic and create a test for that, but IMO that's worth more than the 3 SP for this ticket and I believe there can be more value created by me focusing on other tickets :smile: 

## Media

I had network throttling enabled (Slow 3G) to see the loading states. The layout no longer shifts when picking a new deployment

https://user-images.githubusercontent.com/889383/125618644-2e632ecf-6092-4e8e-8856-3aba0de2d51c.mp4

When there is no deployment picked, the buttons are disabled and there is an error logged in the console:
image
image

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/919/pipeline

Queued 2021-07-14 14:52 CEST

### DIFF
--- a/app/components/PagesList.tsx
+++ b/app/components/PagesList.tsx
@@ -30,7 +30,7 @@ export interface PagesListState {
 }
 export default class PagesList extends Component<PagesListProps, PagesListState> {
     // eslint-disable-next-line react/static-property-placement
-    propTypes = {
+    static propTypes = {
         onPageSelected: PropTypes.func.isRequired,
         onPageRemoved: PropTypes.func.isRequired,
         onPageReorder: PropTypes.func.isRequired,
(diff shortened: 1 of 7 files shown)


stepek (approved):
Code LGTM!

PR #1510, opened 2021-07-16
RD-2324 Add display name to filter widget.

stepek:
image
Long 

Added possibility to change search param in dynamic dropdown, so now widget can take multiple search query params.

Comment: 
Long line handling should be fixed by this ticket  [https://cloudifysource.atlassian.net/browse/RD-1575](RD-1575)


widgets/common/src/DynamicDropdown.tsx
```diff
@@ -58,6 +58,7 @@ export default function DynamicDropdown({
     className = '',
     disabled = false,
     fetchAll = false,
+    searchParam = ['_search'],
```
Gelio, line 61: I would call this property

```suggestion
    searchParams = ['_search'],
```

since it's an array and can contain multiple params :slightly_smiling_face:

widgets/filter/src/Filter.tsx
```diff
@@ -210,6 +211,8 @@ export default class Filter extends React.Component {
                                 entityName: 'Deployment',
                                 filter: blueprintFilter,
                                 pageSize: 20,
+                                textFormatter: item => `${item.display_name} (${item.id})`,
```
Gelio, line 214: I assume we could easily do

```suggestion
                                textFormatter: item => item.display_name === item.id ? item.id : `${item.display_name} (${item.id})`,
```

to already handle the case when both ID and display name are the same, in which case we want to show only one of them :slightly_smiling_face:

widgets/common/src/DynamicDropdown.tsx
```diff
@@ -138,12 +139,18 @@ export default function DynamicDropdown({
...
                 .getManager()
                 .doGet(fetchUrl, {
-                    params: {
-                        _search: searchQuery,
-                        _sort: valueProp,
-                        _size: pageSize,
-                        _offset: nextPage * pageSize
-                    }
+                    params: searchParam.reduce(
+                        (result: Record<string, any>, param) => {
```
Gelio, line 143: Minor:

We know that `searchQuery` is a string, so we can do:

```suggestion
                        (result: Record<string, string>, param) => {
```

widgets/filter/src/Filter.tsx
```diff
@@ -175,6 +175,7 @@ export default class Filter extends React.Component {
                                     .concat(fetchIncludeExtra || [])
                                     .join()
                             )}
+                            searchParam={['_search', '_search_name']}
```
Gelio, line 178: This makes it so that we send the additional `_search_name` query param for resources that do not support it (e.g. for blueprints).

Maybe we could only pass it for deployments (made `searchParam` a parameter of `createDropdown` and only add it on line 214)?

widgets/filter/src/Filter.tsx
```diff
@@ -175,6 +175,7 @@ export default class Filter extends React.Component {
                                     .concat(fetchIncludeExtra || [])
                                     .join()
```
Gelio, line 176: I'm wondering if there should not be 

```suggestion
                                    .join(',')
```

to join using a comma :smile: I did not check, but it looks suspicious to me

Gelio: Nevermind, `,` is a default separator in lodash's `join`: https://lodash.com/docs/4.17.15#join

Optional:

I would still pass it explicitly, as it looks confusing to me :smile:

test/cypress/integration/widgets/filter_spec.ts
```diff
@@ -9,7 +9,10 @@ describe('Filter', () => {
 
     it('fills dropdowns with correct data', () => {
         cy.interceptSp('GET', /blueprints.*state=uploaded/, { fixture: 'filter/blueprints.json' });
-        cy.interceptSp('GET', /deployments.*offset=0/, { fixture: 'filter/deployments0.json' });
+        cy.interceptSp('GET', /deployments.*offset=0&_search=ead&_search_name=ead/, {
+            fixture: 'filter/deployments0.json'
+        }).as('fetchFilteredDeployments');
+        cy.interceptSp('GET', /deployments.*offset=0/, { fixture: 'filter/deployments0.json' }).as('fetchDeployments');
```
Gelio, line 15: Won't this match all the requests from the previous `interceptSp` too? I'm wondering if we want it to be this way, as to me, that's confusing :smile: Maybe we could change the regexp to not match the same requests as the previous intercept?

test/cypress/integration/widgets/filter_spec.ts
```diff
@@ -18,9 +21,17 @@ describe('Filter', () => {
         cy.get('#dynamicDropdown1 > .label').should('have.length', 1);
 
         cy.get('#dynamicDropdown2').click();
```
Gelio, line 23: Can we use

```suggestion
        cy.get('#dynamicDropdown2').within(() => {
          cy.click();
          // other commands that do not start with #dynamicDropdown2
        });
```

to avoid repetition?

test/cypress/integration/widgets/filter_spec.ts
```diff
@@ -18,9 +21,17 @@ describe('Filter', () => {
...
-        cy.get('#dynamicDropdown2 .menu > *:eq(0)').should('have.text', 'app2.2');
-        cy.get('#dynamicDropdown2 .menu > *:eq(1)').should('have.text', 'uuu');
+        cy.get('#dynamicDropdown2 .menu > *:eq(0)').should('have.text', 'App 2.2 (app2.2)');
+        cy.get('#dynamicDropdown2 .menu > *:eq(1)').should('have.text', 'Eadu (uuu)');
         cy.get('#dynamicDropdown2 .menu > *').should('have.length', 2);
```
Gelio, line 26: Minor:

Can we extract `cy.get('#dynamicDropdown2 .menu > *')` to a local function (e.g. `getDropdownItems`) to avoid repetition? It is used 3 times :slightly_smiling_face:

Gelio (commented):
Quite an elegant solution, I like it :+1: Very minimal code changes :smile:

qooban:
Sorry for not providing review within one day. Too many things to do after long vacation. I'll review after the standup meeting.

test/cypress/integration/widgets/filter_spec.ts
```diff
@@ -9,7 +9,10 @@ describe('Filter', () => {
 
     it('fills dropdowns with correct data', () => {
         cy.interceptSp('GET', /blueprints.*state=uploaded/, { fixture: 'filter/blueprints.json' });
-        cy.interceptSp('GET', /deployments.*offset=0/, { fixture: 'filter/deployments0.json' });
+        cy.interceptSp('GET', /deployments.*offset=0&_search=ead&_search_name=ead/, {
```
qooban, line 12: Optional:

Maybe we can start using `stringify` from `query-string` in such URLs? I'm not fully sure it will be better in that particular case, so just suggesting, it is not a blocker by any means.

widgets/filter/src/Filter.tsx
stepek: No, i don't change any in this line, for clean-history let me don't adding anything in this line

widgets/filter/src/Filter.tsx
qooban: Make sense. 

As a side note I'm wondering shouldn't we start with RD-2693 before RD-2324. 
I think that within RD-2693 we could develop a single function which gets `display_name` and `id` values and produces user-friendly text (either `[display_name] ([id])` or just a `[display_name]`). Then we can just reuse it in all the places where we display deployment name and ID.

qooban (commented):
Looks fine. 

@Gelio provided fair comments worth to address, I've just added one general/optional.

widgets/filter/src/Filter.tsx
Gelio: Ok :slightly_smiling_face:

test/cypress/integration/widgets/filter_spec.ts
stepek: In case with regex it can be tricky to create correct URL, but in other cases should works

stepek:
### System Tests 
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/946/

test/cypress/integration/widgets/filter_spec.ts
```diff
@@ -1,5 +1,12 @@
 // @ts-nocheck File not migrated fully to TS
 describe('Filter', () => {
+    const getDropdownItems = (id?: string) => {
+        if (id) {
+            return cy.get(`${id} .menu > *`);
+        }
+        return cy.get('.menu > *');
+    };
```
Gelio, line 8: Minor:

Since the function is only used in a single test, I would keep it local to that test, instead of raising it to some higher scope and pollute the scope for other tests with this function :smile:

Gelio (approved):
Looks good :+1: I did not test it in the UI myself

test/cypress/integration/widgets/filter_spec.ts
```diff
@@ -8,36 +8,59 @@ describe('Filter', () => {
     });
 
     it('fills dropdowns with correct data', () => {
+        const getDropdownItems = (id?: string) => {
+            if (id) {
+                return cy.get(`${id} .menu > *`);
+            }
+            return cy.get('.menu > *');
+        };
```
qooban, line 16: Optional:

Will shorter version:
```suggestion
        const getDropdownItems = (id?: string) => id ? cy.get(`${id} .menu > *`) : cy.get('.menu > *');
```
work here?

widgets/filter/src/Filter.tsx
```diff
@@ -210,6 +212,10 @@ export default class Filter extends React.Component {
                                 entityName: 'Deployment',
                                 filter: blueprintFilter,
                                 pageSize: 20,
+                                textFormatter: item =>
+                                    item.display_name === item.id ? item.id : `${item.display_name} (${item.id})`,
```
qooban, line 216: I assume we will have common function for that, right?

I've mentioned something about it within https://github.com/cloudify-cosmo/cloudify-stage/pull/1510#discussion_r673936244

Are you going to add it within https://github.com/cloudify-cosmo/cloudify-stage/pull/1517 ?

qooban (commented):
Commented.

widgets/filter/src/Filter.tsx
stepek: yes in #1517

Gelio: So, depending on which PR goes in first, this should be changed in the later PR :smile: At least in my view :slightly_smiling_face:

test/cypress/integration/widgets/deployment_actions_spec.ts
```diff
@@ -125,7 +125,9 @@ describe('Deployment Action Buttons widget', () => {
         });
 
         it('adds new label by dropdown selection', () => {
-            cy.interceptSp('GET', '/labels/deployments?_search=exist').as('fetchFilteredKeys');
+            cy.interceptSp('GET', '/labels/deployments?_sort=id&_size=10&_offset=0&_search=exist').as(
```
Gelio, line 128: Minor:

If I'm not mistaken, you could have used the RegExp pattern here to skip specifying all query parameters verbatim :smile: 

```suggestion
            cy.interceptSp('GET', /\/labels\/deployments\?(.*)_search=exist').as(
```

widgets/filter/src/Filter.tsx
stepek: in point of view too

test/cypress/integration/widgets/filter_spec.ts
stepek: changed as you wish

Gelio (commented):
I've reviewed and IMO it looks mostly good to me :smile: Not many further comments :slightly_smiling_face:

test/cypress/integration/widgets/deployment_actions_spec.ts
stepek: I don't change this to regex to have add minimal change, but make sense to have regexp heer

test/cypress/integration/widgets/filter_spec.ts
```diff
@@ -51,6 +58,9 @@ describe('Filter', () => {
 
         cy.get('#dynamicDropdown2 > .label').should('have.length', 0);
         cy.get('#dynamicDropdown2').click();
+
+        // cy.wait('@fetchDeployments');
```
Gelio, line 62: Let's not leave comments in the code :slightly_smiling_face:

Gelio (commented):
One more comment about a comment :slightly_smiling_face:

widgets/filter/src/Filter.tsx
stepek: changed to use new util function

qooban (approved):
Looks good 👍🏻

If tests are fine, I'm OK with merging it.

stepek:
@qooban Thanks

PR #1512, opened 2021-07-19
RD-2769 Fix unbounded resizing of Deployments View panes

Gelio:
1. Add bounds on the minimum widths of panes in the Deployments View widget
when resizing them.
2. Make sure the draggable handle is always visible.
3. Modify system tests to verify that resizing is bounded.

## Video 

https://user-images.githubusercontent.com/889383/126150371-7be2fae6-0617-4730-b652-0132b4352938.mp4

## System tests

https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/933/pipeline

Queued 2021-07-19 13:02 CEST

### DIFF
--- a/widgets/common/src/deploymentsView/DeploymentsView.tsx
+++ b/widgets/common/src/deploymentsView/DeploymentsView.tsx
@@ -39,6 +39,8 @@ export interface DeploymentsViewProps {
     additionalFilterRules?: Stage.Common.Filters.Rule[];
 }
 
+const MIN_PANE_WIDTH = 100;
+
 export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
     toolbox,
     widget,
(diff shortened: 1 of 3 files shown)


widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -39,6 +39,8 @@ export interface DeploymentsViewProps {
     additionalFilterRules?: Stage.Common.Filters.Rule[];
 }
 
+const MIN_PANE_WIDTH = 100;
```
qooban, line 42: I know we don't have consistency across the whole project, but maybe we should follow
https://github.com/airbnb/javascript#naming--uppercase in this case:

> What about all const variables? - This is unnecessary, so uppercasing should not be used for constants within a file. It should be used for exported constants however.

What do you think?

qooban (commented):
Looks good in general 👍🏻  

Added just one minor comment.

widgets/common/src/deploymentsView/DeploymentsView.tsx
Gelio: I don't mind :smile: Yours sounds good, I'll change it to `minPaneWidth`

Gelio: Done :slightly_smiling_face:

qooban (approved):

PR #1515, opened 2021-07-21
RD-2639 Make system tests independent from running executions

qooban:
### Changes 

In this PR I merged `waitUntilNoExecutionIsActive` and `killExecutions` Cypress commands into one `killRunningExecutions`. It should be executed before each test which needs no executions running in the Cloudify Manager.

I found 3 tests that are designed in such a way that no running executions is needed:
* Maintenance Mode Button widget test - `maintenance_mode_button_spec.ts`
* Number of Nodes widget test - `servers_num_spec.ts`
* Snapshots widget test - `snapshots_spec.ts`

and added a call to to `killRunningExecutions` command there.

Especially making the first one (Maintenance Mode Button widget test) more reliable can help us avoid problems stemming from turned on Maintenance Mode (failing Topology widget test, Nodes widget test, Nodes Statistics widget test, Outputs/Capabilities widget test). Of course we could also check if Maintenance Mode is activated before the tests and deactive it in case it is active, but IMHO at this point it is not necessary as the root cause for maintenance mode activated problems were due to running executions.

### Verification

I tested my solution the following way:
* Created simple blueprint with long running `install` workflow (thanks @tehasdf for providing it)
* Deployed and installed that blueprint
* Executed each test (mentioned above) and verified that it passes (and it kills that `install` workflow started in the previous step

### System tests

[Stage-UI-System-Test #942](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/942/pipeline) - deadlock did not happen => cannot verify if the kill execution solution works fine
[Stage-UI-System-Test #943](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/943/pipeline) - no deadlock => no real verification ;-)

### DIFF
--- a/test/cypress/support/executions.ts
+++ b/test/cypress/support/executions.ts
@@ -1,3 +1,4 @@
+import { stringify } from 'query-string';
 import { addCommands, GetCypressChainableFromCommands } from 'cloudify-ui-common/cypress/support';
 import { waitUntilEmpty } from './resource_commons';
 
@@ -8,9 +9,6 @@ declare global {
         export interface Chainable extends GetCypressChainableFromCommands<typeof commands> {}
     }
 }
-
-const getPlannedOrStartedExecutionsUrl = (deploymentId: string) =>
-    `/executions?deployment_id=${deploymentId}&status=scheduled&status=queued&status=pending&status=started`;
 
 const commands = {
     getExecutions: (filter = '') => {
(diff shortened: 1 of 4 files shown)


qooban:
I'm unlucky - in [Stage-UI-System-Test #942](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/942/pipeline) deadlock issue (RD-2598) did not happen (running on branch created from `master`), so we don't see if my solution works ;-) I'll give it one more chance: [Stage-UI-System-Test #943](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/943/pipeline).

test/cypress/support/executions.ts
```diff
@@ -30,17 +28,18 @@ const commands = {
         cy.cfyRequest(`/executions/${executionId}`, 'POST', null, { action: 'kill' });
     },
 
-    killExecutions: (deploymentId: string) => {
-        cy.cfyRequest(getPlannedOrStartedExecutionsUrl(deploymentId), 'GET')
-            .then(response => response.body.items.forEach(({ id }: { id: string }) => cy.killExecution(id)))
-            .then(() => cy.waitUntilNoExecutionIsActive(deploymentId));
-    },
+    killRunningExecutions: () => {
+        const activeExecutionsUrl = `/executions?${stringify({
```
Gelio, line 32: :tada: for using `stringify`. Much better than the concatenation from earlier

Gelio (approved):
Looks good to me :+1: Cool idea, I hope it will cut the cascading failures in the tests :smile: I agree that checking whether maintenance mode is active in each test may be an overkill :slightly_smiling_face: 

I appreciate the thorough PR description :+1:

test/cypress/support/executions.ts
qooban: Yeah, we may consider moving into using it in other parts of our codebase as well.

qooban:
No deadlock in the second run ([Stage-UI-System-Test #943](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/943/pipeline)), so unfortunately I cannot verify it fully.

I'm quite confident that it is working fine (did local tests) and as this is a small change (only in tests) I'm going to merge this PR with single approval.

PR #1516, opened 2021-07-22
RD-2791 Fix warnings reported by `npm install`

qooban:
### Changes

To create this PR I did the following:
```
npm install eslint-plugin-react@^7.21.5 eslint-plugin-jsx-a11y@^6.4.1
npm remove mocha-junit-reporter
npm audit fix
npm install
```
which updated `package.json` and `package-lock.json`.

No warnings are reported by `npm install` right now:
```
audited 3535 packages in 11.528s

156 packages are looking for funding
  run `npm fund` for details

found 13 vulnerabilities (8 moderate, 5 high)
  run `npm audit fix` to fix them, or `npm audit` for details
```
Vulnerabilities have been found only in non-production dependencies.

That went quite smooth, but...
The bigger problem is that `eslint-plugin-react` introduced some changes in `react/destructuring-assignment` and now it looks like it treats all `render()` functions as React component class `render()` function and it seems like it is now checking if the first argument of it (in our case it is `widget`) is destructured. It checks it the same way as props are expected to be destructured in class components. 

I did fix all reported issues, but I'm really not sure it is the right way to go.

I thought that fixing this silly warnings will be quick and painless (just deps changes), but apparently it is not :-/ 
I'd like to get your opinion on the change related to that ESLint rule fixes.

### System Tests

I'll wait for opinions about the PR itself before running tests.


qooban:
Just FYI: This PR is not urgent to do the code review.

test/cypress/fixtures/widgets/fibonacciSequenceWidget.tsx
```diff
@@ -31,7 +31,9 @@ Stage.defineWidget<unknown, unknown, FibonacciSequenceWidgetConfiguration>({
     ],
     render: widget => {
         const { WidgetsList } = Stage.Shared.Widgets;
-        const { targetSequenceLength, sequence = initialSequence } = widget.configuration;
+        const {
+            configuration: { targetSequenceLength, sequence = initialSequence }
+        } = widget;
```
Gelio, line 36: IMO this looks worse than before :smile:

Gelio (commented):
Thanks for a detailed description of the changes you did and of the problem :+1: 

> I did fix all reported issues, but I'm really not sure it is the right way to go.

Yeah, I'm not sure either :smile: IMO the rule makes sense for components, but not for widgets. Thus, my suggestion would be to disable the `react/destructuring-assignment` rule only for the `widgets` directory, and keep the code unchanged :slightly_smiling_face:

test/cypress/fixtures/widgets/fibonacciSequenceWidget.tsx
stepek: me2

widgets/blueprintNum/src/widget.tsx
```diff
@@ -26,7 +26,7 @@ Stage.defineWidget<unknown, unknown, { page?: unknown }>({
     ],
     fetchUrl: '[manager]/blueprints?_include=id&_size=1',
 
-    render(widget, data) {
+    render({ configuration: { page } }, data) {
```
stepek, line 29: Prefer prev version

qooban: Yep, agree, reverted.

test/cypress/fixtures/widgets/fibonacciSequenceWidget.tsx
qooban: Yeah, agree, reverted.

qooban:
> Thus, my suggestion would be to disable the react/destructuring-assignment rule only for the widgets directory, and keep the code unchanged

I disabled it for `widget.{jsx,tsx}` files only as I think in the components' files it still make sense.
Reported a bug on eslint-plugin-react project: https://github.com/yannickcr/eslint-plugin-react/issues/3022 .

widgets/.eslintrc.json
```diff
@@ -34,6 +34,12 @@
                     }
                 ]
             }
+        },
+        {
+            "files": ["./*/src/widget.{jsx,tsx}"],
+            "rules": {
+                "react/destructuring-assignment": "off"
```
Gelio, line 41: Minor/optional:

Can we also add the description of why it's off (similarly to what we added in tsx files)? ESLint supports comments in JSON (we have them in the main eslintrcs)

Gelio (approved):
LGTM :+1: Let's see if the system tests pass

qooban:
System tests started: [Stage-UI-System-Test #950](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/950)

widgets/.eslintrc.json
qooban: Sure, added.

Gelio (approved):

stepek (approved):

qooban:
Thanks for the approvals. 

In the recent system test run ([Stage-UI-System-Test #950](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/950)) I saw 3 widget tests failing:
1. Deployments View widget - known issue (RD-2598)
2. Filters widget - known issue (RD-2664)
3. Deployment Wizard Buttons widget - that's something new, I tested if that can be caused by running execution, but no, locally even with running execution it passes without errors.

I'm not going to report a bug for Deployment Wizard Buttons widget test, yet. For now it seems to me like one-time issue.

I have scheduled one more system test run: [Stage-UI-System-Test #955](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/955)
to be fully confident. I'll update this conversation after it is done.

qooban:
All tests passed, so I'm merging this PR.

PR #1517, opened 2021-07-22
RD-2693 Change method of displaying deployment display name when id and name is equal.

stepek:
Added in deployment and execution modal check to prevent showing two id in modal headers. also changed little bit translation file to move adding brackets for ids to react component itself.


app/translations/en.json
```diff
@@ -433,7 +433,7 @@
                     "header": {
                         "noDeployment": "Execute workflow {{workflowName}}",
                         "singleDeployemnt": "Execute workflow {{workflowName}} on {{deploymentId}}",
-                        "singleNamedDeployemnt": "Execute workflow {{workflowName}} on {{deploymentName}} ({{deploymentId}})",
+                        "singleNamedDeployemnt": "Execute workflow {{workflowName}} on {{deploymentName}} {{ deploymentId }}",
```
Gelio, line 436: Nice idea :+1: We can take it a step further and have only `deploymentName` that we will fill in in runtime with either `name (id)` or just `name` depending on whether `name === id`. Then, we would not have to specify this `{{deploymentName}} {{deploymentId}}` each time

Moreover, the current solution leaves a trailing space at the end if `deploymentId` is `undefined`

Also, then, the logic of computing the string to pass to be inserted as `deploymentName` into i18n can be written once, as an utility function

What do you think?

widgets/common/src/RemoveDeploymentModal.tsx
```diff
@@ -32,7 +32,7 @@ const RemoveDeploymentModal: FunctionComponent<RemoveDeploymentModalProps> = ({
 
     const content = i18n
         .t(`widgets.common.deployments.removeModal.${force ? 'forceDelete' : 'delete'}Description`, {
-            deploymentId,
+            deploymentId: deploymentId === deploymentName ? undefined : `(${deploymentId})`,
```
Gelio, line 35: It would be nice to extract this logic to some utility, maybe in `Stage.Utils` (`formatDeploymentDisplayName`?)

Gelio (commented):
I like the idea :+1:

widgets/common/src/RemoveDeploymentModal.tsx
stepek: Thanks, let me create utils function for that and remove `deploymentName` `deploymentId` from template

app/translations/en.json
stepek: I think about this approach before implementing, and now it's can be better.

stepek:
### System Tests 
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/953/

app/translations/en.json
```diff
@@ -433,7 +433,7 @@
                     "header": {
                         "noDeployment": "Execute workflow {{workflowName}}",
                         "singleDeployemnt": "Execute workflow {{workflowName}} on {{deploymentId}}",
-                        "singleNamedDeployemnt": "Execute workflow {{workflowName}} on {{deploymentName}} ({{deploymentId}})",
+                        "singleNamedDeployemnt": "Execute workflow {{workflowName}} on {{ name }}",
```
Gelio, line 436: Minor:

To me, `name` sounds a bit ambiguous and I would change it to

```suggestion
                        "singleNamedDeployemnt": "Execute workflow {{workflowName}} on {{ deploymentName }}",
```

and apply the same change across all the files in `en.json` :slightly_smiling_face:

app/utils/stageUtils.ts
```diff
@@ -64,6 +64,28 @@ export default class StageUtils {
      */
     static formatLocalTimestamp = TimeUtils.formatLocalTimestamp.bind(TimeUtils);
 
+    /**
+     * Formatting resource id and display name.
+     *
+     * @param data
+     * @returns string
```
Gelio, line 71: Minor:

IMO this is redundant when we have TS :slightly_smiling_face: If you see such comments in the existing code it's probably because they have not been removed yet :smile:

app/utils/stageUtils.ts
```diff
@@ -64,6 +64,28 @@ export default class StageUtils {
...
+     */
+    static formatDisplayName(data: Partial<{ id: string; displayName: string }>): string {
+        if (data.id === undefined) {
+            throw new Error('id is undefined!');
+        }
```
Gelio: We can validate this condition with TypeScript if we do:

```ts
data: { id: string; displayName?: string }
```

in the parameter :smile:

app/utils/stageUtils.ts
```diff
@@ -64,6 +64,28 @@ export default class StageUtils {
      */
     static formatLocalTimestamp = TimeUtils.formatLocalTimestamp.bind(TimeUtils);
 
+    /**
+     * Formatting resource id and display name.
+     *
+     * @param data
+     * @returns string
+     */
+    static formatDisplayName(data: Partial<{ id: string; displayName: string }>): string {
```
Gelio, line 73: Minor:

Since we're using this function only with deployment names for now, I would rename it to

```suggestion
    static formatDeploymentDisplayName(data: Partial<{ id: string; displayName: string }>): string {
```

test/jest/utils/StageUtils.test.ts
```diff
@@ -0,0 +1,43 @@
+import StageUtils from 'utils/stageUtils';
+
+describe('(Utils) StageUtils', () => {
+    describe('formatDisplayName', () => {
```
Gelio, line 4: Very nice unit tests :heart:

widgets/common/src/RemoveDeploymentModal.tsx
```diff
@@ -32,8 +32,7 @@ const RemoveDeploymentModal: FunctionComponent<RemoveDeploymentModalProps> = ({
 
     const content = i18n
         .t(`widgets.common.deployments.removeModal.${force ? 'forceDelete' : 'delete'}Description`, {
-            deploymentId,
-            deploymentName
+            name: Stage.Utils.formatDisplayName({ id: deploymentId, displayName: deploymentName })
```
Gelio, line 35: This looks awesome :tada:

widgets/common/src/UpdateDeploymentModal.tsx
```diff
@@ -220,7 +220,11 @@ function UpdateDeploymentModal({ open, deploymentId, deploymentName, onHide, too
     return (
         <Modal open={open} onClose={onHide} className="updateDeploymentModal">
             <Modal.Header>
-                <Icon name="edit" /> Update deployment {deploymentName} ({deploymentId})
+                <Icon name="edit" />{' '}
+                {`Update deployment ${Stage.Utils.formatDisplayName({
+                    id: deploymentId,
+                    displayName: deploymentName
+                })}`}
```
Gelio, line 227: Minor:

I don't think we need to make this be a template literal :smile: 

```suggestion
                Update deployment {Stage.Utils.formatDisplayName({
                    id: deploymentId,
                    displayName: deploymentName
                })}
```

app/translations/en.json
```diff
@@ -433,7 +433,7 @@
                     "header": {
```
Gelio, line 433: btw. I have noticed that there are some uses of `deploymentId` in `en.json` for which there is no `deploymentName` displayed. Do you think we should create tickets for them?

```sh
10:12 $ rg deploymentId -g en.json
app/translations/en.json
426:                    "fetchError": "Cannot fetch deployment '{{ deploymentId }}'.",
435:                        "singleDeployemnt": "Execute workflow {{workflowName}} on {{deploymentId}}",
442:                        "workflowError": "Workflow '{{ workflowName }}' does not exist for deployment '{{ deploymentId }}'.",
482:                        "deploymentId": {
557:                "fetchingLabelsError": "Cannot fetch labels for deployment '{{deploymentId}}'",
924:            "addHeader": "Add labels for deployment '{{deploymentId}}'",
```

app/utils/stageUtils.ts
```diff
@@ -64,6 +64,28 @@ export default class StageUtils {
...
+        if (data.id === undefined) {
+            return '';
+        }
```
Gelio, line 76: IMO we should not allow for this case at all :smile: We can validate it in TS (make `id` required and `displayName` optional)

Gelio (commented):
Thanks for applying my suggestion :+1: Looks very cool now :smile: :tada:

app/utils/stageUtils.ts
stepek: no, in some cases id is `undefined` don't know why but it is (checked by running system tests and throwing error when is undefined. and it's outdated

app/utils/stageUtils.ts
stepek: imho, No, because this function can be used in more places over only deployment case.

app/translations/en.json
stepek: Probably yee,

app/utils/stageUtils.ts
stepek: No, in some cases `id` is undefined

app/utils/stageUtils.ts
Gelio: Well, then IMO we should fix these places rather than ignore them :smile: If we expect to show deployment names and the function fails, that's a silent error that, to me, is worse than an explicit failure :smile:

Gelio: Also, if you don't exactly apply my suggestion, could you not resolve the conversation to let me answer back :slightly_smiling_face: ? This is the usual flow we follow during code review

app/utils/stageUtils.ts
Gelio: Yeah, but it's not used in other cases at the moment, and I can't see any use for it in the nearest future (I don't know of other resources that will have a display name that's different from ID).

@qooban Do you know if there are plans to add display names to other resources?

app/utils/stageUtils.ts
stepek: Added log.error in this case

Gelio: Hmm I don't know if that's enough :smile: I think we should fix these places instead of having a broken UI (the UI won't display the ID or name in these cases)

widgets/common/src/ExecuteDeploymentModal.tsx
```diff
@@ -240,7 +240,13 @@ export default function ExecuteDeploymentModal({
         <Modal open={open} onClose={onHide} className="executeWorkflowModal">
             <Modal.Header>
                 <Icon name="cogs" />{' '}
-                {t(headerKey, { workflowName, deploymentName, deploymentId: _.head(deploymentsList) })}
+                {t(headerKey, {
+                    workflowName,
+                    deploymentName: Stage.Utils.formatDisplayName({
+                        id: _.head(deploymentsList),
```
qooban, line 246: Optional:

As you are modifing this line anyway, maybe you can import `head` from `lodash` at once according to one of the ongoing refactors -https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Using-explicit-imports-in-widgets ?

app/utils/stageUtils.ts
qooban: > @qooban Do you know if there are plans to add display names to other resources?

No, I am not aware of any plans to add display names to other resources at the moment.

I agree that `formatDeploymentDisplayName` will be more accurate here, but I'm actually OK with having more generic name as well.

app/translations/en.json
qooban: Good catch @Gelio 👍🏻 

@stepek Could you create tickets for them and set the epic to RD-2700?

qooban (approved):
Looks good to me 👍🏻

app/utils/stageUtils.ts
Gelio: Alright, since it's optional, @stepek you can resolve the conversation if you don't plan of making the change :slightly_smiling_face: I don't mind

PR #1530, opened 2021-08-02
RD-2476 Add modal for the blueprint marketplace and use it in Blueprints widget.

stepek:
Initial changes for new blueprint install flow.
in next pr will add menu after clicking upload, system tests and possibility to change tabs in modal
image


widgets/common/src/blueprintMarketplaceModal/index.ts
```diff
@@ -0,0 +1,13 @@
...
+    namespace Stage.Common {
+        // eslint-disable-next-line import/prefer-default-export
+        export { BlueprintMarketplaceModal };
+    }
+}
+
+Stage.defineCommon({
+    name: 'BlueprintMarketplaceModal',
+    common: React.memo(BlueprintMarketplaceModal, _.isEqual)
+});
```
kubama, line 13: Wouldn't it be more compact and at the same time in line with other such components if we only had `widgets/common/src/BlueprintMarketplaceModal.tsx` file?

widgets/blueprints/src/BlueprintsList.tsx
```diff
@@ -190,8 +190,26 @@ export default class BlueprintList extends React.Component<BlueprintListProps, B
...
+                            url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/vm-examples.json'
+                        },
+                        {
+                            name: 'Kubernetes Blueprint Examples',
+                            url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/k8s-examples.json'
+                        },
+                        {
+                            name: 'Orchestrator Blueprint Examples',
+                            url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/orc-examples.json'
+                        }
```
kubama, line 208: Shouldn't it be somehow externalized? At least the tab names moved to translation file?

widgets/blueprints/src/BlueprintsList.tsx
```diff
@@ -190,8 +190,26 @@ export default class BlueprintList extends React.Component<BlueprintListProps, B
...
+                    ]}
+                />
 
-                <UploadBlueprintModal open={showUploadModal} onHide={this.hideUploadModal} toolbox={toolbox} />
+                <UploadBlueprintModal open={false} onHide={this.hideUploadModal} toolbox={toolbox} />
```
kubama, line 212: Can be removed?

widgets/blueprints/src/BlueprintsList.tsx
```diff
@@ -190,8 +190,26 @@ export default class BlueprintList extends React.Component<BlueprintListProps, B
                     onHide={this.hideDeploymentModal}
                     toolbox={toolbox}
                 />
+                <BlueprintMarketplaceModal
+                    open={showUploadModal}
+                    onHide={this.hideUploadModal}
+                    tabs={[
```
stepek, line 196: Will be moved to config later.

widgets/common/src/blueprintMarketplaceModal/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,89 @@
+import type { FunctionComponent } from 'react';
+
+import { noop } from 'lodash';
+
+const i18nPrefix = 'widgets.common.blueprintMarketPlace';
```
kubama, line 5: How about using `Stage.Utils.getT`?

widgets/common/src/blueprintMarketplaceModal/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+    url: string;
+}
+
+interface BlueprintMarketplaceModalProps {
+    open: boolean;
+    tabs: Tab[];
+    onHide: () => void;
+}
+
+const getPageLayout = (value: Tab[]) => ({
```
kubama, line 18: How about a more descriptive name than just `value`?

widgets/common/src/blueprintMarketplaceModal/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+                        isEditMode={false}
+                        // NOTE: No need to handle the events below since edit mode is always off
+                        onWidgetRemoved={noop}
+                        onTabAdded={noop}
+                        onTabRemoved={noop}
+                        onTabUpdated={noop}
+                        onTabMoved={noop}
+                        onWidgetAdded={noop}
+                        onLayoutSectionAdded={noop}
+                        onLayoutSectionRemoved={noop}
```
kubama, line 73: Can't we make these props optional?

widgets/common/src/blueprintMarketplaceModal/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+    return (
+        <div>
+            <Modal open={open} onClose={onHide} className="uploadBlueprintModal">
+                <Modal.Header>
+                    <Icon name="upload" /> {i18n.t(`${i18nPrefix}.modal.header`)}
+                </Modal.Header>
+                <Modal.Content>
+                    <PageContent
+                        page={getPageLayout(tabs) as any}
+                        isEditMode={false}
```
kubama, line 64: Can we make this prop optional?

widgets/common/src/blueprintMarketplaceModal/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const BlueprintMarketplaceModal: FunctionComponent<BlueprintMarketplaceModalProps> = ({ open, onHide, tabs }) => {
+    const { i18n } = Stage;
+
+    const { CancelButton, Icon, Modal } = Stage.Basic;
+    const { PageContent } = Stage.Shared.Widgets;
+
+    return (
+        <div>
+            <Modal open={open} onClose={onHide} className="uploadBlueprintModal">
```
kubama, line 57: Can't see this class name being referenced anywhere

widgets/blueprints/src/BlueprintsList.tsx
stepek: as my comment https://github.com/cloudify-cosmo/cloudify-stage/pull/1530#discussion_r681480510 will be moved to better place.

widgets/blueprints/src/BlueprintsList.tsx
stepek: This will be moved to popup menu.

widgets/blueprints/src/BlueprintsList.tsx
kubama: > Regarding translations probably yes, but don't know if we can set labels from translation files in initial config of widget.

I don't see any reason why it shouldn't work.

kubama (approved):

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+                        y: 0,
+                        width: 12,
+                        height: 24,
+                        definition: 'blueprintCatalog',
+                        configuration: {
+                            jsonPath: tab.url,
+                            displayStyle: 'catalog'
+                        },
+                        drillDownPages: {},
+                        maximized: false
```
qooban, line 36: Hmm... I'm wondering if maximize would work with tabs, it'd probably be good to use `maximized: true` and avoid hardcoding `width` and `height`. WDYT?

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+    name: 'details-pane-widgets',
+    layout: [
+        {
+            type: 'tabs',
+            content: tabs.map((tab, index) => ({
+                name: tab.name,
+                widgets: [
+                    {
+                        id: index.toString(),
+                        showBorder: false,
```
qooban, line 24: Oh, it worked with such override? Great then! 🙂

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+                        name: 'Blueprints Catalog',
+                        x: 0,
+                        y: 0,
+                        width: 12,
+                        height: 24,
```
qooban, line 29: Maybe we can avoid defining some of these (as AFAIR defaults will should be taken) or all of them are necessary to provide?

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+                </Modal.Actions>
+            </Modal>
+        </div>
+    );
+};
+
+BlueprintMarketplaceModal.propTypes = {
+    open: PropTypes.bool.isRequired,
+    onHide: PropTypes.func.isRequired
+};
```
qooban, line 72: I'm not sure about adding this. Is `BlueprintMarketplaceModal` to be used in JS or not fully migrated TS components?

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <div>
+            <Modal open={open} onClose={onHide}>
+                <Modal.Header>
+                    <Icon name="upload" /> {t(`modal.header`)}
+                </Modal.Header>
+                <Modal.Content>
+                    <PageContent page={getPageLayout(tabs) as any} />
+                </Modal.Content>
+                <Modal.Actions>
+                    <CancelButton onClick={onHide} disabled={false} />
```
qooban, line 62: By default `CancelButton` has `Cancel` label AFAIR, shouldn't we have `Close` button here? (by having `content` prop)

(my reasoning: `Cancel` suggests that there should be an option to move forward as well: `Apply`, `OK`, etc.)

widgets/blueprints/src/BlueprintsList.tsx
```diff
@@ -190,8 +190,26 @@ export default class BlueprintList extends React.Component<BlueprintListProps, B
...
+                            url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/k8s-examples.json'
+                        },
+                        {
+                            name: 'Orchestrator Blueprint Examples',
+                            url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/orc-examples.json'
+                        }
+                    ]}
+                />
 
-                <UploadBlueprintModal open={showUploadModal} onHide={this.hideUploadModal} toolbox={toolbox} />
```
qooban, line 194: So right now (before you implement dropdown menu) we won't have an option to upload a blueprint using `UploadBlueprintModal` from Blueprints list? Or am I missing something? 🙂

qooban (commented):
Nicely done. I've added some comments.

Having a screenshot or a video in the description would be great 🙂

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+const getPageLayout = (tabs: Tab[]) => ({
+    name: 'details-pane-widgets',
+    layout: [
+        {
+            type: 'tabs',
+            content: tabs.map((tab, index) => ({
+                name: tab.name,
+                widgets: [
+                    {
+                        id: index.toString(),
```
kubama, line 23: One more comment: wouldn't it be safer if this id was more specific to the context in which it is used? A plain number introduces a risk of conflict if a number is used as an id elsewhere in the application. Not sure if it would be an issue, but since it's an ID I would try to keep it unique.

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const BlueprintMarketplaceModal: FunctionComponent<BlueprintMarketplaceModalProps> = ({ open, onHide, tabs }) => {
+    const { getT } = Stage.Utils;
+    const t = getT('widgets.common.blueprintMarketPlace');
+
+    const { CancelButton, Icon, Modal } = Stage.Basic;
+    const { PageContent } = Stage.Shared.Widgets;
+
+    return (
+        <div>
```
kubama, line 53: Is this wrapping `div` required?

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+interface BlueprintMarketplaceModalProps {
+    open: boolean;
+    tabs: Tab[];
+    onHide: () => void;
+}
+
+const getPageLayout = (tabs: Tab[]) => ({
+    name: 'details-pane-widgets',
```
kubama, line 15: I guess this name is just copied - however I think it's not suitable in this context.

kubama: Actually I think it's not needed at all

widgets/common/src/BlueprintMarketplaceModal.tsx
kubama: I'm facing a similar challenge and it looks like such override does nothing. Or does it?

kubama: Once again, a screenshot would be helpful 🙂.

widgets/blueprints/src/BlueprintsList.tsx
stepek: yes, but done as description in ticket :D and i merge this to feature branch over master.

widgets/common/src/BlueprintMarketplaceModal.tsx
stepek: how about using generated uuid here?

widgets/common/src/BlueprintMarketplaceModal.tsx
stepek: will add, and remove this line.

widgets/common/src/BlueprintMarketplaceModal.tsx
stepek: no

widgets/common/src/BlueprintMarketplaceModal.tsx
stepek: Will change

widgets/common/src/BlueprintMarketplaceModal.tsx
stepek: I can be removed, but as i remember good we agreed with have proptypes validation, isn't it?

widgets/common/src/BlueprintMarketplaceModal.tsx
kubama: How about using camel-cased tab name?

stepek: ok let me change to  camel-cased tab name

widgets/common/src/BlueprintMarketplaceModal.tsx
qooban: There are two points about it in: https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/Cloudify+UI+Code+Style

1. SHOULD: Use propTypes for JS/TS components used in not-fully-migrated TS components
2. MUST: Skip using propTypes for TS components used exclusively in fully-migrated TS components

AFAIR we discussed that we can have `propTypes` for not-fully-migrated TS components, but we didn't see much value in having it in fully-migrated TS components, and the main problem we wanted to avoid is code duplication and additional maintenance cost for taking care of both - TS type and propTypes definition.

@kubama What is your take on that?

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return (
+        <Modal open={open} onClose={onHide}>
+            <Modal.Header>
+                <Icon name="upload" /> {t(`modal.header`)}
+            </Modal.Header>
+            <Modal.Content>
+                <PageContent page={getPageLayout(tabs) as any} />
+            </Modal.Content>
+            <Modal.Actions>
+                <CancelButton onClick={onHide} disabled={false} content="Close" />
```
qooban, line 58: Do we need `disabled={false}`?

```suggestion
                <CancelButton onClick={onHide} content="Close" />
```

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const getPageLayout = (tabs: Tab[]) => ({
+    layout: [
+        {
+            type: 'tabs',
+            content: tabs.map(tab => ({
+                name: tab.name,
+                widgets: [
+                    {
+                        id: tab.name.replace(' ', '-'),
```
qooban, line 22: What if we have 2 or more tabs with the same name? 🙂 
Maybe just add a prefix or suffix with tab number?

qooban (commented):
Added some comments.

widgets/blueprints/src/BlueprintsList.tsx
qooban: Oh, just realized you intend to merge to feature branch instead of `master`.
OK, let's try and see how it goes.

app/components/shared/widgets/PageContent.tsx
```diff
@@ -17,16 +17,16 @@ import EmptyContainerMessage from '../../EmptyContainerMessage';
...
+    onWidgetAdded = _.noop,
+    onTabAdded = _.noop,
+    onTabRemoved = _.noop,
+    onTabUpdated = _.noop,
+    onTabMoved = _.noop,
+    onLayoutSectionAdded = _.noop,
+    onLayoutSectionRemoved = _.noop,
     page,
-    isEditMode
+    isEditMode = false
```
kubama, line 29: 👍

I think there is at least one place where these now optional props can be removed - please see deployments view widget.

widgets/common/src/BlueprintMarketplaceModal.tsx
kubama: @qooban I agree

widgets/common/src/BlueprintMarketplaceModal.tsx
qooban: Also I think we should move "Close" to translation file.

widgets/common/src/BlueprintMarketplaceModal.tsx
stepek: ok

app/translations/en.json
```diff
@@ -655,6 +655,12 @@
                         "placeholder": "Provide the image file URL or click browse to select a file"
                     }
                 }
+            },
+            "blueprintMarketplace": {
+                "modal": {
+                    "header": "Blueprint marketplace",
+                    "cancelButton": "Cancel"
```
qooban, line 662: Didn't we plan to name the button "Close"?

qooban (commented):
Just one from my side.

app/translations/en.json
stepek: oh, right should be close

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+        {
+            type: 'tabs',
+            content: tabs.map((tab, index) => ({
+                name: tab.name,
+                widgets: [
+                    {
+                        id: `blueprint-catalog-${index}`,
+                        name: 'Blueprints Catalog',
+                        x: 0,
+                        y: 0,
```
qooban, line 25: Are those necessary?

PR #1531, opened 2021-08-03
RD-2482 Removed plugin upload confirmation dialog

kubama:
This PR removes confirmation dialog on plugin upload in plugins catalogue widget. The upload process is now started with a single click. Additionally, multiple plugins can be uploaded in parallel, as the operation does not block the UI. A standard spinner indicates which plugins are being uploaded.

image

NOTE: I had a chat with @tehasdf and at the moment there is no way to track the upload process in a way other than waiting for the upload request to complete. So if the user starts an upload, leaves the catalogue and returns back the spinner will be no longer available - if the plugin is still uploading it will be again available for the upload.


kubama:
System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1020/

kubama:
System tests rerun after fixing the test: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1022/

stepek (commented):
tomorrow will do review

qooban:
I'll review tomorrow, too.

widgets/pluginsCatalog/src/Actions.ts
```diff
@@ -45,12 +45,9 @@ export default class Actions {
...
     }
 
-    doUpload(
-        { url: wagonUrl, yamlUrl, icon: iconUrl, title }: { url: string; yamlUrl: string; icon: string; title: string },
-        visibility: string
-    ) {
+    doUpload({ url: wagonUrl, yamlUrl, icon: iconUrl, title }: PluginUploadData) {
         const params = {
-            visibility,
+            visibility: 'tenant',
```
qooban, line 50: Maybe we can have `const defaultVisibility = 'tenant'` defined in widgets consts?
That way @stepek will be able to use it in Blueprints Upload feature as well.

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -60,33 +67,67 @@ export default class PluginsCatalogList extends React.Component<PluginsCatalogLi
...
-        this.setState({ plugin });
-        this.showModal();
+        this.setState(prevState => ({ uploadingPlugins: [...prevState.uploadingPlugins, plugin] }));
+
+        new Actions(toolbox, widget.configuration.jsonPath)
+            .doUpload(plugin)
+            .then(() => {
+                toolbox.getEventBus().trigger('plugins:refresh');
+                this.setState(prevState => ({
+                    successMessages: [...(prevState.successMessages ?? []), `${plugin.title} successfully uploaded`]
```
qooban, line 80: Are you planning to move text to the translation file?
(applies also to the other text labels in this file)

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -60,33 +67,67 @@ export default class PluginsCatalogList extends React.Component<PluginsCatalogLi
...
+            yamlUrl: item.link
+        };
+
+        const uploadInProgress = _.find(uploadingPlugins, pluginUploadData);
+
+        return (
+            <div style={{ position: 'relative' }}>
+                {uploadInProgress && <Stage.Basic.Loader active />}
+                <UploadPluginButton
+                    style={{ visibility: uploadInProgress && 'hidden' }}
```
qooban, line 106: I'm wondering - wouldn't it be easier to use `loading` prop on a button instead of conditionally hiding a button and showing `Loader`?

Ref.: https://react.semantic-ui.com/elements/button/#states-loading

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -105,10 +146,21 @@ export default class PluginsCatalogList extends React.Component<PluginsCatalogLi
 
         return (
             <div>
-                {success && (
-                    <Message color="green" onDismiss={() => this.setState({ success: null })}>
-                        {success}
+                {_.map(successMessages, message => (
+                    <Message
+                        key={message}
+                        color="green"
```
qooban, line 152: I know it was used here before, but maybe we should change it either to `positive` or `success` boolean prop?
We will avoid hardcoding to a specific color then.

Ref.: https://react.semantic-ui.com/collections/message/

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -105,10 +146,21 @@ export default class PluginsCatalogList extends React.Component<PluginsCatalogLi
...
+                        onDismiss={() =>
+                            this.setState(prevState => ({
+                                successMessages: _.without(prevState.successMessages, message)
+                            }))
+                        }
+                    >
+                        {message}
                     </Message>
+                ))}
+                {!_.isEmpty(errorMessages) && (
```
qooban, line 162: Optional:

We can probably import `isEmpty` directly from `lodash` and update other usages of `lodash` in this file. Not a blocker at all.

qooban (commented):
Looks good generally, added some (mostly minor) comments.

1. I see system tests (https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1022/) did not pass unfortunately, probably just re-run needed as it seems like infra issue.
2. Are you planning to add a test for uploading multiple plugins?

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -60,33 +67,67 @@ export default class PluginsCatalogList extends React.Component<PluginsCatalogLi
...
+            url: item.wagon.url,
+            title: item.title,
+            icon: item.icon,
+            yamlUrl: item.link
+        };
+
+        const uploadInProgress = _.find(uploadingPlugins, pluginUploadData);
+
+        return (
+            <div style={{ position: 'relative' }}>
```
stepek, line 103: @qooban @kubama What convention we have regarding inline styles? it's ok to put inline styles in the code insted of use css files?

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -105,10 +146,21 @@ export default class PluginsCatalogList extends React.Component<PluginsCatalogLi
 
         return (
             <div>
-                {success && (
-                    <Message color="green" onDismiss={() => this.setState({ success: null })}>
-                        {success}
+                {_.map(successMessages, message => (
```
stepek, line 149: How about using js api for mapping?

widgets/pluginsCatalog/src/Actions.ts
kubama: Actually there's already such const - I reused it.

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
kubama: Yes, moved.

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
qooban: We had few discussions around that topic and there's a conclusion point in the Cloudify UI Code Style page:
https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/Cloudify+UI+Code+Style#MUST:-Use-CSS-in-JS-over-SCSS-styles

Would you like to share your opinion on that?

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
kubama: Definitely, changed.

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
kubama: Done

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
kubama: Done

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
kubama: I believe we had an agreement that inline styles or OK, especially for simplest cases like this one.

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
kubama: Done

kubama:
> 1. I see system tests (https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1022/) did not pass unfortunately, probably just re-run needed as it seems like infra issue.

Fresh run: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1032/

> 2. Are you planning to add a test for uploading multiple plugins?

I redesigned plugins catalog spec to cover multiple plugins upload.

qooban (approved):
Looks good 👍🏻

PR #1533, opened 2021-08-05
RD-2664 Implemented cy.wait workaround

kubama:
Looks like the issue described in RD-2664 is still there, it happened in two consecutive builds making last nightly build to fail. I'm reopening the ticket and disabling the flaky tests.


qooban (commented):
Uff... This issue (RD-2664) is problematic and annoying :-/

I'm not sure disabling is the best idea as we both already invested some time in the investigation and I don't think the situation will be better very soon. We didn't manage to reproduce the problem locally and that makes fixing it very hard.

If you think the reason for the problem is in `DynamicDropdown`, then I suspect it might get fixed by RD-2591. But I don't think it'll be implemented soon (quite complex, big and risky). Maybe it'd better to implement ugly workaround in https://github.com/cloudify-cosmo/cloudify-stage/blob/7921d24114db51a13a4545602cb0a95a64293aeb/test/cypress/integration/widgets/filters_spec.ts#L419 by having something like `cy.wait(1000)` before waiting for the request. That way we'll probably catch the second one. WDYT?

kubama:
@qooban I'm OK with leaving the tests enabled. In the mean time I can try your suggestion with `cy.wait` - I'd expect it to fix it, although it will be difficult to confirm as the issue seems to be occurring occasionally.

I agree the issue is tricky. RD-2591 may fix the issue or not - hard to say without knowing the actual cause.

I can

kubama:
I reverted the test skip and implemented  the `cy.wait` workaround. Submitted 3 system tests jobs: [#1045](https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1045/), [#1046](https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1046/) and [#1047](https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1047/)

kubama:
System tests passed. Please review.

qooban (approved):

PR #1529, opened 2021-07-29
RD-2191 Fetch less data in Deployments View

Gelio:
This PR changes the parameters when fetching data in the Deployments View widget to only include a specific set of fields in the responses.

This reduces the amount of data received by the UI, since the `workflows`, `policy_triggers`, and other fields are not fetched. Take a look at the diff in `various-statuses.json` to see the difference - it's quite substantial :smile: 

Unfortunately, using some fields with `_include` results in a 500 backend error. I have reported it in RD-2900. Until that is fixed, this branch will not work.

## System tests

I will run them after RD-2900 is fixed since they will be failing at the moment.

PR #1534, opened 2021-08-05
RD-2482 RD-2480 Added marketplace upload option to plugins widget

kubama:
The upload button was converted to a menu dropdown:
image

The new modal window:
image


widgets/plugins/src/MarketplaceModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+                    onWidgetUpdated={_.noop}
+                    onWidgetRemoved={_.noop}
+                    onWidgetAdded={_.noop}
+                    onTabAdded={_.noop}
+                    onTabRemoved={_.noop}
+                    onTabUpdated={_.noop}
+                    onTabMoved={_.noop}
+                    onLayoutSectionAdded={_.noop}
+                    onLayoutSectionRemoved={_.noop}
+                    isEditMode={false}
```
kubama, line 39: Won't be necessary once https://github.com/cloudify-cosmo/cloudify-stage/pull/1530 is merged

kubama:
System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1034/

widgets/plugins/src/PluginsTable.tsx
```diff
@@ -1,8 +1,12 @@
 // @ts-nocheck File not migrated fully to TS
+import MarketplaceModal from './MarketplaceModal';
+
 /**
  * Created by kinneretzin on 02/10/2016.
  */
```
qooban, line 6: We can remove it at once according to ongoing refactors.

Ref.: https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Removing-comments-with-information-about-the-author

widgets/plugins/src/MarketplaceModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+                />
+            </Modal.Content>
+            <Modal.Actions>
+                <CancelButton onClick={onHide} disabled={false} content={t(`upload.close`)} />
```
qooban, line 43: Do we need `disabled={false}` here?

```suggestion
                <CancelButton onClick={onHide} content={t(`upload.close`)} />
```

widgets/plugins/src/MarketplaceModal.tsx
```diff
@@ -0,0 +1,47 @@
+const t = Stage.Utils.getT('widgets.plugins');
+
+const modalContentLayout: { type: 'widgets'; content: any } = {
+    type: 'widgets',
+    content: [
+        {
+            id: 'pluginsCatalog',
+            name: t('upload.catalog'),
+            width: 12,
+            height: 24,
```
qooban, line 10: What do you think about adding `maximized: true` instead?

(it will be aligned with the Blueprint upload changes then - https://github.com/cloudify-cosmo/cloudify-stage/pull/1530)

widgets/plugins/src/MarketplaceModal.tsx
qooban: @stepek As you are pushing to feature branch which means that probably your change related to `PageContent` won't be sooner in `master` branch than this PR - could you take care of removing this part from the code?

widgets/plugins/src/PluginsTable.tsx
```diff
@@ -1,8 +1,12 @@
 // @ts-nocheck File not migrated fully to TS
```
qooban, line 1: Optional:

Maybe worth to migrate this file fully to TS?
I don't know how hard would it be, so that's why this comment is optional.

qooban (commented):
Looks good, just few comments, mostly code-style/refactoring related.

General:

Did you check maybe how hard would it be to remove unnecessary framing (border + widget name) when we are displaying a single widget in the modal? Maybe we should add an option to override widget definition somehow? Or use styling hacks to hide it?

widgets/plugins/src/PluginsTable.tsx
kubama: Right, removed

widgets/plugins/src/MarketplaceModal.tsx
kubama: Right, aligned

widgets/plugins/src/PluginsTable.tsx
kubama: I pushed the migration further, however I was not able to finish it as I don't want to spend too much time on it.

kubama:
> Did you check maybe how hard would it be to remove unnecessary framing (border + widget name) when we are displaying a single widget in the modal? Maybe we should add an option to override widget definition somehow? Or use styling hacks to hide it?

Styling hacks should be relatively easy, introducing a dedicated option would be harder. But I'm wondering if it's something we actually want to do. We don't do it when we have a single widget on a page. We also don't do it for widgets nested in deployments view. So, to be consistent, I think we could leave it here as it currently is.

kubama:
System tests failure seems unrelated to this PR as the same failures could be observed on master on previous build. So I consider it a pass.

qooban (approved):
All good from my perspective.

Thanks for addressing my comments and moving TS migration forward 👍🏻

qooban:
> System tests failure seems unrelated to this PR as the same failures could be observed on master on previous build. So I consider it a pass.

Seems like I had the same issues when running tests for #1529 . I suspect either manager issue or one-timer.

> Styling hacks should be relatively easy, introducing a dedicated option would be harder. But I'm wondering if it's something we actually want to do. We don't do it when we have a single widget on a page. We also don't do it for widgets nested in deployments view. So, to be consistent, I think we could leave it here as it currently is.

Yeah, I understand, valid points. I'm not pushing to change anything right now, I'm OK with having it like it is. I only suspect that there might be a request for a change from the product team. We can see if we get comments on the demo about it.

kubama:
@stepek Waiting for your review

stepek (approved):

PR #1535, opened 2021-08-06
RD-2301 Fixed issue with maximized widgets in tabs

kubama:
There are fixes in two places:
- `userAppCommon.ts` - prevents saving maximized widget state in user apps, required for case when user maximizes a tab contained widget and reloads the app
- other files - unsets maximized state in tab contained widgets when switching through pages


kubama:
System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1040/

stepek (dismissed):
code LGTM

app/actions/widgets.ts
```diff
@@ -47,6 +47,12 @@ export function removeWidget(pageId, widgetId) {
     };
 }
 
+export function minimizeTabWidgets() {
+    return {
+        type: types.MINIMIZE_TAB_WIDGETS
+    };
+}
```
qooban, line 54: I think that the naming is a bit misleading, we actually unmaximize (restore from a maximized state), not minimize.

I realized that we used `MINIMIZE_WIDGETS` type and `minimizeWidgets` function name before, but it was also misleading as actual widgets minimizing (https://cloudifysource.atlassian.net/browse/CYBL-92) has not yet been implemented.

Maybe it's good time to clean it? WDYT? 
I'm not considering it as a blocker, I'd like to hear your opinion.

app/actions/userAppCommon.ts
```diff
@@ -5,7 +5,20 @@ import Internal from '../utils/Internal';
 
 export function saveUserAppData() {
     return (dispatch, getState) => {
-        const body = { appData: _.pick(getState(), 'pages'), version: Consts.APP_VERSION };
+        const appData = _(getState()).pick('pages').cloneDeep();
+
+        // Don't save maximized state for widgets in non-default tabs
```
qooban, line 10: So saving maximized state for widgets in default tabs is working fine, right?

qooban (approved):
The change looks good. 

I've mentioned in the comments that maybe small refactoring would be worth to consider. If you say it's not worth or too time consuming, maybe it'd be good idea to report small ticket (about naming)?

app/actions/userAppCommon.ts
kubama: Yes, it does

kubama:
System tests including the fix: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1050/

app/actions/widgets.ts
kubama: Here we are referencing an operation which is opposite to maximalization. Looking from this perspective minimization seems to be the right term, it's also very obvious what this operation refers to. I think it would start to be misleading only once CYBL-92 was implemented. Since it's not yet there, and it doesn't seem it could be implemented any soon I don't see an issue here.

qooban: > Here we are referencing an operation which is opposite to maximalization. Looking from this perspective minimization seems to be the right term, it's also very obvious what this operation refers to.

Not exactly. IMHO between minimize and maximize there is a restore to the original size (so called [unmaximize](https://en.wiktionary.org/wiki/unmaximize) or [unminimize](https://en.wiktionary.org/wiki/unminimize)).
For me it wasn't very obvious as I remember that we had some attempts to implement minimization in the past and I didn't remember if we did that or not 🙂 

> I think it would start to be misleading only once CYBL-92 was implemented.

Yes, it would definitely be bigger problem then, so I added a note in that issue to rename the type and actions once implementing it. 

I'm OK with keeping the current implementation, so I think we can close that discussion.

qooban (approved):
Looks good 👍🏻

PR #1536, opened 2021-08-10
RD-2922 Add Dynamic Table component.

stepek:
Added Dynamic Table component to edit array of objects 
image
Plus button adds new row with empty strings values of all of object fields. 
to use this editor in widget configuration user needs define columns of table and components for each of field in object.

### TODO: 
[ ] Add some system tests for this feature.

### DIFF
--- a/app/components/EditWidgetModal.tsx
+++ b/app/components/EditWidgetModal.tsx
@@ -110,6 +110,8 @@ EditWidgetModal.propTypes = {
             name: PropTypes.string,
             hidden: PropTypes.bool,
             // eslint-disable-next-line react/forbid-prop-types
+            columns: PropTypes.array,
+            // eslint-disable-next-line react/forbid-prop-types
             value: PropTypes.any
         })
     ).isRequired,
--- a/app/components/EditWidgetModal.tsx
+++ b/app/components/EditWidgetModal.tsx
@@ -75,6 +75,7 @@ export default function EditWidgetModal({ configDef, configuration, show, onHide
                                 name={config.id}
                                 label={config.name}
                                 value={fields[config.id]}
+                                columns={config.columns}
                                 onChange={handleInputChange}
                                 widgetlessToolbox={getToolbox(undefined, undefined, undefined)}
                             />
--- a/app/components/shared/DynamicTable.tsx
+++ b/app/components/shared/DynamicTable.tsx
@@ -9,7 +9,8 @@ interface KeyValueEditorProps
 }
 
 const DynamicTable: FunctionComponent<KeyValueEditorProps> = ({ name, value = [], onChange, columns = [] }) => {
-    const { GenericField, Input, Button, Icon, Table } = Stage.Basic;
+    const { GenericField, Input, Button, Table } = Stage.Basic;
+    const t = Stage.Utils.getT('shared.dynamicTable');
 
     const handleEditRow = (key: string, index: number): ComponentProps<typeof Input>['onChange'] => (event, data) => {
         onChange(event, {
@@ -66,19 +67,21 @@ const DynamicTable: FunctionComponent<KeyValueEditorProps> = ({ name, value = []
                                 </Table.Cell>
                             ))}
                         <Table.Cell textAlign="center">
-                            <Button icon onClick={handleRemoveRow(index)}>
-                                <Icon name="minus" />
-                            </Button>
+                            <Button
+                                basic
+                                icon="trash"
+                                aria-label={t('removeButton')}
+                                title={t('removeButton')}
+                                onClick={handleRemoveRow(index)}
+                            />
                         </Table.Cell>
                     </Table.Row>
                 ))}
             </Table.Body>
             <Table.Footer>
                 <Table.Row>
-                    <Table.HeaderCell colSpan={columns.length + 1} textAlign="center">
-                        <Button icon onClick={handleAddRow}>
-                            <Icon name="plus" />
-                        </Button>
+                    <Table.HeaderCell colSpan={columns.length + 1}>
+                        <Button icon="add" content={t('addButton')} onClick={handleAddRow} />
                     </Table.HeaderCell>
                 </Table.Row>
             </Table.Footer>
--- a/app/components/shared/DynamicTable.tsx
+++ b/app/components/shared/DynamicTable.tsx
@@ -66,7 +66,7 @@ const DynamicTable: FunctionComponent<KeyValueEditorProps> = ({ name, value = []
                                     />
                                 </Table.Cell>
                             ))}
-                        <Table.Cell textAlign="center">
+                        <Table.Cell textAlign="right" width={1}>
                             <Button
                                 basic
                                 icon="trash"
--- a/app/components/shared/DynamicTable.tsx
+++ b/app/components/shared/DynamicTable.tsx
@@ -1,5 +1,4 @@
 import React, { ComponentProps, FunctionComponent } from 'react';
-import { getToolbox } from '../../utils/Toolbox';
 
 interface KeyValueEditorProps
     extends Pick<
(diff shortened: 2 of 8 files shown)


app/components/shared/DynamicTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+                        .map(column => (
+                            <Table.HeaderCell key={column.id}>{column.label}</Table.HeaderCell>
+                        ))}
+                    <Table.HeaderCell />
+                </Table.Row>
+            </Table.Header>
+            <Table.Body>
+                {value.map((val, index) => (
+                    // eslint-disable-next-line react/no-array-index-key
+                    <Table.Row key={index}>
```
stepek, line 49: Disabled because don't have any other uniq key in value object.

kubama:
Looks fine, however there are some inconsistencies with the rest of the application. You may check filter rules form as used in filters widget and use it as a reference point.

app/components/shared/DynamicTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+    columns: Stage.Types.WidgetConfigurationDefinition[];
+}
+
+const DynamicTable: FunctionComponent<KeyValueEditorProps> = ({ name, value = [], onChange, columns = [] }) => {
+    const { GenericField, Input, Button, Icon, Table } = Stage.Basic;
+
+    const handleEditRow = (key: string, index: number): ComponentProps<typeof Input>['onChange'] => (event, data) => {
+        onChange(event, {
+            name,
+            value: [...value.slice(0, index), { ...value[index], [key]: data.value }, ...value.slice(index + 1)]
```
kubama, line 18: `slice` already makes a copy of the array fragment, so do we need `...` here?

app/components/shared/DynamicTable.tsx
```diff
@@ -0,0 +1,91 @@
...
+    };
+    const handleAddRow: ComponentProps<typeof Button>['onClick'] = event => {
+        onChange(event, {
+            name,
+            value: [...value, Object.fromEntries(columns.map(column => [column.id, '']))]
+        });
+    };
+
+    return (
+        <Table>
```
kubama, line 35: I'm not sure if we actually need `Table` component to implement it. Again, you may check `RulesForm` to check if the approach used there would be more appropriate.

qooban:
Please check CI (`npm run build` is failing).

I'll review tomorrow.

app/components/shared/DynamicTable.tsx
stepek: can be changed to concat of multiple arrays, or add new var for that. What do you prefer?

stepek:
@qooban @kubama build failing because `common/common.js` is big over 244 Ki,B how about increasing it ?

app/components/shared/DynamicTable.tsx
kubama: Ah never mind, I misunderstood this line 🙁.

kubama:
> @qooban @kubama build failing because `common/common.js` is big over 244 Ki,B how about increasing it ?

It's just a warning, not the reason behind build failure. There is a compilation error in `DynamicTable.tsx` though.

app/components/shared/DynamicTable.tsx
stepek: Using table we can easy have table header as field name, so i prefer Table over form group

kubama:
> fixed and screen updated in PR description.

Looks much better 🙂. I'm just wondering about margins around the trash icon - I think they are not necessary and it would be better to use this space for wider inputs.

widgets/blueprints/src/BlueprintsList.tsx
```diff
@@ -145,7 +145,7 @@ export default class BlueprintList extends React.Component<BlueprintListProps, B
                 <ErrorMessage error={error} onDismiss={() => this.setState({ error: null })} autoHide />
 
                 <div className="uploadBlueprintButton">
-                    <BlueprintUploadActionsMenu direction="left" toolbox={toolbox} />
+                    <BlueprintUploadActionsMenu direction="left" toolbox={toolbox} widget={widget} />
```
qooban, line 148: Do we need whole `widget` here? I think we just need configuration for `marketplaceTabs` or am I missing something?

widgets/blueprints/src/widget.tsx
```diff
@@ -47,6 +47,40 @@ Stage.defineWidget<unknown, unknown, BlueprintsWidgetConfiguration>({
...
+            columns: [
+                {
+                    id: 'name',
+                    label: 'Tab name',
+                    default: '',
+                    type: Stage.Basic.GenericField.STRING_TYPE
+                },
+                {
+                    id: 'url',
+                    label: 'URL',
```
qooban, line 63: Optional:

I'm wondering if we should add `description` field here or provide more descriptive `label`.
I don't have ready suggestion, so I'm putting that comment as optional to consider.

app/components/shared/DynamicTable.tsx
qooban: Or we can use `Form.Group` solution the following - a bit hacky - way:
```
        <Form.Group widths='equal'>
          <Form.Field label='Tab name' />
          <Form.Field label='URL' />
          <Form.Field />
        </Form.Group>
        <Form.Group widths='equal'>
          <Form.Field control={Input} />
          <Form.Field control={Input} />
          <Form.Button icon='trash' />
        </Form.Group>
        <Form.Group widths='equal'>
          <Form.Field control={Input} />
          <Form.Field control={Input} />
          <Form.Button icon='trash' />
        </Form.Group>
```
which gives:
image

qooban (commented):
Nice way of reusing `GenericField` 👍🏻 
Added some comments. Did not test the UI.

app/components/EditWidgetModal.tsx
```diff
@@ -75,6 +75,7 @@ export default function EditWidgetModal({ configDef, configuration, show, onHide
                                 name={config.id}
                                 label={config.name}
                                 value={fields[config.id]}
+                                columns={config.columns}
```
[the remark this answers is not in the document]
stepek: Yes, but maybe when `EditWidgetModal.tsx` will be migrated to ts. At the moment this can introduce some stupid errors. I prefer to stay with explicit declaration of this prop

app/translations/en.json
```diff
@@ -284,7 +284,11 @@
         "idPopup": {
             "copyButton": "Copy ID",
             "label": "ID"
-        }
+        },
+      "dynamicTable": {
+        "addButton": "Add",
+        "removeButton": "Remove"
+      }
```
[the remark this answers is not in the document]
stepek: yes it's my bad

widgets/blueprints/src/BlueprintsList.tsx
stepek: no i'll put only `marketplaceTabs` config there

stepek:
> > fixed and screen updated in PR description.
> 
> Looks much better 🙂. I'm just wondering about margins around the trash icon - I think they are not necessary and it would be better to use this space for wider inputs.

image

kubama:
Looks cool. Great addition 👍.

kubama (approved):

widgets/blueprints/src/BlueprintsList.tsx
```diff
@@ -145,7 +145,11 @@ export default class BlueprintList extends React.Component<BlueprintListProps, B
                 <ErrorMessage error={error} onDismiss={() => this.setState({ error: null })} autoHide />
 
                 <div className="uploadBlueprintButton">
-                    <BlueprintUploadActionsMenu direction="left" toolbox={toolbox} widget={widget} />
+                    <BlueprintUploadActionsMenu
+                        direction="left"
+                        toolbox={toolbox}
+                        config={widget.configuration.marketplaceTabs}
```
qooban, line 151: Optional:

Maybe better would be to name that prop `marketplaceTabs` instead of generic `config`?

qooban (approved):
Looks cool! 👍🏻 

In the future we may introduce width setting for columns.

widgets/blueprints/src/BlueprintsList.tsx
stepek: sure, will change

PR #1540, opened 2021-08-11
RD-2826 Customize "Number of deployments" widget to allow providing icon, label and filter ID

qooban:
This PR adds more options to the “Number of deployments” widget: 
1. Icon and Image URL - to allow changing the image on the left side
2. Label - to allow changing the text below the number
3. Filter ID - to make it possible to filter deployments
4. Page - on click open a page with already filtered deployments
Full TS and translation support was added to the "Number of deployments" widget.

CI will not pass until:
1. we merge https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1911
2. solve the problem with TS checks (`npm run build` fails) by merging #1544 

System tests will be added as part of RD-2808.

### Video

https://user-images.githubusercontent.com/5202105/129032121-01e5060d-14fa-4449-a185-5550d67204a7.mp4

### System tests

To be started once we resolve CI issues.


widgets/deploymentWizardButtons/src/steps/ConfirmationStep.tsx
```diff
@@ -47,7 +47,7 @@ function ConfirmationStepActions({
 }) {
     async function isUsed(deploymentId) {
         const deploymentActions = new Stage.Common.DeploymentActions(toolbox);
-        const deploymentPromise = () => deploymentActions.doGetDeployments({ _search: deploymentId });
+        const deploymentPromise = () => deploymentActions.doGetDeployments({ _include: 'id', _search: deploymentId });
```
qooban, line 50: I had to adapt it to the changes in `deploymentActions.doGetDeployments` function.

widgets/common/src/DeploymentActions.ts
```diff
@@ -17,7 +17,7 @@ export default class DeploymentActions {
     }
 
     doGetDeployments(params: any) {
-        return this.toolbox.getManager().doGet('/deployments?_include=id', { params });
+        return this.toolbox.getManager().doGet('/deployments', { params });
```
qooban, line 20: It was done to make `doGetDeployments` more flexible.

widgets/common/src/deploymentsView/header/filter/FilterModal.tsx
```diff
@@ -177,6 +180,13 @@ const FilterModal: FunctionComponent<FilterModalProps> = ({
         setInitialFilterRules(newFilterRules);
     }, [JSON.stringify(filterRulesResult.data)]);
 
+    // Update filterRules and filterId when filterId is set in the URL
+    useEffect(() => {
+        if (filterIdFromUrl && !open && filterIdFromUrl === filterId.value) {
+            onSubmit(filterRules.value, filterId.value);
+        }
+    }, [filterRules.value]);
```
qooban, line 188: The main reason for that is to make `FilterModal` users be aware that some filter was selected.
In other words: to make Filter button in Deployments View widget become blue and with filter ID in label.

app/translations/en.json
```diff
@@ -670,6 +670,36 @@
...
+                    "description": {
+                        "text": "Name of the icon displayed on the left side of the deployments count. Available icons list can be found at:",
+                        "linkUrl": "https://react.semantic-ui.com/elements/icon",
+                        "linkTitle": "Icon - Semantic UI React"
+                  }
+                },
+                "imageSrc": {
+                    "name": "Image URL",
+                    "description": "URL of the image displayed on the left side of the deployments count"
+                },
```
kubama, line 692: We can't use `icon` and `imageSrc` at the same time, right? How about noting it in the description?

qooban: Actually we can use both in the current implementation 🙂 I was thinking about not allowing for `icon` and `imageSrc` to be defined at once. I was also thinking of having some kind of priority when both are defined (e.g. show only `imageSrc` when both are defined). I was also considering to have a special input component with text and toggle (to choose between `icon` and `imageSrc`), but ended up with the easiest solution letting user to have full freedom.

I'm open to make a change here if you think it would be better to block showing both - icon and image - at the same time. 
What solution would you suggest?

kubama: I'm OK with allowing both. I just thought it's not possible at the time of writing the comment. But since it is, no issue here for me 🙂.

stepek (approved):
Code lgtm

kubama (approved):

qooban:
System tests passed. Thanks for the approvals. I'll merge this PR once CI pass.

PR #1543, opened 2021-08-12
Add marketplace modal to blueprints widget

stepek:
Added dropdown menu with buttons to open upload from marketplace, file and composer. also added new modal with blueprint marketplace.

Some screens 
image
image
image

### System Tests
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1064/


qooban:
I'll review tomorrow. As we already reviewed the code (in smaller chunks), I'll probably focus more on doing some UI tests.

kubama (dismissed):

widgets/blueprints/src/widget.tsx
```diff
@@ -47,6 +47,40 @@ Stage.defineWidget<unknown, unknown, BlueprintsWidgetConfiguration>({
...
+                },
+                {
+                    name: 'Kubernetes Blueprint Examples',
+                    url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/k8s-examples.json'
+                },
+                {
+                    name: 'Orchestrator Blueprint Examples',
+                    url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/orc-examples.json'
+                }
+            ],
```
qooban, line 81: I think we need translation for this part (as right now we have only one configuration field with translation - `showEditCopyInComposerButton` and the rest without). URLs should also be added to the translation file.

Overall I think we should add translation to the whole file at once.

widgets/common/src/BlueprintMarketplaceModal.tsx
```diff
@@ -0,0 +1,74 @@
...
+const getPageLayout = (tabs: Tab[]) => ({
+    layout: [
+        {
+            type: 'tabs',
+            content: tabs.map((tab, index) => ({
+                name: tab.name,
+                widgets: [
+                    {
+                        id: `blueprint-catalog-${index}`,
+                        name: 'Blueprints Catalog',
```
qooban, line 23: `name` can be added to translation file.

widgets/common/src/BlueprintUploadActionsMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+            // @ts-expect-error UploadBlueprintModal is not converted to TS yet
+            const { UploadBlueprintModal, BlueprintMarketplaceModal } = Stage.Common;
+
+            switch (name) {
+                case 'uploadFromMarketplace':
+                    return <BlueprintMarketplaceModal open onHide={hideModal} tabs={marketplaceTabs} />;
+                case 'uploadFromPackage':
+                    return <UploadBlueprintModal open onHide={hideModal} toolbox={toolbox} />;
+                case 'generateInComposer':
+                    window.open(`/composer/`, '_blank');
```
qooban, line 50: Hmm... Didn't we plan to have common function for that? (as the same function is also used in Composer Link widget)

test/cypress/integration/widgets/blueprints_spec.ts
```diff
@@ -3,19 +3,36 @@ import type { BlueprintsWidgetConfiguration } from '../../../../widgets/blueprin
...
+        },
+        {
+            name: 'Kubernetes Blueprint Examples',
+            url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/k8s-examples.json'
+        },
+        {
+            name: 'Orchestrator Blueprint Examples',
+            url: 'https://repository.cloudifysource.org/cloudify/blueprints/5.1/orc-examples.json'
+        }
+    ];
```
qooban, line 19: I'm wondering, as those are the default values, do we have to set it?
If that is necessary, then if we move tab names and URLs to translation file, then we can maybe reuse it here?

qooban (commented):
Added some comments.

Tested the UI. Looks good! 👍🏻 I've got just one comment about the case when the user does not provide URL:
Screenshot from 2021-08-13 07-49-09
In such case we still present the tab:
Screenshot from 2021-08-13 08-12-09
I would suggest to show tabs only if both values (tab name and URL) are specified.

widgets/blueprints/src/widget.tsx
stepek: sure, will add

test/cypress/integration/widgets/blueprints_spec.ts
stepek: Yes, we should set this there. 
This is only test data and there can by anything, imho this can be heer without translation to easier future debugging

widgets/blueprints/src/widget.tsx
```diff
@@ -20,16 +20,16 @@ Stage.defineWidget<unknown, unknown, BlueprintsWidgetConfiguration>({
         Stage.GenericConfig.PAGE_SIZE_CONFIG(5),
         {
             id: 'clickToDrillDown',
-            name: 'Enable click to drill down',
+            name: Stage.i18n.t('widgets.blueprints.configuration.clickToDrillDown'),
```
qooban, line 23: Optional:

You can use custom function to shorten the lines:
```
const t = StageUtils.getT('widgets.blueprints.configuration');
```

qooban (approved):
Looks good 👍🏻 

Added just one, non-blocking optional comment.

widgets/blueprints/src/widget.tsx
stepek: Sure, but i do this to look like other widget where we have translation in configs.

PR #1545, opened 2021-08-13
RD-2481 Added "Upload all plugins" button

kubama:
The button was added as a standard data table action button:
image

Some implementation notes:
- plugin upload state was moved to redux - this way it can be tracked application wide, the user can switch between pages and the state is not lost
- when plugin upload is completed an event is emitted - as above, when user switches between pages and comes back to the catalog widget he can still see notifications, even if the upload was started in a different widget instance
- plugins list was migrated to functional component, as it makes working with events and redux easier


qooban:
I'll review this PR on Monday.

stepek:
will review on Monday

kubama:
System tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1075/

kubama:
There is one system test failure, but the same test fails on `master`, so I consider it a pass.

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -1,7 +1,9 @@
-import type { ComponentProps, ComponentType } from 'react';
+import type { ComponentProps, ComponentType, FunctionComponent } from 'react';
+import { useState } from 'react';
 import styled from 'styled-components';
-import { compact, isEmpty, isEqual, find, map, without } from 'lodash';
+import { compact, find, isEmpty, map, without } from 'lodash';
 
+import { useDispatch, useSelector } from 'react-redux';
```
qooban, line 6: Hmm... Shouldn't we use `ReactRedux.useDispatch` and `ReactRedux.useSelector` in `widgets`?

widgets/pluginsCatalog/src/widget.tsx
```diff
@@ -56,6 +57,10 @@ Stage.defineWidget<unknown, PluginsCatalogResponse | Error, PluginsCatalogWidget
             formattedData = _.sortBy(data, item => item.pluginDescription.title);
         }
 
-        return <PluginsCatalogList widget={widget} items={formattedData} toolbox={toolbox} />;
+        return (
+            <Provider store={toolbox.getStore()}>
+                <PluginsCatalogList widget={widget} items={formattedData} toolbox={toolbox} />
+            </Provider>
```
qooban, line 63: Is that necessary? Recently I used redux store in "Number of Deployments" widget and did not encounter that need. We also use redux store in "Deployments View" widget.

app/utils/Toolbox.ts
```diff
@@ -73,6 +73,10 @@ class Toolbox implements Stage.Types.Toolbox {
         this.store.dispatch((selectPageByName(pageName, context) as unknown) as AnyAction);
     };
 
+    getStore() {
+        return this.store;
+    }
+
```
qooban, line 79: I'm not sure we need it. See my other two comments about using Redux store in widgets.

qooban (commented):
I did not review `PluginsCatalogList.tsx` and `plugins_catalog_spec.ts` yet. 

I've added 3 comments. All of them are related to the same thing - how to handle redux store in widgets. I'll continue the review once we finalize that topic.

kubama:
> I did not review `PluginsCatalogList.tsx` and `plugins_catalog_spec.ts` yet.
> 
> I've added 3 comments. All of them are related to the same thing - how to handle redux store in widgets. I'll continue the review once we finalize that topic.

I addressed the comments - implementation is now aligned to `ReactRedux`

stepek (dismissed):

test/cypress/integration/widgets/plugins_catalog_spec.ts
```diff
@@ -106,4 +106,26 @@ describe('Plugins Catalog widget', () => {
...
+    it('should upload all plugins', () => {
+        // eslint-disable-next-line security/detect-non-literal-regexp
+        cy.intercept('POST', new RegExp(`console/plugins/upload.*title=AWS`)).as('awsPluginUpload');
+        cy.contains('Upload all plugins').click().should('be.disabled');
+        cy.get('.pluginsCatalogWidget tr button').should('be.disabled');
+        // eslint-disable-next-line security/detect-non-literal-regexp
+        cy.intercept('POST', new RegExp(`console/plugins/upload.*title=Ansible`)).as('ansiblePluginUpload');
+        cy.wait('@awsPluginUpload', { responseTimeout: uploadPluginTimeout });
+        cy.contains('AWS successfully uploaded');
+        cy.contains('.pluginsCatalogWidget tr', 'AWS').find('button.loading').should('not.exist');
```
qooban, line 119: Optional:

Maybe for easier reading we can have:
1. Both intercepts at the top
2. Some spacing between setup, test and assertion blocks
WDYT?

```suggestion
        // eslint-disable-next-line security/detect-non-literal-regexp
        cy.intercept('POST', new RegExp(`console/plugins/upload.*title=AWS`)).as('awsPluginUpload');
        // eslint-disable-next-line security/detect-non-literal-regexp
        cy.intercept('POST', new RegExp(`console/plugins/upload.*title=Ansible`)).as('ansiblePluginUpload');
        
        cy.contains('Upload all plugins').click().should('be.disabled');
        
        cy.get('.pluginsCatalogWidget tr button').should('be.disabled');
        cy.wait('@awsPluginUpload', { responseTimeout: uploadPluginTimeout });
        cy.contains('AWS successfully uploaded');
        cy.contains('.pluginsCatalogWidget tr', 'AWS').find('button.loading').should('not.exist');
```

qooban (approved):
Thanks for addressing my comments.

I think importing Redux was the main reason for widget bundle size increase as right now it dropped to:
```
[2021-08-16T07:43:28.988Z]   Widgets
[2021-08-16T07:43:28.988Z]   Size limit:   3.13 MB
[2021-08-16T07:43:28.988Z]   Size:         3.1 MB
```

test/cypress/integration/widgets/plugins_catalog_spec.ts
kubama: Actually my intension here was to test that the second plugin upload starts after the first one, not in parallel to the first one. Moving the second intercept to the top would make the test to pass regardless of requests timing. I consider it important as launching all plugin uploads at once was causing the backend to freeze for a while making the application unresponsive.

I added empty lines as suggested though. Also added a comment.

qooban: OK, got it, sounds reasonable to have it in that order then. Thanks for adding the comment about it.

qooban (approved):

PR #1547, opened 2021-08-16
RD-2982 Improve CI pipeline by running Audit stage for all branches

qooban:
## Description
This is CI pipeline change. It makes **Audit** stage to be executed for all branches and not fail the whole pipeline in case of failure. It should help us:
1. to detect security audit issues earlier (in PR)
2. to verify security audit issues easier (manual check won't be necessary)

## Screenshots / Videos
Thanks to @yuvlar the **Audit** stage is no longer required in GitHub for PR to be "mergable":
image

## Checklist

## Tests
As we are now having audit issue in Stage I will verify it by checking the result of the CI pipeline for this PR.

## Documentation
I'm not planning to document that change.


kubama:
Seeing how it looks I'm a little bit in doubt.

> 2. to verify security audit issues easier (manual check won't be necessary)

Actually I think this will multiply the number of manual checks that need to be done in case there is a prolonging audit issue. In such case each failure in each PR should be manually investigated to check if there is only the already known issue, or if there possibly is a new additional issue. Seeing just a failure doesn't tell us what's the actual reason.

qooban:
> Actually I think this will multiply the number of manual checks that need to be done in case there is a prolonging audit issue. In such case each failure in each PR should be manually investigated to check if there is only the already known issue, or if there possibly is a new additional issue. Seeing just a failure doesn't tell us what's the actual reason.

By "manual check won't be necessary" I had in mind a situation when you are pushing security fix. With that change it won't be 
necessary to run `npm run audit` in local environment to verify that the fix you are trying to push works fine. You are right that it'd might be necessary to click "Details" button and see if that's the audit issue we are currently experiencing. But I'm wondering :
1. will that be a frequent case that we'll have new issue when there's already one reported? Did we have such cases before?
2. will that be a big problem for us to have that verification step in case of prolonging issue?

I think that it is a bit misleading that we have green audit stage (the stage is marked as successful) right now on non-master branches when the actual check is not executed at all. The benefits I see:
1. Have clear information in all cases - if there is security issue or not - that allows to test audit security fixes easier
2. Have early notification about new security issue in case it was all green before

If you both think it'll be problem for us, we don't need to merge this.
If there would be an option to fully hide some stages for specific branches, then probably it might be an alternative option.

kubama:
> By "manual check won't be necessary" I had in mind a situation when you are pushing security fix.

Looks like I misunderstood then. I thought you were referring to manual checks during daily nightly check and the intention is to replace those checks by checks on dev branches. If it's not the case, then all good, I agree with your reasoning.

kubama (approved):

stepek (approved):

PR #1553, opened 2021-08-17
RD-2477 Add filter by label rules to blueprints widget.

stepek:
## Checklist

- [ ] My code follows the UI Code Style.
- [ ] I verified if that change requires a documentation update.


widgets/blueprints/src/ruleFilterFromString.ts
```diff
@@ -0,0 +1,62 @@
...
+        result = tempValue.replace(/[(^[)(\]$)]/g, '').split(',');
+    } else {
+        result = [tempValue];
+    }
+
+    return result.map(item => item.replace('\x00', ',').replace('\\:', ':'));
+};
+
+const ruleFilterFromString = (rule: string): Stage.Common.Filters.Rule => {
+    const equalMatching = rule.match(/^([\w\-.]+)(=)([^\n\t"]+)$/) || rule.match(/^([\w\-.]+)(!=)([^\n\t"]+)$/);
```
kubama, line 38: Looks like the two expressions can be easily combined into one, same in L48.

test/jest/widgets/blueprints/ruleFilterFromString.test.ts
```diff
@@ -0,0 +1,43 @@
...
+        ],
+        [
+            'name is null',
+            { key: 'name', operator: LabelsOnlyRuleOperator.IsNull, type: FilterRuleType.Label, values: [] }
+        ],
+        [
+            'name is not null',
+            { key: 'name', operator: LabelsOnlyRuleOperator.IsNotNull, type: FilterRuleType.Label, values: [] }
+        ]
+    ];
```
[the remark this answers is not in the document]
kubama: Since string values are intended to by supplied manually I would add some test cases with spaces in string values, for example `'name = Felucia'`

widgets/blueprints/src/ruleFilterFromString.ts
```diff
@@ -0,0 +1,62 @@
...
+    FilterRuleOperator,
+    FilterRuleType,
+    LabelsOnlyRuleOperator
+} from '../../common/src/filters/types';
+
+const OPERATOR_MAPPING: Record<string, any> = {
+    '=': CommonRuleOperator.AnyOf,
+    '!=': CommonRuleOperator.NotAnyOf,
+    'is null': LabelsOnlyRuleOperator.IsNull,
+    'is not null': LabelsOnlyRuleOperator.IsNotNull
```
kubama, line 12: Looking at our code I can see that we support 5 rule operators in total, here we have only 4

widgets/blueprints/src/widget.tsx
```diff
@@ -48,6 +49,12 @@ Stage.defineWidget<unknown, unknown, BlueprintsWidgetConfiguration>({
             name: Stage.i18n.t('widgets.blueprints.configuration.showEditCopyInComposerButton'),
             default: false
         },
+        {
+            id: 'filterRule',
+            name: Stage.i18n.t('widgets.blueprints.configuration.labelFilterRule'),
+            default: '',
+            type: Stage.Basic.GenericField.STRING_TYPE
```
kubama, line 56: Was it considered to use existing components for defining the rule instead of string representation? Should be very easy to accomplish, all the needed components are available.

widgets/blueprints/src/ruleFilterFromString.ts
```diff
@@ -0,0 +1,62 @@
+import {
+    CommonRuleOperator,
+    FilterRuleOperator,
+    FilterRuleType,
+    LabelsOnlyRuleOperator
+} from '../../common/src/filters/types';
```
qooban, line 6: If that's type import, maybe better to use `import type` for clarity.
But personally maybe better would be to move that into `widgets/common/filters` as it seems like strongly connected with filters functionality?

Side note:
With types this is not the case, but in general we shouldn't import from outside of the widget code (mostly because single widget bundle size may be increased unnecessarily). If needed, we should expose data through `Stage` global.

widgets/blueprints/src/widget.tsx
qooban: It was described in the ticket and suggested by Ofer to have free text field.

But looking at the it now (code complexity and maintainability aspects), I think we should reconsider using `RulesForm`. We would probably need some adjustment as:
1. Form works with deployments' labels and blueprints' labels are fetched using different endpoint (https://docs.cloudify.co/api/staging/dev/#get-all-blueprints-labels-keys)
2. The ticket refers only to labels, not attributes, but maybe we can support both - labels and attributes?

widgets/blueprints/src/widget.tsx
```diff
@@ -86,11 +93,17 @@ Stage.defineWidget<unknown, unknown, BlueprintsWidgetConfiguration>({
...
+
         return toolbox
             .getManager()
-            .doGet(
-                '/blueprints?_include=id,updated_at,created_at,description,created_by,visibility,main_file_name,state,error',
-                { params }
+            .doPost(
+                '/searches/blueprints?_include=id,updated_at,created_at,description,created_by,visibility,main_file_name,state,error',
+                {
+                    params,
```
qooban, line 102: We can make it cleaner at once:
```suggestion
                '/searches/blueprints',
                {
                    _include: 'id,updated_at,created_at,description,created_by,visibility,main_file_name,state,error',
                    ...params,
```

widgets/blueprints/src/ruleFilterFromString.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const OPERATOR_MAPPING: Record<string, any> = {
```
qooban, line 8: I think we should have it named using lower camel case:
```suggestion
const operatorMapping: Record<string, any> = {
```

Ref.: https://github.com/airbnb/javascript#naming--uppercase

qooban (commented):
Added some comments.

Most important: Maybe worth to reconsider using `RulesForm` instead of string parsing as @kubama suggests.

app/translations/en.json
```diff
@@ -949,7 +950,6 @@
                 "disabled": "Latest version is already uploaded",
                 "uploading": "Plugin upload in progress"
             },
-            "uploadAllButton": "Upload all plugins",
```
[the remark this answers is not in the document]
stepek: After rebase i think was removed.

test/jest/widgets/blueprints/ruleFilterFromString.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    it('should match test cases', () => {
+        // eslint-disable-next-line no-restricted-syntax
+        for (const testCase of testCases) {
```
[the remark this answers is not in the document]
stepek: can be

test/jest/widgets/blueprints/ruleFilterFromString.test.ts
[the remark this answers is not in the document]
stepek: regarding changing data structure to Object, of Map, this is only test data will change to Array of Objects

Regarding spaces in strings, will add

widgets/blueprints/src/ruleFilterFromString.ts
stepek: I see only 4 in the code `=` `!=` `is null` and `is not null`

widgets/blueprints/src/widget.tsx
stepek: I'm ok with text field as described in ticket.

widgets/blueprints/src/ruleFilterFromString.ts
stepek: > If that's type import, maybe better to use import type for clarity.

No it's not only types

> But personally maybe better would be to move that whole functionality (parsing strings) into widgets/common/filters as it seems like strongly connected with filters functionality?

will move

widgets/blueprints/src/ruleFilterFromString.ts
kubama: I was referring to:
```
enum CommonRuleOperator {
    AnyOf = 'any_of',
    NotAnyOf = 'not_any_of'
}

enum LabelsOnlyRuleOperator {
    IsNot = 'is_not',
    IsNull = 'is_null',
    IsNotNull = 'is_not_null'
}
```

stepek:
This PR is blocked by https://cloudifysource.atlassian.net/browse/RD-3001

widgets/blueprints/src/ruleFilterFromString.ts
stepek: I implemented this based only on CLI implementation of label filter, ware we have only 4 of them only.

kubama: This is strange... 5 supported operators in UI, API and CLI are mentioned in the official docs:
https://docs.cloudify.co/latest/cli/orch_cli/filter-rules/
So if only 4 are supported in CLI then there is either an implementation gap or the doc should be updated to reflect the actual implementation. Could you please check which case it is and report appropriate issue?
Also, if it's a gap in the CLI then I think we shouldn't replicate it and still implement full support.

widgets/blueprints/src/widget.tsx
kubama: I think it's a matter of how the user is expected to provide the rule. If he is expected to copy the string representation from somewhere and paste it then text input is the way to go. In any other case component based solution is by all means better:
1. Higher consistency - we already have a way of specifying label rules in the UI, why should we introduce a second method for doing the same?
2. Low implementation cost - the components are already there, just need to reuse them in different context.
3. Lower code complexity and maintenance cost.

stepek:
@qooban let's summarize, i should reimplement it using existing component for rule filtering?

qooban:
> @qooban let's summarize, i should reimplement it using existing component for rule filtering?

The main goal is to manage to finalize the whole epic (RD-2473) in this sprint.
So could you provide rough estimation for re-implementation? Basing on that we can discuss on our weekly today and decide which path to choose. We can also ask @alexmolev for his opinion on that.

qooban:
As we are finishing the sprint, we need to figure out what to do with this PR. Either finalize it in this shape or have form-based solution.

@stepek Did you get a chance to look into alternative solution (form-based filter instead of text filter)? Did you verified expectations with Alex?

stepek:
@qooban yes, and i rewriting this to use rules filter component

stepek:
Cleaned up PR and force pushed component wat approach.

widgets/blueprints/src/types.ts
```diff
@@ -1,3 +1,5 @@
+import type { FilterRule } from '../../../app/components/shared/RulesFilterForm/types';
```
stepek, line 1: Will be exposed in Shared global object.

kubama:
I only did a brief review, but I can see there is much code copied from `widgets/common`. Is code reuse something you are yet planning to cover?

stepek:
@kubama i plan to remove code from widget common and use there this component.

qooban:
I won't be able to review today. New shape of this PR seems a bit surprising to me, I thought that small refactoring in currently existing `RulesForm` component would be sufficient to handle blueprint labels as they both (deployment labels and blueprint labels) seem to be very similar. I think that's something to discuss in weekly meeting, I'll add a point.

test/cypress/fixtures/blueprints/simple/blueprint.yaml
```diff
@@ -0,0 +1,25 @@
+tosca_definitions_version: cloudify_dsl_1_3
```
kubama, line 1: I think this file is not used anywhere.
There is also `cypress/downloads/26360-Tekst artykułu-54732-1-10-20210115.pdf` which seems to be committed by accident.

stepek: will remove this from pr

qooban (commented):
Oh, it's quite small right now 🙂 That's great 👍🏻 
1. What kind of changes are you planning to implement? Will they be part of this PR or separate one?
2. There's a problem in CI, can you take a look?

I'll review tomorrow.

widgets/common/src/filters/RuleRow.tsx
```diff
@@ -13,14 +13,23 @@ interface RuleRowProps {
     removable: boolean;
     error: boolean;
     rule: FilterRule;
-    toolbox: Stage.Types.Toolbox;
+    collectionName: string;
```
qooban, line 16: 1. I would suggest to have a dedicated type for it, probably an union like `"deployments" | "blueprints"` will allow us to have more strict type checking. What do you think?
2. Maybe instead of generic `collectionName` it'd better to name it `resourceType`? (as blueprints and deployments are Cloudify resources)
3. Just a question: Did you consider having a solution based on React Context API? I suspected that it might be suitable here as it may reduce the number of components which are not using `collectionName`, but just passing it down to another component.

kubama (commented):
I can see there are number of places where the type for `toolbox` is now defined as `Stage.Types.Toolbox | Stage.Types.WidgetlessToolbox`. But `Toolbox` extends from `WidgetlessToolbox`. If so, isn't it sufficient to simply use `Stage.Types.WidgetlessToolbox` as type in all those places?

PR #1559, opened 2021-08-19
RD-1803 Added support for `hidden` inputs

kubama:
## Description

Inputs marked as `hidden` are omitted in the deployment dialog. They are not passed to the manager when creating deployment as well.
NOTE: During testing I discovered an inconsistency in the way manager handles the `hidden` flag - it may lead a valid blueprint to be unable to be deployed using the UI. There is RD-3012 created to address this issue.

## Screenshots / Videos

N/A

## Checklist

## Tests

 

https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1111

## Documentation
N/A


test/cypress/integration/widgets/blueprints_spec.ts
```diff
@@ -84,7 +84,8 @@ describe('Blueprints widget', () => {
 
             const serverIp = '127.0.0.1';
             cy.contains('h4', 'Deployment inputs')
-                .next()
+                .nextUntil('h4')
+                .should('have.length', 1)
```
qooban, line 88: It's hard to tell (without knowing the context) why we do such assertion here.
I would add either a comment or wrap that assertion in the function (which name will properly describe the purpose).

test/cypress/integration/widgets/blueprints_spec.ts
```diff
@@ -84,7 +84,8 @@ describe('Blueprints widget', () => {
...
                 .within(() => {
```
qooban, line 89: I'm not sure if `blueprints_spec.ts` is the right place to add verification for RD-2803 and RD-2801.
There's 'handles inputs of type' section in `deployment_button_spec.ts`. I think it would be more accurate to put it there.

(or even extract deployment inputs displaying - as it is quite broad topic - into separate spec)

qooban (commented):
Added some comments.

test/cypress/integration/widgets/blueprints_spec.ts
kubama: Done

test/cypress/integration/widgets/blueprints_spec.ts
kubama: Done

qooban (approved):
Looks better now 👍🏻 Thanks for addressing the comments.

PR #1555, opened 2021-08-18
RD-2934 Add "Service Button" widget

qooban:
## Description
This PR builds on top of #1554 . New widget created according to details from RD-2934.

Added new components:
1. `SemanticColorDropdown` - Semantic-UI color selector (see 1st video for UI details)
2. `SemanticIconDropdown` - Semantic-UI icon selector (see 1st video for UI details)

Updated "Number of Deployments" widget with new component.

### Configuration

https://user-images.githubusercontent.com/5202105/130065752-9b4d1825-a33f-41b6-95c7-394f6935d55a.mp4

### Behavior

https://user-images.githubusercontent.com/5202105/130065760-409718d2-9aa3-4a97-95c0-fd20b0ee0807.mp4

## Checklist
- [x] My code follows the UI Code Style.
- [x] I verified that all tests and checks have passed.
- [x] I verified if that change requires a documentation update.
- [x] I added proper labels to this PR.

## Tests
System tests for "Service button" widget will be added in the next PR.
System tests runs:
1. [Stage-UI-System-Test/1097](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1097/pipeline) (2021-08-18 11:40:49 CEST) - 1 test failed (Blueprints widget spec, fixed by https://github.com/cloudify-cosmo/cloudify-stage/pull/1555/commits/85c9e2ac876bbb05ed41822b69bc856b642d0643)
2. [Stage-UI-System-Test/1110](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1110/) (2021-08-19 14:36:31 CEST) - 20 tests failed (starting from Maintenance Mode Button widget, seems unrelated, need to re-trigger)
3. [Stage-UI-System-Test/1114](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1114/) (2021-08-20 07:53:09 CEST) - PASSED

## Documentation
To be added by: https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/1922
I'll update `README.md` file once it is merged. 
CI will not pass until we merge it.

PR #1560, opened 2021-08-20
RD-2934 Add system tests for "Service Button" widget

qooban:
## Description
Added system tests for new "Service Button" widget. This PR builds on top of #1555 .

Additional changes:
* Created custom Cypress commands for changing widget configuration fields (boolean, string and searchable dropdown) to `test/cypress/support/editMode.ts`
* Created custom Cypress command for rendering widget with default configuration (`useWidgetWithDefaultConfiguration`) - it would be wise probably to reuse it in other widgets as well, it seems very common case (I can do that within this PR or create another one)
* Updated "Number of Deployments" widget spec to use new Cypress commands

## Screenshots / Videos

https://user-images.githubusercontent.com/5202105/130203745-8db81510-30b4-4ec7-8c15-1c309947c7b6.mp4

## Checklist

## Tests
[Stage-UI-System-Test/1115](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1115/pipeline) - PASSED

## Documentation
N/A


test/cypress/support/commands.ts
```diff
@@ -206,7 +205,7 @@ const commands = {
             );
             if (disableGettingStarted) mockGettingStarted(false);
         });
-        cy.visit('/console');
+        return cy.visit('/console');
```
qooban, line 208: All changes above this line are to make TypeScript happy and not show warnings about returning `void` and using chaining functions. I think we should update all the custom cypress commands to return `Chainable` properly.
What do you think? (if we agree, I'll create a ticket for that and maybe I'll be able to handle it soon)

kubama: Yes, I agree, I can see no reason why it shouldn't be this way.

test/cypress/support/commands.ts
```diff
@@ -319,6 +318,13 @@ const commands = {
             }
         });
     },
+    useWidgetWithDefaultConfiguration: (widgetId: string, widgetConfigurationOverrides: any = {}) =>
+        cy
+            .usePageMock(widgetId, widgetConfigurationOverrides)
+            .mockLogin()
+            // In order to load default configuration for widget
+            // open widget edit configuration modal and save without making any changes
+            .editWidgetConfiguration(widgetId, noop),
```
kubama, line 327: We have RD-1820 to address the issue differently, maybe worth to reference it in the comment here?

kubama (approved):

test/cypress/support/commands.ts
qooban: I wasn't aware of that issue. Definitely, I'm adding a reference to that issue in the comment.

test/cypress/support/commands.ts
qooban: I decided to do better research first, to see if that should be applied for all custom commands. Added a point to UI weekly meeting about it.

kubama (approved):

PR #1563, opened 2021-08-23
RD-2964 Prevented 0 width grid render

kubama:
## Description

Setting 0 for grid width was triggering grid resize event leading to shrinking nested widgets. This PR fixes the issue by skipping grid render when container width is 0. This change should be by all means transparent.

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1134/

## Documentation
N/A


qooban (approved):

kubama:
@qooban Looks like there's an issue with `Grid.test.tsx`. I already spent some time trying to fix it but without success, I'm tempted to remove it. I think the `Grid` component is at the core of virtually every page in the app and its functionality is very well covered by integration tests, so this simple test brings no additional value. WDYT?

qooban:
> @qooban Looks like there's an issue with `Grid.test.tsx`. I already spent some time trying to fix it but without success, I'm tempted to remove it. I think the `Grid` component is at the core of virtually every page in the app and its functionality is very well covered by integration tests, so this simple test brings no additional value. WDYT?

I'm OK with removing it.
What do you think about adding simple test in `edit_mode_spec.ts` checking if switching pages in Edit Mode works fine?

kubama:
> What do you think about adding simple test in `edit_mode_spec.ts` checking if switching pages in Edit Mode works fine?

Yes, good idea. Added.

qooban (approved):

PR #1143, opened 2020-11-24
RD-587 Fixed site name passing on site marker click

kubama:

PR #1568, opened 2021-08-25
RD-2932 Remove confirmation button from blueprint catalog.

stepek:
## Description
Removed confirmation dialog and set name and main_blueprint from json file. Added loader after clicking upload and moving to new page after successfully upload.

## Screenshots / Videos

https://user-images.githubusercontent.com/5058678/130773712-358254d2-5129-4fa1-9119-196c7716b754.mov

## Checklist

- [X] I verified that all tests and checks have passed. (in progress)

## Tests
[#1140](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1140/) (25-Aug-2021 10:23:24)
[#1143](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1143/) 2026-Aug-2021%2012:51 pass
[Build #1146 (27-Aug-2021 10:41:50)](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1146/) 
[Build #1156 (30-Aug-2021 11:50:34)](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1156/)
[Build #1163 (31-Aug-2021 08:05:49)](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1163/) PASS

 

## Documentation 
Will update in https://cloudifysource.atlassian.net/browse/RD-3069


widgets/blueprintCatalog/src/RepositoryList.tsx
```diff
@@ -142,31 +144,38 @@ export default class RepositoryList extends React.Component {
...
+                                successMessages: without(prevState.successMessages, message)
+                            }))
+                        }
+                    >
+                        {message}
+                    </Message>
+                ))}
+                {!isEmpty(errorMessages) && (
+                    <ErrorMessage error={errorMessages} onDismiss={() => this.setState({ errorMessages: null })} />
+                )}
```
kubama, line 162: Isn't it a a copy-paste from plugins? Maybe worth extracting a common component?

qooban:
I'll review tomorrow.

kubama (dismissed):

app/utils/StageAPI.ts
```diff
@@ -115,6 +115,8 @@ interface CommonWidgetDefinition<Params, Data, Configuration> {
     description?: string;
     /** @see https://docs.cloudify.co/developer/writing_widgets/widget-definition/#fetchurl */
     fetchUrl?: string | Record<string, string>;
+    // eslint-disable-next-line camelcase
+    mapGridParams?: (params: Stage.Types.GridParams) => { page?: number; per_page?: number };
```
qooban, line 119: Hmm... I'm not sure it's right to have widget specific return value in `StageAPI`. 
`{ page?: number; per_page?: number }` is only used in Blueprint Catalog widget IIUC. Other widgets can return other params.
Maybe it should return something like `Partial<Params>`?

qooban: Also we don't have a description for `mapGridParams` in https://docs.cloudify.co/latest/developer/writing_widgets/widget-definition/ . I think it is worth to add it there.

qooban (commented):
Big thanks for migrating files to TS 👍🏻 Looks good!

I've got just one comment, which I think is worth to address before my final approval.

app/utils/StageAPI.ts
stepek: I don't know why we have `mapGridParams` in StageAPI used only by one widget (`blueprintCatalog`), maybe some historical debt? maybe there should be som generic type?

stepek: if `mapGridParams` is needed, i think so too

qooban: Looking at the git history it seems like it is there since 2017. It is used by just one widget right now. 
But I think we should keep it. Other widgets can use it, too.

The only thing to fix - at least with my current understanding - is to change return value in the type, because other widgets can return completely different set of parameters. The requirement is that it should return an object with params.

stepek: Ok fixed

kubama (approved):

qooban (approved):

stepek:
@qooban please re approve after test fix.

qooban (commented):
@stepek 

So the only change is that you removed error message: https://github.com/cloudify-cosmo/cloudify-stage/pull/1568/commits/7edd7ac9dec98f6c3e1499e1c215ffb04c295eb1 right? Why you removed it? It was obviously not correct, because as I mentioned in https://github.com/cloudify-cosmo/cloudify-stage/pull/1568#discussion_r697165118 plugin name should be updated, but I think it is still worth to keep it. Or am I missing something?

stepek:
The error message is mockuped and checked in https://github.com/cloudify-cosmo/cloudify-stage/pull/1568/files#diff-60cb8c7cabaf4347e8b6d4e5265f035feb4e30acfbff09c727ee77b15bd6244fR27 previous installing was without mocking the  response, so right now we checking only showing mockuped error message.

qooban (approved):

PR #1233, opened 2021-03-10
Rd 1442 show getting started modal on login

gregcfy:
In this PR modal window with the quick configuration wizard on user login operation is introduced - after first login user is able to select some clouds/technologies and type secrets for them or disable modal permanently.

PR is related with the following task: https://cloudifysource.atlassian.net/browse/RD-1442

Modal disabled information is temporarily kept in local storage. 

Used graphics are located in plugins repository - in the desired version, they will be in svg and embedded

TODO in the future:
	- rethink to change Technology to some other name (Platform, Plugins ?) // technical thing
	- better icons in the first step of modal (when we select technology)
	- add some statuses near installed plugins in the last step (installation)
	- better description for: Don't show next time
	- better exports inside index.ts files // technical thing
	- better close button location (maybe only first and last step)
	- maybe Finish button should be changed into Install?
	- offline mode without getting started modal 
	- move installation process as an independent part of the Summary component // technical thing
	- load current secret values
	- make it possible to display password field
	- make possible to rerun installation on fails
	- add dynamic component loading for Getting Started Modal - it should help to avoid feature loading when it is disabled 

## Quick Demo
1. After the first login into Cloudify Stage (I have selected 2 technologies on screenshot)
image
2. According to the selected technologies it is required to type secrets for them (Selected technologies use same secrets so they are merged into one step):
image
3. In the summary step we can see what should be done (plugins are already installed):
image
4. Finish button click in the previous step starts the installation process.
image

PR #1576, opened 2021-09-01
RD- 2933 Update Dashboard page

qooban:
## Description

1. Updated all the links to examples to point to the latest version - v6.2
2. Updated Dashboard page to show new widgets (see mockup in RD-2698)
3. Adjusted some widgets to be aligned with the rest (see comments in the code)
4. Remove showing tabs in case only one Blueprint Marketplace tab is defined

## Screenshots / Videos

https://user-images.githubusercontent.com/5202105/131962744-0fadeafb-652d-4363-8f4b-55c311238273.mp4

NOTE: I used another filter for K8S, but kubernetes cluster label (`csys-k8s-filter`) is going to be added to Cloudify Manager (see RD-2998).

## Checklist

- [x] My code follows the UI Code Style.
- [x] I verified that all tests and checks have passed.
- [x] I verified if that change requires a documentation update.
- [x] I added proper labels to this PR.

## Tests
1. [Stage-UI-System-Test/1186](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1186/pipeline) (2021-09-02 12:39:32 CEST) - PASSED
2. [Stage-UI-System-Test/1190](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1190/pipeline) (2021-09-03 08:37:47 CEST) - PASSED
3. [Stage-UI-System-Test/1211](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1211/pipeline) (2021-09-07 10:40:55 CEST) - PASSED

I'm also planning to add some tests for Dashboard page once we get final approval for the look'n'feel. Possibly in a separate PR.

## Documentation
https://docs.cloudify.co/latest/working_with/console/pages/dashboard-page/ needs to be updated. There will be another PR for that once we get this PR merged.

PR #1581, opened 2021-09-03
RD-3017 Add support for switching between Catalog and Table view in Blueprints Marketplace modal.

stepek:
## Description
Add possibility to customize display style of blueprint catalog widget.

## Screenshots / Videos
image

## Checklist

## Tests
[Build #1191 (03-Sep-2021 11:10:38)](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1191/)
[Build #1233 (9 Sep 2021, 11:00:55)](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1233/)
[ Build #1235 (9 Sep 2021, 11:36:38)](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1235/)

 

## Documentation
Will add in separate ticket.


test/cypress/integration/widgets/blueprint_catalog_spec.ts
```diff
@@ -34,4 +36,20 @@ describe('Blueprints catalog widget', () => {
         cy.contains('.segment', blueprintName).contains('Upload').click();
         cy.contains('.ui.label.section.active.pageTitle', blueprintName);
     });
+
+    it('should allow to change display style', () => {
+        cy.setDropdownConfigurationField('blueprintCatalog', 'Display style', ['Table']);
+        cy.get('table').should('be.visible');
```
kubama, line 42: I think we should be more specific here, at least `.blueprintCatalog table`.

test/cypress/integration/widgets/blueprint_catalog_spec.ts
```diff
@@ -34,4 +36,20 @@ describe('Blueprints catalog widget', () => {
...
+
+    it('should allow to customize fields to show', () => {
+        cy.setDropdownConfigurationField('blueprintCatalog', 'List of fields to show in the table', [
+            'Name',
+            'Created'
+        ]);
+        cy.contains('Name').should('be.visible');
+        cy.contains('Created').should('be.visible');
+        cy.contains('Updated').should('not.exist');
+        cy.contains('Description').should('not.exist');
```
kubama, line 53: Similarly here, I'd use more specific selectors.

test/cypress/support/commands.ts
```diff
@@ -404,6 +403,14 @@ const commands = {
             .click();
     },
 
+    clearDropdown: (fieldName: string) => {
+        cy.contains('.field', fieldName).within(() => {
+            if (Cypress.$(`.field:contains('${fieldName}') .delete.icon`).length > 0) {
+                cy.get('.delete.icon').click({ multiple: true });
+            }
```
kubama, line 410: What if the the `if` condition is not met? The command will just silently do nothing? Shouldn't it fail instead as all built-in Cypress commands do?
Also, why the `multiple` flag?

widgets/blueprintCatalog/src/widget.tsx
```diff
@@ -59,9 +59,17 @@ Stage.defineWidget<
...
             type: Stage.Basic.GenericField.LIST_TYPE
         },
+        {
+            id: 'fieldsToShow',
+            name: 'List of fields to show in the table',
+            placeHolder: 'Select fields from the list',
+            items: ['Name', 'Description', 'Created', 'Updated'],
+            default: 'Name,Description,Created,Updated',
+            type: Stage.Basic.GenericField.MULTI_SELECT_LIST_TYPE
+        },
```
kubama, line 72: Why no i18n?

widgets/blueprints/src/widget.tsx
```diff
@@ -48,7 +48,31 @@ Stage.defineWidget<unknown, unknown, BlueprintsWidgetConfiguration>({
...
+            ],
+            default: 'table',
+            type: Stage.Basic.GenericField.LIST_TYPE
+        },
+        {
+            id: 'marketplaceColumnsToShow',
+            name: Stage.i18n.t('widgets.blueprints.configuration.marketplaceColumnsToShow.label'),
+            placeHolder: Stage.i18n.t('widgets.blueprints.configuration.marketplaceColumnsToShow.placeHolder'),
+            items: ['Name', 'Description', 'Created', 'Updated'],
+            default: 'Name,Description',
```
kubama, line 73: I guess these could be also i18nized

widgets/blueprintCatalog/src/RepositoryTable.tsx
```diff
@@ -31,10 +32,10 @@ const RepositoryTable: FunctionComponent<RepositoryViewProps> = ({
...
             noDataMessage={noDataMessage}
         >
-            <DataTable.Column label="Name" width="25%" />
-            <DataTable.Column label="Description" width="40%" />
-            <DataTable.Column label="Created" width="12%" />
-            <DataTable.Column label="Updated" width="12%" />
+            <DataTable.Column label="Name" width="25%" show={fieldsToShow.includes('Name')} />
+            <DataTable.Column label="Description" width="40%" show={fieldsToShow.includes('Description')} />
+            <DataTable.Column label="Created" width="12%" show={fieldsToShow.includes('Created')} />
+            <DataTable.Column label="Updated" width="12%" show={fieldsToShow.includes('Updated')} />
```
kubama, line 38: Another place for i18nization.

Idea for future improvement: extend `DataTable` so it handled `fieldsToShow` internally.

test/cypress/support/commands.ts
stepek: This function should clear dropdown when there is any item selected, so when it is clean we can't clear it again

`multiple` flag is for clicking multiple founded items by `get` command.

test/cypress/support/editMode.ts
```diff
@@ -64,6 +64,12 @@ const commands = {
     setSearchableDropdownConfigurationField: (widgetId: string, fieldName: string, value: string) =>
         cy.editWidgetConfiguration(widgetId, () => {
             cy.setSearchableDropdownValue(fieldName, value);
+        }),
+
+    setDropdownConfigurationField: (widgetId: string, fieldName: string, value: string[]) =>
```
qooban, line 69: 1. I think more accurate would be to have 
```suggestion
    setDropdownConfigurationField: (widgetId: string, fieldName: string, values: string[]) =>
```
2. You use it for multiple dropdown, but it will work also for single selection dropdowns, right?
3. Maybe `clearDropdown` should be done inside `setDropdownValues`?

test/cypress/support/commands.ts
```diff
@@ -394,8 +394,7 @@ const commands = {
         }
     },
 
-    clearSearchableDropdown: (fieldName: string) =>
-        cy.contains('.field', fieldName).find('.dropdown.clear.icon').click(),
+    clearSearchableDropdown: (fieldName: string) => cy.contains('.field', fieldName).find('.clear.icon').click(),
```
qooban, line 397: Unrelated change or intentional?

test/cypress/support/commands.ts
qooban: Is it meant to work with single selection dropdown (without `multiple` prop set) as well?
If not, I think we should name it `clearMultipleDropdown` or something like that to be more descriptive about the command purpopse.

widgets/common/src/blueprintMarketplace/BlueprintMarketplaceModal.tsx
```diff
@@ -3,13 +3,15 @@ import { MarketplaceTab } from './types';
 
 interface BlueprintMarketplaceModalProps {
     open: boolean;
-    tabs: MarketplaceTab[];
+    tabs?: MarketplaceTab[];
+    displayStyle?: 'catalog' | 'table';
```
qooban, line 7: Seems like you defined the same union type few times. Maybe better would be to extract it to a reusable type? 
Maybe to `widgets/common/src/blueprintMarketplace/types.ts` and export it to `Stage.Common` similarly to how `MarketplaceTab` is exported?

widgets/common/src/blueprintMarketplace/BlueprintMarketplaceModal.tsx
```diff
@@ -3,13 +3,15 @@ import { MarketplaceTab } from './types';
...
+    columns?: string[];
     onHide: () => void;
 }
 
 const t = Stage.Utils.getT('widgets.common.blueprintMarketplace');
 
-const getPageLayout = (tabs: MarketplaceTab[]) => ({
+const getPageLayout = (tabs: MarketplaceTab[], displayStyle: string, columns: string[]) => ({
```
qooban, line 14: We will have conflict here (see changes in #1576 ), but should be easy to resolve.

widgets/blueprints/src/widget.tsx
```diff
@@ -48,7 +48,31 @@ Stage.defineWidget<unknown, unknown, BlueprintsWidgetConfiguration>({
             name: Stage.i18n.t('widgets.blueprints.configuration.showComposerOptions'),
             default: false
         },
-        Stage.Common.BlueprintMarketplace.tabsConfig
+        Stage.Common.BlueprintMarketplace.tabsConfig,
+        {
+            id: 'marketplaceDisplayStyle',
+            name: Stage.i18n.t('widgets.blueprints.configuration.marketplaceDisplayStyle.label'),
```
qooban, line 54: Lines are getting quite long. What do you think about having `t`-function here?
```
const t = Stage.Utils.getT(`widgets.blueprints...`);
```

In general:
1. Worth to check if the whole file can be translated.
2. Worth to check how difficult would it be to have full TS migration.

widgets/blueprints/src/types.ts
```diff
@@ -8,6 +8,8 @@ export interface BlueprintsWidgetConfiguration {
     hideFailedBlueprints: boolean;
     showComposerOptions: boolean;
     marketplaceTabs: Stage.Common.BlueprintMarketplace.Tab[];
+    marketplaceDisplayStyle: 'table' | 'catalog';
```
qooban, line 11: Maybe we can have something like
```suggestion
    marketplaceDisplayStyle: Stage.Common.BlueprintMarketplace.DisplayStyle;
```
?
(see my comment about extracting a type for `'table' | 'catalog'`)

widgets/blueprintCatalog/src/widget.tsx
```diff
@@ -59,9 +59,17 @@ Stage.defineWidget<
                 { name: 'Table', value: 'table' },
                 { name: 'Catalog', value: 'catalog' }
             ],
-            default: 'table',
+            default: 'catalog',
```
qooban, line 62: This change was intentional?

qooban (commented):
Added some comments.

Are you planning to change display style in "Service Button" widget within this ticket or should I create another one?

test/cypress/support/commands.ts
stepek: Unrelated change, sorry about that.

test/cypress/support/editMode.ts
stepek: 1. fixed 
2. should work
3. no, imho having it separately, keep current implementation of `set Dropdown Values` does not change current behavior of this method

test/cypress/support/commands.ts
kubama: I'm still not sure about the implementation of the command. If a command is intended to do something then I think it should either do it or fail if it's not possible to perform the action. It should also wait for a state change if it's not possible to perform the action at the time the command is executed. This implementation breaks this contract, thus:
- it's not in line with other Cypress commands, so it's behavior is unexpected
- it's error prone - since there is no wait for proper state (availability of delete icons) the result of command execution depends on exact timing
- it makes tests less robust - there are two possible execution paths (the `if`) while in any particular case only one is valid - in a robust test the command should be called in cases where there are items to be cleared, it shouldn't be called at all in cases when there is nothing selected

test/cypress/support/editMode.ts
```diff
@@ -64,6 +64,12 @@ const commands = {
     setSearchableDropdownConfigurationField: (widgetId: string, fieldName: string, value: string) =>
         cy.editWidgetConfiguration(widgetId, () => {
             cy.setSearchableDropdownValue(fieldName, value);
+        }),
+
+    setDropdownConfigurationField: (widgetId: string, fieldName: string, values: string[]) =>
+        cy.editWidgetConfiguration(widgetId, () => {
+            cy.clearMultipleDropdown(fieldName);
+            cy.setDropdownValues(fieldName, values);
```
qooban, line 72: I see you changed `clearDropdown` into `clearMultipleDropdown`, so does it mean it was meant to work only with `Dropdown` component having `multiple` prop on? (as I have asked in https://github.com/cloudify-cosmo/cloudify-stage/pull/1581/files#r701743728)

If so, then I think than now `setDropdownConfigurationField` name is misleading. It shouldn't be used for single selection dropdowns, right?

widgets/blueprints/src/types.ts
```diff
@@ -8,6 +8,8 @@ export interface BlueprintsWidgetConfiguration {
     hideFailedBlueprints: boolean;
     showComposerOptions: boolean;
     marketplaceTabs: Stage.Common.BlueprintMarketplace.Tab[];
+    marketplaceDisplayStyle: Stage.Common.BlueprintMarketplace.DisplayStyle;
+    marketplaceColumnsToShow: string[];
```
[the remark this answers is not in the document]
stepek: are you want add typings for marketplaceColumnsToShow based on translation file?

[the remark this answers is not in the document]
qooban: Hmmm... I'm not sure if that would be the best solution. We would need to make strict connection between `types.ts` and `en.json` file and that is probably something which we don't want to do. 

Maybe just strings union would be sufficient? Will it be valuable or am I wrong?

test/cypress/support/commands.ts
stepek: Ok @kubama let me remove this if statement.

widgets/blueprints/src/types.ts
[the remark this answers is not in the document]
stepek: no, it's not good solution, maybe the column keys should be const and should be separated from translations?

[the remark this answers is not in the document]
qooban: Seeing that it's not trivial and we don't have unlimited time for 6.2 features, let's maybe keep `string[]` for now.

qooban (approved):
I'm OK with the code right now.
Let's wait for @kubama review.

qooban:
@stepek FYI There are conflicts to be resolved.

test/cypress/support/commands.ts
```diff
@@ -375,42 +375,38 @@ const commands = {
...
-        cy.contains('.field', fieldName).within(() => {
-            if (Cypress.$(`.field:contains('${fieldName}') .delete.icon`).length > 0) {
-                cy.get('.delete.icon').click({ multiple: true });
-            }
-        });
-    },
+    clearMultipleDropdown: (fieldName: string) =>
+        cy.getField(fieldName).within(() => {
+            cy.get('.delete.icon').click({ multiple: true });
+        }),
```
kubama, line 409:
```suggestion
        cy.getField(fieldName).find('.delete.icon').click({ multiple: true });
```

kubama (approved):

qooban (approved):

stepek:
https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1235/

kubama (approved):

qooban (approved):

stepek:
Tests passing

PR #1597, opened 2021-09-10
RD-3100 Added `wrapLongLines` prop to `HighlightText` component to prevent off-screen rendering of long text lines

kubama:
## Description
Popup content causing the popup to be wider than viewport size was also disabling automatic positioning of the popup.
The solution is to ensure narrower popup content by adding `wrapLongLines` prop to the nested `HighlightText` component.

## Screenshots / Videos
No change in the actual look or feel.

## Checklist

## Tests
N/A

## Documentation
N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11227,9 +11227,9 @@
       }
     },
     "cloudify-ui-components": {
-      "version": "2.7.1",
-      "resolved": "https://registry.npmjs.org/cloudify-ui-components/-/cloudify-ui-components-2.7.1.tgz",
-      "integrity": "sha512-CEGkBptFsiQ3BcsZMqfM52zsnQnYc2vmAgleOCr9mODkgqvLol5sDJP+1kyJBcgZM0DoxyaIuYt+nTNVanpC9Q==",
+      "version": "2.8.0",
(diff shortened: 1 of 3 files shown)


qooban (commented):
> No change in the actual look or feel.

I believe there is actually a change 🙂 Having a screenshot would be nice to have here.
I think there are two cases:
1. Long lines (WDR-65)
2. Big number of lines (RD-3100)

Does it render properly for both cases?

kubama:
> Does it render properly for both cases?

After playing some more I think the best approach here would be to replace the popup with a modal window. Popup is not the best way to handle variable height, possibly long content - it should stretch vertically to handle it properly, but this would influence its position when referencing elements close to the viewport top. Unfortunately, the `Popup` component we are using doesn't handle such cases well, and there is no API that would allow to control it directly - there is `popperModifiers` prop that allows to customize the underlying implementation (https://popper.js.org/docs/v2/modifiers/) but it's so low-level that I consider it hacky. So, my proposal is to go with a modal - WDYT?

qooban:
> So, my proposal is to go with a modal - WDYT?

I remember we were discussing that it might be necessary to have modal instead of the popup here.
So yes, I think it's the right way to go.

kubama:
> So yes, I think it's the right way to go.

If so I'll submit a PR to revert components link update in the docs, as common-components update won't be necessary here.

PR #1598, opened 2021-09-13
RD-3140 Update Blueprint Marketplace modal to use Table view by default

qooban:
## Description
This PR introduces final alignments in Actionable Dashboard. Blueprint Marketplace modal is displayed right now in Table view with just 'Name' and 'Description' columns visible.

Within this PR I fixed also displaying action items in Table view in Blueprint Catalog widget:
* spinner icon showed on blueprint upload was not spinning before (`loading` prop was never set on `Icon` component)
* blueprint upload icon was not disabled properly when blueprint was already uploaded (`link` prop was always set to `true` on `Icon` component)

## Screenshots / Videos

https://user-images.githubusercontent.com/5202105/133035956-541e82a4-6185-41c4-b1d9-f8cec26c2204.mp4

## Checklist

## Tests
[Stage-UI-System-Test/1248](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1248/) (13.09.2021 08:46:06 CEST) - PASSED

## Documentation
No updates necessary.

### DIFF
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -732,6 +732,10 @@
             }
         },
         "blueprintCatalog": {
+            "actions": {
+                "openDocumentation": "Open documentation",
+                "uploadBlueprint": "Upload blueprint"
+            },
             "configuration": {
                 "jsonPath": {
                     "label": "Blueprints Examples URL",
--- a/.size-limit
+++ b/.size-limit
@@ -13,7 +13,7 @@
     },
     {
         "name": "Widgets",
-        "limit": "3.13 MB",
+        "limit": "3.14 MB",
         "gzip": false,
         "running": false,
         "path": ["dist/appData/widgets", "!dist/appData/widgets/**/*.gz"]
--- a/widgets/blueprintCatalog/src/RepositoryTable.tsx
+++ b/widgets/blueprintCatalog/src/RepositoryTable.tsx
@@ -57,7 +57,10 @@ const RepositoryTable: FunctionComponent<RepositoryViewProps> = ({
             <DataTable.Column width="11%" />
 
             {data.items.map(item => {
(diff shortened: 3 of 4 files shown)


app/translations/en.json
```diff
@@ -732,6 +732,10 @@
             }
         },
         "blueprintCatalog": {
+            "actions": {
+                "openDocumentation": "Open documentation",
```
kubama, line 736: Wouldn't a more appropriate term be 'go to' instead of 'open'?

qooban: Hmm... I'm not sure. Before, the title of the icon was "Blueprint Readme". Clicking on (i) icon opens a modal with `README.md` file from a specific blueprint. "Go to" would probably be more accurate if we were redirecting to another page. Can be also "Show documentation" or "Show readme file", I'm open to the suggestions, but not sure about using "Go to" in that case. WDYT?

kubama: I guess you are right. If so, I'm perfectly fine with any option.

kubama (dismissed):

app/translations/en.json
qooban: In that case I'll leave it as is.

qooban:
I have re-triggered CI, it should pass as `cloudify-ui-components` version is now in line with the docs.

stepek (approved):

qooban:
Thanks for approvals. We need to merge https://github.com/cloudify-cosmo/cloudify-stage/pull/1600 first, then Docs stage will pass in the CI.

qooban:
One more small change unfortunately - I had to increase size limit for widgets. 
I'll consider single approval as a "GO" for merge.

kubama (approved):

PR #1599, opened 2021-09-13
RD-2917 Bumped cloudify-ui-common version

kubama:
## Description
Bumped cloudify-ui-common version to use fixed `getLogger` implementation.

## Screenshots / Videos
N/A

## Checklist

## Tests
N/A

## Documentation
N/A

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -2406,9 +2406,9 @@
       }
     },
     "cloudify-ui-common": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/cloudify-ui-common/-/cloudify-ui-common-4.0.1.tgz",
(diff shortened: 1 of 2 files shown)


kubama:
Hmmm... I messed something with the branch. Just a moment 🙂.

kubama:
Fixed, please review

qooban (approved):
LGTM 👍🏻 

Do you plan to update `cloudify-blueprint-composer` also?
I think we don't need to have that in Composer v6.2, but it'd be good to have it already available in `master`.

kubama:
> Do you plan to update `cloudify-blueprint-composer` also?
> I think we don't need to have that in Composer v6.2, but it'd be good to have it already available in `master`.

I did a quick check and it looks like composer is not prone to the same issue as for each category it obtains logger exactly once. So I'm not planing an update there, as it wouldn't change anything.

PR #1294, opened 2021-04-07
RD-1993 - Fix Sites, Sites Map and Topology widgets' tests

qooban:
This PR cherry-picks fixes for: 
* RD-1358 Prevent overriding app.json when uploading stage package (#1194)
* RD-1120 Fixed blueprint used by topology_spec (#1205)

For more details about where it came from, check `5_1_4_release` slack channel.

PR #1619, opened 2021-09-16
RD-3184 Fixed disable upload button condition.

stepek:
## Description
Fixed disable upload button condition. Now upload button isn't disabled when blueprint state is `invalid`

## Screenshots / Videos
N/A

## Checklist

## Tests
[Build #1294 (Sep 16, 2021, 11:09:05 PM)](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1294/)

 

## Documentation
n/a


widgets/blueprintCatalog/src/widget.tsx
```diff
@@ -99,9 +99,14 @@ Stage.defineWidget<
             widget.configuration.filter,
             widget.configuration.jsonPath
         );
-        return Promise.all([actions.doGetRepos(params), toolbox.getManager().doGet('/blueprints?_include=id', {})])
+        return Promise.all([
+            actions.doGetRepos(params),
+            toolbox.getManager().doGet('/blueprints?_include=id,state', {})
```
qooban, line 104: Do we need to pass `{}`?

Will
```suggestion
            toolbox.getManager().doGet('/blueprints?_include=id,state')
```
work?

Even better would be to reuse `BlueprintActions.doGetBlueprints`.

widgets/blueprintCatalog/src/widget.tsx
```diff
@@ -99,9 +99,14 @@ Stage.defineWidget<
...
+        ])
             .then(([data, blueprintsRes]) => {
-                const uploadedBlueprints = blueprintsRes.items.map(({ id }: Partial<Blueprint>) => id);
+                const uploadedBlueprints = blueprintsRes.items
+                    .filter(({ state }: Partial<Blueprint>) => state !== 'invalid')
```
qooban, line 108: In `BlueprintActions` I see the following states defined:
```
    static CompletedBlueprintStates = {
        Uploaded: 'uploaded',
        FailedUploading: 'failed_uploading',
        FailedExtracting: 'failed_extracting',
        FailedParsing: 'failed_parsing',
        Invalid: 'invalid'
    };
```
which suggests that `uploadedBlueprints` should also filter out other states (all starting from `failed_`).
Looks like there is `BlueprintActions.isUploaded` static method already available for that. Can we use it here?

Or maybe we can have `BlueprintActions.doGetUploadedBlueprints` method and reuse it whenever we call `.doGet('/blueprints...')`? I think the only case when we should return not uploaded blueprints is in Blueprints widget where we list all available and provide information about the blueprint state. In other cases we should only return valid (= uploaded) blueprints.

Do you think you'll be able to fit such fix in this PR?

qooban (commented):
Commented. 

I think it'll be good to take opportunity and unify fetching blueprints (`GET /blueprints` calls) across all widgets. See my comments for details.

kubama:
I don't have anything to add to @qooban's comments, so I'll opt-out from the review.

widgets/blueprintCatalog/src/widget.tsx
stepek: Can be removed

stepek: to use `doGetBlueprints` need to add one more parameter there it's ok?

stepek: nvm fixed

widgets/common/src/BlueprintActions.ts
```diff
@@ -1,4 +1,5 @@
 // eslint-disable-next-line max-classes-per-file
+
```
qooban, line 2: This empty line will probably produce linter error.

widgets/common/src/BlueprintActions.ts
```diff
@@ -55,6 +56,15 @@ export default class BlueprintActions {
         return this.toolbox.getManager().doGet('/blueprints?_include=id', { params });
     }
 
+    doGetUploadedBlueprints(params?: Record<string, any>) {
+        return this.toolbox
+            .getManager()
+            .doGet('/blueprints?_include=id,state', { params })
+            .then(response =>
+                response.items.filter((blueprint: { state: string }) => BlueprintActions.isUploaded(blueprint))
+            );
```
qooban, line 65: Wouldn't it be easier to do the filtering in a call?
```suggestion
            .doGet('/blueprints', { _include: 'id,state', state: BlueprintActions.CompletedBlueprintStates.Uploaded, ...params })
```

qooban (commented):
Commented. What about alignment with other widgets? (not answered comment from my first review)

widgets/blueprintCatalog/src/widget.tsx
```diff
@@ -93,15 +93,17 @@ Stage.defineWidget<
...
             widget.configuration.username,
             widget.configuration.filter,
             widget.configuration.jsonPath
         );
-        return Promise.all([actions.doGetRepos(params), toolbox.getManager().doGet('/blueprints?_include=id', {})])
-            .then(([data, blueprintsRes]) => {
-                const uploadedBlueprints = blueprintsRes.items.map(({ id }: Partial<Blueprint>) => id);
+        return Promise.all([actions.doGetRepos(params), blueprintActions.doGetUploadedBlueprints()])
+            .then(([data, uploadedBlueprintsResp]) => {
+                const uploadedBlueprints = uploadedBlueprintsResp.items.map(({ id }: { id: string }) => id);
```
qooban, line 106: Intentionally not using `Partial<Blueprint>` which was used before?

widgets/common/src/BlueprintActions.ts
```diff
@@ -55,6 +55,12 @@ export default class BlueprintActions {
         return this.toolbox.getManager().doGet('/blueprints?_include=id', { params });
     }
 
+    doGetUploadedBlueprints(params?: Record<string, any>) {
+        return this.toolbox.getManager().doGet('/blueprints?_include=id,state', {
```
qooban, line 59: Intentionally not moving `_include` into params object (as in suggestion)?

qooban (commented):
Commented. 

I don't want to prolong that review, but I see my question not answered, again. What about alignments with other widgets?

qooban (commented):
Change itself looks good. 

The only open item is alignment with other widgets. I'm not sure if you see my question in the previous reviews?

widgets/blueprintActionButtons/src/BlueprintActionButtons.tsx
```diff
@@ -140,7 +141,7 @@ export default class BlueprintActionButtons extends React.Component<
                                     this.setState({ loading: false });
                                 });
                         }}
-                        content="Edit a copy in Composer"
+                        content="Edit a copy in XXXXXXX"
```
qooban, line 144: Probably a mistake 🙂

qooban (approved):

PR #1624, opened 2021-09-22
RD-2993 Added sample Cypress component test

kubama:
## Description
Added sample test for `GettingStartedModal` component and integrated component testing with CI pipeline by extending `test:frontend` npm script.

There is still some more work planned in this area. Please see subtasks in RD-2993 for a complete list.

## Screenshots / Videos
N/A

## Checklist

## Tests
System tests to verify nothing's broken: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1318/

## Documentation
Updated `test/README.md` file


package.json
```diff
@@ -133,9 +136,10 @@
     "webpack-cli": "^3.3.9",
     "webpack-bundle-analyzer": "^3.8.0"
   },
-  "//": "NOTE: Development dependencies NOT required to build the package",
```
qooban, line 136: Unfortunately this comment gets removed by `npm install` because there's entry with the same key (`"//"`). If you could add unique key here and for optional dependencies comment (e.g. `"// devDependencies"` and `"// optionalDependencies"` instead of just `"//"`), I'd be grateful.

package.json
```diff
@@ -28,17 +28,20 @@
...
     "e2e": "cypress run -C node_modules/cloudify-ui-common/cypress/cypress.json",
-    "e2e:open": "cypress open -C node_modules/cloudify-ui-common/cypress/cypress.json",
     "e2e:ci": "npm run beforebuild && cross-env TEST=1 run-system-tests ${TEST_SPEC}",
     "lint": "eslint --cache --ignore-path .gitignore --ext js,jsx,ts,tsx .",
     "size": "[ -d \"dist\" ] && size-limit",
     "test": "npm run test:frontend && npm run test:backend",
-    "test:frontend": "jest test/jest",
+    "test:frontend": "npm run test:frontend:unit && npm run test:frontend:components",
+    "test:frontend:unit": "jest test/jest",
+    "test:frontend:components": "cypress run-ct -C node_modules/cloudify-ui-common/cypress/cypress.json -c nodeVersion=system",
```
qooban, line 44: Nothing bad here, just out of curiosity - why setting `nodeVersion` was necessary? 
Bundled was higher than we use possibly and there were problems with running tests?

test/cypress/components/app/GettingStartedModal_spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { i18nInit } from 'app/app';
+
+describe('GettingStartedModal', () => {
+    it('renders', () => {
+        // TODO: RD-3232 create common setup function
+        cy.viewport(1600, 900);
+        window.React = React;
+        cy.interceptSp('GET', '/users/null', { show_getting_started: true });
+        cy.stub(StageUtils, 'isUserAuthorized').returns(true);
+        i18nInit();
```
qooban, line 20: Does it mean that `app.load` method is not called and we have to load translations manually?

test/cypress/components/app/GettingStartedModal_spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+        mount(
+            <Provider store={createStore(createRootReducer(createBrowserHistory()))}>
+                <ThemeProvider theme={userConfig.whiteLabel}>
```
qooban, line 24: I suspect that it'd make sense to extract `Provider + ThemeProvider` as a common wrapper for component tests. 
Possibly you already had that in mind for the next sub-tasks?

test/cypress/plugins/index.ts
```diff
@@ -12,10 +12,23 @@
 // the project's config changing)
 
 import performCommonSetup from 'cloudify-ui-common/cypress/plugins';
+import { startDevServer } from '@cypress/webpack-dev-server/';
```
qooban, line 15: Very minor:
```suggestion
import { startDevServer } from '@cypress/webpack-dev-server';
```

qooban (commented):
Nicely done!

I tried it locally and looks cool - both runner and console versions.

https://user-images.githubusercontent.com/5202105/134460400-906eaa4c-c5fb-4fca-a693-571cf3cfec78.mp4

I'm wondering about the execution time. I tried:
```
time npm run test:frontend:components
```
and the results are:
```
       Spec                                              Tests  Passing  Failing  Pending  Skipped  
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  app/GettingStartedModal_spec.tsx         461ms        1        1        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        461ms        1        1        -        -        -  


real    0m26,316s
user    0m10,540s
sys     0m1,399s
```

so the test itself is super fast (461ms), but setup takes some time (10s). I'm wondering how will it behave in CI (which is usually slower than our machines). Of course it is not a blocker, I'm just pointing that maybe in the future we would like to improve that area.

package.json
```diff
@@ -28,17 +28,20 @@
     "check-types": "npm run check-types:frontend && npm run check-types:backend",
     "check-types:frontend": "tsc --build",
     "check-types:backend": "npm run --prefix backend check-types",
+    "cy:open": "cypress open -C node_modules/cloudify-ui-common/cypress/cypress.json",
+    "cy:open-ct": "cypress open-ct -C node_modules/cloudify-ui-common/cypress/cypress.json -c nodeVersion=system",
```
stepek, line 32: Maybe this should be named something like `test:frontend:components:open`? same above

test/cypress/.babelrc.json
```diff
@@ -0,0 +1,10 @@
+{
+    "plugins": [
+        ["module-resolver", {
+            "alias": {
+                "app": "./app",
+                "conf": "./conf"
+            }
+        }]
+    ]
+}
```
stepek, line 10: missing new line at the end of file

package.json
kubama: Done

package.json
kubama: There was an issue with `node-sass` which is bound to specific node version. When running webpack from within Cypress it was executed with a different node version than it was installed with. Thus, it was not working without this option.

test/cypress/components/app/GettingStartedModal_spec.tsx
kubama: Yes, nothing from app initialization is executed automatically. If needed, initialization has to be done manually.

test/cypress/components/app/GettingStartedModal_spec.tsx
kubama: Yes, once we have another test it will be pretty clear what the common parts are.

package.json
kubama: It would be a pretty long name for a script that is intended to be run from command line manually.

test/cypress/.babelrc.json
kubama: Fixed

kubama:
> so the test itself is super fast (461ms), but setup takes some time (10s). I'm wondering how will it behave in CI (which is usually slower than our machines). Of course it is not a blocker, I'm just pointing that maybe in the future we would like to improve that area.

I haven't tested yet how it behaves with multiple test suites. We'll see.

> I'm not sure how difficult it would be to add required dependencies to the VM. You can probably take a look at the VM setup for UI-System-Tests. In case of problems, we can always run components tests only within UI-System-Tests job.

I'll try and we'll see.

kubama:
Despite trying hard I was not able to integrate component tests with integration build. For now, integration build just runs `npm run test:frontend:unit`. Created RD-3240 which should be handled by dev-ops.

package.json
qooban: Got it, reasonable solution 👍🏻

test/cypress/components/app/GettingStartedModal_spec.tsx
qooban: OK, so as you noted, we'll need a common setup commands (= Cypress command or some commands) which will make setting up components test easy. That's of course for the later part.

qooban (dismissed):
> Despite trying hard I was not able to integrate component tests with integration build. For now, integration build just runs npm run test:frontend:unit. Created RD-3240 which should be handled by dev-ops.

Thanks for that. 

So right now components tests will be executed as part of `Stage-UI-System-Tests`, right?
(as `e2e:ci` script runs `test:frontend:coverage`, which runs `test:frontend -- --coverage`, which runs `test:frontend:components` 🙂)

kubama:
> So right now components tests will be executed as part of `Stage-UI-System-Tests`, right?
> (as `e2e:ci` script runs `test:frontend:coverage`, which runs `test:frontend -- --coverage`, which runs `test:frontend:components` 🙂)

This is the plan - it's not working at the moment, I need to investigate yet.

package.json
stepek: so maybe need to short all of the names?

kubama:
System tests passed.

There is an issue with job log being bloated with errors. This does not affect the build, just trashes the log. I created RD-3251 so dev-ops could address the issue.

At the moment component tests do not generate coverage report that is merged into the overall report. I created separate subtask to handle it: RD-3236.

@stepek I'm waiting for your approval and we're good to go.

package.json
kubama: The other names are not as long as the one proposed. Personally I like that they are kind of hierarchically structured. Do you have any suggestion how to shorten them and keep the naming structure?

qooban: I would stay with short names, especially that those two scripts are intended to be run from command line manually only, they are not to be used in CI.

qooban (approved):

PR #1626, opened 2021-09-24
RD-2993 Added sample test for widgets component

kubama:
## Description
I'll rebase this to master once #1624 is merged.

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1320/

## Documentation
N/A

### DIFF
--- a/test/cypress/components/app/GettingStartedModal_spec.tsx
+++ b/test/cypress/components/app/GettingStartedModal_spec.tsx
@@ -1,10 +1,11 @@
-import '../initAppContext';
 import React from 'react';
 import { mount } from '@cypress/react';
 import { Provider } from 'react-redux';
 import { createStore } from 'redux';
 import { createBrowserHistory } from 'history';
 import { ThemeProvider } from 'styled-components';
+
+import '../initAppContext';
 import GettingStartedModal from 'app/components/GettingStartedModal';
 import StageUtils from 'app/utils/stageUtils';
 import userConfig from 'conf/userConfig.json';
--- a/test/cypress/components/app/GettingStartedModal_spec.tsx
+++ b/test/cypress/components/app/GettingStartedModal_spec.tsx
@@ -1,3 +1,4 @@
+import '../initAppContext';
 import React from 'react';
 import { mount } from '@cypress/react';
 import { Provider } from 'react-redux';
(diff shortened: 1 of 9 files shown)


test/cypress/components/app/GettingStartedModal_spec.tsx
```diff
@@ -1,3 +1,4 @@
+import '../initAppContext';
```
qooban, line 1: I think that local imports should be below imports from `node_modules` or there's hidden intention here? Same solution (if change needed) should be applied to `test/cypress/components/widgets/filters/RulesForm_spec.tsx`.

(linter does not scream here, so that's another reason for making clean-up with ESLint 🙂)

test/cypress/components/widgets/filters/RulesForm_spec.tsx
```diff
@@ -0,0 +1,16 @@
+import '../../initAppContext';
+import 'widgets/common/src/props/Toolbox';
```
qooban, line 2: Oh, having that on top suggests that it was intentional, am I right? Order of imports matters for Cypress components testing?

qooban (approved):
I added some minor comments about import order.

If that was intentional to do it the way you did it, I'm fine, approved. 
If not, then we probably can fix it and maybe ESLint rule can be reviewed? I'm talking about https://eslint.org/docs/rules/sort-imports which AFAIR was used in our repos, but maybe not everywhere? I don't remember if it was part of some set?

stepek (dismissed):

test/cypress/components/app/GettingStartedModal_spec.tsx
kubama: > I added some minor comments about import order.

I addressed the comments by reordering the imports. Generally: `initAppContext` should be imported before any other application module, but it can be imported after all `node_modules` imports.

> If not, then we probably can fix it and maybe ESLint rule can be reviewed? I'm talking about https://eslint.org/docs/rules/sort-imports which AFAIR was used in our repos, but maybe not everywhere? I don't remember if it was part of some set?

I'll check.

qooban: Thanks 👍🏻

qooban (dismissed):

test/cypress/components/app/GettingStartedModal_spec.tsx
kubama: Running: `npx eslint --print-config test/cypress/components/app/GettingStartedModal_spec.tsx` gives:
```
    "import/order": [
      "error",
      {
        "groups": [
          [
            "builtin",
            "external",
            "internal"
          ]
        ]
      }
    ],
```
So I have no clue why the rule didn't work.

kubama:
@alexmolev
This was already approved by @qooban and @stepek however it needs re-approval after rebase.

alexmolev (approved):

PR #1623, opened 2021-09-22
RD-381 Update DB related code to be compatible with Sequelize v6

qooban:
## Description

This PR migrates all DB-related logic to be compatible with Sequelize v6:
1. Migrates `Sequelize.Promise` uses (as it was removed from v6)
2. Migrates `spread` uses (as Bluebird promises were removed from v6)
3. [not yet pushed, see To Do] DB initializer adjustments to use new models initialization method (compatible with Sequelize v6) and introduced within https://github.com/cloudify-cosmo/cloudify-ui-common/pull/75 (as `sequelize.import` was removed in v6)

I found a lot of inconsistency between model filenames, model names and DB table names, so there are a few file renames and model name changes as well included in this PR. 

### To do

- [x] Merge https://github.com/cloudify-cosmo/cloudify-ui-common/pull/75 and release new major version of `cloudify-ui-common`
- [x] Update `cloudify-ui-common` version
- [x] Update DB initializer to stop passing directory, but to provide array of imported functions:
```
import ApplicationsModel from './models/ApplicationsModel';
import BlueprintAdditionsModel from './models/BlueprintAdditionsModel';
import BlueprintUserDataModel from './models/BlueprintUserDataModel';
import ClientConfigsModel from './models/ClientConfigsModel';
import ResourcesModel from './models/ResourcesModel';
import UserAppsModel from './models/UserAppsModel';
import WidgetBackendsModel from './models/WidgetBackendsModel';

const dbInitializer = new DbInitializer(getConfig().app.db, loggerFactory, [
    ApplicationsModel,
    BlueprintAdditionsModel,
    BlueprintUserDataModel,
    ClientConfigsModel,
    ResourcesModel,
    UserAppsModel,
    WidgetBackendsModel
]);
```

## Screenshots / Videos
N/A

## Checklist
- [x] My code follows the UI Code Style.
- [x] I verified that all tests and checks have passed.
- [x] I verified if that change requires a documentation update.
- [x] I added proper labels to this PR.

## Tests
1. [Stage-UI-System-Test/1306](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1306/) (22.09.2021 10:19:23, before Sequelize version upgrade) - PASSED
2. [Stage-UI-System-Test/1334](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1334/pipeline) (04.10.2021 15:12:04, with all changes) - PASSED

[...]

PR #1628, opened 2021-10-01
RD-3241 Added backend support for reading list of built-in page groups

kubama:
## Description
Includes some refactoring and TS migration.

## Screenshots / Videos
N/A

## Checklist

## Tests
Added unit test.
Audit failure is not related to the changes in this branch.

## Documentation
N/A


stepek (dismissed):

backend/handler/templates/TemplatesHandler.ts
```diff
@@ -76,47 +86,21 @@ export function listTemplates() {
...
         'name',
         null
     );
 }
 
-async function getRole(userSystemRole, groupSystemRoles, tenantsRoles, tenant, token) {
+async function getRole(
+    userSystemRole: string,
+    groupSystemRoles: Record<string, any>,
+    tenantsRoles: Record<string, any>,
```
qooban, line 100: I was curious if we can be more specific here, so I did a quick investigation. Docs for `/users` API are not up to date (BTW I have created an issue for that 3 years ago :-) - https://cloudifysource.atlassian.net/browse/CY-3784), but did some tests and found out that:

`groupSystemRoles` can be mapped to `group_system_roles` from `GET /user?_get_data=true` (see `Auth.ts`). It has the following format:
  
```
  "group_system_roles": {
    "default": [
      "group"
    ]
  }
```
so I think we can have a type for it, sth like:
```
type GroupSystemRoles = Record<string, string[]>
```

`tenantsRoles` can be mapped to `tenants` from `GET /user?_get_data=true` (see `Auth.ts`). It has the following format:

```
  "tenants": {
    "default_tenant": {
      "tenant-role": "manager",
      "roles": [
        "manager"
      ]
    },
    "T2": {
      "tenant-role": "operations",
      "roles": [
        "operations"
      ]
    },
    "T1": {
      "tenant-role": null,
      "roles": [
        "manager"
      ]
    }
  }
```

so we can have sth like:

```
type TenantsRoles = Record<string, { "tenant-role": string | null, roles: string[] }>;
```

WDYT?

backend/routes/index.d.ts
```diff
@@ -0,0 +1,13 @@
+declare global {
+    namespace Express {
+        interface User {
+            username: string;
+            role: string;
+            // eslint-disable-next-line camelcase
+            group_system_roles: Record<string, any>;
+            tenants: Record<string, any>;
```
qooban, line 8: The types I mentioned in one of the previous comments can be introduced here and then reused in other backend routes/handlers.

backend/tsconfig.json
```diff
@@ -3,5 +3,9 @@
     "compilerOptions": {
         "noEmit": true
     },
-    "exclude": ["**/__mocks__/*", "test"]
+    "exclude": ["**/__mocks__/*", "test"],
+
+    "ts-node": {
+        "files": true
```
qooban, line 9: Just out of curiosity: How it actually works? 

I checked ts-node docs and found that:
```
--files Load files, include and exclude from tsconfig.json on startup
```
What does "load" means in this context? Type checking on startup?

qooban (commented):
Good idea to refactor templates/pages handlers 👍🏻 

Added some comments, mostly related to file naming or type improvements.

backend/tsconfig.json
kubama: It loads d.ts files which are ignored without this setting.

backend/routes/Templates.ts
```diff
@@ -1,9 +1,9 @@
-// @ts-nocheck File not migrated fully to TS
-
 import express from 'express';
 import bodyParser from 'body-parser';
 import passport from 'passport';
-import * as TemplateHandler from '../handler/TemplateHandler';
+import * as TemplateHandler from '../handler/templates/TemplatesHandler';
+import * as PageHandler from '../handler/templates/PagesHandler';
+import * as PageGroupsHandler from '../handler/templates/PageGroupsHandler';
```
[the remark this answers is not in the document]
kubama: Done

backend/server.ts
```diff
@@ -2,8 +2,8 @@
 import app from './app';
 import { SERVER_HOST, SERVER_PORT } from './consts';
 import DBConnection from './db/Connection';
-import * as WidgetHandler from './handler/WidgetHandler';
-import * as TemplateHandler from './handler/TemplateHandler';
+import { init as initWidgetHandler } from './handler/WidgetHandler';
```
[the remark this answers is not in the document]
kubama: Done

backend/server.ts
```diff
@@ -2,8 +2,8 @@
...
+import { init as initTemplateHandler } from './handler/templates';
```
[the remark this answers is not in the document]
kubama: Done

backend/handler/templates/index.ts
```diff
@@ -0,0 +1,22 @@
+import { sync as mkdirpSync } from 'mkdirp';
+import { userTemplatesFolder } from './TemplatesHandler';
+import { userPagesFolder } from './PagesHandler';
+import { getLogger } from '../LoggerHandler';
+
+const logger = getLogger('TemplateHandler');
```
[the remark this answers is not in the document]
kubama: Done

backend/handler/templates/TemplatesHandler.ts
kubama: In case of this particular case (L99) `Record<string, any>` for `groupSystemRoles` is more appropriate type as it's less restrictive - here we are using only the object keys, so values type are out of interest. However I introduced the type in `User` as well as introduced `TenantsRoles` in all suggested places.

backend/routes/index.d.ts
kubama: Done

backend/handler/templates/TemplatesHandler.ts
qooban: Got it, make sense then.

backend/tsconfig.json
qooban: Clear, thx for explanation.

qooban (approved):
LGTM 👍🏻 

There's one conflict to be resolved coming from last merge ( #1623 ).

PR #1637, opened 2021-10-13
RD-3241 Added support for in-line edit of page menu items names

kubama:
## Description
Added support for in-line edit of page menu items names. Includes further refactoring around `PagesList` component.

## Screenshots / Videos
image

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1377/

I extended component test to cover name change and remove functionality. Unfortunately I was not able to test items reordering. I tried:
- Cypress drag'n'drop plugin (https://github.com/4teamwork/cypress-drag-drop) - looks like the plugin is frequently reported to not working: https://github.com/4teamwork/cypress-drag-drop/issues/37
- approach taken directly from dnd-kit Cypress test: https://github.com/clauderic/dnd-kit/blob/master/cypress/support/commands.ts

None of the above worked.

## Documentation
Will be covered in RD-3341.

### DIFF
--- a/app/components/sidebar/PagesList.tsx
+++ b/app/components/sidebar/PagesList.tsx
@@ -55,6 +55,8 @@ const RemoveIcon = styled(Icon)`
 `;
 
 const EditIcon = styled(Icon)`
+    position: relative;
+    top: -3px;
     margin-left: 1em !important;
     display: none !important;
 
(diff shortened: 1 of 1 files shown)


qooban (commented):
I went through the code without big attention to detail, didn't find any issues. Good refactoring of page-related actions (unfortunately it made review a bit harder). I hope @stepek would be able to do more detailed review.

I focused on the UI changes and did some tests. In general inline editing works very good. I've got just one question: Can we have page item name truncation for long names in sidebar? (when page name is longer than the container it is not possible to click edit/remove icons)

PR #1639, opened 2021-10-19
RD-3241 Added support for adding/removing page groups to/from templates

kubama:
## Description
Added support for adding/removing page groups to/from templates.

## Screenshots / Videos
image

## Checklist

## Tests
Extended system test.
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1387/

 

## Documentation
Will be updated in separate ticket.


qooban:
I'll review tomorrow. 

It seems that some file renames with content change were recognized as file removals/additions, but going quickly through commits it looks like it will be possible to review commit-by-commit without having that issue.

app/components/templates/CreateTemplateModal.tsx
```diff
@@ -24,15 +23,33 @@ import {
...
+
+const t = StageUtils.getT('templates.createTemplateModal');
+
+interface PageMenuItem {
+    id: string;
+    type: 'page' | 'pageGroup';
+}
+
+function toId(item: PageMenuItem) {
+    return `${item.type}\n${item.id}`;
```
qooban: Can you remind me what was the reason for having line separated ID instead of e.g. concatenating using underscore or hyphen?

app/components/templates/CreateTemplateModal.tsx
```diff
@@ -43,347 +60,318 @@ const SortablePageItem: FunctionComponent<SortablePageItemProps> = ({ name, onRe
...
-    // eslint-disable-next-line react/no-unused-prop-types
-    roles: PropTypes.arrayOf(PropTypes.string),
-    // eslint-disable-next-line react/no-unused-prop-types
-    tenants: PropTypes.arrayOf(PropTypes.string),
-    onCreateTemplate: PropTypes.func.isRequired
-};
+    const tenantOptions = _.map(allAvailableTenants, item => {
+        return { text: item, value: item };
+    });
+    tenantOptions.push({ text: 'All tenants', value: Consts.DEFAULT_ALL });
```
qooban: Maybe we should move "All tenants" to translation file?

qooban (commented):
Added some comments. 

I'm not sure about the naming convention used to describe Pages and Page Groups - Page Menu Item (it might suggest that it refers only to the Pages). I don't have clear suggestion right now.

Did quick UI tests. Overall it works good, have some minor UX related suggestions:

1. Maybe worth to change List of pages and List of page groups to be accordions (and only one expanded at a time) that way user will be aware of both types as by default only pages are displayed (as there are many built-in pages):
image
2. I think we should change icons displayed on the left side of "Available pages" and "Available page groups". Currently we show plus sign, but it might be a page icon for pages and multiple page icon page groups. Something from `file` icon family in SUIR would probably be suitable. WDYT?

app/components/templates/CreateTemplateModal.tsx
kubama: By using `-` or `_` we should consider if the character may or may not occur in the first concatenated string, new line character is a safer option as we can assume it's not present in concatenated strings as long as they are not intended to contain free text.

app/components/templates/PageList.tsx
```diff
@@ -27,37 +30,24 @@ export default function PageList({ custom, onDelete, pages, style }) {
...
                                     }
-                                    content={i18n.t(
-                                        'templates.templateManagement.pageList.removeConfirm',
-                                        'Are you sure to remove this page from template?'
-                                    )}
+                                    content={t('removeConfirm')}
                                     onConfirm={() => onDelete(item)}
+                                    onCancel={undefined}
+                                    onCanConfirm={undefined}
+                                    defaultOpen={undefined}
```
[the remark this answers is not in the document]
kubama: Yes, they are required so TS enforces them.

app/translations/en.json
```diff
@@ -307,22 +307,24 @@
...
             "addPage": "Add page",
+            "addPageGroup": "Add page group",
             "availablePages": "Available pages",
-            "button": "Create template",
+            "availablePageGroups": "Available page groups",
             "create": "Create",
             "createHeader": "Create template",
             "errors": {
-                "page": "Please select page",
+                "pageMenuItem": "Please select at least on page menu item",
```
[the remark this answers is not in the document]
kubama: Fixed

PR #1647, opened 2021-10-21
RD-3241 Added support for displaying built-in page groups

kubama:
## Description

Added support for displaying built-in page groups. Additionally:
- pages list was extended to show number of groups in which each page is used
- usage number indicator was enhanced - it's now blue only when usage count is greater than 0, it stays gray otherwise
- labels indicating where a given resource is used were changed to be clearer what they mean ('Templates' -> 'Used in templates')

## Screenshots / Videos
image
image

## Checklist

## Tests
Added test cases to cover newly added functionality.
All system tests: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1400/
Failed test rerun after the fix: https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1401/

 

## Documentation
Will be covered in separate ticket.


qooban:
Will review tomorrow. Looks promising (from the screenshot).

So the only missing part for the whole story to be finished is creating new page group from Template Management, right?

PR #1663, opened 2021-11-11
RD-3330 Fix negative number in an input

leon-good-life:
## Description
A link to the ticket in Jira: https://cloudifysource.atlassian.net/browse/RD-3330

### Bug description
**The bug is:** Cannot add negative numbers.
**Steps to reproduce:** type "--", then remove it, then type a number, remove it, then try to type a negative number. 
**Expected result:** negative number is in the input.
 **Actual result:** sometimes empty input and sometimes a number without the minus sign.

### Solution description
I added a code that checks that the input value is not NaN. 

## Screenshots / Videos
In the following video I show "before fix" and "after fix". In "before fix" when I type "-" after numbers it deletes everything.  And negative numbers cannot be added. But, after the fix, "-" can be added without problems.

https://user-images.githubusercontent.com/17104212/141311538-780bf5ad-d213-4016-b467-6f7b14751a13.mov

## Checklist

## Tests
System tests:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1481/pipeline

 

## Documentation
N/A

### DIFF
--- a/widgets/common/src/InputsUtils.tsx
+++ b/widgets/common/src/InputsUtils.tsx
@@ -366,6 +366,16 @@ class InputsUtils {
         }
     }
 
+    static evaluateValue(input, inputsState, dataType) {
+        if (input.type === 'integer' && Number.isNaN(inputsState[input.name])) {
+            return '';
+        }
+        if (_.isUndefined(inputsState[input.name])) {
+            return InputsUtils.getInputFieldInitialValue(input.default, input.type, dataType);
+        }
+        return inputsState[input.name];
+    }
+
     static getInputFields(inputs, onChange, inputsState, errorsState, dataTypes) {
         return _(inputs)
             .map((input, name) => ({ name, ...input }))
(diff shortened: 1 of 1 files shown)


widgets/common/src/InputsUtils.tsx
```diff
@@ -366,16 +366,24 @@ class InputsUtils {
         }
     }
 
+    static evaluateValue(input, inputsState, dataType) {
+        if (input.type === 'integer' && Number.isNaN(inputsState[input.name])) {
```
qooban, line 370: Shouldn't we also take care about floating numbers?
```suggestion
        if ((input.type === 'integer' || input.type === 'float') && Number.isNaN(inputsState[input.name])) {
```

Ref.: See possible input types in https://docs.cloudify.co/latest/developer/blueprints/spec-inputs/ - Schema -> Type

widgets/common/src/InputsUtils.tsx
```diff
@@ -366,16 +366,24 @@ class InputsUtils {
         }
     }
 
+    static evaluateValue(input, inputsState, dataType) {
```
qooban, line 369: Optional:

I'm wondering if the function name is accurate. Maybe `normalizeValue` would be more clear in that case?

qooban (commented):
Tested in the UI, seems working fine. Added some comments.

BTW If the fix is in cloudify-stage and not in cloudify-ui-components, then when you'll be working on https://cloudifysource.atlassian.net/browse/RD-3496
you should revert https://github.com/cloudify-cosmo/cloudify-ui-components/pull/95 .

qooban (approved):

PR #1664, opened 2021-11-15
RD-3192 Added support for sidebar collapsing

kubama:
## Description
Added support for sidebar collapsing

## Screenshots / Videos

CPT2111151339-410x318

## Checklist

## Tests
TODO - I'll provide necessary changes to existing tests

 

## Documentation
Will be covered in separate ticket.


app/components/templateManagement/pages/PageManagement.tsx
```diff
@@ -225,7 +225,8 @@ export default function PageManagement({ pageId, isEditMode = false }: PageManag
                 </Sidebar>
             </div>
 
-            <div className="page">
+            {/* Set the margin to default sidebar width (defined by Semantic UI) */}
+            <div className="page" style={{ marginLeft: '13rem' }}>
```
qooban, line 229: Maybe use `expandedSidebarWidth` here? The comment would be not necessary then.

test/cypress/support/commands.ts
```diff
@@ -328,7 +328,9 @@ const commands = {
             .editWidgetConfiguration(widgetId, noop),
     refreshPage: (disableGettingStarted = true) => {
         mockGettingStarted(!disableGettingStarted);
-        return cy.get('.pageMenuItem.active').click({ force: true });
+        cy.get('.pageMenuItem.active').click({ force: true });
+        // Collapse the sidebar
+        return cy.get('.headerBar').click();
```
qooban, line 333: We will have to change that soon as top bar (`headerBar`) will be removed. Maybe we can find better method at once? Maybe `.breadcrumb`?

app/components/sidebar/TenantSelection.tsx
```diff
@@ -53,7 +53,7 @@ const TenantSelection: FunctionComponent = () => {
 
     return (
         <Dropdown trigger={tenantMenuTrigger} onClose={clearSearch} pointing="left" icon={null} fluid>
-            <Dropdown.Menu>
+            <Dropdown.Menu style={{ margin: 0 }}>
```
qooban, line 56: Why do we need this?
(IMHO the space between sidebar and popup make sense)

qooban (commented):
Looks good 👍🏻  Added some small comments. 

Tested UI and have one question: Will it be possible to hide dropdown arrow (in group menu items) when sidebar is expanding? Maybe by manipulating opacity during animation it would be possible? 
(not critical at all, may also be considered when doing Sidebar styling - RD-3482)

Great that you managed to do some migrations to TS and move one file from containers to components 👍🏻

app/components/sidebar/TenantSelection.tsx
kubama: The space was causing `mouseLeave` and then `mouseEnter` events to be triggered when moving mouse pointer from the sidebar to the popup causing the sidebar to shake. So this is a workaround to the issue.

app/components/templateManagement/pages/PageManagement.tsx
kubama: Done

kubama:
>    * when sidebar gets expanded again - maybe selected group should be also expanded?

This is a good point and IMHO it would be a desired behaviour. But if so, then why collapse the group at all?

test/cypress/support/commands.ts
kubama: Ok, changed.

app/components/sidebar/TenantSelection.tsx
qooban: I see. In that case, let's leave that workaround.

qooban:
> > * when sidebar gets expanded again - maybe selected group should be also expanded?
> 
> This is a good point and IMHO it would be a desired behavior. But if so, then why collapse the group at all?
@kubama If it's simple to make icons from a group visible when sidebar is collapsed, then let's try without group collapsing and see how it looks (you can share video/screenshot). I was expecting it might be challenging as those icons are indented. 
There are no mockups for such case, so maybe you can also ask Ilan and Alona about preferred approach?

kubama:
> let's try without group collapsing and see how it looks (you can share video/screenshot)

CPT2111161051-474x388
WDYT @qooban ?

kubama:
> Will it be possible to hide dropdown arrow (in group menu items) when sidebar is expanding? Maybe by manipulating opacity during animation it would be possible?

This is done

qooban:
> > let's try without group collapsing and see how it looks (you can share video/screenshot)
> 
> CPT2111161051-474x388 WDYT @qooban ?

Looks very good 👍🏻  I like it. I think we can keep it. 
With that solution the icon will be highlighted when page item from a group is selected, right?

kubama:
> With that solution the icon will be highlighted when page item from a group is selected, right?

Yes

test/cypress/support/commands.ts
```diff
@@ -209,13 +209,17 @@ const commands = {
...
         if (id) {
             cy.location('pathname').should('be.equal', `/console/page/${id}`);
         }
         return cy.waitUntilPageLoaded();
     },
-    visitTestPage: () => cy.clickPageMenuItem(testPageName),
+    visitTestPage: () => {
+        cy.clickPageMenuItem(testPageName);
+        // Collapse the sidebar
+        cy.get('.breadcrumb').click();
```
qooban, line 221: 1. As there are two uses already - probably we can extract it to a function?
2. As we started returning values in custom Cypress commands, I think it'd be better to return it the same way as  you did in `refreshPage`.

qooban (commented):
Thank you for applying UI changes 👍🏻 

Just one comment related to Cypress tests.

test/cypress/support/commands.ts
kubama: Done

qooban (approved):
Perfect. LGTM 👍🏻

PR #1665, opened 2021-11-16
RD-3216 Hide DrilldownButtons instead of disabling them

NorbertAmartus:
## Description
With this PR comes the change with which DrilldownButtons would be hidden instead of being disabled.

### Before 
Before
### After 
After

## Checklist

## Tests

 

Changes were tested manually and with the usage of scripts declared inside the `package.json` file:
- `e2e:ci`
- `test:frontend`

Additionally, the [system test](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1522/) has been executed

## Documentation

Created [PR](https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2033) to update the documentation inside the [docs.getcloudify.org repository](https://github.com/cloudify-cosmo/docs.getcloudify.org)


widgets/common/src/deploymentsView/detailsPane/header.scss
```diff
@@ -4,17 +4,7 @@
...
-        align-items: baseline;
-        flex-wrap: wrap;
-
-        // NOTE: adds a gap between the tabs headers when the items in the header wrap
-        margin-bottom: 10px;
-    }
-
-    .detailsPaneWidgets {
-        overflow: auto;
+    .centeredButtons {
```
NorbertAmartus, line 7: NOTE: I've added here an additional class, because I couldn't use the `css` method imported from `styled-components`
Everytime that this method tried to be executed, the application was crashing.

I did some research about that and as far as I can tell, this issue is stricty related to the React version.   
Because of the fact, that the change of the core React packages version may take some time, I've created conditional className which is being used inside the  `drilldownButtons/index.ts` file

package.json
```diff
@@ -79,6 +79,7 @@
     "react": "^16.11.0",
     "react-dom": "^16.11.0",
     "react-grid-layout": "^0.16.6",
+    "react-is": "^17.0.2",
```
NorbertAmartus, line 82: I've added this package, because `styled-components` was throwing errors inside the console, which indicates that this package has been missing.

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/SubdeploymentDrilldownButton.tsx
```diff
@@ -40,25 +42,19 @@ const SubdeploymentDrilldownButton: FunctionComponent<SubdeploymentDrilldownButt
         );
     };
 
-    const { Icon } = Stage.Basic;
```
qooban, line 43: Could you explain what was your intention by moving that const out of the component?
(just a question, I'd like to know your reasoning here)

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/index.tsx
```diff
@@ -1,8 +1,8 @@
-import type { FunctionComponent } from 'react';
+import { FunctionComponent } from 'react';
```
qooban, line 1: Any reason for converting `import type` into `import`? 🙂 
(suspecting auto-fix from IDE)

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/common.ts
```diff
@@ -1,6 +1,10 @@
 import { i18nDrillDownPrefix } from '../../common';
+import { SubdeploymentsResult } from './subdeployments-result';
 
 const i18nDrillDownButtonsPrefix = `${i18nDrillDownPrefix}.buttons`;
 
-// eslint-disable-next-line import/prefer-default-export
 export const tDrillDownButtons = Stage.Utils.getT(i18nDrillDownButtonsPrefix);
+
+export const shouldDisplaySubdeploymentButton = (buttonsData: SubdeploymentsResult) => {
+    return !buttonsData.loading && buttonsData.count > 0;
```
qooban, line 9: Optional:

Maybe in that case it would be OK to simply name the variable `result`?
```suggestion
export const shouldDisplaySubdeploymentButton = (result: SubdeploymentsResult) => {
    return !result.loading && result.count > 0;
```

widgets/common/src/deploymentsView/detailsPane/header.scss
qooban: Interesting 🙂 If I had more time, I would have dug into more details, but right now I assume you've chosen the right path 👍🏻

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -1024,7 +1010,7 @@ describe('Deployments View widget', () => {
             getDeploymentsViewMap().should('not.exist');
         });
 
-        it('should render markers for various deployment states', () => {
+        it.only('should render markers for various deployment states', () => {
```
qooban, line 1013: Possible leftover from development phase 🙂 

```suggestion
        it('should render markers for various deployment states', () => {
```

widgets/common/src/deploymentsView/detailsPane/header.tsx
```diff
@@ -2,8 +2,19 @@ import { noop } from 'lodash';
 import { ComponentProps, FunctionComponent, ReactNode, useMemo, useRef } from 'react';
 
 import './header.scss';
+import styled from 'styled-components';
 import { Deployment } from '../types';
 
+const HeaderPaneWrapper = styled.div`
```
qooban, line 8: Optional:

I'm not sure about the name used here. As `DetailsPane` contains `DetailsPaneHeader` (and not `DetailsPaneHeaderPane`), maybe more accurate would be to name it just `HeaderWrapper`?

widgets/common/src/deploymentsView/detailsPane/widgets.tsx
```diff
@@ -1,14 +1,18 @@
 import type { FunctionComponent } from 'react';
-
+import styled from 'styled-components';
 import pageLayout from './pageLayout.json';
 
+const WidgetsPaneWrapper = styled.div`
```
qooban, line 5: Optional:

Similar comment to the one added to `widgets/common/src/deploymentsView/detailsPane/header.tsx`.

As `DetailsPane` contains `DetailsPaneWidgets` (and not `DetailsPaneWidgetsPane`), maybe more accurate would be to name it just `WidgetsWrapper`?

qooban (commented):
Code looks good in general 👍🏻 I have also tested UI, seems working good as well 👍🏻 

I added few comments. Mostly related to naming conventions, small code style improvements.

Reg.
> Changes were tested manually and with the usage of scripts declared inside the package.json file:

I don't expect that the change done within this PR will brake anything outside of `deployments_view_spec`, but it is a good practice to always run full system tests run on a branch (https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/build?delay=0sec) and provide a ling in `Tests` section.

Reg. CI - seems like `npm run size` failed. Please check it out. Maybe addition of `react-is` package to the dependencies caused it. You can compare to `npm run size` executed on `master` branch. You can also follow https://github.com/cloudify-cosmo/cloudify-stage/blob/master/test/README.md#bundle-size-checks - "Bundle size checks" section to check what 'caused the difference. Once you do all checks and ensure that it was expected increase, you can just update `.size-limit` file.

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/SubdeploymentDrilldownButton.tsx
NorbertAmartus: Sure!

When declaring `consts` inside the React Component, we are risking few minor performance issues.
Those issues are related to the fact, that upon the component rerender those variables would be redefined, which would take some of the user's machine resources

That's why I've relocated the const, which is not related to any props/context passed down to the component, so that the memory for it would be allocated just once.

Additionally, using this method we are additionally saving additional resources due to the possibility of using multiple components, which are now having the reference to the same variable (again, they are not redefining them when they'd be rerendering 😁)

As we can tell, the "performance boost" that we are getting from this relocation, won't be noticable from the user perspective, but it's always a little boost, which could be beneficial in the future :)

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/index.tsx
NorbertAmartus: Yup, that wasn't intentional
IDE was boycotting my PR 😁

I will add that declaration along with other changes related to the discussion

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/common.ts
NorbertAmartus: I thought about it, but my reasoning was that the context of the word `result` may not be known for the programmer, and that's why I've added word which was more related to the button itself ... but I forgot about the typing, which gives full context to the programmer 😁

So, thanks for a nice catch!
I'll add that, along with the incoming changes

widgets/common/src/deploymentsView/detailsPane/header.tsx
NorbertAmartus: Hmm, so here I'd like to get your opinion on something, because I may be doing some unnecessary overthinking 😁

So, when we have the component:
```
<HeaderPaneWrapper> -----------------------------------------------> Wrapper of the whole component
    <div style={{ marginRight: '1rem', marginBottom: '1rem' }}> -----> Wrapper of the Header
        <Header>{displayName}</Header>  ----------------------------> Header
    </div>
    {drilldownButtons}
    <Widget
        widget={deploymentActionButtonsWidgetDescription}
        isEditMode={false}
        onWidgetRemoved={noop}
        onWidgetUpdated={noop}
        standalone
    />
</HeaderPaneWrapper>
```

It has some logical parts inside of it (I've tried to comment them out, so that it may be a little bit clearer)
And because of those parts, I thought that putting `Wrapper` after the name of the component (`HeaderPane`) would be a correct choice    

But I may be overthinking to much xD
What do you think?

With the `DetailsPane` we don't have a similar situation:
```
<div className="detailsPane">
    <DetailsPaneHeader
        deployment={deployment}
        drilldownButtons={
            <DrilldownButtons
                deployment={deployment}
                drillDown={drillDown}
                toolbox={toolbox}
                refetchInterval={widget.configuration.customPollingTime * 1000}
                mapOpen={mapOpen}
            />
        }
    />
    <DetailsPaneWidgets />
</div> 
```

NorbertAmartus: Additionally I could add `HeaderWrapper` for the div, which currently have some inline styling in it.
But that's just a side note 😁

Let me know what do you think

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/SubdeploymentDrilldownButton.tsx
qooban: I was expecting it was such motivation. OK, clear! 👍🏻

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/common.ts
qooban: Yes, I think the type name is making the change here, so it is self explanatory 🙂 Thanks for addressing this comment, I'm resolving the discussion.

widgets/common/src/deploymentsView/detailsPane/header.tsx
qooban: > And because of those parts, I thought that putting Wrapper after the name of the component (HeaderPane) would be a correct choice

Hm... But the name of the component is actually `DetailsPaneHeader` which I understand as the header of the details pane. So it's not the same as `HeaderPane`. That is my understanding 🙂 From that comes my concern about the name.

Anyways, I marked the comment as optional, so it's not a blocker for me, especially that styled component is not exported.

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/index.tsx
qooban: Thanks, I see you fixed it already, so I'm resolving this discussion.

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/SubdeploymentDrilldownButton.tsx
```diff
@@ -1,5 +1,4 @@
-import type { FunctionComponent } from 'react';
-
+import { FunctionComponent, useMemo } from 'react';
```
qooban, line 1: There's similar issue with `import type` as we discussed within `drilldownButtons/index.tsx` 🙂

NorbertAmartus: Yup, nice catch!

I'll split this import into the 2 different:
```
import { useMemo } from 'react'; 
```

And additionally I'll get the look on the configuration on my IDE, so that this problem wouldn't happen again ;)

widgets/common/src/deploymentsView/detailsPane/header.tsx
NorbertAmartus: Hmm, it makes sense 
Maybe the comment is optional, but still it's a comment that we can think about 😁

I'll add that modification with the next commit

NorbertAmartus: Additionally, I've look upon the imports of the file and I've discovered the line bellow:
```
import { ComponentProps, FunctionComponent, ReactNode, useMemo, useRef } from 'react';
```

There we are importing some of the types as the "standard" import.
I'll split this line into the:

```
import type { FunctionComponent, ReactNode, ComponentProps } from 'react';
import { useMemo, useRef } from 'react';
```

To be consistent with previously commented discussions ;)

widgets/common/src/deploymentsView/detailsPane/widgets.tsx
NorbertAmartus: As suggestion from (this conversation)[https://github.com/cloudify-cosmo/cloudify-stage/pull/1665#discussion_r751162285] has been applied to code, I did the same here :)

.size-limit
```diff
@@ -16,13 +16,19 @@
         "limit": "3.13 MB",
         "gzip": false,
         "running": false,
-        "path": ["dist/appData/widgets", "!dist/appData/widgets/**/*.gz"]
+        "path": [
+            "dist/appData/widgets",
+            "!dist/appData/widgets/**/*.gz"
+        ]
```
qooban, line 22: Auto-correction from the IDE? :-)

qooban (approved):
Looks good! 👍🏻 

Are you planning to add commit with updated `README.md` for Deployments View widget (after merging https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2033) within this PR or in separate one?

widgets/common/src/deploymentsView/detailsPane/drilldownButtons/SubdeploymentDrilldownButton.tsx
qooban: Thanks for addressing it 👍🏻

widgets/common/src/deploymentsView/detailsPane/header.tsx
qooban: Thanks for addressing it 👍🏻

.size-limit
NorbertAmartus: > Auto-correction from the IDE? :-)

Yup 😁

> Are you planning to add commit with updated README.md for Deployments View widget (after merging cloudify-cosmo/docs.getcloudify.org#2033) within this PR or in separate one?

I think that the seperate commit would be the way to go in this case (I didn't received notification about the review comments, and now I'll try to resolve them 😁)

PR #1676, opened 2021-11-18
RD-3554, RD-3549, RD-3492 - Modified dashboard layout

NorbertAmartus:
## Description

Modifications are mostly related to the modification of the dashboard layout.
Because of this, changes are mostly related to the configuration file.

### Before
Before

### After
After

## Checklist

## Tests

 

 [Link to the system tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1579)

## Documentation

The update of the documentation is blocked by the fact, that the [Dashboard page documentation](https://docs.cloudify.co/latest/working_with/console/pages/dashboard-page/) is up to date to the version `5.1` of the Cloudify Premium   
Problem related to the outdated documentation was mentioned inside [this jira ticket](https://cloudifysource.atlassian.net/browse/RD-2504)


NorbertAmartus:
Just a note:
One of the files with test cases was depending on the previous layout presented on the dashboard
Because of that I removed that file (I've tested all of it's test scenarios, to see if any of them was still relevant).

templates/pages/adminDash.json
```diff
@@ -1,148 +1,142 @@
 {
-  "name": "Dashboard",
```
qooban, line 2: I see lots of whitespace-related changes in this file. Are they intentional?

test/cypress/integration/widgets/sites_map_spec.ts
```diff
@@ -1,71 +0,0 @@
-describe('Sites Map', () => {
```
qooban, line 1: I think we should keep that test. The main thing to change here to make it work is to use `usePageMock`. That will allow you to mock a page to show just a single widget instead of using default templates.

You can easily find a lot of other Cypress tests using that custom command.

qooban (commented):
Added two comments. From the screenshots you shared it looks good, but I didn't tested it in my env, yet.

Reg.
> I added proper labels to this PR.

I see you had confusion about labels, please add a point to UI weekly doc, we'll discuss what to do with it (maybe we'll need `enhancement` label to be officially added to the allowed set).

test/cypress/integration/widgets/sites_map_spec.ts
NorbertAmartus: Okay, seems like a great idea! :)

templates/pages/adminDash.json
NorbertAmartus: Unfortunately not :/
I'll remove those inditation changes and additionally I'll try to modify settings in the IDE, that may cause this sort of behaviour

NorbertAmartus: Previous whitespaces has been restored :)

test/cypress/integration/widgets/sites_map_spec.ts
NorbertAmartus: Test file has been restored and adjusted to your suggestions

test/cypress/integration/widgets/sites_map_spec.ts
```diff
@@ -1,12 +1,14 @@
 describe('Sites Map', () => {
     const testSite = { name: 'Tel-Aviv', location: '32.079991, 34.767291' };
+    const mapSiteName = 'mapSite';
 
-    function refreshPage() {
-        cy.clickPageMenuItem('Dashboard');
-    }
+    const navigateToMapSite = () => {
+        cy.clickPageMenuItem(mapSiteName);
+    };
```
qooban, line 7: You can of course use any value for page name, but "Map site" sounds a bit confusing for me. Your intention (I think) was to create a page for "Sites Map" widget (which stands for a map showing all sites) and not to map a site 🙂 

So I suggest to change `mapSiteName` variable and value, as well as `navigateToMapSite` function name.

qooban (commented):
Nicely done 👍🏻 Just one comment about variable naming in test.

Regarding system tests - I see 4 are failing in the linked system tests run. Have you checked what is the reason? Also, it seems it was executed before the last changes in this PR, so probably worth to trigger a re-run.

NorbertAmartus:
> Nicely done 👍🏻 Just one comment about variable naming in test.
> 
> Regarding system tests - I see 4 are failing in the linked system tests run. Have you checked what is the reason? Also, it seems it was executed before the last changes in this PR, so probably worth to trigger a re-run.

After merging to master, the system test started to fail, so I assumed that the other freshly merged PR was causing this issue.
I'll restart the pipeline and see if I can get the similar log output, as the lastly merged PR, so that I can confirm the theory above.

test/cypress/integration/widgets/sites_map_spec.ts
NorbertAmartus: My intention was to indicate that this name is related to the page with the map widget inside of it
And because of that, there appeared an idea with the `mapSiteName` -> name of the site, that has a "map"

But as it confuses you, it may confuse the other coworkers, so I'll try to redeclare it as a 
```javascript
const siteWithSitesMapWidget = {
  name: 'siteWithSitesMap'
}
```
so that the name would be unambiguous ;)

qooban: Hmm... I'm still confused. Using "site" term when you are referring actually to a "page" might not be the best idea here.
Especially that we are dealing with sites (a type of Cloudify resource) in this test.

Also, AFAIR, you can use spaces when setting page name, you don't have to use camelcase there, it's not page ID :-)

NorbertAmartus: Oh, you're right
I'll try to change my naming methology, to use `page` instead of `page`, because of the fact that the applications has it's own context for the name `page` 😁

Having that in mind, I'll recreate variable, to be:
```
sitesMapWidgetPageName = 'Sites map widget page'
```

And thanks for an additionall tip 😁

NorbertAmartus:
Tests started to pass through, so there was some issue related to them (and possible other branches) which has been resolved ;)

qooban (approved):

PR #1677, opened 2021-11-19
RD-3442 Moved health indicator from top to side bar

kubama:
## Description
Moved health indicator from top to side bar

## Screenshots / Videos
image

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1536/

 

## Documentation
Will be covered in separate ticket


app/components/sidebar/Health.tsx
```diff
@@ -0,0 +1,43 @@
...
+                <SideBarItemIcon name="heartbeat" />
+                {dotColor && <Icon corner="top right" name="circle" color={dotColor} />}
+            </Icon.Group>
+            {t('health')}
+        </SideBarItem>
+    );
+
+    return (
+        <Dropdown trigger={trigger} pointing="left" icon={null} fluid onOpen={() => dispatch(getClusterStatus())}>
+            <Dropdown.Menu style={{ padding: 5, margin: 0, height: 267 }}>
```
qooban: I'm not sure about hardcoding that height. Is there any other way to set the height? 

Also we are using `Dropdown` here probably only to keep the styling as semantically we don't provide any dropdown-like functionalities for health indicator? If so, then my question is: will `Popup` work here?

app/components/sidebar/SystemMenu.tsx
```diff
@@ -1,17 +1,24 @@
...
 
 const SystemMenu: FunctionComponent<SystemMenuProps> = ({ onAboutModalOpen }) => {
+    const mode = useSelector((state: ReduxState) => state.config.mode);
+
     return (
         <>
-            <HelpMenu onAboutModalOpen={onAboutModalOpen} />
             <TenantSelection />
+            <HelpMenu onAboutModalOpen={onAboutModalOpen} />
+            {mode !== Consts.MODE_CUSTOMER && <Health />}
```
qooban: General:

I'm wondering is anybody using `customer` mode for stage-backend. I'd personally like to remove it, but I'm not sure we can. Something to discuss with mgmt.

test/cypress/integration/cluster_status_spec.ts
```diff
@@ -1,4 +1,3 @@
-// @ts-nocheck File not migrated fully to TS
```
qooban, line 1: Great 🙂 👍🏻

test/cypress/integration/widgets/button_link_spec.ts
```diff
@@ -35,35 +35,35 @@ describe('Button link widget', () => {
         it('color', () => {
             const color = 'violet';
             cy.setSearchableDropdownConfigurationField(widgetId, 'Color', color);
-            cy.get('button').should('have.class', color);
+            cy.get('.page button').should('have.class', color);
```
qooban, line 38: Hmm... Maybe extract a function to get a button from Button Link widget?
Probably we can use `widgetId` as part of the CSS path, something like "`${widgetId}widget button`"?

test/cypress/integration/widgets/events_spec.ts
```diff
@@ -13,7 +13,7 @@ describe('Events/logs widget', () => {
             .mockLogin();
     });
     it('should show deployment ID and display name', () => {
-        cy.get('table')
+        cy.get('.page table')
```
qooban, line 16: Maybe we can have a custom command to execute commands within specific widget? 

Something like: 
```
function withinWidget(widgetId: string, fn: (currentSubject: JQuery<HTMLElement>) => void) {
    return cy.get(`${widgetId}widget`).within(fn);
}
```

It relates to other such changes.

qooban (commented):
Added some comments. Did not review the whole PR.

I see you are providing more changes, so I'll wait for information from you that it is finalized, then I'll move on with the review.

app/components/sidebar/Health.tsx
```diff
@@ -0,0 +1,43 @@
...
+                <SystemServicesStatus />
+            </Dropdown.Menu>
+        </Dropdown>
+    );
+};
+
+export default Help;
```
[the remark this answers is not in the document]
kubama: This is now fixed

app/components/sidebar/Health.tsx
kubama: > I'm not sure about hardcoding that height. Is there any other way to set the height?

We could modify `ClusterServicesOverview` so it is rendered also when fetching data, this would effectively make the height of the dropdown menu fixed.

> Also we are using Dropdown here probably only to keep the styling as semantically we don't provide any dropdown-like functionalities for health indicator? If so, then my question is: will Popup work here?

Probably not, or at least not easily. As opposed to `Dropdown`, `Popup` is not rendered as sidebar child component in terms of DOM tree, so moving mouse pointer over it triggers `mouseleave` for the sidebar, which causes it to collapse - I don't think it is desired behaviour.

app/components/sidebar/Health.tsx
```diff
@@ -0,0 +1,43 @@
...
+const Help: FunctionComponent = () => {
+    const dispatch = useDispatch();
+    const systemStatus = useSelector((state: ReduxState) => state.manager.clusterStatus.status);
+    const dotColor = dotColors[systemStatus];
+
+    const trigger = (
+        <SideBarItem>
+            <Icon.Group>
+                <SideBarItemIcon name="heartbeat" />
+                {dotColor && <Icon corner="top right" name="circle" color={dotColor} />}
```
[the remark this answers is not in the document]
kubama: image

test/cypress/integration/widgets/button_link_spec.ts
kubama: Done, added `getWidget` command

test/cypress/integration/widgets/events_spec.ts
kubama: I added `getWidget` command, looks like it's more common to use `getWidget().find()` combination rather than `within`, so I would stay with just `getWidget`.

app/components/sidebar/Health.tsx
leon-good-life: Inline style object causing the component to re-render, because it creates a new instance of an object every time, and React re-render the component whenever the props changing. Source: https://medium.com/welldone-software/why-did-you-render-mr-big-pure-react-component-2a36dd86996f

app/components/sidebar/HealthIndicator.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <SideBarDropdownItem
+            icon={
+                <Icon.Group>
+                    <SideBarItemIcon name="heartbeat" />
+                    {dotColor && <Icon corner="top right" name="circle" color={dotColor} />}
+                </Icon.Group>
+            }
+            label={t('health')}
+            onOpen={() => dispatch(getClusterStatus())}
+            style={{ padding: 5, height: 267 }}
```
leon-good-life, line 34: Inline style object causing the component to re-render, because it creates a new instance of an object every time, and React re-render the component whenever the props changing. Source: https://medium.com/welldone-software/why-did-you-render-mr-big-pure-react-component-2a36dd86996f

leon-good-life (commented):
I added a comment regarding performance

app/components/sidebar/HealthIndicator.tsx
kubama: I don't think we are having any noticeable performance loss by using this pattern, but since cost is close to nothing I extracted the style to module level const.

app/components/sidebar/Health.tsx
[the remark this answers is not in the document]
qooban: I think we will need some small adjustments (for the placement of the top right corner icon), but that will probably be part of RD-3482.

app/components/sidebar/Health.tsx
qooban: > We could modify `ClusterServicesOverview` so it is rendered also when fetching data, this would effectively make the height of the dropdown menu fixed.

Seems like an enhancement in which we could introduce https://react.semantic-ui.com/elements/placeholder/ . I don't want to prolong this PR, but if you find it quick maybe worth to try it out.

qooban (dismissed):
There is a room for more improvements, but main goal of RD-3442 was achieved IMHO, so I'm approving.

Did all system tests passed?

app/components/sidebar/Health.tsx
kubama: I realized setting a fixed height for the content may actually be an issue as it assumes that there are always 3 services displayed. So I removed the `height` prop and instead used `LoadingOverlay` when fetching - it's better suited than the placeholder component is it preserves the height of the underlaying component.

kubama:
> Did all system tests passed?

All green

qooban (approved):
LGTM 👍🏻

PR #1682, opened 2021-11-25
RD-3477 Disabled LDAP functionality in the UserGroups widget, when LDAP is not being enabled

NorbertAmartus:
## Description

The changes will add the functionality to display the `LDAP group` column, displayed by UsersGroup widget table, depending the status of the `isLdapEnable` flag which is being loaded during the application startup

Additionally I've added some internal changes related to [this ticket](https://cloudifysource.atlassian.net/browse/RD-2842), across the whole widget.

### Before

before - 1
before - 2

### After (when the LDAP is not being enabled)
after - 1
after - 2

## Checklist

## Tests

 

[System tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1562/) has been executed
Additionally I've run a locally cypress test (regarding the modified widget: `user_groups_spec.ts`) along with the `frontend` tests
## Documentation

Documentation will be updated with the usage of [this PR](https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2040)
I'll add a proper commit, updating the documentation inside this repository


test/cypress/integration/widgets/user_groups_spec.ts
```diff
@@ -1,8 +1,19 @@
...
     const groupName = 'user_groups_test';
+    const ldapGroupColumnName = 'LDAP group';
+    const userGroupsWidgetName = 'userGroups';
+    const setLdapAvailibility = (isEnabled: boolean) => {
+        const ldapResponse = isEnabled ? 'enabled' : 'disabled';
+        cy.intercept('GET', '/console/sp/ldap', ldapResponse);
+    };
+    const reloadPage = () => {
+        cy.usePageMock(userGroupsWidgetName).mockLogin();
+    };
```
qooban, line 12: 1. You can make it to be one-liner 🙂 
```suggestion
    const reloadPage = () => cy.usePageMock(userGroupsWidgetName).mockLogin();
```

2. The name of the function suggests that we just reload a page, but from the implementation it seems that we are logging in to the application. 
3. Maybe just `cy.reload` would be sufficient for our case? (assuming that `GET /console/sp/ldap` is called on page reload)

test/cypress/integration/widgets/user_groups_spec.ts
```diff
@@ -53,4 +64,16 @@ describe('User group management widget', () => {
...
+        setLdapAvailibility(true);
+        reloadPage();
+        cy.contains(ldapGroupColumnName);
+    });
+
+    it('should hide LDAP group column when LDAP is disabled', () => {
+        setLdapAvailibility(false);
+        reloadPage();
+        cy.contains(ldapGroupColumnName).should('not.exist');
+    });
```
qooban, line 78: Optional:

As `reloadPage` and `setLdapAvailability` functions are always executed one after the other, it seems to me that they can be merged or another function can be created to call both of them. WDYT?

test/cypress/integration/widgets/user_groups_spec.ts
```diff
@@ -1,8 +1,19 @@
...
+        cy.intercept('GET', '/console/sp/ldap', ldapResponse);
+    };
+    const reloadPage = () => {
+        cy.usePageMock(userGroupsWidgetName).mockLogin();
+    };
 
-    before(() => cy.activate('valid_trial_license').usePageMock('userGroups').mockLogin().deleteUserGroup(groupName));
+    before(() => {
+        cy.activate('valid_trial_license').usePageMock(userGroupsWidgetName).mockLogin().deleteUserGroup(groupName);
+    });
```
qooban, line 16: What was the reason for changing return value in this case?

widgets/userGroups/src/CreateModal.tsx
```diff
@@ -3,9 +3,10 @@
  * Created by jakubniezgoda on 03/02/2017.
  */
```
qooban, line 4: You can remove it at once 🙂

widgets/userGroups/src/CreateModal.tsx
```diff
@@ -3,9 +3,10 @@
...
 
+import { UnsafelyTypedFormField } from '../../../app/components/basic';
```
qooban, line 6: We shouldn't use imports from application framework in widgets.

Use `const { UnsafelyTypedFormField } = Stage.Basic;`

(this comment applies to other places as well)

widgets/userGroups/src/CreateModal.tsx
```diff
@@ -52,18 +53,15 @@ export default function CreateModal({ toolbox }) {
 
             <Modal.Content>
                 <Form loading={isLoading} errors={errors} onErrorsDismiss={clearErrors}>
-                    <Form.Field error={errors.groupName}>
-                        <Form.Input name="groupName" placeholder="Group name" value={groupName} onChange={setInput} />
-                    </Form.Field>
+                    <UnsafelyTypedFormField error={errors.groupName} label="Group name">
```
qooban, line 56: What do you think about adding translation for this file at once?
(and maybe to other modified files as well depending on the final size of those changes)

Ref.: 
1. [Ongoing Refactors in Cloudify UI - Adding translation support in new widgets](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Adding-translation-support-in-new-widgets) - I think we need to extend for existing widgets as well (we agreed earlier this year to do that already, but the wording is not covering existing widgets, I'll update it and add a point for UI weekly)
2. RD-438 - that's the long term goal
3. [UI Translation Files Guidelines](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/1875083446/UI+Translation+Files+Guidelines)

widgets/userGroups/src/UserGroupsTable.tsx
```diff
@@ -199,7 +197,7 @@ export default class UserGroupsTable extends React.Component {
                     noDataMessage={NO_DATA_MESSAGE}
                 >
                     <DataTable.Column label="Group" name="name" width="35%" />
-                    <DataTable.Column label="LDAP group" name="ldap_dn" width="20%" />
```
qooban, line 202: So when there's no LDAP group column, the width of other columns looks fine without making any additional changes?

widgets/userGroups/src/UserGroupsTable.tsx
```diff
@@ -1,7 +1,5 @@
 // @ts-nocheck File not migrated fully to TS
-/**
- * Created by jakubniezgoda on 03/02/2017.
- */
+import { connect, ConnectedProps } from 'react-redux';
```
qooban, line 2: Do we need that import?

Please note that we should be careful with importing anything "external" in widgets as they are treated as separate bundles. `Stage` global is the main bridge between app framework and widgets.

qooban (commented):
Tested UI in non-LDAP mode, looks good 👍🏻  Added some comments. 

I don't see any file converted fully to TS, so please try to convert at least one 🙂 
(as we agreed in https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Migrating-files-to-TypeScript)
That's the best way to gradually move our codebase to be fully TypeScripted.

Thanks for addressing RD-2842 in User Groups widget 🎉

app/actions/managers.ts
```diff
@@ -99,10 +99,10 @@ export function getUserData() {
         });
 }
 
-function responseLdap(isLdap) {
+function responseLdap(isLdapEnable) {
```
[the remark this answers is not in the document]
NorbertAmartus: I think, that's the great idea!
I'm embarassed, that I didin't though of it, while adding this change 😁

Good catch!

test/cypress/integration/widgets/user_groups_spec.ts
```diff
@@ -1,8 +1,19 @@
 // @ts-nocheck File not migrated fully to TS
 describe('User group management widget', () => {
     const groupName = 'user_groups_test';
+    const ldapGroupColumnName = 'LDAP group';
+    const userGroupsWidgetName = 'userGroups';
```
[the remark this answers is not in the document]
NorbertAmartus: When it comes to the changes that improve the consistency across the application, I always think that it is a good idea to add them

So if this naming convention is used in other tests, then I think that we should stick to that :)

test/cypress/integration/widgets/user_groups_spec.ts
NorbertAmartus: 1. Yup, but I don't think, that it makes an impact there
And in my opinion this approach may be more suitable, because when you're looking at the name of the variable `reloadPage`, you are able to see the part of the implementation immediately (because it's under each other)

2. Yeah, but I didn't have thought about the better name, but I'll try to figure that out once more. If you'd have any suggestion about the better name, please let mne know.

As I'm thinking about this, I think that the name `reloadPage` it's not that bad, because when it comes to it's usage in tests scenarios, we don't want to know how the "reload" is being implemented, but we just want to somehow refresh the page.

Additionally, from the tests and PR perspective, when we would add some modyfication for the "refresh" mechanism, then we can do that without having to rename all of the occurences of that function.

So eg. we want to change the "refresh" mechanism, so that it may work as you've described in the 3rd point.
Then we just modify the body of that method, without having an impact on the other parts of the application (in that case, the "other parts" would be the tests scenario), because from their perspective, they want the page to be refresh/reloaded, when the method is being invoked but the implementation of that functionality is totally transparent to them.

So we might think about that like some sort of dependency injection 😁

Let me know what you think about that

3. I tried simple `cy.reload`, but I didn't get the mocked pages, but now as I'm looking at it, I should be able to just add `usePageMock`, so that the reloaded page provides me with the pages that I need 😁

widgets/userGroups/src/CreateModal.tsx
NorbertAmartus: Okay, my IDE wasn't that smart to import it that way, but I'll try to correct that accross the modified files

widgets/userGroups/src/UserGroupsTable.tsx
NorbertAmartus: Yup, as shown on the screen attached to the PR

widgets/userGroups/src/UserGroupsTable.tsx
NorbertAmartus: Oh, okay I'll try to check that

Thanks for the information

test/cypress/integration/widgets/user_groups_spec.ts
```diff
@@ -1,8 +1,19 @@
...
+    const setLdapAvailibility = (isEnabled: boolean) => {
```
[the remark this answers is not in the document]
qooban: @NorbertAmartus Just letting you know - I have updated my previous comment.

app/actions/managers.ts
```diff
@@ -99,10 +99,10 @@ export function getUserData() {
...
     return {
         type: types.SET_LDAP,
```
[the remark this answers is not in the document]
NorbertAmartus: I don't think that would be a good idea
The reason is that from the application perspective we have the flag `isLdapEnable` (previously `isLdap`), which is currently of a `boolean` type.

So when we'd like to add a status to that name of the action, we'd need to adjust the name of the property too (to be consistent), but in both cases it wouldn't be a good practice, because:
- Word `Status` may be confusing, as it is used mostly in cases when we could have multiple string/numeric values which would indicate something different. But in our case we just have a 2 values (as far as I know), and in the application itself we have the following conditional statement:
    ```javascript
    responseLdap(data === 'enabled')
    ```
    With the usage of which, we are getting the boolean value
- From the frontend application perspective, we don't care how this value is being generated, because the only think that we are doing is making request to the backend to get the proper value.
   As we currently are using it as a `boolean`, indicating if the ldap is being enable, then I dont think, that would make much sense to label it as a status, because we may say that from the frontend side of the application we wouldn't be able to know it.
   I think, that if we'd like to make that change, then it's more logical to make it on the backend side, as it doesn't need the frontend context of it.
   Additionally, the backend is currently returning the string values (backend application is not depending on the `boolean` value), so it may be consider being a status.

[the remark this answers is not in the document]
NorbertAmartus: One more note 😁
Here is the sample of the code, related to the discussion:
```javascript
function responseLdap(isLdapEnable) {
    return {
        type: types.SET_LDAP,
        isLdapEnable
    };
}

export function getLdap() {
        const manager = new Manager(getState().manager);
        manager.doGet('/ldap').then(data => dispatch(responseLdap(data === 'enabled')));
    };
}
```

As you can see, the `responseLdap` action is being given a `boolean` value, so the only place, when we would consider using the `status` is the line when we are using string `'enabled'`

To make the code better, we could implement the following enum:
```typescript
cosnt enum LdapStatus {
    Enabled = 'enabled',
    Disabled = 'disabled'
}
```

Which could be used as:

```typescript
responseLdap(data === LdapStatus.enabled)
```

Let me think what do you think :)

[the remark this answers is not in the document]
qooban: I appreciate broad description you provided. Let's keep it as is, then. Especially that I marked is optional. I'm resolving this discussion.

test/cypress/integration/widgets/user_groups_spec.ts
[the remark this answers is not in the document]
NorbertAmartus: From the perspective of the test cases, I just want to disable/enable ldap
As mentioned in [this discussion](https://github.com/cloudify-cosmo/cloudify-stage/pull/1682#discussion_r757264471), using the word 'status' would indicate the need of using multiple `numeric`/`string` values, which is not a goal to which I'm aiming

Currently the tests looks like this:
```
    it('should display LDAP group column when LDAP is enabled', () => {
        setLdapAvailibility(true);
        reloadPage();
        cy.contains(ldapGroupColumnName);
    });

    it('should hide LDAP group column when LDAP is disabled', () => {
        setLdapAvailibility(false);
        reloadPage();
        cy.contains(ldapGroupColumnName).should('not.exist');
    });
```

And as you can see, the tests scenarios are related to the cases in which the ldap is enabled or disabled
So from the perspective of those test scenarios, I shouldn't be worried about passing the correct string to the `setLdapAvailibility` as those tests just want to enable/disable LDAP, without the worry about the implementation of this mechanism (so it's similar case to the second point of [that discussion](https://github.com/cloudify-cosmo/cloudify-stage/pull/1682#discussion_r757327182) 😁)

That's why I've created that method (so that the implementation would be somewhere else and additionally to not duplicate the code 😁), and that's why it's being executed with the usage of the `boolean` argument, and not `disabled`/`enabled` strings

widgets/userGroups/src/CreateModal.tsx
NorbertAmartus: Sure, that seems like a great idea :)

NorbertAmartus: Done 😎

test/cypress/integration/widgets/user_groups_spec.ts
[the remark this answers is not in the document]
qooban: Thanks for detailed answer. I had a bit different opinion here, but don't want us to spend time over-discussing it, especially that it is an optional comment.

Please just take a look at the second part of my comment.

widgets/userGroups/src/CreateModal.tsx
qooban: I don't think you can use IDE for this, it's kind of cloudify-stage secret knowledge you have to be aware of 🙂 

I'm sure we talked about it during one of our meetings :-D

NorbertAmartus: Yup, you're right
Just wanted to point out, that I've imported some things, but forgot to double check that the paths, that IDE have chosen, are being imported by the cloudify-stage standards 😁

So that's a nice catch :)

NorbertAmartus: Imports across the widget have been modified :)

widgets/userGroups/src/UserGroupsTable.tsx
NorbertAmartus: That import is not being used accross the file, so I'll remove it

test/cypress/integration/widgets/user_groups_spec.ts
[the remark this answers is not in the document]
NorbertAmartus: Done :)

app/actions/managers.ts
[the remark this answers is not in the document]
NorbertAmartus: Modified all of the names to match `isLdapEnable`, and additionally I've renamed the action to be `SET_LDAP_ENABLE`, rather than `SET_LDAP`, to that we would have more consistency across the system :)

widgets/userGroups/src/CreateModal.tsx
qooban: UPDATE: p. 1 - added a point to the UI weekly agenda and updated refactor description in the linked document.

test/cypress/integration/widgets/user_groups_spec.ts
NorbertAmartus: `cy.reload` works without a problem, so I've included it with the other changes :)

test/cypress/integration/widgets/user_groups_spec.ts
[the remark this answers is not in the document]
NorbertAmartus: I've took a look upon the second part of the comment, and applied suggested changes :)

widgets/userGroups/src/CreateModal.tsx
NorbertAmartus: Translations has been provided :)

test/cypress/integration/widgets/user_groups_spec.ts
qooban: Great, thanks 🙂

app/translations/en.json
```diff
@@ -1250,6 +1250,83 @@
...
+            "columns": {
+                "groupName": "Group",
+                "ldapGroup": "LDAP group",
+                "admin": "Admin",
+                "users": "# Users",
+                "tenants": "# Tenants"
+            },
+            "empty": {
+                "table": "There are no User Groups available. Click \"Add\" to add User Groups."
+            },
```
qooban, line 1264: I see in other widgets we used `no<ResourceName>` key for such cases. What do you think about aligning it with other widget translations?

qooban (commented):
Thanks for all the refactoring and migration efforts 👍🏻 

Looks mostly good, added one comment in the code and have some general:

1. Do I understand "Documentation" section properly that you'll add documentation sync commit to this PR? (once https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2040 is merged) If so, then fine 👍🏻 

2. Please take a look at unresolved discussions and in case you are not planning to change anything, just leave a comment, not to have them unanswered.

3. I found this:
image
Could you fix translation key?

qooban:
I'm adding @kubama as a reviewer, just to have another pair of eyes to scan this PR 🙂

widgets/userGroups/src/MenuAction.tsx
```diff
@@ -1,53 +1,50 @@
...
+interface Item {
+    name: string;
+    tenants: unknown;
+    users: string[];
+}
 
-    static SET_ADMIN_GROUP_ROLE_ACTION = 'set-admin-role';
+interface MenuActionProps {
+    item: Item;
+    onSelectAction: (name: string | undefined, item: Item) => void;
```
kubama, line 16: Is there a benefit in using `| undefined` instead of marking the prop as optional with `?`?

kubama (commented):
I only have minor comments. Good job overall 🙂.

NorbertAmartus:
> Do I understand "Documentation" section properly that you'll add documentation sync commit to this PR? (once [cloudify-cosmo/docs.getcloudify.org#2040](https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2040) is merged) If so, then fine 👍🏻

Yup, when the PR will have the required approves, then I'll merged the linked PR (which would update the docs), and then I'll sync the docs withint the next commit. After that I'll merge this PR aswell ;)

app/translations/en.json
NorbertAmartus: Sure, I think that it would be great idea!
The more consistency, the better 😁

I'm not yet familiar with all the conventions across the application, so sorry for those tiny miastakes
Nonetheless, thanks again for pointing that out!

test/cypress/integration/widgets/user_groups_spec.ts
```diff
@@ -53,4 +64,16 @@ describe('User group management widget', () => {
...
+        setLdapAvailability(true);
+        reloadPage();
+        cy.contains(ldapGroupColumnName);
+    });
+
+    it('should hide LDAP group column when LDAP is disabled', () => {
+        setLdapAvailability(false);
+        reloadPage();
+        cy.contains(ldapGroupColumnName).should('not.exist');
+    });
```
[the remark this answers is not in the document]
NorbertAmartus: I don't see a reason why shouldn't we would use that approach, so I'll implement that :)
Thanks for the suggestion! :)

widgets/userGroups/src/MenuAction.tsx
NorbertAmartus: Yup, if I'd declare type as a:
```typescript
onSelectAction: (name?: string, item: Item) => void
```

Then typescript would throw an error at me, because the first function argument may be provided, but the second one is required, which from the typescript perspective does not make sense

That's why I've explicitly declared that the name may be type of `string` or `undefined`

widgets/userGroups/src/consts.ts
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line import/prefer-default-export
+export const MENU_ACTIONS = {
```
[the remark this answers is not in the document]
NorbertAmartus: Okay, I'll try to adjust the code to that convention :)

widgets/userGroups/src/MenuAction.tsx
kubama: Looking into it further - it doesn't seem to make sense for the `name` argument to allow `undefined`. From what I can see this `name` argument carries action name which is always specified and at the same time expected. Also, this parameter could be named in a more descriptive way like `actionName`, `actionType` or `resourceType`. Lastly, I think it could be typed stronger by allowing a union of string literals instead of any string.

test/cypress/integration/widgets/user_groups_spec.ts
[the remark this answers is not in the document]
NorbertAmartus: Done ;)

widgets/userGroups/src/consts.ts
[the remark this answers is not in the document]
NorbertAmartus: Modifications has been added, along with renaming the properties names, so that they doesn't contain `action` word, which is being used inside of the object containing them 😁

widgets/userGroups/src/MenuAction.tsx
NorbertAmartus: Changing the `name` argument may be indeed a good option
As the method is being named `onSelectAction` I'll rename it as `actionName`

When it comes to the typing `MenuItemProps` is returning a type of `string` | `undefined` for the `name` property
I can try to make an enum from the  `menuActions` (It's `MENU_ACTIONS`, but after renaming it) and use it as:


```typescript
export const enum menuActions {
    EditUsers = 'Users',
    EditTenants = 'tenants',
    Delete = 'delete',
    SetDefaultGroupRole = 'set-default-role',
    SetAdminGroupRole = 'set-admin-role'
}

interface MenuActionProps {
    item: Item;
    onSelectAction: (actionName: menuActionsEnum, item: Item) => void;
}

const MenuAction: FunctionComponent<MenuActionProps> = ({ item, onSelectAction }) => {
    const actionClick = (_proxy: MouseEvent<HTMLAnchorElement>, { name }: MenuItemProps) => {
        onSelectAction(name as menuActionsEnum, item);
    };
    ...
 }
```

But I don't know if they are some cases in which we would not want to cast `name` as a `menuActionsEnum`
But if you'd like to have a strongly typed solution, then I can suggest the approach above 😁

Let me know what you think :)

app/translations/en.json
NorbertAmartus: Done

widgets/userGroups/src/CreateModal.tsx
```diff
@@ -21,7 +20,8 @@ export default function CreateModal({ toolbox }) {
         const { groupName, isAdmin, ldapGroup } = inputs;
 
         if (_.isEmpty(groupName)) {
-            setErrors({ groupName: 'Please provide group name' });
+            const validationMessage = i18n.t('widgets.userGroups.modals.create.validation.groupName');
```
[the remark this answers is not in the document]
NorbertAmartus: Thanks for that suggestion, I was wondering about creating some util function which would resolve this issue, but it's good to hear that there is already one that does the job!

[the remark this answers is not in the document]
NorbertAmartus: Modifications has been applied

widgets/userGroups/src/MenuAction.tsx
kubama: I had another deeper look 🙂. I think an even better approach here would be to remove this `name` argument and have three separate props instead - `onDelete`, `onEditTenants` and `onEditUsers`. However, this goes beyond the refactoring that you did, so I'll leave it to you.

NorbertAmartus: Modifications has been added :)

kubama (approved):
Looks cool!

NorbertAmartus:
I've added docs update, so I'll need a review approval once more 😁

widgets/userGroups/README.md
```diff
@@ -14,7 +14,8 @@ This widget is only available to admin users.
 The widget displays the following information regarding each of the user groups:
 
 * **Name**
-* **LDAP group** When working with an ldap-based external authentication system, this fields identifies the LDAP user group which is connected to the current {{< param product_name >}} user-group.
+* **LDAP group** When working with an LDAP-based external authentication system, this field identifies the LDAP user group which is connected to the current {{< param product_name >}} user-group.   
+   This column will not be visible if the LDAP is disabled in {{< param product_name >}}.
```
kubama, line 18:
```suggestion
   This column will not be visible if LDAP is disabled in {{< param product_name >}}.
```

widgets/userGroups/README.md
```diff
@@ -38,7 +39,8 @@ Please notice that by belonging to several groups, users might be assigned to a
 1. Click **Add** in the User Groups Management widget.
 2. Enter a name for the group.   
    The name must contain at least five alphanumeric characters and begin with a letter. It can also include the following characters: `-`, `_` or `.`.
-3. (Optional) If you want to add user group that is configured in your LDAP/AD system, enter the LDAP Group Distinguished Name (DN). The format must be `CN=GroupName,OU=OUName,DC=Domaincomponent,DC=DomainComponent`, for example `CN=ldapwiki,OU=groups,DC=example,DC=com`.
+3. (Optional) If you want to add user group that is configured in your LDAP/AD system, enter the LDAP Group Distinguished Name (DN). The format must be `CN=GroupName,OU=OUName,DC=Domaincomponent,DC=DomainComponent`, for example `CN=ldapwiki,OU=groups,DC=example,DC=com`.   
+   This field will not be visible if the LDAP is disabled in {{< param product_name >}}.
```
kubama, line 43:
```suggestion
   This field will not be visible if LDAP is disabled in {{< param product_name >}}.
```

NorbertAmartus: Documentation update comes from [pull request](https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2040) which was created on the `docs.getcloudify.org` repository

As I can recall, we shouldn't just add changes on this repository, because after merging into the master, one of the checks would fail (if I recall correctly, that check was named `Documentation`)
This check is comparing documentation from the `docs.getcloudify.org` repository, to one existing on this repository.

Because of that, when it comes to changes, we'd need to add them to the mentioned [docs.getcloudify.org](https://github.com/cloudify-cosmo/docs.getcloudify.org) repository, and then sync the changes

NorbertAmartus: Additionally, I think that the usage of the word `the` is not being misused
As I can see in [the files changes](https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2040/files), `LDAP` was previously prefixed with the word `the`

Moreover, we are referring to the LDAP, which is being used by the manager itself, so we can indicate with the usage of `the` that the meaning of the LDAP in this context is being specific

kubama: Yes, I would correct it in the docs if you think it's worth it - if not, then lets leave it.

NorbertAmartus: I'd leave it as it is
If we'd like to correct that, I think that the seperate ticket would be great for that
Because it would take some time to create an additional PR on the other repository, merge it and then sync changes inside one of the PR's here

kubama: > Moreover, we are referring to the LDAP, which is being used by the manager itself, so we can indicate with the usage of the that the meaning of the LDAP in this context is being specific

I think LDAP is a so called 'proper noun', so prepending it with 'the' is grammatically incorrect.

PR #1684, opened 2021-11-29
RD-3584 Modified the layout of the dashboard

NorbertAmartus:
## Description

Changes bringed with this PR will modify the layout of the dashboard, so that the `execution` and `executionsStatus`  widgets would fill all the available horizontal space   
Additionally, some amount of the comments (related to the creator of the files) has been removed, due to the [ongoing refactors](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Migrating-files-to-TypeScript)

### Before
before

### After
after

## Checklist

## Tests

 

Locally I've run all tests related to the modified widget (the only test related to them was stored in `executions_spec.ts` file) and additionally I've run `npm run test:frontend` command, which has passed succesfully
Moreover, I've run the [System Tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1589)

### DIFF
--- a/templates/pages/adminDash.json
+++ b/templates/pages/adminDash.json
@@ -124,23 +124,23 @@
           "y": 14
         },
         {
+          "name": "Executions Statuses Graph",
+          "width": 12,
+          "height": 24,
+          "definition": "executionsStatus",
+          "x": 0,
+          "y": 38
+        },
+        {
           "name": "Executions",
-          "width": 8,
+          "width": 12,
           "height": 24,
           "definition": "executions",
           "configuration": {
             "pageSize": 10
           },
           "x": 0,
-          "y": 38
-        },
-        {
-          "name": "Executions Statuses Graph",
-          "width": 4,
-          "height": 24,
-          "definition": "executionsStatus",
-          "x": 8,
-          "y": 38
+          "y": 62
         }
       ]
     }


qooban (commented):
@NorbertAmartus It seems that you need to update your branch and check/re-run system tests (there are some failures). I'll re-review once they pass.

I don't have strict opinion here, but have gut feeling that we shouldn't put big amount of unrelated refactorings into such small PR. Maybe it should be part of a separate PR? (as you have already done it, so you can cherry-pick the commit) BTW Are there any remaining comments? @kubama What do you think?

kubama:
> @kubama What do you think?

Yes, I agree. For ongoing refactors we should only address files related to the actual change. Since the refactor in ongoing anybody touching a file is supposed to do the change (here - remove author comment) anyway, so there is no real benefit in doing it upfront.
Here, the git log entry will be a bit misleading, as without going into details and checking all diffs it won't be clear which files needed to be updated to accomplish the ticket.

NorbertAmartus:
Provided arguments are more than enough to discourage me from duplicating this behaviour in the future 😁
I've removed commit in order to keep the modifications related to the task itself.

I may have an additional question related to internal refactors, but I'll ask it on the slack channel

NorbertAmartus:
> @NorbertAmartus It seems that you need to update your branch and check/re-run system tests (there are some failures). I'll re-review once they pass.

I've run again the [system tests](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1589/pipeline/55/) and to my surprise, the tests were failing in scenarios, which are not related to the work that I've done

I've executed same test file, which is failing, on the master branch and the problem seems to appear here.

I've created [this bug](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1589/pipeline/55/) and inform the current UI guardian about that issue.

qooban (approved):
Looks good 👍🏻 

Thanks for taking care of system tests and submitting an issue in JIRA.

PR #1689, opened 2021-12-02
RD-3197 Moved product header to sidebar and removed top bar

kubama:
## Description
Moved product header to sidebar and removed top bar. Includes two minor general sidebar related fixes:
- fixed system menu items shrinking on small vertical resolution
- fixed embedded widgets positioning

## Screenshots / Videos
Collapsed:
image

Expanded:
image

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1614/

 

## Documentation
Will be covered in separate ticket


app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,94 @@
...
+            backgroundColor: '#FFC304',
+            color: '#000000',
+            textAlign: 'center',
+            transform: 'rotate(-90deg)',
+            position: 'absolute',
+            top: 23,
+            left: -23,
+            width: 55,
+            textTransform: 'uppercase',
+            fontFamily: 'JosefinSans-Bold',
```
qooban, line 24: Referencing the font family that way has some flaws:
1. It depends on `LogoLabel` component where that font is actually included. When we'll remove `LogoLabel` component then this font will become unavailable in the package.
2. White-labelling might be difficult (as customers will need to override local styles)

I'm wondering how difficult it would be to make kind of wrapper function, e.g. `getProductFont` which will get all the necessary details - import CSS file, take into account white-labelling and return font name? We could use it in all places where "JosefinSans-Bold" is referenced directly.

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,94 @@
...
+                    textTransform: 'uppercase',
+                    letterSpacing: '0.2em',
+                    marginRight: '1em'
+                }}
+            >
+                {i18n.t('productName')}
+            </span>
+            v<ProductVersion version={productVersion} style={{ color: 'inherit', marginLeft: '-0.3em' }} />
+            {licenseLabelKey && <LicenseLabelWrapper content={t(licenseLabelKey)} linked={!isCommunity} />}
+        </div>
```
qooban, line 90: General:

Having it all hardcoded using `style` prop might be hard for white-labelling.
I don't have strict suggestion here as we still don't have clear requirements for white-labelling, but recently WR contacted mgmt asking for more flexible solution for customization (one of the suggestions was to use CSS classes or variables for styling, instead of styled-components and React `style` prop).

app/components/sidebar/UserInitialsIcon.tsx
```diff
@@ -8,9 +8,9 @@ import { ReduxState } from '../../reducers';
 const StyledLabel = styled(Label)`
     &&&&& {
         float: none;
-        margin-left: -10px;
```
qooban, line 11: I see that you did similar changes in `UserInitialsIcon` and `SideBarItemIcon`.

Is it possible somehow for `UserInitialsIcon` and `SideBarItemIcon` to share common styles to avoid synchronization issues in the future?

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,94 @@
...
+            paddingTop: 2,
+            fontSize: 8,
+            lineHeight: 'normal'
+        }}
+    >
+        {content}
+    </div>
+);
+
+const LicenseLabelWrapper = ({ content, linked }: { content: string; linked: boolean }) => {
```
qooban, line 35: I know that `LicenseLabelWrapper` is only used in `SideBarHeader`, but IMHO for clarity it seems to me reasonable to have it in separate file (L12 - L52 are about half of this file). WDYT?

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const isTrial = useSelector((state: ReduxState) => state.manager.license.data?.trial);
+
+    let licenseLabelKey;
+    if (isCommunity) {
+        licenseLabelKey = 'community';
+    } else if (isExpired) {
+        licenseLabelKey = 'expired';
+    } else if (isTrial) {
+        licenseLabelKey = 'trial';
+    }
```
qooban, line 71: Maybe that is something we can wrap using `useMemo`?

app/components/sidebar/SideBarHeader.tsx
qooban: It seems that the code was copied from `LicenseTag`. I understand that both are required (because  About page or License Management page are not changed yet), so I suggest to extract common parts from both components first and have two version of that tag component.

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    const productVersion = useSelector((state: ReduxState) => state.manager.version.version);
+
+    return (
+        <div style={{ overflow: 'hidden', whiteSpace: 'nowrap', flexShrink: 0 }}>
+            <Logo url={logo} style={{ top: 6, left: licenseLabelKey ? 3 : -3 }} />
+            <span
```
qooban, line 78: Optional:

I think we can make product name a bit bigger and maybe drop letter spacing. WDYT?
As there's separate issue for styling the sidebar I'm adding it as optional.

qooban (commented):
Looks good in general 👍🏻 

Provided some suggestions in comments.

Do you have the same rendering issue with license tag when expanding/collapsing the sidebar?

https://user-images.githubusercontent.com/5202105/144407228-92795183-d5ca-4d23-9f63-0c7e112aaabb.mp4

kubama:
> Do you have the same rendering issue with license tag when expanding/collapsing the sidebar?

Yes, unfortunately I have no clue what's the reason behind.

app/components/sidebar/SideBarHeader.tsx
kubama: If required, we can assign class names to allow custom styling via external CSS file and still use styled-components or style prop.

app/components/sidebar/SideBarHeader.tsx
kubama: Done

app/components/sidebar/SideBarHeader.tsx
kubama: I'd leave it for styling ticket

app/components/sidebar/UserInitialsIcon.tsx
kubama: Done

app/components/sidebar/SideBarHeader.tsx
kubama: > It depends on LogoLabel component where that font is actually included. When we'll remove LogoLabel component then this font will become unavailable in the package.

I moved the font declaration CSS import to top level.

> White-labelling might be difficult (as customers will need to override local styles)

The only minor difficulty is that overriding CSS might need to use `!important` to override local style. But as we are using inline styles everywhere then this is a general thing.
IMO more important issue in terms of white-labelling is that we are not assigning class names or ids that could be referenced in overriding CSS.

> I'm wondering how difficult it would be to make kind of wrapper function, e.g. getProductFont which will get all the necessary details - import CSS file, take into account white-labelling and return font name? We could use it in all places where "JosefinSans-Bold" is referenced directly.

I'm not sure what you mean by "taking white-labelling into account".
It could be possible to define font name as SCSS variable and then import it, but the benefit would be quite minor and it would need another cloudify-ui-common release, so I think it's a bit out of scope for this PR.

app/components/sidebar/SideBarHeader.tsx
kubama: Addressed both comments

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,48 @@
...
+                            fontSize: 8,
+                            lineHeight: 'normal',
+                            textAlign: 'center',
+                            borderRadius: 'unset',
+                            textTransform: 'uppercase',
+                            transform: 'rotate(-90deg)'
+                        }}
+                    />
+                </div>
+                <Logo url={logo} style={{ position: null, height: 33, margin: '11px 8px' }} />
```
qooban, line 32: I didn't spot it before - I'm not sure if passing logo directly is the right thing to do. We will lose white-labelling option here. Please take a look at how `Logo` component works in relation to theming.

Maybe we should introduce new parameter in the theme (in `conf/userConfig.json` - `whiteLabel` section) and/or enhance `Logo` component to get another prop indicating which logo version should be used (for white/dark background)? Or maybe just changing the logo in `cloudify-ui-components` will be sufficient? You need to check if there are any other places using the current logo. I think that Blueprints and Blueprint Catalog widgets are using logo, but it looks like it is part of the widget.

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,48 @@
...
+                <span
+                    style={{
+                        letterSpacing: '0.2em',
+                        marginRight: '1em',
+                        textTransform: 'uppercase'
+                    }}
+                >
+                    {i18n.t('productName')}
+                </span>
```
qooban, line 41: I think you should be able to reuse `LogoLabel` component here (I did it locally slightly extending `LogoLabel` customizability by introducing `style` prop there), then we have only one place with `JosefinSans-Bold` font defined, so we can move CSS import there. WDYT?

My example (not necessarily perfect):
```
<LogoLabel
    color="inherit"
    style={{
        fontSize: '18px',
        marginRight: '5px',
        textTransform: 'uppercase',
        display: 'inline'
    }}
    content={i18n.t('productName')}
/>
```
and in `LogoLabel` I just added `style: CSSProperties` to props.

app/components/sidebar/SideBarHeader.tsx
qooban: > I moved the font declaration CSS import to top level.

I'm not sure about moving it to top level. Maybe we just need to reuse `LogoLabel`? (see my comments in the latest review)

> IMO more important issue in terms of white-labelling is that we are not assigning class names or ids that could be referenced in overriding CSS.

OK, so then if with class names we'll be able to style product header, then let's add them to relevant product header items.

> I'm not sure what you mean by "taking white-labelling into account".
> It could be possible to define font name as SCSS variable and then import it, but the benefit would be quite minor and it would need another cloudify-ui-common release, so I think it's a bit out of scope for this PR.

I didn't have final solution in mind, but right now when I'm thinking about it: maybe we can just add product font name into white-labelling theme `conf/userConfig.json` and use it from there? I'm not sure it is the right solution, I'm looking for the best solution taking into account ease of maintenance and customization flexibility.

app/components/sidebar/SideBarHeader.tsx
qooban: OK.

qooban (commented):
Added some more comments. 

Mostly related to white-labelling which seems a bit tricky. Also - as per my understanding product header is the most common part of customization - let's try to make the customization as easy as possible with the tools we have right now.

If you'll see that it'll be reasonable to handle changes I'm suggesting in a separate ticket, let's create one and try to fit it in this or next sprint.

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Logo, ProductVersion } from '../basic';
+import { ReduxState } from '../../reducers';
+import LicenseLabel from '../LicenseLabel';
+
+const SideBarHeader: FunctionComponent = () => {
+    const productVersion = useSelector((state: ReduxState) => state.manager.version.version);
+
+    return (
+        <div style={{ overflow: 'hidden', whiteSpace: 'nowrap', flexShrink: 0 }}>
+            <span style={{ fontFamily: 'JosefinSans-Bold' }}>
```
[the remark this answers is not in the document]
kubama: After experimenting it turned out this font just looks better in such small size than the main font, so I would keep it.

kubama:
@qooban
My general question after reading your comments is - what's the actual target for customization level? On one hand you are suggesting moving to well defined class names to allow easy CSS overriding, on the other hand you are suggesting extending support for userConfig.json. These two directions are kind of overlapping. If we are going to maintain class names then this will give customers maximum level of flexibility. If this is the case, then is it worth it to spend time on extending userConfig.json support? Is the actual goal for white-labelling well defined at the moment? I think it should be if we are making implementation decisions.

app/components/sidebar/SideBarHeader.tsx
kubama: > I didn't spot it before - I'm not sure if passing logo directly is the right thing to do. We will lose white-labelling option here. Please take a look at how Logo component works in relation to theming.
Maybe we should introduce new parameter in the theme (in conf/userConfig.json - whiteLabel section) and/or enhance Logo component to get another prop indicating which logo version should be used (for white/dark background)? Or maybe just changing the logo in cloudify-ui-components will be sufficient? You need to check if there are any other places using the current logo. I think that Blueprints and Blueprint Catalog widgets are using logo, but it looks like it is part of the widget.

My understanding is that current state with two versions of the logo in use is temporary - IIUC eventually we are going to switch to dark background in other places and only one version of the logo will be used. If so, is it worth to introduce new theme parameter at this point? Wouldn't `theme.logo ?? builtInDarkBackgroundLogo` be sufficient?

app/components/sidebar/SideBarHeader.tsx
kubama: > I think you should be able to reuse LogoLabel component here (I did it locally slightly extending LogoLabel customizability by introducing style prop there), then we have only one place with JosefinSans-Bold font defined, so we can move CSS import there. WDYT?

As stated in another comment I'm intentionally using `JosefinSans-Bold` font in another place.
Also, in the example that you provided you are overriding all `LogoLabel` style props (except font family) and additionally setting `display` prop to override internal `h2` styling. I don't think it is worth it - it's more complicated than simply using a bare `span`.

app/components/sidebar/SideBarHeader.tsx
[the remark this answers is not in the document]
qooban: OK, as we discussed let's keep it.

app/components/sidebar/SideBarHeader.tsx
qooban: OK, as we discussed let's go with `theme.logo ?? builtInDarkBackgroundLogo`.

app/components/sidebar/SideBarHeader.tsx
qooban: OK, let's keep `span` here for now as from the implementation perspective it's not a big gain.

In the future it could probably be improved when we'll have clear overview what are the expected theme changes (e.g. for About Modal and all the so-called logo-pages).

kubama:
I addressed code review comments as agreed during the call.
Created https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2047 to satisfy doc check.

test/cypress/integration/templateManagement/templates_spec.ts
```diff
@@ -81,7 +81,7 @@ describe('Templates segment', () => {
         );
     });
 
-    it('allows users to create and modify templates', () => {
+    it.only('allows users to create and modify templates', () => {
```
qooban, line 84: Probably something to change before merge 🙂

qooban (commented):
Just one small thing to fix before merging.

app/components/LicenseLabel.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (!licenseLabelKey) {
+        return null;
+    }
+
+    const labelComponent = (
+        <Label
+            style={{
+                backgroundColor: '#FFC304',
+                color: '#000000',
+                ...style
```
NorbertAmartus, line 43: As I can see in the [Label typing](https://github.com/Semantic-Org/Semantic-UI-React/blob/61a08733941fb60d6ff09f30c3e2fc13e40448da/src/elements/Label/Label.d.ts#L37) it can be given a `className`

Because of that I think that we should use `styled-components` , so that:
- The `styled` prop would be destructurized with other props (component would be simplier)
- We wouldn't pass a plain object into the `style` prop, which would result in a better performance

Example of styled component:
```typescript
const StyledLabel = styled(Label)`
  backgroundColor: '#FFC304',
  color: '#000000'
`;
```

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,62 @@
...
+                    style={{
+                        letterSpacing: '0.2em',
+                        marginRight: '1em',
+                        textTransform: 'uppercase',
+                        fontFamily: productFont
+                    }}
+                >
+                    {i18n.t('productName')}
+                </span>
+                v<ProductVersion version={productVersion} style={{ color: 'inherit', marginLeft: '-0.3em' }} />
```
NorbertAmartus, line 56: I think that this `v`, at the beggining of the line, might be a typo 😁

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -0,0 +1,62 @@
...
+            </StyledLink>
+        </div>
+    );
+};
```
NorbertAmartus, line 60: Every element (inside this component) is using `style` prop for styling purposes.
I think that it would be more advised to use `styled-components`, so that:
- Whole component would be more performant (simple rerender would add a cost of recreating all of those plain objects)
- Name of the newly created `styled-components` components would be useful in terms of readability of the component (to know what is the purpose of 2 divs at the top of the return statement, I'd have to go through the rest of the `return` declaration, to understand that)

Let me know what do you think

app/components/sidebar/UserInitialsIcon.tsx
```diff
@@ -4,22 +4,23 @@ import { useSelector } from 'react-redux';
...
     }
 `;
 
 const UserInitialsIcon: React.FunctionComponent = () => {
     const username = useSelector((state: ReduxState) => state.manager.username);
 
-    return <StyledLabel circular>{username.substr(0, 2).toUpperCase()}</StyledLabel>;
+    return (
+        <StyledLabel style={defaultStyle} circular>
+            {username.substr(0, 2).toUpperCase()}
```
NorbertAmartus, line 21: We could extract that to another, memoized variable, so that it would be explicitly declared what is the purpose of that line

Without that variable, me as an another developer, might look into the application (or guess by the file name), to understand why this line is used like it is 😁

I think that it should be explicitly said, so that we wouldn't have to guess.

Implementation example:

```typescript
const userInitials = useMemo(() => {
  return username.substr(0, 2).toUpperCase();
}, [username]);

return (
  <StyledLabel style={defaultStyle} circular>
    {userInitials}
  <StyledLabel />
)
```

app/components/templateManagement/pages/PageManagement.tsx
```diff
@@ -215,7 +223,7 @@ export default function PageManagement({ pageId, isEditMode = false }: PageManag
     return (
         <div className="main">
             <div className="sidebarContainer">
-                <Sidebar visible as={Menu} vertical size="small">
+                <Sidebar visible as={Menu} vertical size="small" style={{ width: expandedSidebarWidth }}>
```
NorbertAmartus, line 226: Sidebar component supports `className` property
Because of that I think that we may favour the declaration which may be using `styled-components` (for performance and readability reasons, which was mentioned in the previous comments)

Implementation example:
```typescript
const StyledSidebar = styled(Sidebar)`
  width: ${expandedSidebarWidth}
`
```

NorbertAmartus (commented):
Looks good 💪
Just added some comments mostly related to usage of the inline `style` prop

test/cypress/integration/templateManagement/templates_spec.ts
kubama: Done

app/components/sidebar/SideBarHeader.tsx
kubama: Actually it's there by design 🙂. It's a prefix before the version number, for example 'v6.3'.

app/components/sidebar/UserInitialsIcon.tsx
kubama: Done

kubama:
@NorbertAmartus

Regarding your styling comments - at the moment our established approach is that inline styling as well as styled-components are equally good approaches, please see https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/Cloudify+UI+Code+Style#Rules. If you think it is something worth changing and that we should prefer one approach over the other then lets discuss your points on our weekly.

qooban (approved):

app/components/sidebar/SideBarHeader.tsx
NorbertAmartus: Okay, thanks for explaining :)

NorbertAmartus:
> @NorbertAmartus
> 
> Regarding your styling comments - at the moment our established approach is that inline styling as well as styled-components are equally good approaches, please see https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/Cloudify+UI+Code+Style#Rules. If you think it is something worth changing and that we should prefer one approach over the other then lets discuss your points on our weekly.

Okay, sure
In that case I'm resolving the issues related to the `style` prop

app/components/templateManagement/pages/PageManagement.tsx
NorbertAmartus: As commented [here](https://github.com/cloudify-cosmo/cloudify-stage/pull/1689#issuecomment-986705037), I'm resolving this issue

app/components/sidebar/SideBarHeader.tsx
NorbertAmartus: As commented [here](https://github.com/cloudify-cosmo/cloudify-stage/pull/1689#issuecomment-986705037), I'm resolving this issue

app/components/LicenseLabel.tsx
NorbertAmartus: As commented [here](https://github.com/cloudify-cosmo/cloudify-stage/pull/1689#issuecomment-986705037), I'm resolving this issue

NorbertAmartus (approved):

PR #1700, opened 2021-12-10
RD-3730 Modified execution of disabling the popup

NorbertAmartus:
## Description

Modified the way, that the popup is not being shown.
Additionally, I've addressed our refactor guidelines, by migrating file (on which I was working with) to typescript 🎉

## Screenshots / Videos

N/A

## Checklist

## Tests

 

I've executed the [System tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1652)
Additonally, I've run the `npm run test:frontend` command, which has been executed successfully

Additionally, the `executions_tests` are looking unstable and I couldn't fix them easily, that's why I've created [this ticket](https://cloudifysource.atlassian.net/browse/RD-3762) in which I've addressed the issue that I've encountered.

## Documentation

N/A


NorbertAmartus:
Just a note,
I've addressed our refactor guidelines, by migrating file (on which I was working with) to typescript 🎉

app/components/shared/ExecutionStatus.tsx
```diff
@@ -27,108 +26,79 @@ import ExecutionUtils from '../../utils/shared/ExecutionUtils';
...
-export default function ExecutionStatus({
+
+interface Execution {
+    // eslint-disable-next-line camelcase
+    status_display?: string;
+    status?: string;
+    // eslint-disable-next-line camelcase
+    scheduled_for?: string;
+    // eslint-disable-next-line camelcase
+    workflow_id?: string;
```
qooban, line 37: Optional:

Sometimes we use multiple `// eslint-disable-next-line camelcase`, sometimes we close part not using camelcase within `/* eslint-disable camelcase */` and `/* eslint-enable camelcase */` statements. I prefer the second, it seems more clear to me, especially that in backend response there can be more parameters using underscore so it is justified to disable that rule for the whole list.

```suggestion
    /* eslint-disable camelcase */
    status_display?: string;
    status?: string;
    scheduled_for?: string;
    workflow_id?: string;
    /* eslint-enable camelcase */
```

Anyways, it's optional comment.

app/components/shared/ExecutionStatus.tsx
```diff
@@ -27,108 +26,79 @@ import ExecutionUtils from '../../utils/shared/ExecutionUtils';
...
-    labelAttached: PropTypes.string,
-
-    /**
-     * If set to true, then execution status will be added to label
-     */
-    showLabel: PropTypes.bool,
-
-    /**
-     * If set to true, then workflow ID will be added to label
-     */
```
qooban, line 123: Optional:

I'm wondering maybe we should move all these comments to `ExecutionStatusProps`?

qooban (dismissed):
Looks good, I've added two optional non-blocking comments. Did not verify UI. 

I assume that you decided it's not worth to add a test case for such a specific problem?

> Additonally, I've run the npm run test:frontend command, which has been executed successfully

Just a hint `test:frontend:unit` is executed as part of Stage-UI-Multibranch and `test:frontend:components` is executed as part of Stage-UI-SystemTests. As `test:frontend` is defined the following way:
```
"test:frontend": "npm run test:frontend:unit && npm run test:frontend:components"
``` 
you don't need to run it manually when both jobs are passing.

> Additionally, the executions_tests are looking unstable and I couldn't fix them easily, that's why I've created this ticket in which I've addressed the issue that I've encountered.

Strange issue, I've added a comment in the ticket.

app/components/shared/ExecutionStatus.tsx
NorbertAmartus: We may do that, just for the component descriptivity
I've looked if placing comments inside the `ExecutionStatusProps`  interface would result in passing them, but it didn't seem to work that way 😢 

So just for the descriptivity, I think we may be able to do that 😁

app/components/shared/ExecutionStatus.tsx
NorbertAmartus: Fixed :wink

app/components/shared/ExecutionStatus.tsx
NorbertAmartus: Fixed 🎉

NorbertAmartus:
> I assume that you decided it's not worth to add a test case for such a specific problem?

That's a correct assumption 😁

qooban (approved):
LGTM 👍🏻

PR #1658, opened 2021-11-02
RD-3242 Added edit mode support for page menu items icons

kubama:
## Description
Added edit mode support for page menu items icons

## Screenshots / Videos
image

## Checklist

- [X] My code follows the UI Code Style.
- [x] I verified that all tests and checks have passed.
- [X] I verified if that change requires a documentation update.
- [X] I added proper labels to this PR.

## Tests
https://jenkins.cloudify.co/job/Stage-UI-System-Test/1442/

 

## Documentation
Will be updated in separate ticket

PR #1656, opened 2021-10-29
RD-3242 Added read-only support for menu icons

kubama:
## Description
Added read-only support for menu icons.

## Screenshots / Videos
image

## Checklist

- [X] My code follows the UI Code Style.
- [x] I verified that all tests and checks have passed.
- [X] I verified if that change requires a documentation update.
- [X] I added proper labels to this PR.

## Tests
I'll add tests in upcoming PRs.

 

## Documentation
Will be covered in separate ticket.

PR #1701, opened 2021-12-13
RD-3482 Add styling to sidebar

qooban:
## Description
Changed sidebar styling according to [UI guidelines @ Zeplin](https://app.zeplin.io/project/612e8cb57788ff10babc7900/screen/615daeeb7e36c23ab8f47d91).

## Screenshots / Videos

https://user-images.githubusercontent.com/5202105/145954398-7f2f514e-ae95-4d2e-809c-a22ab12ca970.mp4

## Checklist

## Tests
1. [Stage-UI-System-Test #1656](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1656/pipeline) (14.12.2021 08:45 CEST) - 1 test failed (`flows/admin_spec.ts`)
2. [Stage-UI-System-Test #1658](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1658/pipeline) (14.12.2021 11:36 CEST) - PASSED

## Documentation
There's RD-3341 to cover documentation update.


app/components/sidebar/SideBar.tsx
```diff
@@ -27,7 +28,19 @@ const ThemedSidebar = styled(Sidebar)`
...
         background-color: ${props => props.theme.sidebarHoverActiveColor} !important;
-        color: ${props => props.theme.sidebarTextColor} !important;
+        color: ${props => props.theme.sidebarHoverActiveTextColor} !important;
+    }
+    :after {
+        position: fixed;
+        content: '';
+        width: 0.5em;
+        height: 0.5em !important;
+        -webkit-transform: rotate(45deg);
```
NorbertAmartus, line 38: As I remember, the `styled-components` comes with the "Automatic vendor prefixing" feature [Link to the information from the official docs](https://styled-components.com/docs/basics)

So I tihnk, that the `-webkit-transform: rotate(45deg);` could be replaced with `transform: rotate(45deg);` 😉 

```suggestion
        transform: rotate(45deg);
```

app/components/sidebar/SideBar.tsx
```diff
@@ -11,9 +11,10 @@ import { useBoolean, useResettableState } from '../../utils/hooks';
 import SideBarHeader from './SideBarHeader';
 
 export const collapsedSidebarWidth = '4.3rem';
-export const expandedSidebarWidth = '15rem';
+export const expandedSidebarWidth = '18rem';
 
 const ThemedSidebar = styled(Sidebar)`
+    border: 0 !important;
```
NorbertAmartus, line 17: Just a question, if putting the border inside the below `&&& {}` would  remove the `!important` flag? 😁

app/components/sidebar/SideBarHeader.tsx
```diff
@@ -11,9 +11,9 @@ import Consts from '../../utils/consts';
 import { productFont } from '../fonts';
 
 const StyledLink = styled(Link)`
-    color: inherit !important;
+    color: white !important;
     &:hover {
-        color: inherit !important;
+        color: white !important;
```
NorbertAmartus, line 16: `white` color shouldn't come from the `sidebarTextColor` variable located inside the `conf/userConfig.json` file?

app/components/sidebar/UserInitialsIcon.tsx
```diff
@@ -7,6 +7,7 @@ import { ReduxState } from '../../reducers';
 import { defaultStyle } from './SideBarItemIcon';
 
 const StyledLabel = styled(Label)`
+    background: #65adff !important;
```
NorbertAmartus, line 10: Just a question, if putting the background inside the below `&&&&& {}`  would remove the !important flag? 😁

app/components/sidebar/SideBar.tsx
```diff
@@ -27,7 +28,19 @@ const ThemedSidebar = styled(Sidebar)`
...
+        color: ${props => props.theme.sidebarHoverActiveTextColor} !important;
+    }
+    :after {
+        position: fixed;
+        content: '';
+        width: 0.5em;
+        height: 0.5em !important;
+        -webkit-transform: rotate(45deg);
+        transform: rotate(45deg);
+        background-color: ${props => (props.$expanded ? 'white' : '#050229')} !important;
```
kubama, line 40: Shouldn't `#050229` be set to `sidebarColor` as defined in `userConfig.json`?

app/components/sidebar/SideBar.tsx
qooban: Good to know that. I'll remove this line as the code is already in L39.

app/components/sidebar/UserInitialsIcon.tsx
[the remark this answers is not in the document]
qooban: Hmm... I don't think so. It's a background of user initials icon, so it's not a text color of hovered/active sidebar.

app/components/sidebar/SideBarHeader.tsx
qooban: Good point, can be. I'll change it.

app/components/sidebar/SideBar.tsx
qooban: Right, I'll change that.

app/components/sidebar/UserInitialsIcon.tsx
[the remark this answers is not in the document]
kubama: It's not, but it's in line intentionally. Since there is no separate property for this particular icon background I think we should use `sidebarHoverActiveTextColor` here, otherwise styling the sidebar background may make the icon unreadable without the ability to alter its color.

app/components/sidebar/SideBar.tsx
qooban: Works fine, so I'll move it.

app/components/sidebar/UserInitialsIcon.tsx
qooban: Works fine, so I'll move it and remove `!important` flag.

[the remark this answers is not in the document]
qooban: What if `sidebarHoverActiveTextColor` will be the same as or similar to `sidebarTextColor`? (it's not very unlikely I believe)

I'm still not sure it's the right way to go, but I think it should be done when we get to changing the whole theme. 
I would expect that hardcoded color values will be moved to a predefined colors list map, then.

Also `#65adff` (let's call it "new blue color") may become `theme.mainColor` and then we could use it here.
If I'll use `sidebarHoverActiveTextColor` here, it's more likely that we'll forget about this place as potential place to use "new blue color".

kubama (approved):

qooban:
Thanks for the reviews. 
There's one system test failing - `flows/admin_spec.ts`. I'll check it out and update once I have it working.

NorbertAmartus (approved):

kubama (approved):

PR #1659, opened 2021-11-03
RD-3419 Fixed system test

kubama:
## Description
Fixed system test that's failing after #1657 was merged.

## Screenshots / Videos
N/A

## Checklist

- [X] My code follows the UI Code Style.
- [X] I verified that all tests and checks have passed.
- [X] I verified if that change requires a documentation update.
- [X] I added proper labels to this PR.

## Tests
Verified locally

 

## Documentation
N/A

PR #1703, opened 2021-12-14
RD-3760 Improve Maintenance Mode Button widget test to be more robust

qooban:
## Description
The problem is described in RD-3760. Unfortunately I couldn't reproduce it in my local environment, but I suspect that it might be due to some intermediate state of the Cloudify Manager. Thus, I have added additional verification in Maintenance Mode Button widget test.

## Screenshots / Videos
N/A

## Checklist

## Tests
[Stage-UI-System-Test #1659 (only `maintenance_mode_button_spec.ts`)](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1659/pipeline) (14.12.2021 11:56 CEST) - PASSED

## Documentation
N/A


test/cypress/integration/widgets/maintenance_mode_button_spec.ts
```diff
@@ -1,20 +1,25 @@
...
-import { waitUntilEmpty } from '../../support/resource_commons';
+import { waitUntil } from '../../support/resource_commons';
 
 describe('Maintenance mode button widget', () => {
     before(() => cy.activate('valid_trial_license'));
     beforeEach(() => cy.usePageMock('maintenanceModeButton', { pollingTime: 2 }).mockLogin());
 
     const getActivateButton = () => cy.contains('Activate Maintenance Mode');
     const getDeactivateButton = () => cy.contains('Deactivate Maintenance Mode');
+    const waitUntilMaintenanceModeIs = (status: 'activated' | 'deactivated') =>
```
kubama, line 9: Quite an unusual way to name a function, how about sticking to a common sense and using something like `waitForMaintenanceModeState`?

qooban: For me it's similar pattern to e.g. `orderBy`, `differenceWith`, etc., where the last portion of the name refers to the argument. That makes the function name + function arguments read like a sentence in natural language.

But I'm OK with changing it, I would use `Status` suffix instead of `State`, so it'd be `waitForMaintenanceModeStatus`.

kubama (approved):

qooban:
As it is small change, I'm merging it with just a single approval.

PR #1707, opened 2021-12-15
RD-382 Migrate files using request package to TypeScript

qooban:
## Description
This PR is the first in the long run of replacing `request` package with `axios`. 
Decided to first migrate all stage backend files using `request` package to TypeScript. With type checking in place I believe that the transition to new library will be a bit easier.

## Screenshots / Videos
N/A

## Checklist

## Tests
TODO

## Documentation
N/A

### DIFF
--- a/backend/handler/ManagerHandler.ts
+++ b/backend/handler/ManagerHandler.ts
@@ -1,19 +1,19 @@
-// @ts-nocheck File not migrated fully to TS
-
 import _ from 'lodash';
 import fs from 'fs-extra';
+import type { CoreOptions, Headers, Response } from 'request';
+
 import { getLogger } from './LoggerHandler';
 import { getConfig } from '../config';
 import * as RequestHandler from './RequestHandler';
 
 const logger = getLogger('ManagerHandler');
 
-let caFile = null;
+let caFile: Buffer | null = null;
 try {
     const caPath = _.get(getConfig(), 'app.ssl.ca');
     caFile = caPath ? fs.readFileSync(caPath) : null;
 } catch (e) {
-    throw new Error('Could not setup ssl ca, error loading file.', e);
+    throw new Error('Could not setup ssl ca, error loading file.');
 }
 
 export function getManagerUrl() {
@@ -24,7 +24,7 @@ export function getApiUrl() {
     return `${getConfig().managerUrl}/api/${getConfig().manager.apiVersion}`;
 }
 
-export function updateOptions(options, method, timeout, headers, data) {
+export function updateOptions(options: CoreOptions, method: string, timeout?: number, headers?: Headers, data?: any) {
     if (caFile) {
         logger.debug('Adding CA file to Agent Options');
         options.agentOptions = {
@@ -49,7 +49,15 @@ export function updateOptions(options, method, timeout, headers, data) {
     }
 }
 
-export function request(method, url, headers, data, onSuccess, onError, timeout) {
+export function request(
+    method: string,
+    url: string,
+    headers: Headers,
+    data: any,
+    onSuccess: (response: Response) => void,
+    onError: (error: Error) => void,
+    timeout?: number
+) {
     const requestUrl = getApiUrl() + (_.startsWith(url, '/') ? url : `/${url}`);
     const requestOptions = {};
     updateOptions(requestOptions, method, timeout, headers, data);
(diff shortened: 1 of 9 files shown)


backend/routes/Plugins.ts
```diff
@@ -95,50 +103,62 @@ router.get('/icons/:pluginId', (req, res) => {
...
+        const files = getFiles(req);
+        getPluginYaml = Promise.resolve(files.yaml_file[0].buffer);
     }
 
+    // eslint-disable-next-line camelcase
+    type PluginYaml = { plugins: Record<string, { package_name: string }> };
     getPluginYaml
-        .then(yaml.safeLoad)
-        .then(pluginYamlData =>
+        .then(pluginYamlString => yaml.load(pluginYamlString) as PluginYaml)
```
qooban, line 117: There is no `safeLoad` in newest version of `js-yaml` and in TS declarations, so I moved to use `load` method.

backend/routes/ServerProxy.ts
```diff
@@ -10,7 +10,15 @@ import { getLogger } from '../handler/LoggerHandler';
...
 const logger = getLogger('ServerProxy');
 
-function errorHandler(url, res, err) {
+declare global {
+    namespace Express {
+        interface Request {
+            su: string;
+        }
+    }
+}
```
qooban, line 19: I'm not sure about that solution. It changes Express.Request type globally, I'd like to have it changed only locally, but I couldn't find better. Do you have any ideas?

backend/routes/GitHub.ts
```diff
@@ -13,11 +13,19 @@ const logger = getLogger('GitHub');
...
 const authList = {};
 
-function getSecretName(secretName) {
+declare global {
+    namespace Express {
+        interface Response {
+            data: string;
+        }
+    }
+}
```
qooban, line 22: I'm not sure about that solution. It changes Express.Request type globally, I'd like to have it changed only locally, but I couldn't find better. Do you have any ideas?

On the other hand - I think it will not be necessary after migration to `axios`.

backend/handler/ManagerHandler.ts
```diff
@@ -1,19 +1,19 @@
...
 const logger = getLogger('ManagerHandler');
 
-let caFile = null;
+let caFile: Buffer | null = null;
 try {
     const caPath = _.get(getConfig(), 'app.ssl.ca');
     caFile = caPath ? fs.readFileSync(caPath) : null;
 } catch (e) {
-    throw new Error('Could not setup ssl ca, error loading file.', e);
+    throw new Error('Could not setup ssl ca, error loading file.');
```
kubama, line 16: I would apply https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#rethrowing_an_error_with_a_cause here

backend/routes/ServerProxy.ts
kubama: Can't we use casting locally instead of extending the type globally?

backend/handler/ManagerHandler.ts
qooban: I'd like to do so (thanks for sharing that link), but it seems TypeScript is not happy with it:
```
TS2554: Expected 0-1 arguments, but got 2.
```
When I dag deeper it seems that it is quite new feature - https://tc39.es/proposal-error-cause/ - seems to be introduced 5 days ago within ES2022 - https://tc39.es/ecma262/ . So I think we should wait probably with introducing it. Let me know what do you think.

backend/routes/ServerProxy.ts
qooban: Oh, of course we can. Thanks for enlightening me 🙂

backend/handler/ManagerHandler.ts
kubama: I was not aware it's such a bleeding edge 🙂. If so, lets leave it as it is.

backend/routes/GitHub.ts
NorbertAmartus: I think that this solution is fine, because we'll migrate to `axios` shortly
One think to notice is that, we should leave some sort of comment, that'd indicate a need for checking that behavior in the future (either in [this task](https://cloudifysource.atlassian.net/browse/RD-3802) or in the code itself 😉 )

NorbertAmartus (dismissed):
I have nothing to comment on, pretty good job! 💪

backend/routes/GitHub.ts
qooban: I actually managed to find better solution thanks to @kubama - see https://github.com/cloudify-cosmo/cloudify-stage/pull/1707#discussion_r769585422 discussion 🙂

qooban:
If you approve this PR, feel free to merge it during my absence. In case of comments, I'll address them when I'm back.

NorbertAmartus (approved):

PR #1709, opened 2021-12-16
RD-3806 Adjust height of Deployments View widget in Services and Environments pages

qooban:
## Description
See RD-3806 description for reasoning. It's another "bandage" solution. I did not change the height 10x because I think it is more than enough changing the height from 28 to 100 in our page JSON files. All widgets in the right pane in all tabs are visible and there is still much space left. There is no scrollbar on the right side. So both main requirements for RD-3806 are met.

In addition to the change I fixed minor styling issues.

## Screenshots / Videos

Adding screenshots only for Services page, the same change was done in Environments page.

### Before
image

### After
image

## Checklist

## Tests
[Stage-UI-System-Test #1677](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1677/) (16.12.2021 08:46) - PASSED

## Documentation
N/A

### DIFF
--- a/templates/pages/environments.json
+++ b/templates/pages/environments.json
@@ -16,7 +16,7 @@
         {
           "name": "Environments",
           "width": 12,
-          "height": 28,
+          "height": 100,
           "definition": "deploymentsView",
           "configuration": {
             "filterId": "csys-environment-filter"
--- a/templates/pages/services.json
+++ b/templates/pages/services.json
@@ -16,7 +16,7 @@
         {
           "name": "Services",
           "width": 12,
-          "height": 28,
+          "height": 100,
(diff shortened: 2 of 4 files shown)


kubama (approved):
I understand that now we will have double scrollbar on low resolution screens, but I guess it's a trade off and we can't solve it properly with current widget/greed solution, right?

qooban:
> I understand that now we will have double scrollbar on low resolution screens, but I guess it's a trade off and we can't solve it properly with current widget/greed solution, right?

Yes. There will be scrollbar for the window (not inside the widget) even for high resolutions, but this is what is expected by Ilan.

PR #1729, opened 2021-12-31
RD-3661 Added support for Terraform template URL authentication

kubama:
## Description
Added support for Terraform template URL authentication

## Screenshots / Videos
image

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1733/ - there are two failing tests, which are failing on master as well, so I consider it a pass

 

## Documentation
Will be updated as part of separate ticket.


backend/routes/Terraform.ts
```diff
@@ -22,47 +22,53 @@ router.use(bodyParser.json());
...
-                    .sort()
-                    .value();
+            decompress(body)
+                .then((files: File[]) => {
+                    const modules = _(files)
+                        .filter({ type: 'directory' })
+                        .map('path')
+                        .filter(
+                            directory =>
+                                !!find(
```
NorbertAmartus, line 52: ### An optional comment
We could use here the method called `some`, which is the native `Array` method or this which comes from the [lodash](https://lodash.com/docs/4.17.15#some) itself

This method is doing the same thing, the difference is that it's returning the `boolean` by default 😁

widgets/blueprints/src/TerraformActions.ts
```diff
@@ -7,7 +7,15 @@ export default class TerraformActions {
         return this.toolbox.getInternal().doPost('/terraform/blueprint', { body });
     }
 
-    doGetTemplateModules(templateZipUrl: string) {
-        return this.toolbox.getInternal().doPost('/terraform/resources', { params: { zipUrl: templateZipUrl } });
+    doGetTemplateModules(templateZipUrl: string, username?: string, password?: string) {
+        let headers;
+        if (username) {
+            headers = { Authorization: `Basic ${btoa(`${username}:${password}`)}` };
+        }
```
NorbertAmartus, line 14:
```suggestion
        const headers = username ?
            { Authorization: `Basic ${btoa(`${username}:${password}`)}` } :
            undefined;
```

What would you think about the suggestion above?
Just wanted to comment on that, so that we wouldn't have a situation when someone would override the `let` variable, when it shouldn't be necessary 😉 
(currently the `headers` variable is being assigned only 2 values. Because of that I thought that we could reassign it as a `const` 😉 )

Please let me know what you think

widgets/blueprints/src/TerraformModal.tsx
```diff
@@ -287,14 +289,22 @@ export default function TerraformModal({
...
+        setUrlAuthentication(checked);
+        if (!checked) {
+            clearUsername();
+            clearPassword();
+        }
+    }
+
     function handleTemplateUrlBlur() {
-        if (!Stage.Utils.Url.isUrl(templateUrl)) {
+        if (!Stage.Utils.Url.isUrl(templateUrl) || (urlAuthentication && (!username || !password))) {
```
NorbertAmartus, line 301:
```suggestion
    const isAuthenticated = urlAuthentication && (!username || !password);
    if (!Stage.Utils.Url.isUrl(templateUrl) || isAuthenticated ) {
```

We could extract the variable from the `if`, so that we don't have so much logic inside of it
Also, this change would make the function more readable 💪

NorbertAmartus (commented):
Nice one!
I've added some minor comments, but overall everything looks neat! 🎉

backend/routes/Terraform.ts
kubama: Done

widgets/blueprints/src/TerraformActions.ts
kubama: > Just wanted to comment on that, so that we wouldn't have a situation when someone would override the let variable, when it shouldn't be necessary 😉

There is nothing easier than changing `const` to `let` in case anybody wanted to reassign the variable, I don't think `const` actually protects from anything. It's usage is enforced by eslint rule in all cases where it can be used in place of `let`, so in terms of understanding the code it doesn't have the semantics of "don't reassign".
But even if it did, why do you think we should try to guard against reassigning? I think it's impossible to say if it would be (or not) a desirable thing without knowing what the future change would be.

widgets/blueprints/src/TerraformModal.tsx
kubama: Done, although I used a more accurate variable name.

widgets/blueprints/src/TerraformActions.ts
NorbertAmartus: You may be right about it
But still I think that there exists problem related to using `let` and assigning a single value to it ... just in another block of code 😁 
I think that this operation may be confusing, because why should we make seperate block lines just to assign one value to a variable?

Currently the situation is not as confusing, because we keep the lines underneath each other
But still I think that the question "where is the value of that `headers`?" may be unnecessarily being asked by a developer to himself

NorbertAmartus: But I think that's not the "deal breaker", so if you'd find that modification unnecessary, we may skip it
Just let me know 😉

kubama: > I think that this operation may be confusing, because why should we make seperate block lines just to assign one value to a variable?

Yes, this is a good point after all. Updated.

NorbertAmartus: I see, that you've added changes related to this issue (thanks for addressing it! 💪 )
Because of that, I'm resolving this discussion

NorbertAmartus (approved):
Thanks for addressing all of the discussions, looks neat! 💪

backend/routes/Terraform.ts
```diff
@@ -22,47 +22,50 @@ router.use(bodyParser.json());
...
+                if (zipRes) {
+                    const statusCode = zipRes.statusCode >= 400 && zipRes.statusCode < 500 ? zipRes.statusCode : 400;
+                    res.status(statusCode).send({
+                        message: `The URL is not accessible - Error ${zipRes.statusCode} ${
+                            STATUS_CODES[zipRes.statusCode]
+                        }`
+                    });
+                } else {
+                    res.status(400).send({
+                        message: `The URL is not accessible`
```
qooban, line 39: Minor:

```suggestion
                        message: 'The URL is not accessible'
```

NorbertAmartus (approved):

backend/routes/Terraform.ts
```diff
@@ -22,47 +22,50 @@ router.use(bodyParser.json());
...
-            .catch((decompressErr: any) => {
-                logger.error(`Error while decompressing zip file: ${decompressErr}`);
-                res.status(400).send({ message: 'The URL does not point to a valid ZIP file' });
-            });
-    });
+                    if (modules.length) res.send(modules);
+                    else res.status(400).send({ message: "Couldn't find a Terraform module in the provided package" });
+                })
+                .catch((decompressErr: any) => {
+                    logger.error(`Error while decompressing zip file: ${decompressErr}`);
```
qooban, line 64: Optional:

I'm not sure if with such construct it will be rendered properly (I didn't check what actually `decompress` throws). I remember seeing sometimes "[object Object]" to be displayed in such cases.

Maybe we should go with:
```suggestion
                    logger.error('Error while decompressing zip file:', decompressErr);
```
? (see https://github.com/cloudify-cosmo/cloudify-ui-common/blob/bf4f27da339fa417570926ac8021ba25e17a7105/backend/logger.js)

It's optional, just a suggestion, or maybe something to check deeper. I don't have strong opinion here.

backend/test/routes/Terraform.test.ts
```diff
@@ -43,11 +43,16 @@ describe('/terraform/blueprint endpoint', () => {
...
-        nock(/test/)
+        const authorizationHeaderValue = 'auth';
+
+        nock(/test/, { reqheaders: { authorization: authorizationHeaderValue } })
             .get(`/test.zip`)
             .reply(200, readFileSync(resolve(__dirname, 'fixtures/terraform/template.zip'), null));
 
-        const response = await request(app).post('/console/terraform/resources?zipUrl=http://test/test.zip').send();
+        const response = await request(app)
+            .post('/console/terraform/resources?zipUrl=http://test/test.zip')
```
qooban, line 53: Wouldn't it be better to add another test instead of changing existing? 
To have 2 tests - one for URL with authorization and the other for URL without authorization?

widgets/blueprints/src/TerraformModal.tsx
```diff
@@ -374,6 +385,35 @@ export default function TerraformModal({
...
+                            <UnsafelyTypedFormField style={{ display: 'inline-block' }} width={5}>
+                                <Form.Input
+                                    disabled={!urlAuthentication}
+                                    value={password}
+                                    onChange={setPassword}
+                                    label={t(`password`)}
+                                    onBlur={handleTemplateUrlBlur}
+                                    required={urlAuthentication}
+                                />
+                            </UnsafelyTypedFormField>
```
qooban, line 416: Can we use `Form.Group` (`UnsafelyTypedFormGroup`) component here?
It might reduce the need of adding custom style for each field component.

test/cypress/integration/widgets/blueprints_spec.ts
```diff
@@ -563,6 +563,38 @@ describe('Blueprints widget', () => {
...
+                    method: 'POST',
+                    pathname: '/console/terraform/resources',
+                    query: { zipUrl: singleModuleTerraformTemplateUrl }
+                },
+                { statusCode: 401 }
+            );
+
+            openTerraformModal();
+
+            cy.getField('URL to your Terraform template (zip or git)')
```
qooban, line 578: Side question:

What actually git means in that context? Can we point to e.g. "https://github.com/cloudify-cosmo/cloudify-stage.git" or "git@github.com:cloudify-cosmo/cloudify-stage.git"?

qooban (commented):
Added few comments. 

Tested quickly UI (without using real URL requiring auth). I think on error dismissal we should clear field errors, see video (the top error message is showing up unexpectedly I think):
https://user-images.githubusercontent.com/5202105/147927264-6904b639-11cb-46ca-a4bd-1d95d6e71616.mp4

test/cypress/integration/widgets/blueprints_spec.ts
kubama: We don't support anything like that on the code level. This label was updated on @MichaelShn's request, maybe worth to clarify with him?

backend/test/routes/Terraform.test.ts
kubama: What the code does and the test checks is blindly passing the authorization header. So, the test for "no authorization" scenario would be just a subset of the existing test (no header defined, no header check). It wouldn't increase code coverage as there is no logic behind. So I don't see a value in explicitly adding it.

backend/routes/Terraform.ts
kubama: Is there any actual advantage? Performance maybe?

backend/routes/Terraform.ts
kubama: I guess it's always safer as you suggested, applied.

backend/routes/Terraform.ts
qooban: Don't know, maybe. When adding that comment I had only consistency in mind.

backend/test/routes/Terraform.test.ts
qooban: OK, let's keep it as is then.

widgets/blueprints/src/TerraformModal.tsx
kubama: Works perfectly 🙂.

kubama:
> I think on error dismissal we should clear field errors

Fixed

test/cypress/integration/widgets/blueprints_spec.ts
qooban: Yep, started discussion with Michael on Slack.

qooban (approved):
Looks good 👍🏻 

Thanks for addressing the comments.

PR #1730, opened 2022-01-03
RD-3731: Hidded help popup when none of the help informations should …

NorbertAmartus:
## Description

Changes consists of returning `null`, instead of `<div />` , when none of the HelpProperties should be shown.

### Before 
Bez tytułu

### After 
localhost_4000_console_page_services

## Checklist

## Tests

 

[System tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1739/) has been executed
Additionally, I've run `deployments_view_spec.ts` cypress test locally

## Documentation

N/A


NorbertAmartus:
Just a note regarding the [ongoing refactors](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI):

Regarding migrating files to ts (upon which I've been working on), [file](https://github.com/cloudify-cosmo/cloudify-stage/blob/c2c933c0563a1f9822a037dc6c330c973c55d8b3/widgets/common/src/InputsUtils.tsx) (which was modified by me) is large, that's why I didn't migrate it

widgets/common/src/InputsUtils.tsx
```diff
@@ -197,11 +203,11 @@ class InputsUtils {
                 />
                 <HelpProperty
                     name={!_.isUndefined(defaultValue) ? 'Default Value' : 'Example'}
-                    show={!_.isUndefined(defaultValue) || !_.isUndefined(dataType)}
+                    show={showDefaultValue}
                     value={<ParameterValue value={example} />}
                 />
```
qooban, line 208: Optional: 

I'm wondering what should be correct names here. This component shows either a default value of an input or an example - object shape (created based on the passed `dataType`). In L206 we use `showDefaultValue` name, in the other (L207) we use `example` ;-)
Maybe (thus optional) something to think about and fix, but not a blocker I suspect.

qooban (approved):
Looks good! 👍🏻 

Added just one optional comment.

widgets/common/src/InputsUtils.tsx
NorbertAmartus: Hmm, then I think that we may stick into the `showExample` variable as the `default value` may be too considered as an example :)

Please let me know, what do you think about it 😉

qooban: OK, let's go with `showExample` then 👍🏻

NorbertAmartus: Done 😉

qooban (approved):
Looks good 👍🏻

PR #1735, opened 2022-01-06
RD-3623 Change the DropDown Button value to Deploy on selecting Deploy

leon-good-life:
## Description
This PR changes the functionality of the drop-down so "Deploy" or "Install" values can be selected first. And then appear in a button. This PR is the follow up to: https://github.com/cloudify-cosmo/cloudify-stage/pull/1723

## Screenshots / Videos
https://user-images.githubusercontent.com/17104212/148372225-6ffe0021-2bb4-4f3f-829b-b39125c2862d.mov

## Checklist

- [x] My code follows the UI Code Style.
- [x] I verified that all tests and checks have passed.
- [x] I verified if that change requires a documentation update.
- [x] I added proper labels to this PR.

## Tests
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1763/pipeline

 

## Documentation
Another ticket: https://cloudifysource.atlassian.net/browse/RD-3799

PR #1736, opened 2022-01-10
RD-2866 Optimizing tenants fetching

NorbertAmartus:
## Description

Changes consists of modyfing the current logic underneath the fetching tenants mechanism.
Currently the tenants list for the menu dropdown will be downloaded only when the user would click the corresponding menu item.   

Moreover, regarding our [internal code style guidelines](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/Cloudify+UI+Code+Style), I've migrated:
- [one file](https://github.com/cloudify-cosmo/cloudify-stage/pull/1736/commits/4bcc49326a70d83db2ec035e39cee89e4fc8dfbe) to the typescript
- [files](https://github.com/cloudify-cosmo/cloudify-stage/pull/1736/commits/3052385dfaa0dee3300f36a10577f9b3ccf5df97), on which I've been working on, to use translation file instead of using just a plain magic strings as labels

### After changes
CPT2201110837-621x367

## Checklist

## Tests

 

[System tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1776/) has been executed.
(2 tests are failing due to the problems which were detected on the master branch)

Additional [test scenario](https://github.com/cloudify-cosmo/cloudify-stage/pull/1736/commits/2f978fac2f461cb7dab7759ad9644b22d3abbd1d) has been added 🚀.

## Documentation

N/A


NorbertAmartus:
## Just a note:
I've updated the repository documentation (via `npm run docWidgets` command), because some pipelines were failing due to the changes, that [has been made](https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2066/files) 2 days ago on the [docs.getcloudify.org](https://github.com/cloudify-cosmo/docs.getcloudify.org) branch, without mirroring the corresponding changes inside this repository

widgets/tenants/src/CreateModal.tsx
```diff
@@ -31,27 +36,26 @@ export default function CreateModal({ toolbox }) {
...
     return (
         <Modal trigger={addButton} open={isOpen} onOpen={doOpen} onClose={doClose}>
             <Modal.Header>
-                <Icon name="add user" /> Add tenant
+                <Icon name="add user" /> {t('header')}
             </Modal.Header>
 
             <Modal.Content>
                 <Form loading={isLoading} errors={errors} onErrorsDismiss={clearErrors}>
+                    {/* @ts-ignore Form.Field is not migrated yet */}
```
qooban, line 54: Optional:

Alternatively you can use `UnsafelyTypedFormField`.

widgets/tenants/src/TenantsTable.tsx
```diff
@@ -8,6 +8,8 @@ import TenantDetails from './TenantDetails';
 import UsersModal from './UsersModal';
```
qooban, line 8: Optional:

Am I assuming properly that it's hard to migrate it to TS at once?

test/cypress/integration/user_menu_spec.ts
```diff
@@ -48,4 +50,22 @@ describe('User Menu', () => {
...
+        cy.log('Showing spinner while fetching data');
+        cy.contains('Loading');
+
+        cy.log('New tenant is visible in the dropdown');
+        cy.contains(newTenantName);
+    });
+
+    after(() => {
+        cy.deleteTenant(newTenantName);
+    });
```
qooban, line 70: Do we need it? 

1. Having `deleteAllUsersAndTenants()` in `before` hook seems to cover tenant deletion or am I wrong?
2. Whenever it's possible we should avoid using `after` - ref.: https://docs.cypress.io/guides/references/best-practices#Using-after-or-afterEach-hooks

test/cypress/integration/user_menu_spec.ts
```diff
@@ -48,4 +50,22 @@ describe('User Menu', () => {
...
+
+        cy.log('Adding new tenant');
+        cy.addTenant(newTenantName);
+        cy.contains(defaultTenantName).click();
+
+        cy.log('Showing spinner while fetching data');
+        cy.contains('Loading');
+
+        cy.log('New tenant is visible in the dropdown');
+        cy.contains(newTenantName);
```
qooban, line 65: What do you think about wrapping it inside something like `cy.contains('Tenant selection').parent().within(() => { ... })` Didn't check the correctness. My point is to be more specific which may let us avoid accidentally getting another loader with `cy.contains('Loading');`.

test/cypress/integration/user_menu_spec.ts
```diff
@@ -20,7 +22,7 @@ describe('User Menu', () => {
         cy.activate()
             .deleteAllUsersAndTenants()
             .addUser(nonAdminUsername, nonAdminPassword, false)
-            .addUserToTenant(nonAdminUsername, 'default_tenant', 'viewer');
+            .addUserToTenant(nonAdminUsername, defaultTenantName, 'viewer');
```
qooban, line 25: What do you think about extracting it globally or better - reusing what we already have in `app`?
There's `DEFAULT_TENANT` constant in `app/utils/consts.ts`, so that's a good candidate to be used here.
Probably you could extend `test/cypress/support/app_commons.ts`.

There are other places using "default_tenant" string in system tests, seems reasonable to update them as well.

app/components/sidebar/TenantSelection.tsx
```diff
@@ -13,60 +13,72 @@ import SideBarDropdownItem from './SideBarDropdownItem';
...
-                        text={tenant.name}
-                        selected={tenant.name === selectedTenant}
-                        active={tenant.name === selectedTenant}
-                        onClick={() => onTenantSelected(tenant)}
-                    />
-                ))}
+                {isLoadingTenants ? (
+                    <StyledLoaderWrapper>
+                        <Loading />
+                    </StyledLoaderWrapper>
```
qooban, line 60: Optional:

Is this wrapper necessary? Wouldn't it be simpler to just have something like:
```suggestion
                    <Loading style={{ height: 80 }} />
```
I assume with a proper answer we need to wait for your investigation on React styling, am I correct?

qooban (commented):
Looks good in general. Added some minor comments.
Tested UI, looks good.

Reg. widget docs update - @leon-good-life is aware of the issue. He also added a commit for that in #1735 . So the fix will be part of either yours or his PR.

widgets/tenants/src/CreateModal.tsx
NorbertAmartus: Ooo, I'll try to do so
Thanks for the tip! 😉

widgets/tenants/src/TenantsTable.tsx
NorbertAmartus: Yeah, as I've quickly investigated it may take some time time to migrate this file as there are some types (like `tenant`) to be checked and implemented in a few places   
That's why I've migrated a smaller file 😁

test/cypress/integration/user_menu_spec.ts
NorbertAmartus: I think, that's a neat idea!
I'll most certainly add it

Thanks for pointing that out! 🎉

app/components/sidebar/TenantSelection.tsx
NorbertAmartus: As for styling the component with the `styled-components`, it's required (without the wrapper, it's hard to determine the specifity of the css rule, without the `!important` flag)   

But I'm not blocked to use the coding methods which are related to our [UI Code Style](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/Cloudify+UI+Code+Style)

And because of that, and the fact the we are introducing here unnecessary wrapper (which you've pointed out 💪 ), I'll introduce the solution suggested by you 😉

NorbertAmartus: Done 🎉

test/cypress/integration/user_menu_spec.ts
NorbertAmartus: Hmm, that sounds like a great idea!
So I'll try to implement one of the solutions that you've proposed 🚀

test/cypress/integration/user_menu_spec.ts
NorbertAmartus: Done 🎉

widgets/tenants/src/CreateModal.tsx
NorbertAmartus: Done 🎉

test/cypress/integration/user_menu_spec.ts
NorbertAmartus: I've found all the strings inside the tests files from the `test/cypress` directory and applied to them  `DEFAULT_TENANT ` const 😉 

(I've did that over extending the `app_comons.ts` file, because in most cases the consts are being used from the `app/utils/consts.ts`, which you've pointed out 😉 )

There are some left occurences of the `default_tenant` inside the backend tests but I've found out that importing the consts from the different root file than the `backend` is a little more complicated. That's why I didn't add mentioned modifications across that area

test/cypress/integration/user_menu_spec.ts
qooban: If that specific tenant is used only by that test and having it present does not affect other tests, then I suggest to remove `after` hook.

qooban (commented):
Just one thing - regarding one of your comments:
> (I've did that over extending the app_comons.ts file, because in most cases the consts are being used from the app/utils/consts.ts, which you've pointed out wink )

I'm not sure I understand that sentence.

Anyways, in general I don't have a strong opinion here, but I thought that it may be wise to not have too many files referencing application source code from the test code (that was my thinking when I was introducing `app_commons.ts`). That could help us to keep some kind of interface between test code and application code. It might make it easier (lower number of files affected) to do any changes application code used by test code. But as I said - I don't have strong opinion here. 

I didn't dig deep enough to be fully sure my approach is better, but what I think we should do is to be consistent - either extend `app_commons` or remove it completely and import constants directly. What do you think?

test/cypress/integration/user_menu_spec.ts
NorbertAmartus: Alright 😉

NorbertAmartus: Done 🚀

NorbertAmartus:
I think, in that case we could remove the `app_commons.ts` file to have just one file containing consts
This file would be our Single Source of Truth, regarding the consts variables 😉

qooban (commented):
> I think, in that case we could remove the app_commons.ts file to have just one file containing consts
> This file would be our Single Source of Truth, regarding the consts variables wink

In that case, let's remove it and keep the code consistent.

NorbertAmartus:
Mentioned file has been removed and corresponding imports has been updated 🚀

qooban (approved):
Thanks for addressing all comments.
If all system tests are passing (except known failing tests), then feel free to merge it 👍🏻

PR #1745, opened 2022-01-14
RD-3944 Updated Terraform wrapper blueprint

kubama:
## Description
Updated Terraform wrapper blueprint

## Screenshots / Videos
N/A

## Checklist

## Tests

- Verified blueprints spec locally
- Executed [Stage-UI-System-Test/1829](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1829/) (18.01.2022, 11:22) - only `blueprints_spec` - PASSED

 

## Documentation
Separate ticket


backend/routes/Terraform.ts
```diff
@@ -70,9 +70,11 @@ router.post('/resources', (req, res) => {
 
 router.post('/blueprint', (req, res) => {
     const {
+        blueprintName,
```
NorbertAmartus, line 73: Every other field have a default value    
Because of that, I just wanted to confirm, that providing the default value for the `blueprintName` property wouldn't make any difference in the functionality 😁

NorbertAmartus (commented):
Looks good 💪 
Just commented on one thing, to make sure that it should be a proper behavior

widgets/blueprints/src/TerraformModal.tsx
```diff
@@ -276,6 +290,14 @@ export default function TerraformModal({
             });
 
             setProcessPhase('upload');
+
+            if (urlAuthentication) {
+                const secretActions = new Stage.Common.SecretActions(toolbox);
+                const { defaultVisibility } = Stage.Common.Consts;
+                await secretActions.doCreate(`${blueprintName}.username`, username, defaultVisibility, false);
+                await secretActions.doCreate(`${blueprintName}.password`, password, defaultVisibility, false);
```
qooban, line 298: I'm wondering how will those calls behave when the secret is already present?

backend/routes/Terraform.ts
qooban: There's also `urlAuthentication` without default value.

But looking at how `TerraformModal` component and blueprint template (`blueprint.ejs`) has changed recently, I think that actually we can go the other way - make:
```
        terraformVersion = '',
        terraformTemplate = '',
        resourceLocation = '',
```
to not have default values as the resulting blueprint won't make much sense without them.

qooban (commented):
Added two comments. I did not validate how it works in the UI.

qooban:
As per Alex request I'm merging this PR without addressing comments right now.
@kubama Please take a look at discussions started in this PR when you are back. I'll leave you a message on Slack about it as well.

backend/routes/Terraform.ts
kubama: Will be addressed in https://github.com/cloudify-cosmo/cloudify-stage/pull/1771

widgets/blueprints/src/TerraformModal.tsx
kubama: > 1. I'm wondering how will those calls behave when the secret is already present?

They will throw an exception which will be handled by surrounding `catch` block and error message will be displayed - all fine here IMO.

> 2. I think that we also need to extend system tests to check if secrets were created properly.

Will be addressed in https://github.com/cloudify-cosmo/cloudify-stage/pull/1771

PR #1758, opened 2022-01-18
RD-3828 Change rules for setting Secure flag on token cookie

qooban:
## Description
As a result of discussion in RD-3828 a change in `Auth.ts` had to be implemented to satisfy different environment needs. `getCookieOptions` function had to be updated to check both - `X-Scheme` and `X-Force-Secure` headers - to determine if token cookie should have `Secure` flag set.
In addition to that change few backend files were migrated to TypeScript.

## Screenshots / Videos
N/A

## Checklist

## Tests
1. Updated backend unit tests.
2. [Stage-UI-System-Test/1834](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1834/tests) - only one test failing, it's reported already as RD-3760, not related to this PR, so I'm considering it as PASSED

## Documentation
N/A

### DIFF
--- a/backend/handler/AuthHandler.ts
+++ b/backend/handler/AuthHandler.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck File not migrated fully to TS
 import _ from 'lodash';
 import { EDITION } from '../consts';
 import { jsonRequest } from './ManagerHandler';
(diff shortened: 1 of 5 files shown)


backend/routes/Auth.ts
```diff
@@ -16,13 +15,13 @@ const logger = getLogger('Auth');
 router.use(bodyParser.json());
 router.use(bodyParser.urlencoded({ extended: true }));
 
-function getCookieOptions(req) {
-    const httpsUsed = req.header('X-Scheme')?.includes('https');
-    return { sameSite: 'strict', secure: httpsUsed };
+function getCookieOptions(req: Request) {
+    const httpsUsed = req.header('X-Scheme') === 'https' || req.header('X-Force-Secure') === 'true';
```
qooban, line 19: This line is the only functional change provided by this PR.

The rest of the changes are either test updates or TypeScript migrations.

backend/handler/AuthHandler.ts
```diff
@@ -8,44 +7,80 @@ import { getLogger } from './LoggerHandler';
...
+        {},
+        {
+            'saml-response': samlResponse
+        }
+    );
 }
 
-export function getAndCacheConfig(token) {
-    return jsonRequest('GET', '/config', {
+export function getAndCacheConfig(token: string | undefined) {
```
NorbertAmartus, line 82: Optional:
We can shorten that typing with a `?`:  syntax

```suggestion
export function getAndCacheConfig(token?: string) {
```

NorbertAmartus (dismissed):
Looks good 💪 
I've been searching for the built-in consts/enums, which could be used to replace the string `https` but it seems like that kind of fireworks doesn't exists yet 😁   

I've left just one optional comment

Thank you for taking your time to migrate multiple files to the typescript 🎉
Because of you, our weekly TS chart will grow faster as ever! 😎

backend/handler/AuthHandler.ts
qooban: Sure, changed.

NorbertAmartus (approved):

PR #1759, opened 2022-01-18
RD-3978: Terraform Modal modifications

NorbertAmartus:
## Description

Changes are consisting of adding an additional validation to the TerraformModal's static variable values.
Along with it comes the modifications related to the labels.

### Before
Naming in the table - before
Errors in the form - before

### After
Naming in the table - after
Errors in the form - after

## Checklist

## Tests

 

[System tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1840) has been executed
(2 of them has failed, which is currently known to be expected on the master branch. Fix for them has been introduced in a [seperate PR](https://github.com/cloudify-cosmo/cloudify-stage/pull/1760))

## Documentation

N/A


widgets/blueprints/src/TerraformModal.tsx
```diff
@@ -50,6 +50,7 @@ function getDynamicTableDropdown(options: DropdownProps['options']) {
 }
 
 const cloudifyResourceRegexp = /^[a-zA-Z][a-zA-Z0-9._-]*$/;
+const cloudifyStaticValueRegexp = /^[a-zA-Z0-9._-]*$/;
```
qooban, line 53: Actually I think the better name would be something relating to Terraform as static value is directly passed to `resource_config` property in `cloudify.nodes.terraform.Module` node in the resulting blueprint. In the other cases we are using the value as Cloudify resource (either secret key or input name).

```suggestion
const terraformResourceRegexp = /^[a-zA-Z0-9._-]*$/;
```

app/translations/en.json
```diff
@@ -870,15 +870,17 @@
...
                     "environmentVariable": {
                         "nameMissing": "Please provide environment variable name",
                         "nameInvalid": "Please provide valid environment variable name - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
                         "nameDuplicated": "Environment variables must be unique, duplicates are not allowed",
                         "sourceMissing": "Please provide environment variable source",
                         "valueMissing": "Please provide environment variable value",
-                        "valueInvalid": "Please provide valid environment variable value - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed"
+                        "valueInvalid": "Please provide valid environment variable value - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
+                        "staticValueInvalid": "Please provide valid environment variable value - allowed only letters, digits and the characters \"-\", \".\" and \"_\".  If special characters or complex structures are needed please provide this value inside a secret"
```
qooban, line 883: Similar small issue with double space 🙂

widgets/blueprints/src/TerraformModal.tsx
```diff
@@ -211,7 +212,21 @@ export default function TerraformModal({
...
+            ) {
+                formErrors[`${errorPrefix}ValueInvalid`] = tVariableError('staticValueInvalid');
+            }
+            if (
+                find(
+                    variablesList,
+                    variable =>
+                        !isEmpty(variable.value) &&
+                        variable.source !== 'static' &&
+                        !variable.value.match(cloudifyResourceRegexp)
```
qooban, line 229: Regarding our discussion on Slack - I think it asks for a small refactoring as it becomes harder and harder to read, but I think it is not critical, so let's keep it as is for now.

qooban (commented):
Added some minor comments.

app/translations/en.json
```diff
@@ -870,15 +870,17 @@
                         "nameDuplicated": "Variables must be unique, duplicates are not allowed",
                         "sourceMissing": "Please provide variable source",
                         "valueMissing": "Please provide variable value",
-                        "valueInvalid": "Please provide valid variable value - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed"
+                        "valueInvalid": "Please provide valid variable value - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
+                        "staticValueInvalid": "Please provide valid variable value - allowed only letters, digits and the characters \"-\", \".\" and \"_\".  If special characters or complex structures are needed please provide this value inside a secret"
```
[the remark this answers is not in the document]
NorbertAmartus: Yup, you're right 😁

widgets/blueprints/src/TerraformModal.tsx
NorbertAmartus: Okay 😉

app/translations/en.json
NorbertAmartus: Same label, same problems 😁 
I'll resolve it in the upcoming commit

widgets/blueprints/src/TerraformModal.tsx
NorbertAmartus: Hmm, that might be a valuable hint
But I think that this perspective is related to the other parts of the system     

It may get confusing, when somebody would have to get into this component to modify the validation for the `static values` and see that under the hood they are just a terraform resource   

I think that the knowledge about what is happening under the hood is valuable, but I don't think that connecting the names across distant system parts may be a good option 🤔 

What do you think? 🤔

NorbertAmartus:
@qooban 
> BTW No tests to add/update?

Inside the [blueprints_spec.ts](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/test/cypress/integration/widgets/blueprints_spec.ts#L465) file there are some test regarding this functionality   
During my work, I've checked if they need some updates, and they were good as they were 😁 

I was thinking about adding an extra one just to be more precisely "validate newly added validation", but then the thought fled away 😁    

I'll try to add a test scenario in the morning, thanks for rising that topic 😎

app/translations/en.json
NorbertAmartus: Whitespace has been eradicated 😎

widgets/blueprints/src/TerraformModal.tsx
qooban: On one hand yes (good argument with connecting names across distant parts of the system), on the other no (the whole feature is about converting cloudify resources into terraform resources). 

Anyways, I'm OK with keeping `StaticValue` part, but I think we shouldn't use `cloudify` prefix in the name, let's call it just `staticValueRegexp`.

NorbertAmartus: Okay, I'm perfectly fine with it 😉 
I'll add it with the next commit

Thanks for showing your perspective upon that matter! 🎉

NorbertAmartus: Commit with the changes has arrived 🎉

NorbertAmartus:
Test scenario has been added 😉

test/cypress/integration/widgets/blueprints_spec.ts
```diff
@@ -50,6 +50,11 @@ describe('Blueprints widget', () => {
         return cy.get(`#blueprintsTable_${blueprintName}`);
     }
 
+    function selectTerraformVariableSource(source: string) {
+        cy.get('td:eq(1) .selection').click();
+        cy.contains('.item', source).click();
+    }
```
qooban, line 56: Shouldn't we put it inside 'should open upload from Terraform module modal and' context?
Then I think name change `selectVariableSource` -> `selectTerraformVariableSource` won't be necessary.

qooban (commented):
Thanks for addressing already discussed issues and adding more tests. I've added just one more.
Unfortunately I don't have time to validate the UI and tests in my environment this time.

test/cypress/integration/widgets/blueprints_spec.ts
NorbertAmartus: Hmm, you're right!

I was lagged upon the thought, that every shared method should be at the top of the test file (after being inside multiple files it was sounding quite natural 😁)

Thanks for mentioning it, I'll modify it right away! 🚀

NorbertAmartus: Function has been renamed and relocated 🎉

qooban (approved):
Thanks for addressing all the comments. LGTM! 👍🏻

PR #1760, opened 2022-01-19
RD-2866: Fixed tests related to the TenantSelection component

NorbertAmartus:
## Description

Added intercepting of the tenants into the tests logic, as it wasn't implemented before

## Screenshots / Videos

N/A

## Checklist

- [ ] I verified if that change requires a documentation update.

## Tests

 

As the changes are only related to the `TenantSelection_spec.tsx`, I've locally executed `npm run cy:open-ct` command and successfully run (multiple times) the `TenantSelection_spec.txt` test scenarios

## Documentation

N/A

### DIFF
--- a/test/cypress/components/app/sidebar/TenantSelection_spec.tsx
+++ b/test/cypress/components/app/sidebar/TenantSelection_spec.tsx
@@ -1,25 +1,25 @@
 import React from 'react';
-import type { RouteMatcher } from 'cypress/types/net-stubbing';
 
 import '../../initAppContext';
 import TenantSelection from 'app/components/sidebar/TenantSelection';
 import { mountWithProvider } from '../../utils';
 
-const tenantListRouteMatcher: RouteMatcher = {
-    pathname: '/console/sp/tenants',
-    query: { _include: 'name', _get_all_results: 'true' }
-};
-
-const fetchTenants = (tenants: any[]) => {
-    cy.intercept(tenantListRouteMatcher, {
-        items: tenants
-    });
+const mockTenantsResponse = (tenants: any[]) => {
+    cy.intercept(
+        {
+            pathname: '/console/sp/tenants',
+            query: { _include: 'name', _get_all_results: 'true' }
+        },
+        {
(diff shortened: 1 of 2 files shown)


test/cypress/components/app/sidebar/TenantSelection_spec.tsx
```diff
@@ -1,13 +1,27 @@
...
+const tenantListRouteMatcher: RouteMatcher = {
+    pathname: '/console/sp/tenants',
+    query: { _include: 'name', _get_all_results: 'true' }
+};
+
+const fetchTenants = (tenants: any[]) => {
+    cy.intercept(tenantListRouteMatcher, {
+        items: tenants
+    });
+};
```
qooban, line 17: 1. Isn't having `tenantListRouteMatcher` const extracted and not used anywhere else than in `fetchTenants` kind of overengineering? :-) IMHO it is unnecessary, but I'm open to hear reasoning behind that. Performance? :-)
2. `fetchTenants` seems not accurate as it is just intercepting a call and it does not fetch anything. Looking at the implementation it is more like mocking tenants response. Am I right?

NorbertAmartus: Ad. 1
This time is not a performance, but just something unnecessary 😁
While doing some experiments, I've extracted it as a seperate variable (not to copy this route matcher every time).
As it didn't seem to cause any harm, I didn't remove it

But as you pointed out, it may be unnecessary, so I'll remove it 

Ad 2.
Yeah, you are right
I was locked on the thought, that the request is being sent to fetch the tenants
Buuuuut I was thinking from the component perspective, rather than from the perspective of the test itself 😁 

So I'll modify that too!
Thanks for the comments! 🎉

NorbertAmartus: Modifications has been added 🎉

qooban (approved):
Thanks for addressing the comment 👍🏻 Looks better IMHO.

PR #1761, opened 2022-01-19
RD-4032 Add token authentication to backend API routes

qooban:
## Description
The goal of this PR is to provide token-based authentication for stage backend API routes that missed it and that does not require functional changes in the source code to have them with authentication. Those are: `/external`, `/github` and partially `/plugins` routes.

The following routes are left without authentication requirement (due to the current application architecture):
1. `/auth` - some endpoints defined inside that route have to be accessible without authentication (e.g. `/auth/login` used on login credentials submission)
2. `/sp` - it requires a lot more changes in source code and system test to make it work (it'll be part of RD-3827), but it works as a proxy to Cloudify REST API, so there will be token check anyway
3. `/style` - it is used to get stylesheet, e.g. in Login page
4. `/config` - it is used to get white-labelling configuration required e.g. in Login page

In addition to the main change the following was implemented:
1. Unified way of getting authentication middleware for routes
2. Improved backend test coverage by introducing tests for `/file` and `/wb` routes 
3. Improved `waitForExecutionToEnd` custom command to be more flexible
4. Improved `user_spec` test to be more robust and less prone to timing-issues
5. Migrated some files to TypeScript

## Screenshots / Videos
N/A

## Checklist

## Tests
1. [Stage-UI-System-Test/1846](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1846/tests) (19.01.2022 14:58 CEST) - only `page_spec` failed
2. [Stage-UI-System-Test/1849](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1849/tests) (20.01.2022 08:58 CEST) - only `flows/user_spec` failed
3. [Stage-UI-System-Test/1854](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1854/tests) (20.01.2022 13:15 CEST) - PASSED (except `TopologySelection` component, but that is reported as RD-4040, so not related to changes coming from this PR)
4. [Stage-UI-System-Test/1889](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1889/tests) (25.01.2022 08:15 CEST) - PASSED

## Documentation
At this point I don't see any place requiring update in the documentation.


backend/test/routes/File.test.ts
```diff
@@ -0,0 +1,33 @@
+import request from 'supertest';
+import app from 'app';
+import path from 'path';
+
+describe('/file endpoint', () => {
```
qooban, line 5: Added test to satisfy backend test coverage check 🙂

backend/test/routes/WidgetBackend.test.ts
```diff
@@ -0,0 +1,24 @@
+import request from 'supertest';
+import app from 'app';
+import { callService } from 'handler/BackendHandler';
+
+jest.mock('handler/BackendHandler');
+(<jest.Mock>callService).mockImplementation(
+    (_serviceName, _method, _req, res, _next) => new Promise(() => res.send(''))
+);
+
+describe('/wb endpoint', () => {
```
qooban, line 10: Added test to satisfy backend test coverage check 🙂

test/cypress/integration/flows/user_spec.ts
```diff
@@ -65,6 +65,7 @@ describe('User flow', () => {
         cy.clickButton('Execute');
         cy.get('.modal', { timeout: minutesToMs(1) }).should('not.exist');
 
+        cy.waitForExecutionToEnd('install', { deploymentDisplayName: resourceName });
```
qooban, line 68: Not related strictly to the PR, but I faced the problem in this test when running system test and decided to apply quick fix to improve test robustness.

backend/app.ts
```diff
@@ -95,35 +96,37 @@ app.use(`${contextPath}/userData/${translationsOverrides}`, (req, res) => {
...
-app.use(`${contextPath}/plugins`, Plugins);
-app.use(`${contextPath}/source`, SourceBrowser);
-app.use(`${contextPath}/sp`, ServerProxy);
 app.use(`${contextPath}/style`, Style);
-app.use(`${contextPath}/templates`, Templates);
-app.use(`${contextPath}/terraform`, Terraform);
-app.use(`${contextPath}/ua`, UserApp);
-app.use(`${contextPath}/wb`, WidgetBackend);
-app.use(`${contextPath}/widgets`, Widgets);
+app.use(`${contextPath}/sp`, ServerProxy);
```
NorbertAmartus, line 129: Just a comment:
We could extract it in the future (as some sort of `AppRouter` or `RootRouter`), so that we would have a dedicated page for the application routing   
Additionally, currently it may be not obvious when this kind of root routing is being stored, but after extracting it I think that it would be more obvious, as it would be visible even from the file system 😁

NorbertAmartus (approved):
Looks amazing! 🚀    
Lots of losing weight across the files, which is great 🎉 

Added some optional comments, but as they are not deal breakers, I'm approving this PR.

Nice one! 💪

test/cypress/support/executions.ts
```diff
@@ -39,8 +39,15 @@ const commands = {
...
-        const deploymentExecutionsUrl = `executions?_include=id,workflow_id,ended_at&deployment_id=${deploymentId}&workflow_id=${workflowId}`;
+    waitForExecutionToEnd: (
+        workflowId: string,
+        options: { deploymentId?: string; deploymentDisplayName?: string } = {}
+    ) => {
+        const { deploymentId, deploymentDisplayName } = options;
+
+        let deploymentExecutionsUrl = `executions?_include=id,workflow_id,ended_at&workflow_id=${workflowId}`;
+        if (deploymentId) deploymentExecutionsUrl += `&deployment_id=${deploymentId}`;
+        if (deploymentDisplayName) deploymentExecutionsUrl += `&deployment_display_name=${deploymentDisplayName}`;
```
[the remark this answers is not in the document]
qooban: Did you have something like that in mind (using IIFE):
```
const deploymentExecutionsUrl = function getDeploymentExecutionsUrl() {
    let url = `executions?_include=id,workflow_id,ended_at&workflow_id=${workflowId}`;
    if (deploymentId) url += `&deployment_id=${deploymentId}`;
    if (deploymentDisplayName) url += `&deployment_display_name=${deploymentDisplayName}`;
    return url;
}();
```
?

If not, but extracting external function, then we would need to extract type for it (options) and I think it may be an overkill.

WDYT?

backend/app.ts
qooban: Make sense, I'll add a note with your suggesting in the follow-up issue - RD-3827 and I would leave it as is in this PR.

NorbertAmartus: Sounds great!

test/cypress/support/executions.ts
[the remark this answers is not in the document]
NorbertAmartus: I was thinking about the second solution, but as you've pointed out, there would be a need to extract a type, which would require some extra work      
So I think, that we could proceed without it then 😉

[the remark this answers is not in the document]
qooban: OK 🙂

NorbertAmartus (dismissed):

backend/app.ts
```diff
@@ -95,35 +96,37 @@ app.use(`${contextPath}/userData/${translationsOverrides}`, (req, res) => {
...
+app.use(`${contextPath}/filters`, authenticateWithToken, Filters);
+app.use(`${contextPath}/github`, authenticateWithToken, GitHub);
+app.use(`${contextPath}/maps`, authenticateWithToken, Maps);
+app.use(`${contextPath}/plugins`, authenticateWithToken, Plugins);
+app.use(`${contextPath}/source`, authenticateWithToken, SourceBrowser);
+app.use(`${contextPath}/templates`, authenticateWithToken, Templates);
+app.use(`${contextPath}/terraform`, authenticateWithToken, Terraform);
+app.use(`${contextPath}/ua`, authenticateWithToken, UserApp);
+app.use(`${contextPath}/wb`, authenticateWithToken, WidgetBackend);
+app.use(`${contextPath}/widgets`, authenticateWithToken, Widgets);
```
kubama, line 121: How about creating a map (path -> controller) and using it in a loop here to save on some repeating?

backend/auth/AuthMiddlewares.ts
```diff
@@ -0,0 +1,17 @@
...
+export function authenticateWithCookie(req: Request, res: Response, next: NextFunction) {
+    return authenticateWith('cookie')(req, res, next);
+}
+
+export function authenticateWithToken(req: Request, res: Response, next: NextFunction) {
+    return authenticateWith('token')(req, res, next);
+}
+
+export function authenticateWithSaml(req: Request, res: Response, next: NextFunction) {
+    return authenticateWith('token')(req, res, next);
```
kubama, line 16: A bug? Shouldn't the argument be `'saml'`?

qooban: Great catch! 👍🏻 Definitely. I'll fix it.

backend/app.ts
qooban: Nice idea. I'll try to implement it.

backend/auth/AuthMiddlewares.ts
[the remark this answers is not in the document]
qooban: I'm not sure I understand. Could you elaborate more on that?

Are you thinking of having extended `req` (with `user` field) as a result of such type addition?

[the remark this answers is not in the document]
kubama: I was thinking about something like:
```
type authenticate = (req: Request, res: Response, next: NextFunction) => ReturnType<typeof authenticateWith>;
export const authenticateWithCookie : authenticate = (req, res, next) => {
    return authenticateWith('cookie')(req, res, next);
}
...
```

backend/app.ts
qooban: Done.

backend/auth/AuthMiddlewares.ts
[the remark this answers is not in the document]
qooban: Done.

qooban: Fixed.

qooban:
I have addressed all the PR comments and triggered another system test execution (see **Tests** section).

backend/auth/AuthMiddlewares.ts
```diff
@@ -1,17 +1,19 @@
-import passport from 'passport';
+import { authenticate } from 'passport';
 import type { Request, Response, NextFunction } from 'express';
 
+type AuthenticateMiddleware = (req: Request, res: Response, next: NextFunction) => ReturnType<typeof authenticateWith>;
```
kubama, line 4: Looks like it compiles, however it's not accurate - the middleware function returns `athenticateWith()()`, so it should be something like `ReturnType<ReturnType<typeof authenticateWith>>` (no idea if this is correct syntax).

qooban: Seems like `ReturnType<ReturnType<typeof authenticateWith>>` is valid, so changed.

kubama (approved):

PR #1767, opened 2022-01-21
RD-3979: Modified resourceLocation regexp

NorbertAmartus:
## Description

Changes consists of modifying resourceLocation regexp, so that it may cover the initial slash of the resource path.

## Screenshots / Videos

N/A

## Checklist

## Tests

 

I've executed [System tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1887)

## Documentation

N/A

### DIFF
--- a/test/cypress/integration/widgets/blueprints_spec.ts
+++ b/test/cypress/integration/widgets/blueprints_spec.ts
@@ -660,7 +660,7 @@ describe('Blueprints widget', () => {
                 });
                 cy.clickButton('Execute');
 
-                cy.waitForExecutionToEnd(deploymentId, 'install');
+                cy.waitForExecutionToEnd(deploymentId, 'install', { numberOfRetriesLeft: 120 });
                 cy.getDeployment(deploymentId).then(response => {
                     expect(response.body.latest_execution_status).to.be.equal('completed');
                 });
--- a/widgets/blueprints/src/TerraformModal.tsx
+++ b/widgets/blueprints/src/TerraformModal.tsx
@@ -289,7 +289,7 @@ export default function TerraformModal({
             ) {
                 return resourceLocation;
             }
-            return resourceLocation.replace(/^[^/]*\//, '');
+            return resourceLocation.replace(/^[^/]*[/]?/, '');
         }
 
         try {
--- a/widgets/blueprints/src/TerraformModal.tsx
+++ b/widgets/blueprints/src/TerraformModal.tsx
@@ -289,7 +289,7 @@ export default function TerraformModal({
             ) {
                 return resourceLocation;
             }
-            return resourceLocation.replace(/^[^/]*/, '.');
+            return resourceLocation.replace(/^[^/]*\//, '');
         }
 
         try {
--- a/test/cypress/components/app/sidebar/TenantSelection_spec.tsx
+++ b/test/cypress/components/app/sidebar/TenantSelection_spec.tsx
@@ -4,14 +4,19 @@ import '../../initAppContext';
 import TenantSelection from 'app/components/sidebar/TenantSelection';
 import { mountWithProvider } from '../../utils';
 
-const mockTenantsResponse = (tenants: any[]) => {
+const mockTenantsResponse = (tenants: any[], withDelay = false) => {
+    const defaultDelay = 250;
+
     cy.intercept(
         {
             pathname: '/console/sp/tenants',
             query: { _include: 'name', _get_all_results: 'true' }
         },
         {
-            items: tenants
+            body: {
+                items: tenants
+            },
+            delay: withDelay ? defaultDelay : 0
         }
     );
 };
@@ -19,7 +24,7 @@ const mockTenantsResponse = (tenants: any[]) => {
 describe('TenantSelection', () => {
     it('renders loader when fetching tenants', () => {
         mountWithProvider(<TenantSelection />, { manager: { tenants: { items: [] } } });
-        mockTenantsResponse([]);
+        mockTenantsResponse([], true);
 
         cy.contains('No Tenants').click();
         cy.get('div.loader').should('be.visible');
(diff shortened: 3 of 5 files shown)


NorbertAmartus:
@qooban what do you think, should I provide an additional test scenario?
Or it would be optional, as currently there was no tests related to generating `resourceLocation`?

widgets/blueprints/src/TerraformModal.tsx
```diff
@@ -289,7 +289,7 @@ export default function TerraformModal({
             ) {
                 return resourceLocation;
             }
-            return resourceLocation.replace(/^[^/]*/, '.');
+            return resourceLocation.replace(/^[^/]*\//, '');
```
qooban, line 292: So basically right now all the starting characters except `/` + `/` will be removed from the `resourceLocation`, right?

I'm wondering what was the reason when the original solution was introduced.
We can check with @kubama I think.

qooban (approved):
> @qooban what do you think, should I provide an additional test scenario?
> Or it would be optional, as currently there were no tests related to generating resourceLocation?

Seems like `getResourceLocation` function would be good candidate for unit testing, but as it is not exported function, used only in one place I think we can treat adding test for this case as optional.

I'm good to go with merging this PR.

widgets/blueprints/src/TerraformModal.tsx
NorbertAmartus: Yup, something like that   
I'd give a thumbs up for checking that, so that we'd be sure that an additionall `/` is not breaking any of the requirements, which we may not be aware of  😁

NorbertAmartus:
Some tests scenarios are failing, I'll investigate upon that matter

widgets/blueprints/src/TerraformModal.tsx
kubama: I don't think this is the right change. The `.` should stay for a scenario when there is no following subdirectory. I'm guessing it may be the reason behind system tests failure as there is a dedicated test for this case.

kubama (changes requested):
Please see https://github.com/cloudify-cosmo/cloudify-stage/pull/1767/files#r790598831

NorbertAmartus:
I've created another branch, on which I've made some tests regarding fixing the discovered issues
I've good some promising results which resulted in the [completly green system tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1885/) 🎉 

Because of that I'll migrate those changes to this branch.
Changes consists of:
- increasing the number of retries inside the `waitForExecutionToEnd` function
- modifying regex so that the added `/` is being optional (in cases that we have trouble with -> in downloaded `zip` file there is only one terraform module, placed inside it's root directory), so that the resource location would equal to an empty string (which works the same as `./`, looking at my observation and Trammell suggestion), not the name of that single directory e.g. `single` (currently regexp was demending `/` at the end of the string, to be removed), which doesn't work 😁

test/cypress/integration/widgets/blueprints_spec.ts
```diff
@@ -660,7 +660,7 @@ describe('Blueprints widget', () => {
                 });
                 cy.clickButton('Execute');
 
-                cy.waitForExecutionToEnd(deploymentId, 'install');
+                cy.waitForExecutionToEnd(deploymentId, 'install', { numberOfRetriesLeft: 120 });
```
NorbertAmartus, line 663: The default number of retries is `60`, I've doubled it just to be sure that the tests are having enought time to be executed successfully (even on the local machines, not only on jenkins), which helped with one of the failing test scenarios 🎉

qooban (dismissed):
I'm OK with changes, but let's wait for final confirmation from Kuba M.

kubama (commented):
Thanks for spending your time on in-depth investigation and testing.

I've got one more suggestion though. I would advice to add explicit test cases for `getResourceLocation` function that would document its behaviour (I think it could be refactored as a top-level function for easy independent testing). Without it I think it will be easily forgotten what's the purpose of the regexp.
I admit I should have added such test myself when this function was initially created 🙂.

NorbertAmartus:
> Thanks for spending your time on in-depth investigation and testing.
> 
> I've got one more suggestion though. I would advice to add explicit test cases for `getResourceLocation` function that would document its behaviour (I think it could be refactored as a top-level function for easy independent testing). Without it I think it will be easily forgotten what's the purpose of the regexp. I admit I should have added such test myself when this function was initially created 🙂.

Sure, it seems like a great idea!
To be honest, now I'm a little busy with updating the documentation, so I'd merge this PR (as there are no comments on the functionality), to meet the sprint requirements 😁    
In the near future, I'll create a PR which would contain an additional test scenario 😉 

@qooban, that would be okay?
(I've never tried that sort of actions before, so just want to make sure, that it's "legal" 😁 )

widgets/blueprints/src/TerraformModal.tsx
NorbertAmartus: As the mentioned functionality has been modified, I'm resolving this conversation

kubama (approved):

kubama:
> Sure, it seems like a great idea! To be honest, now I'm a little busy with updating the documentation, so I'd merge this PR (as there are no comments on the functionality), to meet the sprint requirements 😁 In the near future, I'll create a PR which would contain an additional test scenario 😉

Cool, approved then 👍.

qooban:
> In the near future, I'll create a PR which would contain an additional test scenario wink
> @qooban, that would be okay?

Sure, if you plan to do that this week, then perfectly fine 👍🏻  In other case let's have a dedicated ticket for that.

NorbertAmartus:
Okay, I'm planning to start adding that functionality, before starting working on the tasks from the next sprint, so that we'd have completed this functionality (and that we'd be able to sleep peacefully 😁 )

NorbertAmartus:
Resolved conflicts with the `master` branch, if that wouldn't be a problem, would you approve once more @qooban, just so the PR could be able to be merged again? 😁

kubama (approved):

qooban (approved):

PR #1739, opened 2022-01-12
RD-3624 (Part 3) Move execution parameters to deploy modal

leon-good-life:
## Description
This PR moves execution parameters from execution modal to deploy modal.

## Screenshots / Videos
Screen Shot 2022-01-12 at 12 53 20

## Checklist

- [x] My code follows the UI Code Style.
- [ ] I verified that all tests and checks have passed.
- [x] I verified if that change requires a documentation update.
- [x] I added proper labels to this PR.

## Tests
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1855/pipeline

deployments_view_spec after fix: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1861/pipeline

 

## Documentation
https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2076

PR #1727, opened 2021-12-29
RD-3624 (part 1) refactor before implementation

leon-good-life:
## Description
The GenericDeployModal is over 700 lines of code. And was not migrated to TypeScript yet.
The purpose of this pull-request is to extract some of the code in GenericDeployModal to separate component files.

## Screenshots / Videos
N/A

## Checklist

- [x] My code follows the UI Code Style.
- [x] I verified that all tests and checks have passed.
- [x] I verified if that change requires a documentation update.
- [x] I added proper labels to this PR.

## Documentation
N/A

PR #1770, opened 2022-01-24
RD-3624 (part 2)Extract shared code

leon-good-life:
## Description

New main PR of the task - part 3: https://github.com/cloudify-cosmo/cloudify-stage/pull/1812
This PR is created to reduce the amount of the code in the old main PR of the task (https://github.com/cloudify-cosmo/cloudify-stage/pull/1739). (Old Part 3)

It is the second refactoring PR. The first one was merged: https://github.com/cloudify-cosmo/cloudify-stage/pull/1727 (Part 1)

## Screenshots / Videos
N/A

## Checklist

## Tests
I fixed `spire_manager_spec` and started system tests again (26 JAN 2022): https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1911/pipeline

System tests after addressing review comments (25 JAN 2022):
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1899/pipeline

System tests after addressing review comments (24 JAN 2022):
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1880/pipeline

Initially I run:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1879/pipeline

 

## Documentation
Will be made in the main PR.


widgets/common/src/ExecuteWorkflow.tsx
```diff
@@ -0,0 +1,262 @@
...
+    handleYamlFileChange: (file: File) => void;
+    fileLoading: boolean;
+    handleExecuteInputChange: (event: Event, field: Field) => void;
+    showInstallOptions: boolean;
+    force: boolean;
+    dryRun: boolean;
+    queue: boolean;
+    schedule: boolean;
+    scheduledTime: string;
+    createChangeEvent: (fieldName: string) => (event: Event, field: Field) => void;
```
kubama, line 197: I would follow the common convention and start callbacks naming with 'on'

widgets/common/src/ExecuteDeploymentModal.tsx
```diff
@@ -236,6 +153,35 @@ export default function ExecuteDeploymentModal({
...
+                    setSchedule(field.checked);
+                    break;
+                case 'scheduledTime':
+                    setScheduledTime(field.value);
+                    break;
+                default:
+                    break;
+            }
+        };
+    }
```
kubama, line 183: Looking at the function name I would expect it to return an event object, but it actually returns another function.

widgets/common/src/ExecuteWorkflow.tsx
```diff
@@ -0,0 +1,262 @@
...
+}
+
+const ExecuteWorkflow: FunctionComponent<ExecuteWorkflowProps> = ({
```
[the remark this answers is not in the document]
leon-good-life: I renamed it: 53078da

widgets/common/src/ExecuteWorkflow.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+export function isWorkflowName(workflow: Workflow) {
+    return typeof workflow === 'string';
+}
+
+export function getWorkflowName(workflow: Workflow) {
+    return typeof workflow === 'string' ? workflow : workflow.name;
+}
+
+export const createExecuteWorkflowFunction = ({
```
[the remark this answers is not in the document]
leon-good-life: Because I'm planning to use it in GenericDeployModal in another PR in the following way:
Screen Shot 2022-01-24 at 15 29 19

widgets/common/src/ExecuteWorkflow.tsx
leon-good-life: I added `on` prefix: 05dc17a

widgets/common/src/ExecuteDeploymentModal.tsx
leon-good-life: 👍🏽 I changed it to `createOnChangeEvent` (05dc17a), I wonder if is it still not clear that it returns a function?

widgets/common/src/ExecuteWorkflow.tsx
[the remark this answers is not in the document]
leon-good-life: actually I can change that. I will do it.

widgets/common/src/ExecuteDeploymentModal.tsx
leon-good-life: I'm changing it to `onChange`

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,262 @@
...
+    fileLoading,
+    handleExecuteInputChange,
+    errors,
+    showInstallOptions,
+    force,
+    dryRun,
+    queue,
+    schedule,
+    scheduledTime,
+    createOnChangeEvent
```
kubama, line 212: I think this prop should be actually split into 5 separate props for each input: `onForceChange`, `onDryRynChange`, ...

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,262 @@
...
+interface ExecuteWorkflowInputsProps {
+    baseWorkflowParams: WorkflowParams;
+    userWorkflowParams: WorkflowParams;
+    errors: {
+        errors?: string;
+        scheduledTime?: string;
+        yamlFile?: string;
+        [inputName: string]: undefined | string;
+    };
+    handleYamlFileChange: (file: File) => void;
```
kubama, line 187:
```suggestion
    onYamlFileChange: (file: File) => void;
```

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,262 @@
...
+    fileLoading: boolean;
+    handleExecuteInputChange: (event: Event, field: Field) => void;
```
kubama, line 189: This callback is related to inputs provided in `baseWorkflowParams` prop so I think it's name should somehow indicate that. It should also start with 'on'.

widgets/common/src/ExecuteDeploymentModal.tsx
kubama: Please see https://github.com/cloudify-cosmo/cloudify-stage/pull/1770#discussion_r790776463, after digging deeper I actually think it should be refactored.

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I changed that: b15b90d

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I think that the disadvantage of splitting into multiple props is that it will increase the number of props, right now there are 13 props already.

Screen Shot 2022-01-24 at 16 36 01

Screen Shot 2022-01-24 at 16 40 48
Screen Shot 2022-01-24 at 16 41 36
Screen Shot 2022-01-24 at 16 42 08

I can push those changes.

kubama: Seeing how function passed as `createOnChangeEvent` prop is implemented (the `switch`) I think it's still worth it as it will simplify the usage of the component.

leon-good-life:
I started system tests:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1880/pipeline

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,257 @@
...
+    const { YamlFileButton, InputsHeader, InputsUtils } = Stage.Common;
+    return (
+        <>
+            {!_.isEmpty(baseWorkflowParams) && (
+                <YamlFileButton onChange={onYamlFileChange} dataType="execution parameters" fileLoading={fileLoading} />
+            )}
+            <InputsHeader header={t('paramsHeader')} compact />
+            {_.isEmpty(baseWorkflowParams) && <Message content={t('noParams')} />}
+
+            {InputsUtils.getInputFields(baseWorkflowParams, onWorkflowInputChange, userWorkflowParams, errors)}
```
kubama, line 224: Something confusing is happening here. `baseWorkflowParams` are passed as `inputs` argument and `userWorkflowParams` as `inputsState` argument. However, both props have same type and similar naming further suggesting that both are collections of same types of entities.

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+interface ExecuteWorkflowInputsProps {
+    baseWorkflowParams: WorkflowParams;
+    userWorkflowParams: WorkflowParams;
+    errors: {
+        errors?: string;
+        scheduledTime?: string;
+        yamlFile?: string;
+        [inputName: string]: undefined | string;
+    };
```
kubama, line 186: I don't see any usage of `errors.errors` or `errors.yamlFile` properties. Also, `inputName` should be restricted to what's passed in `baseWorkflowParams` (although I'm not sure if/how it could be achieved).

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I changed that: 36fe0e1

widgets/common/src/ExecuteDeploymentModal.tsx
leon-good-life: I addressed it: 36fe0e1

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: In another PR I will call `<ExecuteWorkflowInputs ... errors={this.state.errors} />` in `GenericDeployModal`. In `GenericDeployModal`, `yamlFile` is added to the errors state: 
Screen Shot 2022-01-24 at 17 49 32

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I moved this line from `ExecuteDeploymentModal.tsx`:
Screen Shot 2022-01-24 at 17 58 29

I agree that it is confusing that `baseWorkflowParams` are passed as `inputs`:
Screen Shot 2022-01-24 at 18 00 38

This function is also used here:
Screen Shot 2022-01-24 at 18 03 07

I'm not sure 🤔 how to rename the arguments.

leon-good-life: I think maybe `getInputFields` is a generic function, so the names of arguments don't match.

widgets/common/src/ExecuteDeploymentModal.tsx
```diff
@@ -236,6 +155,35 @@ export default function ExecuteDeploymentModal({
...
+        setQueue(field.checked);
+    };
+    const onScheduleChange = (event: Event, field: Field) => {
+        clearErrors();
+        clearQueue();
+        setSchedule(field.checked);
+    };
+    const onScheduledTimeChange = (event: Event, field: Field) => {
+        clearErrors();
+        clearQueue();
```
NorbertAmartus, line 183: In 5 functions we are using those 2 lines of code   
It would be beneficial if we could extract it as a separate function, which would be reused across those functions

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,261 @@
...
+        onHide();
+        return true;
+    }
+
+    if (_.isEmpty(deploymentsList)) {
+        setErrors(t('errors.missingDeployment'));
+        return false;
+    }
+
+    setLoading();
```
NorbertAmartus, line 111: In the `line 67` the same function is being executed, but between those lines I can't see a case which the loading would need a "re-setting" 😁    
Am I missing something (maybe it's necessary to have 2 times declared `setLoading()`) or one of them should be omitted?

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,261 @@
...
+    clearErrors: () => void;
+    onExecute: () => void;
+    onHide: () => void;
+}) => {
+    setLoading();
+
+    const { InputsUtils, DeploymentActions } = Stage.Common;
+    const validationErrors: Record<string, string> = {};
+    const deployments: any[] = [];
+    const deploymentsList: string[] = _.isEmpty(deployments) ? _.compact([deploymentId]) : deployments;
```
NorbertAmartus, line 72: deployments is being initialized as the empty array and is never being modified
Because of that, the condition inside the deploymentsList will always return _.compact([deploymentId])

I don't know if the deployments array was meant to be updated, but if it was (by eg. line 131, which contains the following code: toolbox.getEventBus().trigger('deployments:refresh', id);) then we should adjust that variable, to not be just a constantly empty array 😁

In other case, we could get rid of it, as it has no purpose 😉

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,261 @@
...
+    force: boolean;
+    dryRun: boolean;
+    queue: boolean;
+    schedule: boolean;
+    scheduledTime: string;
+    onForceChange: (event: Event, field: Field) => void;
+    onDryRynChange: (event: Event, field: Field) => void;
+    onQueueChange: (event: Event, field: Field) => void;
+    onScheduleChange: (event: Event, field: Field) => void;
+    onScheduledTimeChange: (event: Event, field: Field) => void;
```
NorbertAmartus, line 196: We could reuse typing across 6 functions here `(event: Event, field: Field)`

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,261 @@
...
+                            </>
+                        )}
+                    </UnsafelyTypedFormField>
+                </>
+            )}
+        </>
+    );
+};
+
+export default React.memo(ExecuteWorkflowInputs);
```
NorbertAmartus, line 261: This file is huuuuuuuuuuge, it has **261 lines** of code   
At the one of the previous [PRs](https://github.com/cloudify-cosmo/cloudify-stage/pull/1739) files were much smaller, why we've marged them into one? 😁

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,261 @@
...
+                queue,
+                scheduledTime: schedule ? moment(scheduledTime).format('YYYYMMDDHHmmZ') : undefined
+            })
+            .then(() => {
+                unsetLoading();
+                clearErrors();
+                onHide();
+                toolbox.getEventBus().trigger('executions:refresh');
+                // NOTE: pass id to keep the current deployment selected
+                toolbox.getEventBus().trigger('deployments:refresh', id);
```
NorbertAmartus, line 128: Is this TODO not completed in the line below it? 😁

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,261 @@
...
+    setLoading();
+    const actions = new DeploymentActions(toolbox);
+
+    const executePromises = _.map(deploymentsList, id => {
+        return actions
+            .doExecute(id, name, workflowParameters, {
+                force,
+                dryRun,
+                queue,
+                scheduledTime: schedule ? moment(scheduledTime).format('YYYYMMDDHHmmZ') : undefined
```
NorbertAmartus, line 120: We are duplicating here logic from the `line 100`, we should be able to extract this logic as a variable

qooban (commented):
Added some comments, but seeing that @NorbertAmartus and @kubama did detailed reviews, I did just quick scan.

widgets/common/src/ExecuteWorkflowInputs.tsx
NorbertAmartus: Oh yeah 😁 
Sorry, my bad! 

(I've have tricked myself due to the previous PR, where TODO was added in that place 😁 )

app/translations/en.json
```diff
@@ -490,7 +490,7 @@
                 "workflowsMenu": {
                     "tooltip": "Execute workflow"
                 },
-                "executeModal": {
+                "execute": {
```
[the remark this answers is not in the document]
leon-good-life: The `"execute"` contains both the "install" section in GenericDeployModal and the executeModal. Probably it would be better to change in the other PR. But I think I can leave it as is because I'm planning to change it to "execute" in the other PR.

widgets/common/src/ExecuteDeploymentModal.tsx
leon-good-life: I extracted and reused: ccffc69

widgets/common/src/ExecuteWorkflowInputs.tsx
kubama: I looked into `getInputFields` implementation and here are my findings:
* `inputsState` (passed in as `userInputParams`) is a map holding input values, so `userInputParams` should be named accordingly and also its type should be aligned
* `inputs` (passed in as `baseWorkflowParams`) is a map of input descriptors, I think `baseWorkflowParams` should be renamed to better reflect what it is, its currently defined type seems to be fine

widgets/common/src/ExecuteWorkflowInputs.tsx
kubama: I think this type could be reused from SUI, as this is where it's defined for all 'onChange' handlers

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: It is unnecessary, I removed it: ffd5fa9

widgets/common/src/ExecuteWorkflowInputs.tsx
kubama: But in this component `yamlFile` is not used, so why should we require it to be passed in?

I also have a general comment here - if this relatively small PR extracts a common component then I think it should include all its intended usages. Without it it's hard to say if the scope if the extraction is correct, as we only see one place where it is being used.

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I changed it: 3f10e81, 5d407e5

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+type WorkflowParams = {
+    [key: string]: {
+        type?: string;
+        default?: string;
+        constraints?: {
+            pattern: string;
+        }[];
+    };
+};
```
kubama, line 152: Wouldn't it be simpler to use `Record` here?

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I addressed that, I use now real `deployments`: 5262add

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: 👍🏽 I will change to:
```
type WorkflowParams = Record<string, {
        type?: string;
        default?: string;
        constraints?: {
            pattern: string;
        }[];
    }>;
```

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: Regarding your comment, I agree, I think I need to provide more details in the description of the PR about how it will be used in other places. I created this PR because I got a comment that my previous PR is too large, and I wanted to reduce the amount of changes in a PR (split implementation into multiple PRs) otherwise there would be too much changes in a single PR.

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I addressed it: e6f528f, 7711fba

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I addressed it: 1524332

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I split this file into two files: b725e2b

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I renamed `baseWorkflowParams` to `baseWorkflowInputs`, and I removed the wrong type from `userInputParams`: 1c611b5

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I addressed `errors` type: 23949d2

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I also renamed `userWorkflowParams` to `userWorkflowInputsState`: 8b11405

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I named the function better: 37803d9

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: Regarding the type of `userWorkflowInputsState`, I wonder if `Record<string, string | number | boolean | null | undefined>` is the right type?

leon-good-life: I addressed this review

widgets/common/src/ExecuteDeploymentModal.tsx
```diff
@@ -1,32 +1,11 @@
 // @ts-nocheck File not migrated fully to TS
```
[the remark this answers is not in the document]
leon-good-life: I agree, good idea

leon-good-life:
I started system tests: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1899/pipeline

leon-good-life:
I'm checking sprite_manager_spec (the only system test that didn't pass):
Spire Manager widget -- allows to do bulk workflow execution on spire deployments (failed)

widgets/common/src/ExecuteWorkflowInputs.tsx
NorbertAmartus: @kubama  I think that it's not as easy
I've looked upon the e.g. `Dropdown.d.ts` typpe file, that comes from the SUI, and here is the definition of the `onChange` method:
```

  /**
   * Called when the user attempts to change the value.
   *
   * @param {SyntheticEvent} event - React's original SyntheticEvent.
   * @param {object} data - All props and proposed value.
   */
  onChange?: (event: React.SyntheticEvent<HTMLElement>, data: DropdownProps) => void
  ```
  
  So I think that we don't have a "ready for use" typing for that :/

Comment (qooban) [reply widgets/common/src/ExecuteWorkflowInputs.tsx]: If you use `ComponentProps` you can get type of `onChange` prop. 
Eg.: `ComponentProps<typeof Dropdown>['onChange']`

File: widgets/common/src/ExecuteDeploymentModal.tsx [hunk 20]
```diff
@@ -236,6 +152,36 @@ export default function ExecuteDeploymentModal({
...
+        clearSchedule();
+        clearScheduleTime();
+        clearErrors();
+        setQueue(field.checked);
+    };
+    const onScheduleChange = (event: Event, field: Field) => {
+        clearErrorsAndQueue();
+        setSchedule(field.checked);
+    };
+    const onScheduledTimeChange = (event: Event, field: Field) => {
```
NorbertAmartus, line 180: Almost all of those methods are using the same typing (`(event: Event, field: Field)`) as the one that you've created inside the `ExecuteWorkflowInput` file. This typing is currently being stored as:
```
type OnChange = (event: Event, field: Field) => void;
```

We could reuse it here 😁

leon-good-life: done: 60296cd

leon-good-life:
I fixed `spire_manager_spec` and started system tests again: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1911/pipeline

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I changed the type of `OnChange` to `ComponentProps<typeof Dropdown>['onChange']`: d5439f8

NorbertAmartus (dismissed):
I can't find anything else, which could be bringed into the discussion

Nicely done! 🎉

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,164 @@
...
+        default?: string;
+        constraints?: {
+            pattern: string;
+        }[];
+    }
+>;
+
+export type UserWorkflowInputsState = Record<string, string | number | boolean | null | undefined>;
+
+export type Workflow =
```
qooban, line 27: That name is misleading. It seems to me that you are providing a type for a Blueprint resource, not a Workflow. `plan` is only part of a Blueprint resource AFAIR.

Take a look at:
1. https://docs.cloudify.co/api/staging/dev/#the-workflow-resource
2. `Stage.PropTypes.Workflows`

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,107 @@
...
+    showInstallOptions: boolean;
+    force: boolean;
+    dryRun: boolean;
+    queue: boolean;
+    schedule: boolean;
+    scheduledTime: string;
+    onForceChange: OnChange;
+    onDryRunChange: OnChange;
+    onQueueChange: OnChange;
+    onScheduleChange: OnChange;
```
qooban, line 41: The type here is invalid. `OnChange` right now is defined for a dropdown, and AFAIR `force`, `dryRun`, `queue` and `schedule` are checkboxes.

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,107 @@
...
+    onScheduledTimeChange: OnChange;
```
qooban, line 42: The type here is invalid. `OnChange` right now is defined for a dropdown, and `scheduleTime` is associated with `DateTime` component.

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+const ExecuteWorkflowInputs: FunctionComponent<ExecuteWorkflowInputsProps> = ({
```
qooban, line 45: I like the name of the new extracted component.

And I think that we could also change `ExecuteDeploymentModal` to be more aligned with Cloudify resources naming. We actually are not executing deployment, but executing workflow on a deployment, so what do you think about changing `ExecuteDeploymentModal` into `ExecuteWorkflowModal`?

Ref.: https://docs.cloudify.co/latest/working_with/manager/execute-workflow/

I found that the current name is used 17 times:
image
so shouldn't be hard to change it I believe.

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I addressed it: 86ad18d

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I addressed it: 86ad18d

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,164 @@
...
+const t = Stage.Utils.getT('widgets.common.deployments.execute');
+
+export type Field = {
+    name: string;
+    value: unknown;
+    type: string;
+    checked?: string;
+};
+
+export type OnChange = ComponentProps<typeof Stage.Basic.Dropdown>['onChange'];
```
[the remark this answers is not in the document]
leon-good-life: I addressed it: 86ad18d

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I addressed it: a1be50f

widgets/common/src/ExecuteDeploymentModal.tsx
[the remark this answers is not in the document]
leon-good-life: I partially migrated it: a1be50f

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,164 @@
+import type { ComponentProps } from 'react';
+
```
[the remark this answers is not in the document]
leon-good-life: I think I can move the `OnCheckboxChange` type to another place. Other types are related to `workflow`

widgets/common/src/executeWorkflow.ts
leon-good-life: I fixed the `Workflow` type: e10540c

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,164 @@
...
+    {
+        type?: string;
+        default?: string;
+        constraints?: {
+            pattern: string;
+        }[];
+    }
+>;
+
+export type UserWorkflowInputsState = Record<string, string | number | boolean | null | undefined>;
```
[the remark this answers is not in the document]
leon-good-life: I don't remember, I think I made `console.log` and looked what's inside or just by looking in the code. I now realize that it is probably not a good way to check the type because there may be other types.

I read the documentation and added ` | string[] | Record<string, string>`: 47f1f35
The documentation doesn't specify if the array (list) is of strings or other types. And if the dict has only string values or other types.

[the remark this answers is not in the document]
leon-good-life: I changed to:
```

export type UserWorkflowInputsState = Record<
    string,
    | string
    | number
    | boolean
    | string[]
    | number[]
    | boolean[]
    | Record<string, string | number | boolean>
    | null
    | undefined
>;
```

cda860a

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+function renderCheckboxField(name: string, checked: boolean, onChange: OnCheckboxChange) {
+    const { Field } = Stage.Basic.Form;
+    return <Field>{renderActionCheckbox(name, checked, onChange)}</Field>;
+}
+
+interface ExecuteWorkflowInputsProps {
+    baseWorkflowInputs: BaseWorkflowInputs;
+    userWorkflowInputsState: UserWorkflowInputsState;
+    errors: Record<string, string>;
```
qooban, line 33: Isn't type for `errors` already defined in `executeWorkflow.ts` file? (`Errors` is `string | Record<string, string>`, but if we are sure in this case we'll have only `Record<string, string>` maybe worth to extract sub-type for this union?

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,112 @@
+import type { FunctionComponent } from 'react';
+import type {
+    BaseWorkflowInputs,
+    UserWorkflowInputsState,
+    OnCheckboxChange,
+    OnDropDownChange
+} from './executeWorkflow';
```
qooban, line 7: Is there any case where all those 4 types are used without `ExecuteWorkflowInputs`? If not, maybe that's a good argument to put all 4 types either directly to `ExecuteWorfklowInputs` or to something like `ExecuteWorkflow.types.ts`?

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,176 @@
...
+            force: boolean;
+            dryRun: boolean;
+            queue: boolean;
+            scheduledTime: string;
+        }
+    ) => void;
+    onHide: () => void;
+}) => {
+    const { InputsUtils, DeploymentActions } = Stage.Common;
+    const validationErrors: Record<string, string> = {};
```
qooban, line 113: Isn't type for errors already defined? `Errors` is `string | Record<string, string>`, but if we are sure in this case we'll have only `Record<string, string>` maybe worth to extract sub-type for this union? (I've added similar comment in other place as well)

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,176 @@
...
+export type Workflow =
+    | string
+    | {
+          name: string;
+          plugin: string;
+          operation?: string;
+          // eslint-disable-next-line camelcase
+          is_cascading?: boolean;
+          parameters?: Record<string, string>;
+      };
```
qooban, line 48: 1. Are we using `operation` and `is_cascading`?

2. I found that there is:
```
type Workflow = {
    name: string;
    parameters: Record<string, { description?: string; default?: any; type?: any }>;
    plugin: string;
};
```
in `widgets/common/src/deploymentsView/header/RunWorkflowModal.tsx`.
Probably worth to extract it and re-use it.

widgets/common/src/ExecuteWorkflowModal.tsx
```diff
@@ -0,0 +1,261 @@
+// @ts-nocheck File not migrated fully to TS
+import type { FunctionComponent } from 'react';
+import ExecuteWorkflowInputs from './ExecuteWorkflowInputs';
+import { executeWorkflow, isWorkflowName, getWorkflowName } from './executeWorkflow';
+import type { OnCheckboxChange, Workflow } from './executeWorkflow';
+
+const t = Stage.Utils.getT('widgets.common.deployments.execute');
+
+interface ExecuteWorkflowModalProps {
+    deploymentId?: string;
```
qooban, line 10: In the previous implementation, we had the following validation:
```
    deploymentId: (props, propName, componentName) => {
        const { [propName]: deploymentId, workflow } = props;

        if (_.isString(workflow) && !(_.isString(deploymentId) && deploymentId)) {
            return new Error(
                `Invalid prop \`deploymentId\` supplied to \`${componentName}\`. ` +
                    `When \`workflow\` prop is specified as a string, \`deploymentId\` must be provided. ` +
                    `Validation failed.`
            );
        }
        return null;
    },
```
Is it covered in the new implementation?

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,176 @@
...
+    scheduledTime,
+    force,
+    dryRun,
+    queue,
+    setErrors,
+    unsetLoading,
+    clearErrors,
+    onExecute,
+    onHide = () => {}
+}: {
```
qooban, line 86: Optional:

`executeWorkflow` and `ExecuteWorkflowInputs` props have common part.:
```
    baseWorkflowInputs: BaseWorkflowInputs;
    userWorkflowInputsState: UserWorkflowInputsState;
    force: boolean;
    dryRun: boolean;
    queue: boolean;
    schedule: boolean;
    scheduledTime: string;
```

Maybe worth to think about having it defined in one place and relate to it in the other place?

widgets/common/src/executeWorkflow.ts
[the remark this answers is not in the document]
qooban: I see you resolved the discussion here, but for me it is unresolved. There are two topics I brought up:

1. Isn't the file name misleading?
2. Shouldn't types be in different place? (either new types files or in `ExecuteWorkflowInputs` component)

qooban (commented):
I added more comments. I think you should schedule a meeting and we should discuss it before making any additional changes, there's a lot of open issues.

widgets/common/src/executeWorkflow.ts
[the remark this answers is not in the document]
leon-good-life: > I'm wondering if the name of the new file isn't misleading. `executeWorkflow` would make sense if `executeWorkflow` function was the only exported function from this file, but I see there are other exported functions here as well.

Other than `executeWorkflow`, there are 2 other functions `isWorkflowName` and `getWorkflowName`. All those functions are used in more than one place. So they need to be placed in some file, I placed them here. I'm not sure how to rename this file. Do you have a suggestion? I thought about `executeWorkflowUtils` maybe. What do you think?
> 
> There are also multiple types exported from this file. If the intention is to reuse them in many files maybe worth to have separate file for those types as well?

I can move the types to be under `Stage.Types` namespace.  Is it the right place?

> 
> I'm asking questions, I'm not going to provide final solution.

Just advices and some hints would help a lot 🙂

widgets/common/src/ExecuteWorkflowModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+        clearSchedule();
+        clearScheduleTime();
+        clearErrors();
+        setQueue(field.checked);
+    };
+    const onScheduleChange: OnCheckboxChange = (event, field) => {
+        clearErrorsAndQueue();
+        setSchedule(field.checked);
+    };
+    const onScheduledTimeChange: OnCheckboxChange = (event, field) => {
```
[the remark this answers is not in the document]
leon-good-life: I changed to `OnChange` type because `ComponentProps<typeof Stage.Basic.DateInput>['onChange'];` is causing an error `Type 'typeof DateInput' does not satisfy the constraint 'JSXElementConstructor<any> | keyof IntrinsicElements'.`.

48d627c

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,176 @@
+import type { ComponentProps } from 'react';
+
+const t = Stage.Utils.getT('widgets.common.deployments.execute');
+
+export type Field = {
+    name: string;
+    value: unknown;
+    type: string;
+    checked?: string;
+};
```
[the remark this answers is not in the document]
leon-good-life: I removed it. faae747

widgets/common/src/executeWorkflow.ts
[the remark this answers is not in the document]
leon-good-life: I'm moving the types to `ExecuteWorkflowInputs ` as you suggested in another comment.

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,176 @@
...
+    | string
+    | number
+    | boolean
+    | string[]
+    | number[]
+    | boolean[]
+    | Record<string, string | number | boolean>
+    | null
+    | undefined
+>;
```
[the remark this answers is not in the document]
leon-good-life: I moved them: e6e8532

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,176 @@
...
+const t = Stage.Utils.getT('widgets.common.deployments.execute');
+
+export type Field = {
+    name: string;
+    value: unknown;
+    type: string;
+    checked?: string;
+};
+
+export type OnDropDownChange = ComponentProps<typeof Stage.Basic.Dropdown>['onChange'];
```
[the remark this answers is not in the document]
leon-good-life: I renamed it: d8c7f4e

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: I fixed the type of onScheduledTimeChange: 48d627c

widgets/common/src/executeWorkflow.ts
leon-good-life: I extracted those props to a common interface and then extended the other interfaces: 807e32f

widgets/common/src/executeWorkflow.ts
leon-good-life: I changed to type `Errors`: 0b0d1de

widgets/common/src/executeWorkflow.ts
leon-good-life: 1. There is no usage for `operation` and `is_cascading `. I removed them.
2. I extracted and reused it.

git commit: 5ef832e

widgets/common/src/ExecuteWorkflowModal.tsx
leon-good-life: I added validation for `deploymentId` prop in `ExecuteWorkflowModal`:  df6a4d1

widgets/common/src/executeWorkflow.ts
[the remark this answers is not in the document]
leon-good-life: I moved types to a new `types.ts` file: 366f769

Can I resolve this comment?

widgets/common/src/ExecuteWorkflowModal.tsx
```diff
@@ -0,0 +1,273 @@
...
+const t = Stage.Utils.getT('widgets.common.deployments.execute');
+
+interface ExecuteWorkflowModalProps {
+    deploymentId?: string;
+    deploymentName?: string;
+    deployments?: string[];
+    hideOptions?: boolean;
+    onExecute?: (
+        installWorkflowParameters: InstallWorkflowParameters,
+        installWorkflowOptions: InstallWorkflowOptions
```
qooban, line 17: Both - parameter name and parameter type should not refer to `install` workflow as this component is used for any workflow.

widgets/common/src/ExecuteWorkflowInputs.tsx
```diff
@@ -0,0 +1,116 @@
...
+    userWorkflowInputsState: UserWorkflowInputsState;
+    force: boolean;
+    dryRun: boolean;
+    queue: boolean;
+    schedule: boolean;
+    scheduledTime: string;
+}
+
+interface ExecuteWorkflowInputsProps extends CommonExecuteWorflowProps {
+    errors: Record<string, string>;
```
qooban, line 42: I'm not sure, but would it be possible to reuse already created `Errors` type here?

widgets/common/src/ExecuteWorkflowModal.tsx
```diff
@@ -0,0 +1,273 @@
+// @ts-nocheck File not migrated fully to TS
+import type { FunctionComponent } from 'react';
+import ExecuteWorkflowInputs from './ExecuteWorkflowInputs';
+import type { Workflow, InstallWorkflowOptions, InstallWorkflowParameters } from './types';
+import type { OnChange, OnCheckboxChange } from './ExecuteWorkflowInputs';
```
qooban, line 5: I don't see any exported types from `ExecuteWorkflowInputs`.

widgets/common/src/ExecuteWorkflowModal.tsx
```diff
@@ -0,0 +1,273 @@
+// @ts-nocheck File not migrated fully to TS
```
qooban, line 1: I'm wondering how hard would it be to remove that line? 🙂

widgets/common/src/executeWorkflow.ts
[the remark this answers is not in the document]
qooban: Let's resolve and continue discussion about filename for `executeWorkflow.ts` in another discussion.

widgets/common/src/types.ts
```diff
@@ -12,4 +12,30 @@ export type Workflow = {
...
+    | string
+    | number
+    | boolean
+    | string[]
+    | number[]
+    | boolean[]
+    | Record<string, string | number | boolean>
+    | null
+    | undefined
+>;
```
qooban, line 40: Have you checked that all types are actually supported by the UI?

widgets/common/src/types.ts
```diff
@@ -12,4 +12,30 @@ export type Workflow = {
     plugin: string;
 };
 
+export type InstallWorkflowParameters = Record<string, string>;
+export type InstallWorkflowOptions = { force: boolean; dryRun: boolean; queue: boolean; scheduledTime: string };
```
qooban, line 16: Seeing where it is used, I think we shouldn't  use names referring `install` workflow. I suggest to just remove `Install` prefix.

qooban (commented):
Added some comments for today's discussion.

Regarding directory structure and file names, I suggest to go with something like we already have for `labels` and `filters` (see those directories).

We could end up with something like:
```
/widgets/common/src
  /executeWorkflow
    ExecuteWorkflowInputs.tsx
    ExecuteWorkflowModal.tsx
    common.ts (old executeWorkflow.ts)
    types.ts (only types relevant for workflow execution)
    index.ts (grouping file with `Stage.defineCommon` and extending `Stage.Common` namespace)
```

widgets/common/src/executeWorkflow.ts
```diff
@@ -0,0 +1,176 @@
...
+
+export type UserWorkflowInputsState = Record<
+    string,
+    | string
+    | number
+    | boolean
+    | string[]
+    | number[]
+    | boolean[]
+    | Record<string, string | number | boolean>
```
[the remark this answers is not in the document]
qooban: I've added another comment about it, so I'm resolving this one.

widgets/common/src/ExecuteWorkflowInputs.tsx
qooban: Similar comment added in the latest version. Resolving.

widgets/common/src/ExecuteWorkflowInputs.tsx
qooban: Not relevant anymore

widgets/common/src/ExecuteWorkflowModal.tsx
leon-good-life: I remved `install` prefix: d5575e2

widgets/common/src/ExecuteWorkflowModal.tsx
leon-good-life: I fixed the import: d5575e2

widgets/common/src/types.ts
leon-good-life: I removed `install` prefix: d5575e2

widgets/common/src/types.ts
leon-good-life: Screen Shot 2022-02-10 at 11 26 40
Screen Shot 2022-02-10 at 11 29 24
type of `value` of `DropDown` in `semantic-ui-react`:
Screen Shot 2022-02-10 at 11 34 54

So I will change to:
```
export type UserWorkflowInputsState = Record<
    string,
    undefined
    | boolean
    | number
    | string
    | (boolean | number | string)[]
>;
```

leon-good-life: I changed the type: 9fada5e

leon-good-life:
I started system tests: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/2007/pipeline

widgets/common/src/ExecuteWorkflowInputs.tsx
leon-good-life: fully migrated in multiple commits

widgets/common/src/ExecuteWorkflowModal.tsx
leon-good-life: fully migrated in multiple commits

leon-good-life:
I'm working on fixing deployment_button_spec

leon-good-life:
I started system tests:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/2008/pipeline

qooban:
> Only `maintenance_mode_button_spec` failed. I'm working on fixing it.

Is it related to RD-3760? If yes, then you don't have to bother about it.

Thanks for applying changes, I'll review the PR tomorrow morning.

In the meantime, please take a look at linter errors.

leon-good-life:
> Is it related to [RD-3760](https://cloudifysource.atlassian.net/browse/RD-3760)? If yes, then you don't have to bother about it.

I verified now that `maintenance_mode_button_spec` is the same as [RD-3760](https://cloudifysource.atlassian.net/browse/RD-3760).

widgets/common/src/DeploymentActions.ts
```diff
@@ -1,3 +1,5 @@
+import type { Workflow } from './executeWorkflow';
```
qooban, line 1: It's invalid, there's no such file.

widgets/common/src/types.ts
```diff
@@ -1,15 +0,0 @@
-import { ComponentProps } from 'react';
```
qooban, line 1: Why you removed this file? I think that types should be placed a bit differently.

It seems like:
1. `Workflow` type should be moved to `executeWorkflow/types.ts` as it is specific to workflows execution, so that's fine.
2. `DropdownValue` and `Field` are used outside of `executeWorkflow` folder, so they should probably stay here.
3. Regarding `OnChange` types - they seems to be generic, but right now used only in `executeWorfklow` folder? If so, then let's keep it there.

widgets/common/src/executeWorkflow/common.ts
```diff
@@ -0,0 +1,123 @@
+import type { CommonExecuteWorflowProps } from './ExecuteWorkflowInputs';
+import type { Workflow } from './types';
+
+const t = Stage.Utils.getT('widgets.common.deployments.execute');
+
+export type Errors = string | Record<string, string>;
```
qooban, line 6: If it is exported, then I assume it is used outside of this file, right? If so, then maybe `types.ts` is the right place for it?
If it is not specific only to workflow execution, then probably `widgets/common/src/types.ts` and not `widgets/common/src/executeWorkflow/types.ts`.

widgets/common/src/deployModal/DeploymentInputsSection.tsx
```diff
@@ -1,6 +1,6 @@
 import type { FunctionComponent } from 'react';
 import type { FullBlueprintData } from '../BlueprintActions';
-import type { Field } from '../types';
+import type { Field } from '../executeWorkflow/types';
```
qooban, line 3: We shouldn't change that (`Field` is not a type specific only for workflow execution). See other comment about types scope.

widgets/common/src/executeWorkflow/ExecuteWorkflowModal.tsx
```diff
@@ -0,0 +1,275 @@
...
+import type {
+    Workflow,
+    WorkflowOptions,
+    WorkflowParameters,
+    OnDateInputChange,
+    OnCheckboxChange,
+    OnDropdownChange
+} from './types';
+import ExecuteWorkflowInputs from './ExecuteWorkflowInputs';
+import { Errors, executeWorkflow, getWorkflowName } from './common';
```
qooban, line 11: `Errors` should be imported with `import type`, but please verify if the place you have chosen is the right for `Errors` type (see other comment).

qooban (commented):
Good that we are closer to have clean solution. Still, I think few changes have to be done. See my comments.

widgets/common/src/deployModal/DeploymentInputsSection.tsx
leon-good-life: I moved it back to the `common/src/types.ts`: 5fd4d41

widgets/common/src/executeWorkflow/common.ts
leon-good-life: I think I import it only inside `widgets/common/src/executeWorkflow/` directory, so I moved it to `widgets/common/src/executeWorkflow/types.ts`: 5fd4d41

widgets/common/src/executeWorkflow/ExecuteWorkflowModal.tsx
leon-good-life: I moved to types file, and I'm importing using `import type`: 5fd4d41e4ffb93708ba20c72d450bb4c97fe27f6

widgets/common/src/types.ts
leon-good-life: I addressed all the points: 5fd4d41e4ffb93708ba20c72d450bb4c97fe27f6

qooban (approved):
Looks good 👍🏻 If system tests are passing, then I'm OK with merging it.

BTW If you are working today, then more important to unblock others is RD-2946. See discussion in #cfydevs-ui slack channel.

NorbertAmartus (approved):

PR #1773, opened 2022-01-26
RD-4086 Fix displaying Blueprints and Plugins logos

qooban:
## Description
As part of RD-4032 two routes:
* `GET /external/content`
* `GET /ba/:blueprint`

were changed to require authentication. The first one is among others used to display blueprint images in Blueprint Marketplace. The second one is used to display blueprint images in blueprints widget. 

Within this PR I'm changing `/ba` and `/external` routes to be unauthenticated by default in backend (again). Inside `/ba` router only `GET` endpoint was changed to be unauthenticated.

#### Blueprints marketplace
image

#### Blueprints widget
image

#### Blueprints marketplace
image

#### Blueprints widget
image

## Checklist

## Tests

[Stage-UI-System-Test #1916](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1916/tests) (26.01.2022 12:19 CET) - PASSED

## Documentation
N/A

### DIFF
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -102,13 +102,11 @@ app.use(
         indexFromEmptyFile: false
     })
 );
-// API Routes (with authentication)
+// API Routes with authentication
 const authenticatedApiRoutes: Record<string, Router> = {
     applications: Applications,
-    ba: BlueprintAdditions,
     bud: BlueprintUserData,
     clientConfig: ClientConfig,
-    external: External,
     file: File,
     filters: Filters,
     github: GitHub,
@@ -125,11 +123,15 @@ Object.entries(authenticatedApiRoutes).forEach(([routePath, router]) =>
     app.use(`${contextPath}/${routePath}`, authenticateWithToken, router)
 );
 
-// API Routes (without authentication)
+// API Routes with authentication only for some endpoints (see routers for details)
 app.use(`${contextPath}/auth`, Auth);
+app.use(`${contextPath}/ba`, BlueprintAdditions);
+
+// API Routes without authentication
 app.use(`${contextPath}/config`, (req, res) => {
     res.send(getClientConfig(getMode()));
 });
+app.use(`${contextPath}/external`, External);
 app.use(`${contextPath}/style`, Style);
 app.use(`${contextPath}/sp`, ServerProxy);
 
--- a/backend/app.ts
+++ b/backend/app.ts
@@ -123,17 +123,19 @@ Object.entries(authenticatedApiRoutes).forEach(([routePath, router]) =>
     app.use(`${contextPath}/${routePath}`, authenticateWithToken, router)
 );
 
+// TODO(RD-3827): All API routes should be authenticated
 // API Routes with authentication only for some endpoints (see routers for details)
 app.use(`${contextPath}/auth`, Auth);
 app.use(`${contextPath}/ba`, BlueprintAdditions);
 
 // API Routes without authentication
-app.use(`${contextPath}/config`, (req, res) => {
+const Config = (req, res) => {
     res.send(getClientConfig(getMode()));
-});
-app.use(`${contextPath}/external`, External);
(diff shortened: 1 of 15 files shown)


backend/app.ts
```diff
@@ -125,11 +123,15 @@ Object.entries(authenticatedApiRoutes).forEach(([routePath, router]) =>
     app.use(`${contextPath}/${routePath}`, authenticateWithToken, router)
 );
 
-// API Routes (without authentication)
+// API Routes with authentication only for some endpoints (see routers for details)
```
NorbertAmartus, line 126: ### Optional

I'd improve this comment, cause in the `BlueprintAdditions` Router we are able to see some details, stated as:
```typescript
// This path returns image resource so there is no point to secure that
// (if yes all credentials should be passed in the query string)
```

But when it comes to the `Auth`, then we have no description regarding this comment
As a developer I'd think that `see smth for details` would mean that somewhere there is an explaination why the certain decision is being made (after going through the `Auth` file I got that it was referring to the `/login` endpoint, but I think that scanning the full file would be done without seeing the `see routers for details` 😁 ), so I think that it's a little missleading

NorbertAmartus (approved):
Looks good 🎉 
Just added one optional comment, regarding one of the comments

kubama (commented):
Was this issue caused by the fact that the images are fetched directly by the browser and not through `Internal` class which sets the required header? If yes, then the question is why are we using a header that needs to be explicitly set and not a cookie which would be set automatically by the browser? This will have to be addressed anyway once `HttpOnly` flag is implemented as it will prevent reading the token cookie to set the header. Once this is addressed I think these endpoints could and should be authenticated, so maybe worth leaving TODO here?

qooban:
@kubama 

> Was this issue caused by the fact that the images are fetched directly by the browser and not through `Internal` class which sets the required header? 

Yes, they are directly fetched by the browser.

In Blueprints widget we use the following way to display images:
```
<img src="/console/ba/image/GCP-Basics-VM-Setup" width="30px" height="auto" class="ui inline image">
```

In Blueprints Marketplace we use the following way to display images:
```
<img src="/console/external/content?url=https%3A%2F%2Fraw.githubusercontent.com%2Fcloudify-community%2Fcloudify-catalog%2Fmaster%2Faws%2Fdynamodb%2Fdynamodb.png" width="30px" height="auto" class="ui inline image">
```

We could probably add cookie-based authentication for them, but I'm not sure it is the right time to do it. I think the whole stage backend authentication should be reviewed and clarified what is expected (especially in terms of security), but it should be planned for the period after the release.

> If yes, then the question is why are we using a header that needs to be explicitly set and not a cookie which would be set automatically by the browser? 

I may ask you the same question 🙂 Cookie and token authentications were introduced long time ago and I was not involved in designing that solution. 

> This will have to be addressed anyway once HttpOnly flag is implemented as it will prevent reading the token cookie to set the header. Once this is addressed I think these endpoints could and should be authenticated, so maybe worth leaving TODO here?

The goal of RD-3827 is to make all API routes to be authenticated. RD-4032 was meant to be only a preparation for the proper fix. See my comment there. If you have any findings, please add it there.

I can add "TODO(RD-3827)" to the code.

kubama:
> I can add "TODO([RD-3827](https://cloudifysource.atlassian.net/browse/RD-3827))" to the code.

To prevent future confusion I would also add a note explaining why these endpoints are left unauthenticated.

qooban (commented):
Added some clarifications in comments and TODOs.

kubama (approved):

qooban:
I have started stage system tests (see link in **Tests** section). Once everything passes, I'm going to merge this PR and create cherry-pick-PR for `6.3.0-build` branch.

NorbertAmartus (approved):
Looks good to me! 💪 
Thanks for addressing all of the comments 🎉

PR #1782, opened 2022-01-26
RD-2570: Make sure that videos are removed for passed tests

jawamartus:
## Description

The PR provide addon to cypress which removes artifacts from passed tests. It looks like working - we don't have artifacts for tests run on this branch: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1938/artifacts

## Checklist

## Tests
1. https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/1938/pipeline
2. https://jenkins.cloudify.co/job/Stage-UI-System-Test/1945/


test/cypress/plugins/index.ts
```diff
@@ -12,12 +12,26 @@
...
 
+    // Delete the recorded video for specs that had no retry attempts
+    // https://github.com/cypress-io/cypress/issues/16377
+    on('after:spec', async (_spec, results) => {
+        if (results && results.video) {
+            // Do we have failures for any retry attempts?
+            const failures = results.tests?.some(test => test.attempts.some(attempt => attempt?.state === 'failed'));
+            if (!failures) {
+                // delete the video if the spec passed and no tests retried
+                await del(results.video);
```
qooban, line 30: I think the code is self-descriptive enough, so we don't need additional comment here. WDYT?

test/cypress/plugins/index.ts
```diff
@@ -12,12 +12,26 @@
...
 
 const setupPluginsAndConfig: Cypress.PluginConfig = (on, config) => {
     config.baseUrl = 'http://localhost:4000';
 
+    // Delete the recorded video for specs that had no retry attempts
+    // https://github.com/cypress-io/cypress/issues/16377
+    on('after:spec', async (_spec, results) => {
+        if (results && results.video) {
+            // Do we have failures for any retry attempts?
+            const failures = results.tests?.some(test => test.attempts.some(attempt => attempt?.state === 'failed'));
```
qooban, line 27: If you think the code is not self-descriptive, I suggest to change the name of the const to describe it's meaning better rather than adding a comment. WDYT?

qooban (commented):
I like the direction overall, not sure if the code placement is right. See comments.

package.json
```diff
@@ -60,6 +60,7 @@
     "connected-react-router": "^6.5.2",
     "d3": "^3.5.17",
     "d3-format": "^1.3.2",
+    "del": "^6.0.0",
```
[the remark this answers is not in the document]
jawamartus: 1. Good point! Thanks.
2. I prefer to use this library. It's more safe than native node function (del prevents some cases like removing too much). I believe it's very light and safe to use. The code is taken partly from official cypress documentation and they are using del there. It's also very popular - for example "webpack-dev-server" is also using del library as a requirement. 
Therefore it's not a big dependency and we shouldn't be afraid of using it. However I understand your point of view and generally I like the guideline that native solutions > more dependencies.

test/cypress/plugins/index.ts
```diff
@@ -12,12 +12,26 @@
...
 
 import performCommonSetup from 'cloudify-ui-common/cypress/plugins';
+import del from 'del';
 // @ts-ignore Webpack config not in TS
 import getWebpackConfig from '../../../webpack.config';
 
 const setupPluginsAndConfig: Cypress.PluginConfig = (on, config) => {
     config.baseUrl = 'http://localhost:4000';
 
+    // Delete the recorded video for specs that had no retry attempts
```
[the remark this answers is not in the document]
jawamartus: It was good together feedback here and make some tests. I agree and I will place it in ui-common in another PR.

test/cypress/plugins/index.ts
jawamartus: You are right sir! :)

test/cypress/plugins/index.ts
jawamartus: You are right sir! :) The comment is redundant.

PR #1788, opened 2022-01-28
RD-4095 Fixed terraform topology rendering when dependency node is missing

kubama:
## Description
Fixed terraform topology rendering when dependency node is missing

## Screenshots / Videos
N/A

## Checklist

## Tests
This fix is transparent for any case that might by used as part of any test, so no need to run tests.

 

## Documentation
N/A

### DIFF
--- a/widgets/topology/src/Topology.tsx
+++ b/widgets/topology/src/Topology.tsx
@@ -305,8 +305,13 @@ export default class Topology extends React.Component<TopologyProps, TopologySta
                     .uniq()
                     .forEach((dependency: string) => {
                         const [dependencyName] = dependency.split('.').slice(-1);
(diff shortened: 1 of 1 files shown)


kubama:
@qooban If you are thinking about E2E test then it would be very difficult if possible at all - the blueprint that is attached to the ticket would have to be stripped of external resources usage and I'm not sure if it would allow to reproduce the same scenario then.
In case of unit test it would be possible but it would first require some refactoring around `Topology.tsx` that would extract the logic so it could be tested individually - not sure if it's worth it at this point.

qooban:
@kubama Thanks for clarification. I suspected it won't be easy, but just wanted to check. No blockers from my side, so let's have it in `master` and in `6.3.0-build` then.

PR #1771, opened 2022-01-25
RD-3944 Follow up to PR #1745

kubama:
## Description
Addresses unresolved comments in #1745

## Screenshots / Videos
N/A

## Checklist

- [X] My code follows the UI Code Style.
- [x] I verified that all tests and checks have passed.
- [X] I verified if that change requires a documentation update.
- [X] I added proper labels to this PR.

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1918/
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1925/ - failed spec rerun after providing a fix - PASS

 

## Documentation
N/A

PR #1789, opened 2022-01-28
RD-4098: Removed unnecessary plugin upload

NorbertAmartus:
## Description

Changes consists of removing the upload of the Terraform plugin, because it's being already covered.

## Screenshots / Videos

N/A

## Checklist

## Tests

 

I've executed [System tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/1948) only related to the test file, which was related to the issue mentioned in the [jira ticket](https://cloudifysource.atlassian.net/browse/RD-4098).

## Documentation

N/A

### DIFF
--- a/test/cypress/integration/widgets/blueprints_spec.ts
+++ b/test/cypress/integration/widgets/blueprints_spec.ts
@@ -666,8 +666,6 @@ describe('Blueprints widget', () => {
         });
 
         describe('create installable blueprint on submit from', () => {
-            before(() => cy.uploadPluginFromCatalog('Terraform'));
-
             beforeEach(openTerraformModal);
 
             function testBlueprintGeneration(terraformTemplateUrl: string, modulePath: string) {


NorbertAmartus:
Tests are passing now flawlessly 🎉 

Referring to our discussion on slack @qooban, It's hard to determine the speed gain upon blueprint update, as the cypress is not showing timestamps for every step of the test scenarios

Neither I could compare it, by running multiple tests with/without the functionality (that has been removed during this PR), because without the `before` method one of the test scenarios are depending on the other
And comparing 2 tests to one seems like a crazy idea 😁 

Additionally, after merging this PR I'd still have to do a cherry-pick to the `6.3.0-build` branch if I'm not mistaken?

qooban (approved):
LGTM.

> Referring to our discussion on slack @qooban, It's hard to determine the speed gain upon blueprint update, as the cypress is not showing timestamps for every step of the test scenarios

I was suggesting to check if blueprint upload (without explicitly uploading Terraform plugin before) takes longer 🙂 And if necessary adapt the timeouts, but if it works in Jenkins right now, let's keep it as it is.

> Additionally, after merging this PR I'd still have to do a cherry-pick to the 6.3.0-build branch if I'm not mistaken?

I don't think so. Updated `blueprint_spec.ts` is not in `6.3.0-build` branch as #1771 was not merged there (which is OK I think).

We can push your PR to `6.3.0-build`, but probably only with #1771 .

I've added @kubama to take a look at the change.

kubama:
I'm not sure I understand what's happening here.

> Changes consists of removing the upload of the Terraform plugin, because it's being already covered.

Where is it covered? I don't see any other place in the spec that uploads the plugin.

qooban:
> Where is it covered? I don't see any other place in the spec that uploads the plugin.

@kubama I assume it is covered by plugin auto-upload mechanism which is triggered by the core during blueprint upload (if blueprint relies on plugins).

kubama:
Ah, I forgot about this feature.

If so, then I think we should be able to remove explicit plugin installation from many (all?) other places where they are installed prior to blueprint deployment. Can we do it at once?

qooban:
> If so, then I think we should be able to remove explicit plugin installation from many (all?) other places where they are installed prior to blueprint deployment. Can we do it at once?

I believe so, but I would leave it to @NorbertAmartus .
Probably it requires just reviewing all `uploadPluginFromCatalog` uses?
image

NorbertAmartus:
https://github.com/cloudify-cosmo/cloudify-stage/pull/1789#issuecomment-1024211082

> Ah, I forgot about this feature.
> 
> If so, then I think we should be able to remove explicit plugin installation from many (all?) other places where they are installed prior to blueprint deployment. Can we do it at once?

I think that it'd be a better idea to complete it as a part of a separate ticket, because we'd still need to execute System tests, which would postpone merge of that PR to the start of the next week
And because of that we'd get a red system tests while proceeding with the release  

What's your opinion on that, @kubama?

NorbertAmartus:
> Probably it requires just reviewing all `uploadPluginFromCatalog` uses?

Yeah it would come down to do just that
If there wouldn't be a need for executing System Tests (if I could just test it locally 😁 ), then I'd gladly add this feature
But in the other case, I think that the better solution would be to do it in the next week

kubama:
@NorbertAmartus OK, if there is time constraint then lets proceed right away.

kubama (approved):

NorbertAmartus:
Okay, I'll create a separate PR soon, to modify mentioned behavior 😁

PR #1797, opened 2022-02-01
RD-4048: Added component which is used to animate menu subitems

NorbertAmartus:
## Description

Changes consists of adding transition to the menu subitems, so that they won't "jump", while the sidebar is being expanded.

## Screenshots / Videos

Same before/after effect has been achieved as shown in [jira task](https://cloudifysource.atlassian.net/browse/RD-4048)

## Checklist

## Tests

 

N/A, as the changes are strictly styling related 🚀 

## Documentation

N/A

### DIFF
--- a/app/components/sidebar/SideBarItem.tsx
+++ b/app/components/sidebar/SideBarItem.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
-import type { FunctionComponent, ReactNode } from 'react';
-import styled from 'styled-components';
+import type { FunctionComponent, ReactNode, CSSProperties } from 'react';
+import styled, { css } from 'styled-components';
 import type { MenuItemProps, SemanticICONS } from 'semantic-ui-react';
 import { Icon, Menu } from '../basic';
 import { expandedSidebarWidth } from './SideBar';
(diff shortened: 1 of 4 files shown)


app/components/sidebar/SideBarItem.tsx
```diff
@@ -36,28 +48,30 @@ const SideBarItem: FunctionComponent<SideBarItemProps> = ({
...
         height: '100%',
         width: expandedSidebarWidth,
         paddingTop: 13,
-        paddingLeft: subItem && 25,
+        paddingLeft: subItem && 15,
         ...style
     };
 
     return (
-        <SideBarItemWrapper className="sidebarItemWrapper">
```
NorbertAmartus, line 48: Just a note:
I've double checked and this `className` is not being used anywhere in the project

qooban (dismissed):
Checked UI - looks very good! :-)

Styling is not my strong area, so I'd like to someone else review it. I'll nominate @kubama 🙂 

Please check bundle size limit as it seems exceeded.

kubama (commented):
Looks very nice.

I only have one question - did you consider using https://react.semantic-ui.com/modules/transition/? I don't have much experience with this SUI component, but it looks like if it was targeted to solve similar kind of issues.

NorbertAmartus:
> Looks very nice.
> 
> I only have one question - did you consider using https://react.semantic-ui.com/modules/transition/? I don't have much experience with this SUI component, but it looks like if it was targeted to solve similar kind of issues.

I did not, thanks for letting me know about this feature 🥳 
As I'm now looking at it, implementing that functionality would require us to restructure `Sidebar` component to propagate the `$expanded` prop across it's children

Because only with having that information we could determine which stage of the animation should be visible on the screen

Currently that issue is being bypassed by overriding components styles by the `Sidebar` component, depending on the inner `$expanded`  prop value

kubama (approved):

kubama (approved):

PR #1800, opened 2022-02-02
RD-1293 Added support for TS in widget backend

kubama:
## Description
Added support for TS in widget backend

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/1971/

 

## Documentation
I'll update https://docs.cloudify.co/latest/developer/writing_widgets/widget-backend/ in separate PR


backend/package.json
```diff
@@ -29,6 +29,7 @@
     "delay": 2000
   },
   "dependencies": {
+    "@tsconfig/node12": "^1.0.9",
```
qooban, line 32: Maybe worth to use node v14 (as that is our current version we support)?

qooban (commented):
Added one comment.
As I'll be not available till the end of this week, please don't wait for my approval.

jawamartus (approved):
LGFM

backend/package.json
kubama: Done

backend/package.json
```diff
@@ -29,6 +29,7 @@
     "delay": 2000
   },
   "dependencies": {
+    "@tsconfig/node14": "^1.0.1",
```
jawamartus, line 32: What about package-lock.json? Do I think correctly that we should update it as well?

jawamartus (approved):
LGFM with exception for one doubt about package-lock.json

backend/package.json
kubama: I used `npm i` for adding this dependency so I assume `package-lock.json` would be updated if it was required.
Also, this package was already present as `ts-node` dependency (I added it as it is now used directly in an `import` statement) so I think all is good here.

PR #1802, opened 2022-02-04
RD-3240 Modified CI build to include component tests

kubama:
## Description
Modified CI build to include component tests

## Screenshots / Videos
N/A

## Checklist

## Tests
See CI build status

 

## Documentation
N/A


kubama:
The build passes but there is an issue on github side - it's waiting for 'Test frontend' check to be completed, but it was renamed to 'Test frontend unit'. I asked devops for assistance to resolve this issue.

jenkins/Jenkinsfile
```diff
@@ -56,11 +56,13 @@ pipeline {
...
+              echo "Install required packages"
+              sh 'rm -rf /var/lib/apt/lists/* && apt update && apt-get -y install postgresql-client && apt-get -y install jq'
               dir("${STAGE_DIR}/conf") {
                 withVault([configuration: configuration, vaultSecrets: secrets]) {
-                  sh """
-                    jq '.maps.accessToken = "${env.MAPS_ACCESS_TOKEN}"' app.json > app.json.tmp && mv app.json.tmp app.json
-                  """
+                  sh '''
+                    jq ".maps.accessToken = \\"$MAPS_ACCESS_TOKEN\\"" app.json > app.json.tmp && mv app.json.tmp app.json
+                  '''
```
NorbertAmartus, line 65: As there are other places which are using the old notation (with `"""`) I think we should stick to it or implement that in those places (L92-L96 😉) 

As I can see in the other Jenkins file, we are using the notation that you've used (`'''`), so it may be more advised to apply it to the lines mentioned above, so that we'd be consistent across all of the Jenkins files 🥳 

What do you think?

NorbertAmartus (commented):
Looks good 💪 
Added just one minor comment, regarding consistency across the Jenkins files

Also, thanks for explaining the pipeline issue!

jenkins/Jenkinsfile
kubama: The whole `Jenkinsfile` is a groovy script and as opposed to JS in groovy single and double quoted strings are not semantically equal. The differences are outlined here: https://groovy-lang.org/syntax.html#all-strings. As this is the case I don't see a good reason for trying to stick to one notation while the other one has its own features.

NorbertAmartus: Sure, in that case it wouldn't be a proper thing to do 😁 
Thanks for the explaination!

NorbertAmartus (approved):

PR #1803, opened 2022-02-07
RD-2946 Add hot reloading

leon-good-life:
## Description
This PR adds hot-reloading.

## Screenshots / Videos
https://user-images.githubusercontent.com/17104212/152801045-9a994ef8-f238-4f8c-92e8-a84692030174.mov

## Checklist

## Tests
I tested on the login page.

**TODO:**
- Check on some other pages, just to make sure. Currently I have an issue to log in - not related to this PR.

 

## Documentation
N/A


devServer.ts
```diff
@@ -16,9 +16,10 @@ const proxyOptions = {
 };
 
 const options = {
```
NorbertAmartus, line 18: **Optional**:
It would be a good opportunity to add typings for the `webpack-dev-server`

It would be done by executing this line:
```
npm install @types/webpack-dev-server@3.11.0 --save-dev
```

Then we could redeclare this line as 
```ts
const options: WebpackDevServer.Configuration = {
```

I think it could be beneficial, as typescript helped us with this issue by showing the configuration properties description 😁 
So if we'd have to reconfigure it in the future, we'd have an additional support 😉 

But that's just an optional suggestion

NorbertAmartus (commented):
Looks good to me 😉 
Added one **optional** comment related to the typescript support 

There is one minor thing, that none on the commits have a ticket number prefix, in our case it would be a `RD-2946`

As I can see, we are trying to keep that convention, so that after merging PRs, we're still able to jump quickly into the changes details, using the that number

If I'm not mistaken, while merging PR you've the ability to set the name of the squashed commit, so that it may contain the mentioned ticket number

So I'd say that there are 2 convenient ways of adding that prefix:
- by setting that during commits squash
- by renaming last commit 😁

devServer.ts
leon-good-life: Added: bc62016

NorbertAmartus (approved):
Thanks for addressing all the comments! :)
Looks simple yet great! 🥳 

I suspect that you''re taking a first [mentioned approach](https://github.com/cloudify-cosmo/cloudify-stage/pull/1803#pullrequestreview-874731257) of adding the jira ticket number as the commit message prefix and because of that I'm approving this PR 🎉

jawamartus (commented):
Well done! 

I tested it for widgets. It makes changes in widgets also visible after auto-refresh.

PR #1845, opened 2022-02-23
RD-4317 Add NPM script for building single widget 

qooban:
## Description

Custom widget developers are struggling with no easy method for widget building.
I added simple NPM script:
```
npm run build:widget <widget-name>
```
which builds only specific (`<widget-name>`) widget and also creates ZIP archive:
```
dist
├── executions.zip
└── widgets
    └── executions
        ├── backend.ts
        ├── README.md
        ├── widget.css
        ├── widget.js
        ├── widget.js.gz
        └── widget.png
```
Documentation updated accordingly.

## Screenshots / Videos
N/A

## Checklist

## Tests
1. I did mainly manual tests on different widgets (with/without widget backend).
2. [Stage-UI-System-Test #2132](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2132/) - one test failed, but seems like not related, so I consider is as PASSED

## Documentation
* Updated `README.md` files
* Updated user documention - https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2117
* Added deprecation notice to https://github.com/cloudify-cosmo/Cloudify-UI-Widget-boilerplate repository - https://github.com/cloudify-cosmo/Cloudify-UI-Widget-boilerplate/pull/5


package.json
```diff
@@ -23,6 +23,7 @@
     "build:analyse:widgets": "npm run build -- --env.analyse=widgets",
     "build:analyse:widgets-common": "npm run build -- --env.analyse=widgets-common",
     "build:coverage": "cross-env COVERAGE_CHECK=1 npm run build",
+    "build:widget": "run(){ npm run build -- --widget=$1 && cd dist/widgets && bestzip ../$1.zip $1; }; run",
```
kubama, line 26: What's the purpose of the wrapping `run`?

webpack.config.js
```diff
@@ -289,5 +300,17 @@ module.exports = (env, argv) => {
...
+        const widgetsConfiguration = configuration[1];
+        const widgetPath = path.join(__dirname, `./widgets/${widgetName}`);
+        if (fs.existsSync(widgetPath)) {
+            console.log('Building widget', widgetName);
+        } else {
+            exitWithError(`Invalid widget name provided. Widget directory "${widgetPath}" does not exist.`);
+        }
+
+        return widgetsConfiguration;
+    }
```
kubama, line 313: I think it would be a little bit cleaner if this section was moved before `configuration` array construction. We could first construct `widgetsConfiguration` object and then use it in `configuration` array.

widgets/README.md
```diff
@@ -1,15 +1,33 @@
...
+
 All built-in widgets resides in [this](.) directory.
 
-Documentation for them is available [here](https://docs.cloudify.co/latest/working_with/console/widgets/).
+Documentation for them is available [here](https://docs.cloudify.co/staging/dev/working_with/console/widgets/).
+
+## Custom widgets
 
-## Custom widgets development
+This section describes how to create new widget.
```
kubama, line 11:
```suggestion
This section describes how to create a new widget.
```

widgets/README.md
```diff
@@ -1,15 +1,33 @@
...
+
+### Creating widget
```
kubama, line 13:
```suggestion
### Creating a widget
```

widgets/README.md
```diff
@@ -1,15 +1,33 @@
...
 
 To create your own widget:
-1. Create new widget structure following the instructions from [here](https://docs.cloudify.co/latest/developer/writing_widgets/) 
-1. Place widget in [widgets](./widgets) directory
-1. Restart webpack dev server (`npm run devServer` from main folder)
-1. Install new widget (read [this](https://docs.cloudify.co/latest/working_with/console/configure-display/) for details)
+1. Create new widget structure following the instructions from [here](https://docs.cloudify.co/staging.dev/developer/writing_widgets/) 
+2. Place widget in [widgets](./widgets) directory
+
+### Testing widget
```
kubama, line 19:
```suggestion
### Testing a widget
```

widgets/README.md
```diff
@@ -1,15 +1,33 @@
...
+
+To test your own widget:
+1. Start the application (see [Application start section in main README file](../README.md#Setup))
+2. Add your widget to the page using **Edit Mode** (check [this](https://docs.cloudify.co/staging/dev/working_with/customization/edit-mode/#adding-widgets) for details)
 
 Since now, every single change in widget's source code should be reflected by webpack dev server after page reload.
+
+### Creating widget package
+
+To use your custom widget outside of the development environment, you need to create widget ZIP package.
```
kubama, line 29:
```suggestion
To use your custom widget outside of the development environment, you need to create a widget ZIP package.
```

widgets/README.md
```diff
@@ -1,15 +1,33 @@
...
+Assuming you are working on widget named `myWidget` just run `npm run build:widget myWidget` 
+and your custom widget package will be available in `dist/widgets` directory. 
+
+You can install widget ZIP package on Cloudify Manager (check [Adding widget section in Edit Mode page](https://docs.cloudify.co/staging/dev/working_with/customization/edit-mode/#adding-widgets) for details)
```
kubama, line 33:
```suggestion
You can install widget ZIP package on Cloudify Manager (check [Adding widgets](https://docs.cloudify.co/staging/dev/working_with/customization/edit-mode/#adding-widgets) for details).
```
Additionally, when I'm opening the link I'm getting kind of broken page (looks like lacking CSS):
image

qooban: Good catch. It should be https://docs.cloudify.co/staging/dev/working_with/console/customization/edit-mode/#adding-widgets .

package.json
qooban: That was the way I found to make use of argument passed in CLI to NPM script (see `$1`).
I'm open to any other solution. Maybe I missed something.

webpack.config.js
qooban: I thought about that originally, but wouldn't we have then unnecessary code duplication then?
I gave it a try - see [webpack.config.js.zip](https://github.com/cloudify-cosmo/cloudify-stage/files/8124043/webpack.config.js.zip) and let me know.

There are differences between configuration for single widget and for all widgets in `entry`, `output` and `plugins` (and they are on deep level).

BTW I found that copying widgets content (all except `src` folder) is done in configuration for app bundles, so I'll move it to widget configuration.

package.json
kubama: I have no other solution, I was just curious of the intent 🙂.

webpack.config.js
kubama: I'm not sure I understand the code duplication issue. Looks like I didn't go deep enough - I'll leave it up to you then.

kubama (approved):

jawamartus (commented):
1. Building whole project last about 20 seconds. 
a) How much time we save building one widget only?
b) Is it worth to create our code more complex and harder to maintain which saves a few seconds of our time? Do we get other benefits by building one widget only?
c) What is the purpose of the script? Would it be used inside of our CI or only locally before the module delivery to the customer?


2. The new script inside of package.json is not very readable. What do you think about extracting it to the .sh file?

3. Do we need additional dependency `bestzip`? What do you think about using `zip` system app? It is worth to consider if we plan to use it only locally.

Note1: Adding `build single widget` functionality create additional conditions inside of our complex webpack file. It can make our webpack file harder to update and mantain in the future. I consider building whole project and packing one widget from the whole build.

Note2: I created similar script during my development of windrive widget. Feel free to copy and modify it if you think that it is usable or somehow helpful.
https://github.com/cloudify-cosmo/cloudify-stage/blob/windriver/deployment-list/widgets/deploymentListWidgetZip.sh

qooban:
@jawamartus Thanks for your comments. Please find responses below.

Some background - we had https://github.com/cloudify-cosmo/Cloudify-UI-Widget-boilerplate , but it is not maintained anymore. So I wanted to provide alternative for custom widget developers. 

> 1. Building whole project last about 20 seconds.
>    a) How much time we save building one widget only?

On my laptop it takes ~30 seconds. In Jenkins it takes almost 2 minutes.
Building just one widget takes ~4 seconds on my laptop. So IMHO there's benefit.

>    b) Is it worth to create our code more complex and harder to maintain which saves a few seconds of our time? Do we get other benefits by building one widget only?
>    c) What is the purpose of the script? Would it be used inside of our CI or only locally before the module delivery to the customer?

I don't think we'll often use it, it's mainly for custom widget developers.

> 2. The new script inside of package.json is not very readable. What do you think about extracting it to the .sh file?

I'm not sure having `.sh` file for this one-liner is necessary, but I can consider that. The only benefit I can see is that I can add more explanation comments.

> 3. Do we need additional dependency `bestzip`? What do you think about using `zip` system app? It is worth to consider if we plan to use it only locally.

Not always you have `zip` installed. `bestzip` is a cross-platform solution. Also, I added it as `optionalDependency` only. So I don't think it is a big problem for us.

> Note1: Adding `build single widget` functionality create additional conditions inside of our complex webpack file. It can make our webpack file harder to update and mantain in the future. I consider building whole project and packing one widget from the whole build.

Agree that it gets a bit more complex, maybe there's some room for improvement in webpack file. Still I think it's better to have it there and not require building the whole project.

> Note2: I created similar script during my development of windrive widget. Feel free to copy and modify it if you think that it is usable or somehow helpful. https://github.com/cloudify-cosmo/cloudify-stage/blob/windriver/deployment-list/widgets/deploymentListWidgetZip.sh

It's not self-sufficient (you have to have `zip` installed). I prefer having cross-platform and universal solution to allow custom widget developers do their job with just a single command:
```
npm run build:widget myCustomWidget
```
and that's it.

scripts/buildWidget.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+WIDGET_NAME=$1
+bestzip --version > /dev/null || npm install -g bestzip
```
qooban, line 4: `npm install -g bestzip` will only be called when `bestzip` is not installed (so in most of the cases - the first time this script is executed).

qooban (commented):
Improved the solution a bit:
1. Created `scripts/buildWidget.sh` and added there conditional installation of `bestzip` package
2. Removed `bestzip` package from `optionalDependencies`

@jawamartus What do you think? 🙂

jawamartus (commented):
Generally it looks good. I made a few comments.

1. One question. I don't understand who are the custom widget developers. Do people outside the Cloudify develop own widgets? Does it concern only people from UI team?

2. What do you think about creating separate configuration object for single widget build inside of webpack.config.js? I see that now we check `isSingleWidget...` many times. Also we are checking isProduction, what is redundant in the case when it's SingleWidgetBuild (we know that it's production because we use always `npm run build`).  The same case with `devtool` variable. We may not need to use this variable for webpack configuration for building single object.

scripts/buildWidget.sh
qooban: No, it does not end up in `dist` directory. It ends up in the main folder. Please check yourself.
I prefer to keep my solution as it does the job properly and doesn't introduce a need to add `cd ...`.

webpack.config.js
```diff
@@ -16,6 +16,8 @@ const CONTEXT_PATH = '/console';
 module.exports = (env, argv) => {
     const isProduction = argv.mode === 'production';
     const isDevelopment = argv.mode === 'development';
+    const isSingleWidgetBuild = !!argv.widget;
+    const widgetName = argv.widget;
     const mode = isProduction ? 'production' : 'development';
```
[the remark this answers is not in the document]
qooban: 1. It's not semantically the same (`!isProduction` can also mean than `argv.mode` is not necessarily "development", it can be anything different than "production")
2. It's out of scope of this PR

so I'd like not to change it within this PR.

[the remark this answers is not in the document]
jawamartus: And that inconsistency is suspicious. Currently you can have isDevelopment set for false and mode equal 'development'.

But as I said it's not the scope of this PR. I just for the first first time was checking out the file and noticed it.

qooban:
@jawamartus 

> Do people outside the Cloudify develop own widgets? 

Yes. There are few customers creating their own widgets. In practice - anyone can create a widget, `cloudify-stage` is open sourced.

> Does it concern only people from UI team?

No.

> What do you think about creating separate configuration object for single widget build inside of webpack.config.js? I see that now we check `isSingleWidget...` many times. Also we are checking isProduction, what is redundant in the case when it's SingleWidgetBuild (we know that it's production because we use always `npm run build`).  The same case with `devtool` variable. We may not need to use this variable for webpack configuration for building single object.

1. Good catch with `isProduction`, I'll update L265 in webpack  configuration as with the current shape of `buildWidget.sh` - `isSingleWidgetBuild === true` means `isProduction === true`
2. Regarding separate configuration - I'm not conviced by the arguments you brought. I'm afraid of unnecessary code duplication. See this discussion - https://github.com/cloudify-cosmo/cloudify-stage/pull/1845#discussion_r812747815 . There's also an example of a change attached. I personally don't like it, it's longer and I think that it is more prone to introducing errors in single widget build configuration as we'll be mainly focused on normal configuration.

jawamartus:
> 2\. Regarding separate configuration - I'm not conviced by the arguments you brought. I'm afraid of unnecessary code duplication. See this discussion - https://github.com/cloudify-cosmo/cloudify-stage/pull/1845#discussion_r812747815 . There's also an example of a change attached. I personally don't like it, it's longer and I think that it is more prone to introducing errors in single widget build configuration as we'll be mainly focused on normal configuration.

I think it's not unnecessary code duplication. The code changes from different reasons in both cases. We don't need to check condition `isSingleWidgetBuild...` 7-8 times every time we run our `webpack.config.js` for production build.

What if we would add script for example for building single component in the future? How the code would looks like?

It's more about Single Responsibility Principle than DRY. Both configurations may be changed from different reasons. Currently changes in the configuration object which is responsible for building single widget affect our production build command. Currently the developer who is working on improving configuration of `SingleWidgetBuild...` may affects the other developer who is working on changing production webpack configuration.

scripts/buildWidget.sh
jawamartus: My mistake. Probably I was in the `dist` directory when I was checking it.

jawamartus (approved):

qooban:
@kubama @jawamartus As both of you were suggesting the same change, I changed the code accordingly 🙂 I'm still not sure it's better, but I'm OK with having it like that for now 🙂 

The main goal is to provide single widget build functionality for custom widget developers and both approaches are serving that purpose.

kubama (approved):

PR #1846, opened 2022-02-23
RD-2792: StageAPI - making type field ENUM type

jawamartus:
## Description

https://cloudifysource.atlassian.net/browse/RD-2792 (point 2)

The PR contains adjustment for typing related to GenericField component typescript migration. It also provides a refactoring in a few places.

## Checklist

## Tests

 

https://jenkins.cloudify.co/job/Stage-UI-System-Test/2173/ [passed]
 [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2253/)
 
 Tests fail due to two known issues present in the master as well.

## Documentation

AFAIK there is nothing to update except of the components library version: https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2129

### DIFF
--- a/app/components/EditWidgetModal.tsx
+++ b/app/components/EditWidgetModal.tsx
@@ -2,57 +2,55 @@
 import i18n from 'i18next';
 import _ from 'lodash';
 import PropTypes from 'prop-types';
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { getToolbox } from '../utils/Toolbox';
 
 import { ApproveButton, CancelButton, Form, GenericField, Message, Modal } from './basic';
 
+const getInitialFields = (configDef, configuration) =>
+    configDef
+        .filter(config => !config.hidden)
+        .reduce((fields, config) => {
+            fields[config.id] = _.get(configuration, `[${config.id}]`, config.value || config.default);
+            return fields;
+        }, {});
+
+function EditWidgetGenericField({ fields, setFields, ...restProps }) {
+    const handleInputChange = (proxy, field) => {
+        const { name } = field;
+        const value = GenericField.formatValue(
+            restProps.type,
+            restProps.type === GenericField.BOOLEAN_TYPE ? field.checked : field.value
+        );
+
+        setFields({ ...fields, [name]: value });
+    };
+    return <GenericField onChange={handleInputChange} {...restProps} />;
+}
+
 export default function EditWidgetModal({ configDef, configuration, show, onHideConfig, onWidgetEdited, widget }) {
-    function getInitialFields() {
-        const fields = {};
-
-        configDef
-            .filter(config => !config.hidden)
-            .forEach(config => {
-                const currValue = _.get(configuration, `[${config.id}]`, config.value || config.default);
-                fields[config.id] = currValue;
-            });
-
-        return fields;
-    }
-
-    const [fields, setFields] = useState(getInitialFields());
+    const [fields, setFields] = useState(getInitialFields(configDef, configuration));
 
     useEffect(() => {
-        if (show) setFields(getInitialFields());
+        if (show) setFields(getInitialFields(configDef, configuration));
     }, [show]);
 
-    function onApprove() {
+    const onApprove = useCallback(() => {
         if (configuration) {
             onWidgetEdited(widget.id, { configuration: { ...configuration, ...fields } });
         }
 
         onHideConfig();
         return true;
-    }
+    }, [fields]);
 
-    function onDeny() {
+    const onDeny = useCallback(() => {
         onHideConfig();
         return true;
-    }
-
-    function handleInputChange(proxy, field) {
-        const { name } = field;
-        const value = GenericField.formatValue(
-            field.genericType,
-            field.genericType === GenericField.BOOLEAN_TYPE ? field.checked : field.value
-        );
-
-        setFields({ ...fields, [name]: value });
-    }
+    }, []);
 
     return (
-        <Modal open={show} onClose={() => onHideConfig()} className="editWidgetModal">
+        <Modal open={show} onClose={onHideConfig} className="editWidgetModal">
             <Modal.Header>Configure Widget</Modal.Header>
 
             <Modal.Content>
@@ -60,7 +58,9 @@ export default function EditWidgetModal({ configDef, configuration, show, onHide
                     {configDef
                         .filter(config => !config.hidden)
                         .map(config => (
-                            <GenericField
+                            <EditWidgetGenericField
+                                fields={fields}
+                                setFields={setFields}
                                 component={config.component}
                                 default={config.default}
                                 description={config.description}
@@ -73,7 +73,6 @@ export default function EditWidgetModal({ configDef, configuration, show, onHide
                                 label={config.name}
                                 value={fields[config.id]}
                                 columns={config.columns}
-                                onChange={handleInputChange}
                                 widgetlessToolbox={getToolbox(undefined, undefined, undefined)}
                             />
                         ))}
--- a/app/utils/StageAPI.ts
+++ b/app/utils/StageAPI.ts
@@ -1,7 +1,8 @@
-import type { ComponentType, ReactElement, ReactNode } from 'react';
+import type { JSXElementConstructor, ReactElement, ReactNode, SyntheticEvent } from 'react';
 // NOTE: the file contains only types and is undetectable for ESLint
 // eslint-disable-next-line import/no-unresolved
 import type { SemanticCOLORS } from 'semantic-ui-react';
+import type { GenericFieldType } from 'cloudify-ui-components';
 import type * as BasicComponents from '../components/basic';
 import type * as SharedComponents from '../components/shared';
 import type * as StagePropTypes from './props';
@@ -83,7 +84,14 @@ export type { StageWidgetDefinition as WidgetDefinition };
 interface StageCustomConfigurationComponentProps<T> {
     name: string;
     value: T;
-    onChange: (event: unknown, field: { name: string; value: T }) => void;
+    onChange: (
+        event: SyntheticEvent<HTMLElement, Event> | undefined,
+        field: {
+            name?: string;
+            value: T;
+            checked?: boolean;
+        }
+    ) => void;
     widgetlessToolbox: StageWidgetlessToolbox;
 }
 export type { StageCustomConfigurationComponentProps as CustomConfigurationComponentProps };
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -58,7 +58,7 @@
     "blueimp-md5": "^2.7.0",
     "cloudify-blueprint-topology": "^2.5.5",
     "cloudify-ui-common": "^5.2.0",
-    "cloudify-ui-components": "^2.12.1",
+    "cloudify-ui-components": "file:cloudify-ui-components-2.12.1.tgz",
```
kubama, line 61: Committed by mistake I guess.

app/utils/StageAPI.ts
```diff
@@ -1,7 +1,8 @@
-import type { ComponentType, ReactElement, ReactNode } from 'react';
+import type { JSXElementConstructor, ReactElement, ReactNode, SyntheticEvent } from 'react';
 // NOTE: the file contains only types and is undetectable for ESLint
 // eslint-disable-next-line import/no-unresolved
 import type { SemanticCOLORS } from 'semantic-ui-react';
+import type { GenericFieldEnum } from 'cloudify-ui-components';
```
kubama, line 5: This is in fact comment to https://github.com/cloudify-cosmo/cloudify-ui-components/pull/125
I think a more accurate type name would by `GenericFieldType`.

package.json
jawamartus: It's a draft. Sorry, I didn't mention it clearly in the PR description. The changes in the following PR are made in the coherence with the changes of  cloudify-ui,components. The .tgz file contains the codebase from this PR. https://github.com/cloudify-cosmo/cloudify-ui-components/pull/125
When the new version of cloudify-ui-components will be released I replace it with a new version instead of tgz package created by `npm pack` command.

app/utils/StageAPI.ts
jawamartus: I prefer to keep this name because enum works slightly different in typescript than type and interface. enum still exist after transpilation to javascript. You can use it as a type but you can also use it in your javascript code like this: `GenericFieldEnum.BOOLEAN_LIST_TYPE`. It's exactly like enum from other programming languages. The word enum has more narrowed meaning and inform developers what it is exactly.

jawamartus: Sorry, I don't follow your reasoning with an antipattern. Is everything what we don't use is antipattern?

Some people use the Interface suffix for interfaces. It's very popular approach.

Me as a developer I want to know what I'm using. Currently in typescript there is almost no difference between type and interface. However there is big difference between enum and type or interface. The name with a suffix Type can suggest that it is declared by `type` or `interface` and would disappear after transpilation.

On the other hand I was thinking about GenericFieldTypeEnum. However the word type is redundant here because Enum it's enumerated list and there is nothing else to enumerate than GenericField type.

kubama: > Is everything what we don't use is antipattern?

I have no idea where this conclusion comes from.
I think this is an anti-pattern because adding 'Interface' or 'Enum' postfixes is most commonly a workaround for poorly named types which do not describe precisely what they are. And I think this is actually the case here - if you strip the `Enum` part then we are left with `GenericField`, which does not describe exactly what it is.

As I wrote previously, the phrase 'Type' in `GenericFieldType` would refer to a type of the field, not TS type. To be honest, I would never think of TS type when reading 'GenericFieldType'. On the other hand if the type is named `GenericFieldEnum` then it means it enumerates generic fields. But what is a generic field? The type of a generic field is something different than a generic field itself. And in this case we are referring to the type of a generic field.
Following this reasoning the name could be then `GenericFieldTypeEnum`, but as you pointed out it would be redundant - adding `Enum` is no longer necessary as the base part of the type already describes it precisely.

The last thing - if you look at the values of the enum then each of them is postfixed with `_TYPE`. This is unnecessary duplication (or actually multiplication 😄). Why not move it to the type name and delete it from the individual values names?

It's good to discuss, but it's still a minor thing, so if we still disagree then lets allow @qooban to have the final word 🙂.

jawamartus: Is that the reason for you to no approve the PR?

kubama: No, but it's marked as a draft.

qooban: Good discussion 🙂 👍🏻 

I like the reasoning provided by @kubama and agree that `GenericFieldType` seems to be a better fit here.

jawamartus: I have a different opinion. Is that enum name is a blocker for the PR to be approved?

qooban: @jawamartus What's the reason of discussing it and providing strong arguments that 2 devs agreed upon and then end up with just approving the PR without a change?

jawamartus: But one have different opinion. Programming is not a democracy. I don't think that it should be the case for not approving the PR.

jawamartus: The phrase `Enum` I used refer to the data type. I don't see anything invalid in that name what should be a blocker for the PR.

kubama: I don't think the purpose of code review is sharing opinions. Rather than that I would expect exchanging arguments supporting different solutions to choose the best one. I gave my arguments in https://github.com/cloudify-cosmo/cloudify-stage/pull/1846#discussion_r818515364. As I can see you didn't refer to them anyhow and silently ignored them. This is not only discouraging - it looks like I just wasted my time on commenting here. This puts in question the whole idea of having code review 🙄. To be clear - as a reviewer I don't expect PR author to agree with me on everything. From my experience an argument supported discussions where both sides refer to one another's arguments almost always lead to final agreement on the topic (which doesn't necessarily mean agreeing with the initial reviewer's comment).

You might say that it's a minor thing and not a blocker - you are right, it is and we can leave it as it is if you insist. However please consider my comment as general thoughts on the topic.

jawamartus: You shared your opinion without any arguments here: ` I think a more accurate type name would be GenericFieldType.` which was very discouraging and blocking my PR. Please provide arguments in your comments next time if it is blocking your approval.

Let me recap it. I think that we have some misunderstanding.

You said:
`The phrase Type in GenericFieldType would refer to a type of the field, not TS type.`

After rethinking I'm saying that:

`The phrase Enum in GenericFieldEnum refer to both the type in typescript and to the module content after transpilation. To the TS type and to the type of the data stored inside`. 

The meaning of Enum it's not only the typescript data type like interface or type. It also contains information that it is an object with enumerated values. Which have more precise and narrow meaning. `Type` can be confusing and misleading and also can be associated wrongly with the type of GenericField component. That's my arguments.

However I will change it because I want to close the PR fast. Other things are waiting and we have a lot of things to deliver to the customers next sprint.

jawamartus: `It's good to discuss, but it's still a minor thing, so if we still disagree then lets allow @qooban to have the final word 🙂.` Sorry, programming is not a democracy. I don't understand this approach.

jawamartus: `The last thing - if you look at the values of the enum then each of them is postfixed with _TYPE. This is unnecessary duplication (or actually multiplication 😄). Why not move it to the type name and delete it from the individual values names?` - In the case of using GenericFieldType we can of course remove it. However the change from Enum to Type doesn't provide a business value to the customers. I will do it because seemingly there is no other way to get approval fast.

qooban: I see there are many misunderstandings. Let's have a call on Monday and try to understand each other.

> Sorry, programming is not a democracy. I don't understand this approach.

I don't understand your approach as well. Imagine 4 out of 5 developers not agreeing with you, but they will still have to use your code in the future - will you still be insisting to use your approach? Does it make sense?

> However the change from Enum to Type doesn't provide a business value to the customers.

This is not an argument for me. Most of the things we are discussing in code review are not providing business value (especially in tech-debt tickets). What is the value for customers (I'm not talking about custom widget developers) to have TypeScripted `GenericField`?

kubama: > You shared your opinion without any arguments here

I provided arguments in the following comments. I don't see a point in providing all possible arguments upfront. Detailed arguments can be always provided when there is a disagreement.

> which was very discouraging and blocking my PR

Well, the purpose of a CR comment is to discuss and possibly improve certain code fragment, it's not a surprise that it blocks PR from being merged for a while 🙂.

> Sorry, programming is not a democracy. I don't understand this approach.

No idea what's the relation to democracy 🙄. We are not voting here - I just think it's always good to have another voice, especially when two parties disagree.

> However the change from Enum to Type doesn't provide a business value to the customers.

Of course it doesn't, it's a refactoring. But the whole PR is about refactoring, isn't it?

kubama: > I see there are many misunderstandings. Let's have a call on Monday and try to understand each other.

Good idea, thanks for proposing this 👍.

app/components/EditWidgetModal.tsx
```diff
@@ -2,65 +2,65 @@
 import i18n from 'i18next';
```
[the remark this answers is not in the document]
jawamartus: Thanks for you suggestion. I prefer to keep the scope of the PR as it is.

qooban (commented):
You asked for a review in this PR, but is it ready for review? I still see `cloudify-ui-components-2.12.1.tgz` test file. 

Could you also clean the description a bit? Do we need any documentation change in relation to this PR?

jawamartus:
The tests failed because of this PR is still not merged: https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2129

package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cloudify-stage",
   "version": "6.4.0-.dev1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
qooban, line 4: Probably something we should not do 🙂

jawamartus:
```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```

What is our current production version of node and npm ? I use:
```
Now using node v14.18.1 (npm v6.14.15)
```

jawamartus: Fixed with npm v8.5.4

jawamartus:
>You asked for a review in this PR, but is it ready for review? I still see cloudify-ui-components-2.12.1.tgz test file.
>Could you also clean the description a bit? Do we need any documentation change in relation to this PR?

The PR is ready and tests are passed except the known issues. Description has been cleaned a little bit. I'm not aware about any required documentation change in relation to this PR except other related PR related to cloudify-ui-components version change.

qooban (approved):
Looks good from my perspective 👍🏻

jawamartus:
Changes in doc merged. Test rerun: https://jenkins.cloudify.co/blue/organizations/jenkins/cloudify-stage/detail/internal%2Frd-2792/22/pipeline

PR #1853, opened 2022-02-25
RD-4367: Made all checkboxes required

NorbertAmartus:
## Description

Changes consists of requiring one of the checkboxes, which was optional till this day 😁 

## Screenshots / Videos

N/A

## Checklist

## Tests

 

[System test](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2141) (related to the modified functionality) has been executed 

## Documentation

N/A
(as I've investigated, the modal is not documented, but that's just a note 😁)


NorbertAmartus:
As I'm currently unable to run the application, I've investigated all of the places which was correlated with the modal (both on frontend and backend) and made necessary changes

Tests has passes without a problem, so I think that everything is working as expected, but if it would be okay with you @qooban, I'd ask you for your help when it comes to taking a look on the UI side, if the functionality is working as expected

 I suspect that everything is working well, but it's better to be sure, rather that to speculate 😁

NorbertAmartus:
As the feature should go with the `6.3.1` version, after merging this PR I'll do another one for that exact purpose 🚀

qooban (approved):

PR #1862, opened 2022-03-04
RD-382 Migrate from `request` to `axios` in stage backend

kubama:
## Description
Migrate from `request` to `axios` in stage backend

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/2214/

 

## Documentation
N/A


NorbertAmartus:
I won't be able to make a review today
I'll get to it tomorrow morning

backend/handler/ArchiveHelper.ts
```diff
@@ -5,7 +5,7 @@ import pathlib from 'path';
 import sanitize from 'sanitize-filename';
 import decompress from 'decompress';
 import multer from 'multer';
-import { Request } from 'express';
+import { AxiosRequestConfig, AxiosResponse } from 'axios';
```
NorbertAmartus, line 8:
```suggestion
import type { AxiosRequestConfig, AxiosResponse } from 'axios';
```

backend/handler/AuthHandler.ts
```diff
@@ -81,7 +81,7 @@ export function getTokenViaSamlResponse(samlResponse: string) {
 
 export function getAndCacheConfig(token?: string) {
     return jsonRequest<ConfigResponse>('GET', '/config', {
-        'Authentication-Token': token
+        'Authentication-Token': token as string
```
NorbertAmartus, line 84: Method `getAndCacheConfig` is being used in 3 places:
- backend/routes/Auth.ts
- backend/routes/ServerProxy.ts
- backend/handler/AuthHandler.ts

In all of those places, the token is being passed as a string
Because of that, I think that we should be able to modify it's parameter to always require a string
It'd slightly simplify this piece of code 😁 

 ```suggestion
 export function getAndCacheConfig(token: string) {
    return jsonRequest<ConfigResponse>('GET', '/config', {
        'Authentication-Token': token
 ```

backend/handler/ManagerHandler.ts
```diff
@@ -25,69 +27,41 @@ export function getApiUrl() {
...
+    timeout?: number
+) {
+    return request(method, url, { headers, data }, timeout)
+        .catch(err => {
+            if (err.response) {
+                throw err.response.data;
+            }
+            throw err;
+        })
+        .then(res => <ResponseBody>res.data);
```
NorbertAmartus, line 66: Just out of curiosity, the order of `catch` and `then` matters in that case?
Usually it's being used in the different order, but there is a [small difference](https://stackoverflow.com/questions/42013104/placement-of-catch-before-and-after-then#:~:text=If%20you%20return%20a%20normal,catch()%20will%20be%20called.), which may favor one over another 
And just wanted to ask if the order is more of a coincidence or it was intentional 😁

backend/handler/RequestHandler.ts
```diff
@@ -1,36 +1,32 @@
-import req from 'request';
-import type { Response } from 'request';
+import axios, { AxiosResponse } from 'axios';
+import type { AxiosRequestConfig } from 'axios';
```
NorbertAmartus, line 2:
```suggestion
import axios from 'axios';
import type { AxiosResponse, AxiosRequestConfig } from 'axios';
```

backend/routes/Plugins.ts
```diff
@@ -3,14 +3,13 @@ import type { Request, Response, NextFunction } from 'express';
 import multer from 'multer';
 import yaml from 'js-yaml';
 import _ from 'lodash';
-import request from 'request';
+import axios, { AxiosRequestHeaders } from 'axios';
```
NorbertAmartus, line 6:
```suggestion
import axios from 'axios';
import type { AxiosRequestHeaders } from 'axios'
```

NorbertAmartus (commented):
Extremely good job! 🎉 
Definitelly there are some key differences between using axios and request (as you've said on the daily) and it was nice to look on them from the practical point of view 💪 

I've left some minor suggestions/comments, but looking at the amount of work related to that task, it may be named as "nothing" 😄 

Once more, well done! 🥳

backend/handler/ArchiveHelper.ts
```diff
@@ -121,16 +122,17 @@ export function saveDataFromUrl(url, targetDir, req?: Request) {
...
+            if (req) {
+                options.headers = { ...req.headers, ...userAgentHeader };
+            }
+            getRequest = ManagerHandler.request('GET', archiveUrl, options);
         }
 
-        if (req) {
-            req.pipe(getRequest);
-        }
+        getRequest.then(onSuccessFetch).catch(onErrorFetch);
```
qooban, line 135: I'm wondering if that change solves RD-571?

backend/routes/ServerProxy.ts
```diff
@@ -1,74 +1,42 @@
...
 
 const router = express.Router();
-const logger = getLogger('ServerProxy');
+setUpRequestForwarding(router);
 
-interface RequestWithServerUrl extends Request {
-    su?: string;
-}
-
-function errorHandler(url: string, res: Response, err: any) {
```
qooban, line 17: Just checking if my understanding is correct. After removing that function, we will rely on `axios` error handler and error messages provided by that library, right?

qooban (commented):
Very well done! API is now much more clear for me and we don't need to do all Promise wrappers we used to do with `request`. I like that part:
image

Added just few questions, no blockers from my side.

backend/handler/ArchiveHelper.ts
kubama: I think there is no change in context of RD-571

backend/routes/ServerProxy.ts
kubama: Kind of, instead of the message returned by this function I'm using `err.code` and `err.message` as the final error message. I think it makes this function just not needed, as messages returned by axios are descriptive yet short. I noticed that in some cases `err.message` is missing, but then `err.code` provides a common socket error code which is commonly recognizable.

qooban: Sounds great 👍🏻

backend/handler/ArchiveHelper.ts
qooban: OK, I just hoped we might kill two birds with one stone, but apparently not 🙂

backend/handler/ManagerHandler.ts
kubama: It was not intentional. I think it makes no difference in this particular case, but I decided to fix it as it indeed looked surprising.

backend/handler/AuthHandler.ts
kubama: Yes. Looks like this function doesn't event make much sense with nullish token.
Applied.

qooban (approved):

PR #1865, opened 2022-03-08
RD-4253: Display Form.TextArea on textarea type + refactoring

jawamartus:
https://cloudifysource.atlassian.net/browse/RD-4253

## Description

N/A
## Screenshots / Videos

image

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2198/)

## Documentation

N/A

### DIFF
--- a/test/jest/widgets/common/InputUtils.test.tsx
+++ b/test/jest/widgets/common/InputUtils.test.tsx
@@ -1,4 +1,3 @@
-import React from 'react';
 import { shallow } from 'enzyme';
 
 import InputsUtils from 'common/src/InputsUtils';
--- a/test/jest/widgets/common/InputUtils.test.tsx
+++ b/test/jest/widgets/common/InputUtils.test.tsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import InputsUtils from 'common/src/InputsUtils';
+import { noop } from 'lodash';
+import { Form } from 'cloudify-ui-components';
+
+describe('InputsUtils.getInputField', () => {
+    let input: Record<string, any>;
+    const value = 'value';
+    const onChange = noop;
+    const error = '';
+
+    beforeEach(() => {
+        input = {
+            name: 'name',
+            default: 'defaultName',
+            type: 'textarea',
+            constraints: '',
+            display: {
+                rows: 5
+            }
+        };
+        InputsUtils.getRevertToDefaultIcon = () => undefined;
+    });
+
+    it('Renders Textarea with default rows value', () => {
+        delete input.display.rows;
+        const wrapper = shallow(InputsUtils.getInputField(input, value, onChange, error));
+        const wrapperTextArea = wrapper.find(Form.TextArea);
+        expect(wrapper.find(Form.TextArea)).toHaveLength(1);
+        expect(wrapperTextArea.props().rows).toEqual(InputsUtils.DEFAULT_TEXTAREA_ROWS);
+    });
+
+    it('Renders Textarea with appropriate rows prop', () => {
+        const wrapper = shallow(InputsUtils.getInputField(input, value, onChange, error));
+        const wrapperTextArea = wrapper.find(Form.TextArea);
+        expect(wrapperTextArea.props().rows).toEqual(5);
+    });
+});
--- a/widgets/common/src/InputsUtils.tsx
+++ b/widgets/common/src/InputsUtils.tsx
@@ -339,10 +339,8 @@ class InputsUtils {
                         </div>
                     </div>
                 );
-            case 'textarea':
-                const {
-                    display: { rows = InputsUtils.DEFAULT_TEXTAREA_ROWS }
-                } = input;
+            case 'textarea': {
+                const rows = input?.display?.rows ?? InputsUtils.DEFAULT_TEXTAREA_ROWS;
 
                 return (
                     <div style={{ position: 'relative' }}>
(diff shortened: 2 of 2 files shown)


test/jest/widgets/common/InputUtils.test.tsx
```diff
@@ -0,0 +1,39 @@
+import { shallow } from 'enzyme';
+
+import InputsUtils from 'common/src/InputsUtils';
+import { noop } from 'lodash';
+import { Form } from 'cloudify-ui-components';
+
+describe('InputsUtils.getInputField', () => {
```
qooban, line 7: I'm wondering why you decided to use `enzyme` test and not Cypress component test?

It's not a blocker at all, but AFAIR we introduced Cypress components test to use it more instead of `enzyme` based tests. Maybe something to talk about during weekly or improve our testing guidelines. I'll add a point.

widgets/common/src/InputsUtils.tsx
```diff
@@ -338,7 +339,18 @@ class InputsUtils {
...
+            case 'textarea': {
+                const rows = input?.display?.rows ?? InputsUtils.DEFAULT_TEXTAREA_ROWS;
 
+                return (
+                    <div style={{ position: 'relative' }}>
+                        <Form.TextArea name={name} value={value} onChange={onChange} rows={rows} />
+                        <div style={{ position: 'absolute', top: 10, right: 10 }}>
+                            {InputsUtils.getRevertToDefaultIcon(name, value, defaultValue, onChange)}
+                        </div>
+                    </div>
```
qooban, line 351: Seeing that almost the same 5 lines are in L357 (the only difference I see is `rows` prop), maybe worth to extract it to separate component?

widgets/common/src/InputsUtils.tsx
```diff
@@ -310,9 +301,19 @@ class InputsUtils {
...
+            case 'float': {
+                // Limit numerical values if at least one of range limitation constraints is set
+                const inRange = getConstraintValue('in_range');
+                const greaterThan = getConstraintValue('greater_than');
+                const greaterOrEqual = getConstraintValue('greater_or_equal');
+                const lessThan = getConstraintValue('less_than');
+                const lessOrEqual = getConstraintValue('less_or_equal');
+
+                const min = _.max([inRange ? inRange[0] : null, greaterThan, greaterOrEqual]);
+                const max = _.min([inRange ? inRange[1] : null, lessThan, lessOrEqual]);
```
qooban, line 315: I understand the change and from my perspective it's fine, but AFAIK there's no need to provide type for an input. Maybe not for all cases, I don't know.

Could you just verify if it's possible to define this kind of constraints without specifying `integer` or `float` type?

widgets/common/src/InputsUtils.tsx
```diff
@@ -23,6 +23,8 @@ HelpProperty.defaultProps = {
 };
```
qooban, line 23: Have you considered full migration to TS?

Not having it is not a blocker for this PR, but I think it is inevitable in the near future (as you'll be working in this area).

qooban (commented):
Nicely done 👍🏻 

Didn't test the UI. Added few comments.

test/jest/widgets/common/InputUtils.test.tsx
jawamartus: Unit Tests are much faster and shorten feedback loop.

widgets/common/src/InputsUtils.tsx
jawamartus: Thanks for the suggestions. I decided to keep it as it is. If you want to get to know detailed reasons for that, please have a call with me. It would be much faster.

widgets/common/src/InputsUtils.tsx
jawamartus: Good idea. I think there is plenty to do with this component. However I decided to limit the scope of the changes.

widgets/common/src/InputsUtils.tsx
jawamartus: > Could you just verify if it's possible to define this kind of constraints without specifying integer or float type?

Sorry, I don't understand the comment. I didn't make a change here IIUC. The behaviour is the same as previously. What should I verify? What do you mean by `constraints without specifying integer or float type`?

widgets/common/src/InputsUtils.tsx
qooban: It's OK not to do it right now, but as I said - as you'll be working here (New Input Parameters epic), I think you should do it in one of the next PRs. Can be done partially, but IMHO New Input Parameters epic is a great chance to have it done 🙂

widgets/common/src/InputsUtils.tsx
qooban: If you don't want to provide reasoning here, then yes, let's discuss it then during our call.

test/jest/widgets/common/InputUtils.test.tsx
qooban: Reasonable for me. Let's keep it.

widgets/common/src/InputsUtils.tsx
jawamartus: Resolved.

qooban (approved):
LGTM 👍🏻 

Once you verify all is fine in tests, feel free to merge it.

jawamartus:
System tests failed because of known issue only. I'm merging.

PR #1879, opened 2022-03-21
RD-4118 RD-4125 RD-4528 Added deployment_id and blueprint_id to inputs and workflow parameters. Bug fix - lack of support for .yml files

jawamartus:
## Description

The PR contains:
1. Adding support for new blueprint inputs and workflow parameters:
 - deployment_id
 - blueprint_id
2. Fix a bug related to lack of support for extensions on frontend side:
- .yml, .YML, .YAML. (RD-4528)

As agreed with Jakub N., tests would be provided in the separated PR.

## Screenshots / Videos

blueprint_id:
image

deployment_id:
image

Workflow parameters:
image

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2271/)


widgets/common/src/InputsUtils.tsx
```diff
@@ -351,6 +347,40 @@ class InputsUtils {
...
                 );
             }
+            case 'deployment_id': {
+                const fetchUrl = '/searches/deployments?_include=id';
+
+                return (
+                    <DynamicDropdown
+                        name={name}
+                        error={!!error}
+                        icon={InputsUtils.getRevertToDefaultIcon(name, value, defaultValue, onChange)}
```
qooban, line 357: I'm wondering will the `default` parameter be used in deployment inputs or workflow parameters when using `deployment_id` type. This type is kind of dynamic, so I'm wondering what will happen if there's no default deployment_id available in the system? Will `DynamicDropdown` handle that?

Don't need to change anything right now, maybe worth to consider having a test for that (as we agreed - in the next PRs).

widgets/common/src/BlueprintActions.ts
```diff
@@ -241,7 +241,8 @@ export default class BlueprintActions {
 
         let promise;
         if (file) {
-            const compressFile = _.endsWith(file.name, '.yaml') || _.endsWith(file.name, '.yml');
+            const compressFile =
+                _.endsWith(file.name.toLowerCase(), '.yaml') || _.endsWith(file.name.toLowerCase(), '.yml');
```
qooban, line 245: Since we don't have a way to reuse code between stage frontend and stage backend (except using cloudify-ui-common), I'm OK with having a duplicate right now. We'll probably be able to fix it once we implement RD-1103. 

But looking only at frontend part - we are using the same code in `BlueprintActions` and in `BlueprintSources`, so maybe worth to extract it to a simple function somewhere in `Stage.Common`? Sth like:
```
import { endsWith } from 'lodash';
function isYamlFile(filename: string) {
  const lowerCasedFilename = filename.toLowerCase();
  return endsWith(lowerCasedFilename, '.yaml') || endsWith(lowerCasedFilename, '.yml')
}
```

As @kubama is working on `Stage.Common` refactoring (RD-3027) worth to consult with him.

widgets/common/src/DynamicDropdown.tsx
```diff
@@ -167,23 +169,35 @@ export default function DynamicDropdown({
...
-                        },
-                        {
-                            _sort: valueProp,
-                            _size: pageSize,
-                            _offset: nextPage * pageSize
-                        }
-                    )
-                })
+            if (!_.isUndefined(constraints)) {
+                const constraintsObject = _.isArray(constraints) ? Object.assign({}, ...constraints) : {};
```
qooban, line 187: Minor thing - as we are trying to avoid accessing `lodash` functions with `_`, can we import `isUndefined` and `isArray` and use it directly here?

Ref.: [Ongoing Refactors in Cloudify UI - Using explicit imports in widgets](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Using-explicit-imports-in-widgets)

qooban (commented):
Looks good! I added just a few comments. Did not test the UI, I assume we will check that anyway soon in the tests.

Regarding documentation - I think we don't have anything to update in relation to newly added input types, so we should be good without documentation changes for RD-4118, RD-4125 and RD-4528.

widgets/common/src/InputsUtils.tsx
jawamartus: I think it's a question to our PMs.

widgets/common/src/BlueprintActions.ts
jawamartus: Sure, I may do it. However please consider that it will take additional ~two-three days to accomplish that.
It requires:
1. Consulting with Jakub M.
2. Agree for the solution.
3. Fix a PR.
4. Ask for rereview (+1 day).
5. Ask for review of Jakub M.
6. Some time for adding potential improvements.
7. Potential problems with system tests and running system tests again.

Are you sure that you want me to spend my time for that? It would put at risk other items assigned to sprint. However it's up to you.

Worth to notice that we had here duplicity before.

IMHO it's not worth an effort to extract this two command to separate function now. What is more, every function call it's another brick in the call stack.

qooban: ~2-3 days for extracting a function? I don't believe you are serious 🤯 

> Worth to notice that we had here duplicity before.

Yes, exactly, and the bug you discovered was probably due to that duplication 🙂 We updated one place and forgot about the other, so IMHO the proper fix would take fixing root cause into account.

> IMHO it's not worth an effort to extract this two command to separate function now. What is more, every function call it's another brick in the call stack.

I don't understand such argument. If we would take care more about number of function calls in call stack, then we would never extract common functions 🙂 

Anyways - I don't want to spend time on that - I don't agree with your approach, but I'm not going to block this PR. Let's leave it as is.

widgets/common/src/InputsUtils.tsx
qooban: > I think it's a question to our PMs. Do we want to allow specify default value for deployment_id and blueprint_id?

It's not a question to PMs only. I thought that you added it to allow resetting the value to the default (by having an `undo` icon in the right side of the input field, you can check it for other input fields). Does it work?

> BTW I see some strange thing. Why icon props contain default value? What default value have common with icon? It's odd.

I don't understand your question. It's you who passed `getRevertToDefaultIcon` as an `icon` prop 🙂 So I can ask you the same question. Why you did it?

widgets/common/src/DynamicDropdown.tsx
jawamartus: Done!

qooban (approved):

widgets/common/src/BlueprintActions.ts
jawamartus: Yes, 2-3 days. Asynchronous communication is costly. One single system test last around 1 hour. Validating system tests currently is also time consuming. A few runs may be required before potential rereviews. Testing the project locally again also require time. It looks that we don't have a tests for accepted file extension. Also I would love to consult the PR and wait for Jakub M. opinion because as you said he is working on something related. Also some potential merge conflicts to resolve may appear. Everything here is time consuming.

Of course I would be able to start something in the meantime of waiting for responses in tests. But I assume that the PR would be closed in the next 2-3 days with the requested changes.

So, It's up to you. I'm happy with your decision whatever it is.

jawamartus:
Topology spec test rerun: https://jenkins.cloudify.co/job/Stage-UI-System-Test/
[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2276/)

jawamartus:
It failed. I need to investigate it deeper tomorrow.

jawamartus:
The branch wasn't existing. Running tests again for topology spec: https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/2277/pipeline

jawamartus:
Passed! It looks that it was rarely occurring error in our topology spec test on Stage.

PR #1884, opened 2022-03-23
RD-4577: Fix for improper Content-Type header in requests in Safari and Chrome browsers

jawamartus:
## Description

https://cloudifysource.atlassian.net/browse/RD-4577

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2279/)


app/utils/External.ts
```diff
@@ -30,7 +30,7 @@ interface RequestOptions {
 }
 
 function getContentType(type?: string) {
-    return { 'content-type': type || 'application/json' };
+    return { 'Content-Type': type || 'application/json' };
```
qooban, line 33: According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2 header field names are case-insensitive. We didn't have problems due to that before.

Is that really a problem? Can you point out a place where Safari and Chrome browsers are not supporting lower-cased headers?

jawamartus: It's not related to the size of the letters in HTTP request header field name. Please read the bug description. I tested the PR it removes the bug.

jawamartus: In the future we may consider a refactoring of this part of our app where we pass json to options.headers. We may use Headers object instead which is dedicated to this purpose.

It may require a lot of work and testing and it's not in the scope of this PR I think.

qooban (changes requested):

app/utils/External.ts
qooban: > It may require a lot of work and testing and it's not in the scope of this PR I think.

Please create ticket for that then and let me know about the ID.

jawamartus: I think it would still make a space for the other related bugs. What do you think about using Headers object instead of passing json to the headers?

jawamartus: The problem has been solved in proposed solution. What "whole problem" do you mean?

jawamartus: > Is that really a problem? Can you point out a place where Safari and Chrome browsers are not supporting lower-cased headers?

Answering your question. Safari and Chrome are supporting case-insensitive headers. However in different way.

Regarding first question. What do you mean by "that"? The proposed solution? Yes it's fixing the bug.

qooban: I guess it'd be easier to have a chat f2f tomorrow morning to discuss all of that 🙂 
Scheduled it (let me know if the time is OK).

jawamartus: I created a ticket you requested: https://cloudifysource.atlassian.net/browse/RD-4579

widgets/common/src/DynamicDropdown.tsx
```diff
@@ -199,9 +199,6 @@ export default function DynamicDropdown({
                 const constraintsObject = isArray(constraints) ? Object.assign({}, ...constraints) : {};
                 fetchPromise = toolbox.getManager().doPost(fetchUrl, {
                     params,
-                    headers: {
```
qooban, line 202: Oh, so no need to pass it? Even better I think 🙂

qooban (approved):

app/utils/External.ts
jawamartus: Solution changed. Tested manually in Firefox and Chrome.

qooban:
@jawamartus Once https://github.com/cloudify-cosmo/cloudify-stage/pull/1886 is merged, you should be able to get green CI.

PR #1889, opened 2022-03-24
RD-3827 Add cookie-based authentication for some UI backend endpoints

qooban:
## Description

This PR aims to add authenticaton for all stage-backend routes which are not used before logging in.

The following changes were applied:
* Removed token-based authentication for stage-backend routes and applied cookie-based authentication in most of the backend routes (not always possible, see comments in `backend/app.ts`)
* Removed all uses of `Authentication-Token` header from the frontend side and Cypress tests
* Improved typing and logging in passport strategies files
* Improved types exposed for authentication routes
* Added more backend tests to satisfy code coverage check

TODO:
- [x] Check Okta login if it is possible to add authentication to `/auth/manager` endpoint

## Screenshots / Videos
N/A

## Checklist

## Tests
1. [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2313/) (28.03.2022 06:14 GMT) - some failures, but only known issues, so I'm considering it as PASSED
2. [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2316/) (28.03.2022, 13:10 GMT) - only Deployment Info widget test failed (known issue - RD-4462), so PASSED

## Documentation
N/A


backend/auth/CookieStrategy.ts
```diff
@@ -1,12 +1,19 @@
...
+const logger = LoggerHandler.getLogger('Passport');
 
 export default () => {
-    return new Strategy({ cookieName: TOKEN_COOKIE_NAME }, (token, done) =>
+    return new Strategy({ cookieName: TOKEN_COOKIE_NAME }, (token: string, done: PassportDoneFunction) =>
         getUser(token)
             .then(user => done(null, user))
-            .catch(err => done(null, false, err + token))
+            .catch(err => {
+                logger.error(`User authentication failed with token "${token}". Error:`, err);
```
kubama, line 15: Should we reveal token value in the logs?

backend/routes/Auth.types.ts
```diff
@@ -0,0 +1,55 @@
...
+    groups: any;
+    role: string;
+    group_system_roles: GroupSystemRoles;
+    active: boolean;
+    first_login_at: string;
+    last_login_at: string;
+    is_locked: boolean;
+    show_getting_started: boolean;
+}
+/* eslint-enable camelcase */
```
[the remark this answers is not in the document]
qooban: Can you elaborate a bit more? At which level would such converter work?

backend/auth/CookieStrategy.ts
qooban: Right, we shouldn't reveal it in the logs. I removed it.

app/utils/auth.ts
```diff
@@ -8,24 +6,26 @@ import StageUtils from './stageUtils';
 import External from './External';
 import Internal from './Internal';
 import encodeTextToBase64 from './encodeTextToBase64';
+import type { ManagerData } from '../reducers/managerReducer';
+import type { LicenseObject, LicenseData, LicenseStatus } from '../reducers/licenseReducer';
 
 export default class Auth {
```
qooban, line 12: Decided to migrate auth utils (with `license` reducer) on frontend side as it is somehow related to this PR and it was quick to do.

backend/utils.ts
```diff
@@ -12,24 +14,41 @@ export function getResourcePath(path: string, isUserData: boolean) {
...
     Object.keys(obj).forEach(i => {
         if (typeof obj[i] === 'object') {
-            objects = objects.concat(this.getValuesWithPaths(obj[i], key, [...arr, i]));
+            objects = objects.concat(getValuesWithPaths(obj[i], key, [...arr, i]));
         } else if (i === key) {
             objects.push({ [obj[i]]: arr });
         }
     });
     return objects;
 }
```
NorbertAmartus, line 27: ### Optional

While being here, it may be a good time to modify the naming of the function parameters / parameters inside `forEach` method, so that it would be much cleaner what the function is doing

For now it seems like reading it is a complicated thing to do and it's taking some time to understand what is going here 😁 

I think, that modifying parameter names would reduce that issue

NorbertAmartus (approved):
Nice piece of work! 🎉 
I've left one optional comment and as there are no other comments from my side, I'm clicking the legendary `Approve` button 😁

backend/routes/Auth.types.ts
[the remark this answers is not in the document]
kubama: I was thinking about converter that would translate manager response before it is handed to stage backend logic and further to the frontend. I'm not sure if would be the best idea for `/sp` endpoint as it is an explicit proxy to the manager (so maybe it should preserve original property names), but I would consider such solution for other endpoints that are passing manager responses further.

kubama (approved):

backend/utils.ts
qooban: Good point. I checked a bit deeper and it seems like this code is used only by Deployment Wizard Buttons widget (which as we know is not widely used and is "unofficially not supported" ;-) ) So... I decided not to spent more time in this area and leave it as it is.

NorbertAmartus: Seems reasonable 😁 
Thanks for taking your time to check that out!

PR #1899, opened 2022-03-29
RD-4558: Provided additional test scenarios

NorbertAmartus:
## Description

Changes consists of adding additionall test scenarios for the `installation_status` filter values.

Because of using the static option list, for the [InstallationStatusValueInput](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/common/src/filters/inputs/InstallationStatusValueInput.tsx), there appeared a need to adjust automatic tests generation as it was only compatible with `DynamicDropdown`

## Screenshots / Videos

N/A

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2321/)
(tests have been executed successfully, the failure label is coming from the coverage check - which was expected to fail)

## Documentation

N/A


test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -449,7 +449,8 @@ describe('Filters widget', () => {
         function selectRuleAttributeValues(
             ruleRowType: FilterRuleRowType,
             values: RuleValueObject[],
-            withAutocomplete = false
+            withAutocomplete = false,
+            multipleValues = true
```
qooban, line 453: As we've got more than 1 optional parameter maybe the right time to introduce options object pattern here?
So that we would have sth like (it's just a pseudo-code to show the idea):
```
 function selectRuleAttributeValues(
            ruleRowType: FilterRuleRowType,
            values: RuleValueObject[],
            options: Options = { withAutocomplete: false, multipleValues: true })
```

Ref.: https://www.codereadability.com/what-are-javascript-options-objects/

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -593,10 +615,15 @@ describe('Filters widget', () => {
             cy.get('.modal').within(() => verifyRulesForm(testRules));
         }
 
-        interface ExtendedFilterRule extends FilterRule {
+        interface NonDynamicDropdownFilterRules {
+            useNonDynamicDropdown?: boolean;
+        }
```
qooban, line 620: The name of the interface is a bit misleading as it contains `useNonDynamicDropdown`, which can be set to `false` and then the interface is pointing to actually `DynamicDropdownFilterRule` 🙂 

What do you think about something like:
```suggestion
        interface FilterRuleValueInputOptions {
            useStaticDropdown?: boolean;
        }
```
? It can also be named shorter I guess - `ValueInputOptions`.

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -550,16 +553,29 @@ describe('Filters widget', () => {
             });
         }
 
+        function getRulesWithoutNonDynamicDropdownFlags(
```
qooban, line 556: "WithoutNonDynamic" is a double negation, so may be hard to understand.

What about something like:
```suggestion
        function getRulesWithStaticValueInputs(
```
?

test/cypress/integration/widgets/filters_spec.ts
```diff
@@ -550,16 +553,29 @@ describe('Filters widget', () => {
...
         function verifyRequestRules(request: CyHttpMessages.IncomingRequest, testRules: ExtendedFilterRule[]) {
             const requestRules = request.body.filter_rules;
             expect(requestRules).to.have.length(testRules.length);
-            testRules.forEach((_rule, index: number) => {
-                expect(requestRules[index]).to.deep.equal(_.omit(testRules[index], ['newKey', 'newValues']));
+
+            const mappedTestRules = getRulesWithoutNonDynamicDropdownFlags(testRules);
+
+            mappedTestRules.forEach((_rule, index: number) => {
+                expect(requestRules[index]).to.deep.equal(_.omit(mappedTestRules[index], ['newKey', 'newValues']));
```
qooban, line 572: What do you think about doing that filtering in `saveAndVerifyFilter`?

I guess that it would be a bit cleaner to pass only relevant rules `verifyRequestRules` instead of doing filtering here.

test/cypress/integration/widgets/filters_spec.ts
NorbertAmartus: Surely, seems like a great idea!
Thanks for pointing that out 🎉

NorbertAmartus: Applied necessary modifications 🚀

test/cypress/integration/widgets/filters_spec.ts
NorbertAmartus: Sure, seems reasonable 😉

test/cypress/integration/widgets/filters_spec.ts
NorbertAmartus: I was not sure about the naming, so thanks a lot for that comment
I like the suggestion, even the longer one (because it's being specific about the input purpose), so I'm applying it right away 🚀

test/cypress/integration/widgets/filters_spec.ts
NorbertAmartus: I think that the `getRulesWithStaticValueInputs` is a bit unclear, as it's hard do establish which input values are `static` 😁 
What do you think about `getRulesWithoutInputOptions`?

test/cypress/integration/widgets/filters_spec.ts
NorbertAmartus: Necessary modifications has been applied 🎉

test/cypress/integration/widgets/filters_spec.ts
qooban: Actually I did a mistake in my comment as in fact the function seems to do: `getRulesWithoutStaticValueInputs` or `getRulesWithDynamicValueInputs` 🙂 

Two points:

1. Function implementation

Looking at the function implementation it seems like it does not exclude rules with static values as it is just removing `useStaticDropdown` field, but what if that field is already set to `true`? 🤔 Shouldn't we filter out rules having `useStaticDropdown : true` instead of taking rules which does not have it defined? I suspect it'll be more clear if we do so. Probably then `FilterRuleValueInputOptions` interface won't be necessary.

2. Function name

> What do you think about `getRulesWithoutInputOptions`?

The implementation suggests that, but I think the goal of this function and the context where it is used suggests something different. I suspect that the main goal of this function is to get only filter rules which are using HTTP requests to fetch data for value inputs and filter out rules which are using statically defined valid values. Am I right? If so, then maybe `getRulesUsingExternalDataSources` or `getRulesUsingDynamicDropdowns` or `getRulesNotUsingStaticDropdowns` or whatever showing the actual purpose? 🙂

qooban (commented):
Thanks for applying changes. I have shared my thoughts in the last outstanding discussion.

test/cypress/integration/widgets/filters_spec.ts
qooban: Oh, so if that's the intention, then why just not extend L580:
`expect(requestRules[index]).to.deep.equal(_.omit(testRules[index], ['newKey', 'newValues']))`
to omit also `useStaticDropdown` and have:
`expect(requestRules[index]).to.deep.equal(_.omit(testRules[index], ['newKey', 'newValues', 'useStaticDropdown']))`
?

That way you won't need neither `getRulesWithoutNonDynamicDropdownFlags` function nor `FilterRuleValueInputOptions` interface.

qooban (approved):
Looks good 👍🏻

test/cypress/integration/widgets/filters_spec.ts
qooban: All clarified in online meeting, so I'm resolving this discussion 🙂

NorbertAmartus: Seems like a good idea 💪 
Added the necessary adjustments 🎉

PR #1903, opened 2022-03-31
RD-3672 Edit Mode spec adjustments

qooban:
## Description

As we all know Edit Mode spec is quite flaky. The flakiness comes from verification of `POST /console/ua` calls response body.
I think we can live without those checks without losing much, so I removed all those assertions and removed test retries as they were not helping to solve the flakiness.

Ref. RD-3672

## Screenshots / Videos
N/A

## Checklist

## Tests
1. [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2329/) - only Edit Mode spec - PASSED
2. [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2330/) - only Edit Mode spec - PASSED

## Documentation
N/A

### DIFF
--- a/test/cypress/integration/edit_mode_spec.ts
+++ b/test/cypress/integration/edit_mode_spec.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck File not migrated fully to TS
 describe('Edit mode', () => {
     before(() => cy.activate('valid_trial_license').removeCustomWidgets());
 
@@ -12,7 +11,7 @@ describe('Edit mode', () => {
     it('should allow to edit widget settings', () => {
         cy.get('.blueprintsWidget .setting').click({ force: true });
 
-        cy.get('.pollingTime input').type(0);
+        cy.get('.pollingTime input').type('0');
 
         cy.contains('Save').click();
         cy.wait('@updateUserApps');
@@ -66,7 +65,7 @@ describe('Edit mode', () => {
         cy.wait('@updateUserApps');
 
         cy.get('.editModeButton .edit:eq(0)').click();
-        cy.get('.modal input[type=text]').type(2);
+        cy.get('.modal input[type=text]').type('2');
         cy.get('.modal .toggle').click();
 
         cy.contains('Save').click();
@@ -142,7 +141,12 @@ describe('Edit mode', () => {
             }
         }
 
-        function submitInvalidWidget(widgetName, expectedError, expectedStatus, expectedDelete = false) {
+        function submitInvalidWidget(
+            widgetName: string,
+            expectedError: string,
+            expectedStatus: number,
+            expectedDelete = false
+        ) {
             submitWidget(widgetName);
 
             cy.wait('@installWidget').its('response.statusCode').should('equal', expectedStatus);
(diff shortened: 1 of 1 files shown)


test/cypress/integration/edit_mode_spec.ts
```diff
@@ -267,12 +252,13 @@ describe('Edit mode', () => {
             const widgetName = 'Backend';
             const widgetId = `testWidget${widgetName}`;
             const url = 'https://this-page-intentionally-left-blank.org/';
-            const verifyRequest = service =>
+            const verifyRequest = (service: string) =>
                 cy.wait(service).then(({ request, response }) => {
                     expect(request.headers).contain({ 'widget-id': widgetId });
-                    expect(response.statusCode).equals(200);
+                    expect(response!.statusCode).equals(200);
```
kubama, line 258: I think it is always cleaner:
```suggestion
                    expect(response?.statusCode).equals(200);
```

kubama (commented):
As I understand removing the checks does not guarantee that the issue is fixed. Maybe it would be good to just comment out the checks and provide additional comment(s) explaining why the checks are disabled? In case somebody wanted to again introduce missing checks, but also in case it turned out that checks are not the issue but the actual endpoint response. It's also possible that the root cause of the issue is identified/fixed in the future, so it will be possible to re-enable the checks easily.

NorbertAmartus (approved):

qooban:
@kubama 
> As I understand removing the checks does not guarantee that the issue is fixed. Maybe it would be good to just comment out the checks and provide additional comment(s) explaining why the checks are disabled? In case somebody wanted to again introduce missing checks, but also in case it turned out that checks are not the issue but the actual endpoint response. It's also possible that the root cause of the issue is identified/fixed in the future, so it will be possible to re-enable the checks easily.

Hmm... I'm not sure. I have compared videos/screenshots for passing and failed tests I don't see a way how could it be possible that Edit Mode spec will fail in the same places without response body verification. I can revert checks and comment it out, but will that be valuable for us? We have Git history.

kubama:
> Hmm... I'm not sure. I have compared videos/screenshots for passing and failed tests I don't see a way how could it be possible that Edit Mode spec will fail in the same places without response body verification.

My thinking is that it can still fail in subsequent places if it's an issue with actual endpoint response and not just the verification.

> I can revert checks and comment it out, but will that be valuable for us? We have Git history.

Having it commented out with additional comment explaining the reason will keep the awareness of the issue, which is beneficial for the reasons I mentioned in my previous comment.

This is optional comment, so I'm approving regardless of your final decision.

qooban:
> My thinking is that it can still fail in subsequent places if it's an issue with actual endpoint response and not just the verification.

From what I have investigated that's very unlikely, 'cause it seems like the proper request is sent anyway. The problem is - as I suspect - that sometimes it is being sent twice (there are 2 updates instead of 1). But all in all the result in the UI view is OK.

> This is optional comment, so I'm approving regardless of your final decision.

Thanks. I'll keep it as is for now.

PR #1906, opened 2022-04-04
RD-3918 Turn on HttpOnly flag on token cookie

qooban:
## Description

In order to set `httpOnly` flag on token cookie the following changes/refactorings were applied:
* Refactored `auth` object in `managerReducer` to contain also `username` (`managerReducer.username`), `error` (`managerReducer.err`) and `state` (combination of `managerReducer.isLoggingIn` and new `isLoggedIn`).
* Extracted `auth` object to a separate reducer file
* Updated `Auth` spec and removed `sinon` package uses
* RD-2314 Used object instead of multiple optional parameters in `cy.login` and `cy.mockLogin` (RD-2314)
* Improved `AuthRoutes` and `Routes` components to handle Okta login properly (also migrated to TypeScript at once)
* RD-4457 Fixed Login spec by providing more reliable method to check SSO redirection and making sure that login screen is refreshed whenever it is necessary

### `managerReducer` structure
Screenshots taken from Redux DevTools:
### Before
image

### After
image

## Checklist

## Tests
1. [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2363/) (6.4.2022, 12:08 CEST) - all PASSED
2. [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2378/) (7.4.2022 13:52 CEST) - with SAML login fix - Community spec failed (need to fix it) & Snapshots widget spec failed (not related, known issue)
3. [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2383/) (8.4.2022 07:55 CEST) - only Login spec failed (seems like related to RD-4457)

## Documentation
N/A


kubama (dismissed):

app/reducers/managerReducer/authReducer.ts
```diff
@@ -0,0 +1,53 @@
+import type { Reducer } from 'redux';
+import * as types from '../../actions/types';
+import emptyState from './emptyState';
+
+export type AuthState = 'loggedOut' | 'loggingIn' | 'loggedIn';
```
NorbertAmartus, line 5: What do you think about creating object, which would contain all of the available options?
We could extract it's keys/values as the type, which would be used similarly to this one

The benefit that we'd get from that action would be the maintainability and decreased chances for typos
(as we'd be using object values across the system)

Btw, I cannot kudos that enough, how happy I'm about seeing migration from the multiple boolean properties to the one property but with a multiple states!
Great adjustment! 🎉

NorbertAmartus (commented):
Great work! 🎉 
Not only related to the required changes, but also when it comes to the refactors and code adjustments!

Really great piece of work! 🚀 

Left some minor comments regarding imports and login state idea

app/reducers/managerReducer/authReducer.ts
qooban: I thought about that when I was introducing that type as we have types defined the way you suggest - object with consts and then type created from object values.
But I realized that TypeScript is smart enough to handle typo issues without it 🙂 

See:

https://user-images.githubusercontent.com/5202105/162129956-cef4b335-557f-481a-952b-fb2ccc2d9c31.mp4

If you think there are other advantages of having const object let me know 🙂

app/containers/LoginPage.ts
```diff
@@ -1,28 +1,28 @@
-// @ts-nocheck File not migrated fully to TS
-
 import _ from 'lodash';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 
+import type { Dispatch } from 'redux';
 import LoginPage from '../components/LoginPage';
 import { login } from '../actions/managers';
+import { ReduxState } from '../reducers';
```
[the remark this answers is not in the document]
qooban: Good catch 👍🏻  Applied.

I was fighting with myself to migrate `LoginPage` (container and component), but was afraid that it will make this PR too big, so decided to leave it for now.

kubama (approved):

qooban:
I have detected an issue with Okta login, so there's a bit more to do within this PR. I'll let you know once it is ready for re-review.

app/reducers/managerReducer/authReducer.ts
NorbertAmartus: I think that there are 2 advantages:

### Related to refactoring
When using regex for modifying all instances of the specific string, there may be some related to e.g. labels, and because of that it should be double checked, to make sure that nothing "extra" has been modified

When it comes to the object, we can just modify one of the corresponding value, without double checking if something extra has been changed

And when it comes to modifying object property names, it's easier to modify them across files, because we can search it with regex (like `.loggingIn` or `authState.loggingIn`), having almost 100% certainty that we'll modify only the things that we wanted 🎉 

### Related to typos
Typescript is smart enough, but developer may not be 😁 
What I mean by that is that we can try to quickly modify the string to the other value, e.g.: `loggingin` and switch back to the browser, thinking "why this is not working" 😄 

It's a small thing, but kind of related to the DX 😁 
<hr />

Anyway, just wanted to share with you my point of view
The change is not required, as the type safety is enough for our use cases 💪 

But if you'd think that it would bring more benefits to our code, feel free to do so 🥳 
In other way, I'd be more than happy to have a little brain storming about it, on some sort of meeting, to clarify our view on that aspect
It may be a small thing, but I think that discussions about those aspects are leading to a better understanding of eachother and the application architecture 🚀

<hr />

I'm approving this PR, as the change is not required
Thanks again for an amazing job and for all of the refactor that you've done! 🎉

NorbertAmartus (approved):
As stated in [one of the discussions](https://github.com/cloudify-cosmo/cloudify-stage/pull/1906#discussion_r844947683), I'm approving this PR 🚀 

Great one! 🥳

qooban:
Okta issue fixed. I have triggered another full system tests run to verify all is fine. If all good, I'll request re-review.

app/reducers/managerReducer/authReducer.ts
qooban: Thanks for explanation. After reading it I realized that I had the same or very similar point of view before I started using TypeScript 🙂 

Let me comment on some parts of the explanation:

> Related to refactoring

> When it comes to the object, we can just modify one of the corresponding value, without double checking if something extra has been changed

That's true, but in most of the cases you would also change the name of the property in the object, so all the uses of that property needs to be updated as well - so it may be the same number of updated LOCs as when using string values.

> And when it comes to modifying object property names, it's easier to modify them across files, because we can search it with regex (like .loggingIn or authState.loggingIn), having almost 100% certainty that we'll modify only the things that we wanted tada

With string values you can also search using regex (it's just "loggingIn" and it is exactly that string, it cannot be part of a different string). BTW WebStorm has also that nice feature to search only in strings.

> Related to typos

> Typescript is smart enough, but developer may not be grin
> What I mean by that is that we can try to quickly modify the string to the other value, e.g.: loggingin and switch back to the browser, thinking "why this is not working" smile

OK, true, but in that case Webpack (dev server) will throw an error, so I guess it won't be very long way for a developer to fix it.

> It's a small thing, but kind of related to the DX grin

OK, agree, small DX thing 🙂 

<hr />

Summing up - thanks again for the explanation, I appreciate that. 
I'd like to leave the code unchanged, and we'll see in the future how it goes with this approach.

qooban:
Fixed Community spec (the only failing test) and some minor TS config related issues. 
I triggered one more full system tests run, but I don't expect any more changes there.
PR is ready for final review 🙂

app/components/AuthRoutes.tsx
```diff
@@ -51,14 +42,15 @@ export default function AuthRoutes({ isSamlEnabled }) {
     useEffect(() => {
         if (isProductOperational && isManagerDataFetched) {
             dispatch(getUserData())
+                // @ts-ignore Type of tenantsRoles is not relevant here
```
kubama, line 45: Wouldn't it be cleaner to instead set the type to `any` if it's not relevant?

app/reducers/managerReducer/authReducer.ts
NorbertAmartus: No problem, thanks for sharing your point of view!
All points are valid and it's nice to see a different perspective on the subject 🙂

app/components/AuthRoutes.tsx
```diff
@@ -13,36 +11,29 @@ import { useBoolean } from '../utils/hooks';
...
+import { getManagerData, getUserData, logout } from '../actions/managers';
 import Layout from '../containers/layout/Layout';
 import LicensePage from '../containers/LicensePage';
 import MaintenanceMode from '../containers/maintenance/MaintenanceModePageMessage';
 import SplashLoadingScreen from '../utils/SplashLoadingScreen';
 
-export default function AuthRoutes({ isSamlEnabled }) {
+const AuthRoutes: FunctionComponent = () => {
     const [isManagerDataFetched, setManagerDataFetched /* , unsetManagerDataFetched */] = useBoolean();
     const [isUserDataFetched, setUserDataFetched /* , unsetUserDataFetched */] = useBoolean();
```
NorbertAmartus, line 22: I think that there are some development leftovers 😁 

```suggestion
    const [isManagerDataFetched, setManagerDataFetched] = useBoolean();
    const [isUserDataFetched, setUserDataFetched] = useBoolean();
```

qooban: Probably 🙂 Removed.

app/components/AuthRoutes.tsx
qooban: Yeah, make sense. Changed it.

kubama (approved):

NorbertAmartus (approved):

PR #1907, opened 2022-04-04
RD-3917: Added support for using terraform module from git files

NorbertAmartus:
## Description

Changes consists of modifying backend `terraform/resources` endpoint, to support scanning git files for the terraform modules

Additionally, one of the bugs was fixed on the frontend side. 
Bug was related to the errors occurring on the `TerraformModal.tsx`.
After showing first error message, related to the template URL, it wouldn't be changed (e.g. when we'd provide incorrect URL, and after that change it to the URL that is requiring an authentication, it would still show the message about incorrect URL not the autorization issue) or even cleared, when the proper URL would be provided.

## Screenshots / Videos

N/A

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2460/)
(All test scenarios has passed successfully, but the test coverage step has failed - as expected)

## Documentation

Documentation will be updated with the help of [this PR](https://github.com/cloudify-cosmo/cloudify-stage/pull/1907) 
The synchronization will be performed as a part of this PR.


NorbertAmartus:
Tests are passing locally, but I've got some trouble when it comes to running them on Jenkins
To not waste any time, I've decided to open this PR and resolve the Jenkins issues in the meantime

app/translations/en.json
```diff
@@ -879,8 +879,8 @@
                 "blueprintName": "Blueprint name",
                 "terraformVersion": "Terraform version",
                 "blueprintInformation": "Blueprint main information",
-                "template": "URL to a zip archive that contains the Terraform module",
-                "resourceLocation": "Terraform folder in the archive",
+                "template": "Terraform module source - URL to a zip archive or a GIT repository",
```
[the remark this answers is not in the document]
NorbertAmartus: That label has been sent to me by Michael, but I'll double check with him, if the `GIT` should be consisting only of capital letters

When I'll get a response from him, I'll come back to this comment 🚀 

Nice catch, btw 💪

backend/test/routes/Terraform.test.ts
```diff
@@ -50,11 +52,18 @@ describe('/terraform/resources endpoint', () => {
             .reply(200, readFileSync(resolve(__dirname, 'fixtures/terraform/template.zip'), null));
 
         const response = await request(app)
-            .post('/console/terraform/resources?zipUrl=http://test/test.zip')
+            .post(`${endpointUrl}?templateUrl=http://test/test.zip`)
```
qooban, line 55: Optional:

Maybe a small improvement here 🙂 
```suggestion
            .post(endpointUrl)
            .query({ templateUrl: 'http://test/test.zip' })
```

Similar can be applied in L65.

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -369,6 +369,9 @@ export default function TerraformModal({
                 setResourceLocation(
                     find(loadedTemplateModules, module => module.indexOf('terraform') >= 0 || module.indexOf('tf') >= 0)
                 );
+
+                const { template, ...modalErrors } = errors;
+                setErrors(modalErrors);
```
qooban, line 374: Can you explain why that change is necessary?

qooban (commented):
Nice refactoring in Terrafrom backend route. Added some comments.

Quickly tested UI part and got one question - would it be possible to distinguish between repo authentication and repo existence errors?

I have tried not existing repo URL and received the following:
image

app/translations/en.json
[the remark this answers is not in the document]
NorbertAmartus: He've said that it should be indeed `Git` 🎉

widgets/common/src/terraformModal/TerraformModal.tsx
NorbertAmartus: Sure, it's clearing the template errors, in the following case:
1. User has entered invalid template URL (and he's getting an error message)
2. User has corrected template URL (and without the changes, the previous error message would still be visible)

test/cypress/integration/widgets/blueprints_spec.ts
```diff
@@ -440,7 +439,7 @@ describe('Blueprints widget', () => {
         });
     });
 
-    describe('should open upload from Terraform module modal and', () => {
+    describe.only('should open upload from Terraform module modal and', () => {
```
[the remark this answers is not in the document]
NorbertAmartus: It surely is 😁 
Thanks for catching it! 🎉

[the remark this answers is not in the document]
NorbertAmartus: That line has been modified with a laser precision, accordingly to your suggestion 😁

backend/test/routes/Terraform.test.ts
NorbertAmartus: Sure, seems neat!
Thanks for pointing that out! 🚀

NorbertAmartus:
@qooban 

>Quickly tested UI part and got one question - would it be possible to distinguish between repo authentication and repo existence errors?

Currently as we can see, it's possible but not for all cases, e.g. when we'd like to access the following git file:
``` bash
https://github.com/cloudify-cosmo/docs.getcloudify.org/2.git
```

We'd get an error saying that `The URL is not accessible`, but when accessing some routes for the private users/organizations it seems like the git is still requiring some sort of authentication, to ensures the access to those paths (in those places, node git is returning error message related to the authentication - just double checked that for the URL that you've shown on the screenshot)

We can even see the difference between those links with a little experiment 😁 
Open with the browser:
1. https://github.com/cloudify-cosmo/docs.getcloudify.org/2.git
2. https://github.com/cloudify-community/tddf-source.git

Both of them are invalid, but the information about that is being displayed in a different way 🤯 
(for the first, the TerraformModal is displaying a proper `The URL is not accessible` error message 😉)

So I think that one of the github mechanisms is blocking us from being able to display to the user the most accurate error message every time 😢  

I just did a little experiment, when it comes to cloning those git files, which were attached above
For https://github.com/cloudify-cosmo/docs.getcloudify.org/2.git the console displayed to me simple `Not found` message, but for the `https://github.com/cloudify-community/tddf-source.git` it has displayed an authentication popup 🤯 

Which confirms the hypothesis above 😁

widgets/common/src/terraformModal/TerraformModal.tsx
qooban: Got it, thx for explanation.

qooban:
> So I think that one of the github mechanisms is blocking us from being able to display to the user the most accurate error message every time 

Got it, thx for testing it. Seems not worth to change anything then in the error message 👍🏻

NorbertAmartus:
I've replaced the library used for cloning the git repo, from `node-git` to `simple-git`, so that it wouldn't require update on jenkins pipelines (as we've got some trouble there with a `node-git`) and to be more secure (as `node-git` has some security vulnerabilities)

backend/routes/Terraform.ts
```diff
@@ -75,32 +77,27 @@ router.post('/resources', async (req: ResourcesRequest, res) => {
...
-                    }
-                }
-            });
-            // @ts-ignore-next-line nodegit library ensures that the occured error would be in a shape of the GitError type
+            const gitUrl = getGitUrl();
+            await simpleGit().clone(gitUrl, repositoryPath, [disableGitAuthenticationPromptOption]);
+            // @ts-ignore-next-line simple-git library ensures that the occured error would be in a shape of the GitError type
         } catch (error: GitError) {
-            const isAuthenticationIssue = error.message.includes('authentication');
+            const isAuthenticationIssue = error.message.includes('Authentication failed');
```
kubama, line 98: What if the error caught here is not an instance of `GitError` (for exmaple it's thrown from `getGitUrl`)?

NorbertAmartus: Then it would be an error comming from the string operations failure, which would have a corresponding error message (If I'm not mistaken)
But that scenario is highly unlikely to happen, while we are keeping the mechanism of sending credentials the same in both frontend and backend

But as there exists a possibility, that the error wouldn't be of the `GitError` type I can adjust a code to that 😉 
WDYT?

qooban (approved):
Checked UI quickly, works good 👍🏻 

Thanks for addressing my previous comments, I don't have more.

backend/routes/Terraform.ts
kubama: Nothing specific to `GitError` is being used here - the `message` property is part of generic `Error` interface. So I think it makes more sense to use just `Error` without the need for `ts-ignore`. The same applies to L115.

NorbertAmartus: Okay, seems fine by me 😉 
I'll implement it right away! 🚀

kubama (approved):

NorbertAmartus:
As there are no other comments, I've synchronized the [documentation](https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2161)

kubama (approved):

PR #1908, opened 2022-04-05
RD 273 Adding upload RPM build artifact to s3 bucket

bartoszkosciug:
RD-273
https://cloudifysource.atlassian.net/browse/RD-273

## Description
The change that was need is to add the update artifacts to s3 bucket functionality to the pipeline workflow.

## Screenshots / Videos
N/A

## Checklist

## Tests
N/A

## Documentation
N/A


jenkins/Jenkinsfile
```diff
@@ -94,7 +99,7 @@ pipeline {
               sh """
                 cd ~/rpmbuild
                 mkdir -p BUILD && cp -rf ${WORKSPACE}/${STAGE_DIR}/. BUILD && cd BUILD
-                curl -fO "${CFY_MANAGER_URL}/${BRANCH}/packaging/version_info"
```
yuvlar, line 97: why change this one? the version should be fetched according to the running branch.
The thing is though, that if there's no such branch in cloudify-manager repo, it'll fail - but it should enter this stage only if it's *-build | master, so it shouldn't be a problem

jenkins/jenkins2
```diff
@@ -0,0 +1,133 @@
+@Library('pipeline-shared-library') _
```
yuvlar, line 1: I believe this file was only for testing?

yuvlar (changes requested):
Looks good, left some comments for changes

jenkins/Jenkinsfile
bartoszkosciug: back to the original way of fetching [fdd2a59]

jenkins/jenkins2
bartoszkosciug: removed from the pr

jenkins/build-pod.yaml
```diff
@@ -6,6 +6,15 @@ spec:
       persistentVolumeClaim:
         claimName: pipelines-efs
   containers:
+  - name: awscli
```
qooban, line 9: I assume adding new container won't affect overall job performance?
(as it will be used only specific branches, we don't want to have slow down on development branches)

jenkins/Jenkinsfile
```diff
@@ -31,11 +31,16 @@ pipeline {
 
   environment {
     BRANCH = "${env.BRANCH_NAME}"
+    VERSION = getVersion(PROJECT,"${env.BRANCH_NAME}").trim()
+    PRERELEASE = getPreRelease(PROJECT,"${env.BRANCH_NAME}").trim()
```
qooban, line 35: In "Build RPM" stage there is different method used to get version and release. 

I guess it works like that:
1. Fetch `version_info` - `curl -fO "${CFY_MANAGER_URL}/${BRANCH}/packaging/version_info"`
2. Source `version_info` - line with `cd ~/rpmbuild/BUILD && source ./version_info`
3. Use env vars created during sourcing - `CLOUDIFY_VERSION`, `CLOUDIFY_PACKAGE_RELEASE`

Shouldn't we align it and use only one method in both cases?

qooban (commented):
Added some comments.

Regarding checklist and:

> the change is not UI related

we often add non-UI related changes to this repo and we always fill the checklist, so I think it would be good to have it still. Labels, documentation updates and tests are not only related to frontend code updates. See other PRs.

jenkins/build-pod.yaml
yuvlar: Shouldn't interfere, like you said, it'll be used only once and won't affect the overall job performance.

jenkins/Jenkinsfile
yuvlar: @qooban I agree with you that there's no benefit in using two different methods to get the same vars.
But in order to have those env vars available for RPM uploading, it's best to pull them in the beginning of the pipeline, and update the "Build RPM" stage to use the $VERSION & $PRERELEASE that we created in the beginning.
@bartoszkosciug Please change it accordingly

jenkins/build-pod.yaml
qooban: Good 👍🏻

jenkins/Jenkinsfile
qooban: Yeah, make sense. Thanks for making the changes.

qooban (approved):
From my perspective looks good. Let's wait for @yuvlar confirmation.

jenkins/Jenkinsfile
```diff
@@ -132,6 +137,13 @@ pipeline {
               }
             }
           }
+          post {
```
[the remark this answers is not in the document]
bartoszkosciug: in examples the path will be cloudify-release-eu/cloudfy/6.4.0/.dev1-build/cloudify-stage/13 
and cloudify-release-eu/cloudfy/6.4.0/.dev1-build/cloudify-stage/latest

yuvlar (approved):

PR #1924, opened 2022-04-13
RD-4720 Refactor input component factory functions

kubama:
## Description
Refactored component factory functions to regular components, extracted standalone input field components

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/2539/

 

## Documentation
N/A


jawamartus:
Do you plan to move new created components to the common package?

kubama:
> Do you plan to move created components to the common package?

Yes, I plan to do this as part of https://cloudifysource.atlassian.net/browse/RD-4691.

widgets/common/src/inputs/fields/BlueprintIdInputField.tsx
```diff
@@ -0,0 +1,26 @@
+import DynamicDropdown from '../../components/DynamicDropdown';
+import { DynamicDropdownInputFieldProps } from './types';
```
NorbertAmartus, line 2:
```suggestion
import type { DynamicDropdownInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/CapabilityValueInputField.tsx
```diff
@@ -0,0 +1,33 @@
+import DynamicDropdown from '../../components/DynamicDropdown';
+import { DynamicDropdownInputFieldProps } from './types';
```
NorbertAmartus, line 2:
```suggestion
import type { DynamicDropdownInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/DeploymentIdInputField.tsx
```diff
@@ -0,0 +1,29 @@
+import DynamicDropdown from '../../components/DynamicDropdown';
+import { DynamicDropdownInputFieldProps } from './types';
```
NorbertAmartus, line 2:
```suggestion
import type { DynamicDropdownInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/GenericInputField.tsx
```diff
@@ -0,0 +1,13 @@
+import { PositionedRevertToDefaultIcon } from './RevertToDefaultIcon';
+import { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
```
NorbertAmartus, line 2:
```suggestion
import type { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/types.ts
```diff
@@ -0,0 +1,20 @@
+import { InputProps } from 'semantic-ui-react';
+import { DynamicDropdownProps } from '../../components/DynamicDropdown';
```
NorbertAmartus, line 2:
```suggestion
import type { InputProps } from 'semantic-ui-react';
import type { DynamicDropdownProps } from '../../components/DynamicDropdown';
```

widgets/common/src/inputs/fields/types.ts
```diff
@@ -0,0 +1,20 @@
...
+import { OnChange } from '../types';
```
NorbertAmartus, line 3:
```suggestion
import type { InputProps } from 'semantic-ui-react';
import type { DynamicDropdownProps } from '../../components/DynamicDropdown';
import type { OnChange } from '../types';
```

widgets/common/src/inputs/fields/ValueListInputField.tsx
```diff
@@ -0,0 +1,33 @@
+import { PositionedRevertToDefaultIcon } from './RevertToDefaultIcon';
+import { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
```
NorbertAmartus, line 2:
```suggestion
import type { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/TextareaInputField.tsx
```diff
@@ -0,0 +1,19 @@
+import { DEFAULT_TEXTAREA_ROWS } from '../consts';
+import { PositionedRevertToDefaultIcon } from './RevertToDefaultIcon';
+import { RevertableInputFieldProps } from './types';
```
NorbertAmartus, line 3:
```suggestion
import type { RevertableInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/StringInputField.tsx
```diff
@@ -0,0 +1,23 @@
+import RevertToDefaultIcon, { PositionedRevertToDefaultIcon } from './RevertToDefaultIcon';
+import { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
```
NorbertAmartus, line 2:
```suggestion
import type { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/SecretKeyInputField.tsx
```diff
@@ -0,0 +1,27 @@
+import DynamicDropdown from '../../components/DynamicDropdown';
+import { DynamicDropdownInputFieldProps } from './types';
```
NorbertAmartus, line 2:
```suggestion
import type { DynamicDropdownInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/RevertToDefaultIcon.tsx
```diff
@@ -0,0 +1,39 @@
+import GenericRevertToDefaultIcon from '../../components/RevertToDefaultIcon';
+import { STRING_VALUE_SURROUND_CHAR } from '../consts';
+import getInputFieldInitialValue from '../utils/getInputFieldInitialValue';
+import { RevertableInputFieldProps } from './types';
```
NorbertAmartus, line 4:
```suggestion
import type { RevertableInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/NumberInputField.tsx
```diff
@@ -0,0 +1,39 @@
+import RevertToDefaultIcon from './RevertToDefaultIcon';
+import { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
```
NorbertAmartus, line 2:
```suggestion
import type { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
```

widgets/common/src/inputs/fields/BooleanInputField.tsx
```diff
@@ -0,0 +1,25 @@
+import RevertToDefaultIcon from './RevertToDefaultIcon';
+import { RevertableInputFieldProps } from './types';
+
+export default function BooleanInputField(props: RevertableInputFieldProps) {
+    const { Form } = Stage.Basic;
+    const { name, value, onChange } = props;
```
NorbertAmartus, line 6: We could destructurize props as it has been done in other, extracted input fields 😉 

```suggestion
export default function BooleanInputField({
    name,
    value,
    onChange
}: RevertableInputFieldProps) {
    const { Form } = Stage.Basic;
```

widgets/common/src/inputs/fields/NumberInputField.tsx
```diff
@@ -0,0 +1,39 @@
...
+            greaterThan: number;
+            greaterOrEqual: number;
+            lessThan: number;
+            lessOrEqual: number;
+        };
+    };
+
+export default function NumberInputField(props: NumberInputField) {
+    const { Form } = Stage.Basic;
+    const { name, value, onChange, type, error, constraints } = props;
```
NorbertAmartus, line 18: We can destructurize props as it has been done with the other, extracted input fields 😉 

```suggestion
export default function NumberInputField({
    name,
    value,
    onChange,
    type,
    error,
    constraints
}: NumberInputField) {
    const { Form } = Stage.Basic;
```

jawamartus:
> Yes, I plan to do this as part of https://cloudifysource.atlassian.net/browse/RD-4691.

DynamicDropdown component is strictly related to stage. For example it requires toolbox. (I assume that you want to move it to components repository as well.) AFAIK we don't have toolbox object inside of composer. Also it depends on `useEventListener` and `useUpdateEffect`. AFAIK we don't have those hooks inside of composer. `useEventListener` reguires toolbox which is abstraction created only inside of stage and shouldn't be moved outside IMHO. It also has other dependencies specific for stage repository.

What is more moving those components to components repository make modifying them for the purpose of stage more time consuming.

How do you plan solve the problems described above? Do you have some architectural solution for this problem in your mind?

widgets/common/src/inputs/fields/types.ts
kubama: Done

widgets/common/src/inputs/fields/BooleanInputField.tsx
kubama: See L22 - `props` object is used there directly.

widgets/common/src/inputs/fields/NumberInputField.tsx
kubama: See L35 - `props` object is used there directly.

kubama:
> DynamicDropdown component is strictly related to stage. For example it requires toolbox. (I assume that you want to move it to components repository as well.) AFAIK we don't have toolbox object inside of composer. Also it depends on `useEventListener` and `useUpdateEffect`. AFAIK we don't have those hooks inside of composer. `useEventListener` reguires toolbox which is abstraction created only inside of stage and shouldn't be moved outside IMHO. It also has other dependencies specific for stage repository.

I'm not sure if `DynamicDropdown` based components are to be reused at all - probably they will stay. As you pointed out their implementation heavily depends on stage. Additionally, whenever manager resource id needs to be entered in composer it may reference resource that does not exist at the time it's being referenced, so it makes applicability of such components quite limited.

> What is more moving those components to components repository make modifying them for the purpose of stage more time consuming.

This comment is true for any other piece of code that is being shared as part of `cloudify-ui-common`, it's nothing specific to this particular case.

> How do you plan solve the problems described above? Do you have some architectural solution for this problem in your mind?

I don't have any solution yet as it is not yet known which components are to be reused and what would be their final props interface. This PR only sets ground to future work. The final solution will emerge once work on [RD-4691](https://cloudifysource.atlassian.net/browse/RD-4691) progresses.

widgets/common/src/inputs/fields/BooleanInputField.tsx
NorbertAmartus: Oh yea, you're right!
Thanks for the clarification :)

NorbertAmartus: Oh yea, you're right!
Thanks for the clarification :)

widgets/common/src/inputs/fields/NumberInputField.tsx
NorbertAmartus: Oh yea, you're right!
Thanks for the clarification :)

NorbertAmartus (approved):
Great chunk of work!

jawamartus:
> I'm not sure if DynamicDropdown based components are to be reused at all - probably they will stay. As you pointed out their implementation heavily depends on stage.

Is there any component you are sure that you would move to the components repository?

kubama:
> Is there any component (refactored or created in the PR) you are sure that you would move to the components repository?

No.

jawamartus (commented):
> No.

Then I don't see the profit from adding lines 200 of boilerplate code to our codebase. Also If we are not certain that we would move (related to the changes) component to components repository I think it's better to keep *InputField components more private and limit its usage to getInputField function (or InputFIeld component).

However I see two things that may be refactored here, but those things are not in the scope of mentioned task.

1. [getInputFields.tsx](https://github.com/cloudify-cosmo/cloudify-stage/pull/1924/files#diff-ea9966e5995c385e844b10996dceea40ee9fcb813384220dd397201d6046b236) and getInputField have a switch where we are checking the input.type. It may be refactored. Do we really need to check the input.type twice and have two switches? What is the reason for the second switch in getInputField if we checked input.type previously inside of getInputFields?

2. We are using <div style={{ position: 'relative' }}> wrapper in many input fields. Can we unify the usage of this wrapper across all input fields?

kubama:
> Then I don't see the profit from adding 200 lines of boilerplate code to our codebase.

Where do you see boilerplate code here?

> Also If we are not certain that we would move (related to the changes) component to components repository I think it's better to keep *InputField components more private and limit its usage to getInputField function (or InputFIeld component).

This is the situation right now.

> 1. [getInputFields.tsx](https://github.com/cloudify-cosmo/cloudify-stage/pull/1924/files#diff-ea9966e5995c385e844b10996dceea40ee9fcb813384220dd397201d6046b236) and getInputField have a switch where we are checking the input.type. It may be refactored. Do we really need to check the input.type twice and have two switches? What is the reason for the second switch in getInputField if we checked input.type previously inside of getInputFields?
> 2. We are using <div style={{ position: 'relative' }}> wrapper in many input fields. Can we unify the usage of this wrapper across all input fields?

I'll try to address the comments above.

jawamartus:
> This is the situation right now.

Remember about YAGNI principle.

kubama:
> Remember about YAGNI principle.

Possible future reuse is not the only goal of this ticket. In fact it's a continuation of widgets/common refactoring I was working on in this and previous sprint. IMO the code quality after this refactoring is much better, so even if no reuse happened in the future there are still noticeable benefits.

widgets/common/src/inputs/InputField.tsx
```diff
@@ -0,0 +1,94 @@
...
+            return <DeploymentIdInputField {...commonProps} />;
+        case 'blueprint_id':
+            return <BlueprintIdInputField {...commonProps} />;
+        case 'capability_value':
+            return <CapabilityValueInputField {...commonProps} constraints={constraints} />;
+        case 'secret_key':
+            return <SecretKeyInputField {...commonProps} />;
+        case 'string':
+        case 'regex':
+            return <StringInputField {...commonProps} />;
```
jawamartus, line 90: Why do we pass toolbox here?

widgets/common/src/inputs/InputField.tsx
```diff
@@ -0,0 +1,94 @@
...
+                <NumberInputField
+                    {...commonProps}
+                    constraints={{ inRange, greaterThan, greaterOrEqual, lessThan, lessOrEqual }}
+                    type={type}
+                />
+            );
+        }
+        case 'textarea': {
+            const rows = input?.display?.rows ?? DEFAULT_TEXTAREA_ROWS;
+            return <TextareaInputField rows={rows} {...commonProps} />;
```
jawamartus, line 78: Why do we pass toolbox here as a prop? Why do we pass toolbox to other components which don't need it?

widgets/common/src/inputs/InputField.tsx
kubama: For sake of simplicity there is just one `commonProps` object

widgets/common/src/inputs/InputField.tsx
kubama: For sake of simplicity there is just one `commonProps` object

kubama:
> 1. [getInputFields.tsx](https://github.com/cloudify-cosmo/cloudify-stage/pull/1924/files#diff-ea9966e5995c385e844b10996dceea40ee9fcb813384220dd397201d6046b236) and getInputField have a switch where we are checking the input.type. It may be refactored. Do we really need to check the input.type twice and have two switches? What is the reason for the second switch in getInputField if we checked input.type previously inside of getInputFields?

You are referring to a removed file. It was transformed into `InputFields` component, which no longer contains `switch` statement.

kubama:
> 2\. We are using <div style={{ position: 'relative' }}> wrapper in many input fields. Can we unify the usage of this wrapper across all input fields?

Looks like this wrapping div was not necessary at all - removed.

widgets/common/src/inputs/InputField.tsx
jawamartus: Because the PR is not related to the fix nor feature, I would consider additional changes to make our codebase and product better.

1. We may take benefit of new components and use React.memo on them to limit rerenders and increase the performance of our app.
2. We should always pass only necessary props. It's a good practice in react related to performance optimisation and code readability. There are at least 4 risks in my head related to passing more props than are necessary:
    a) when the child component would use the spread operator to pass additional props as html attributes, for example toolbox would be generated and visible in DOM as something like toolbox="[object Object]" and some console warning/errors may be generated (like: HTML A element doesn't have attribute `toolbox`).
    b) In some cases it may generate hard to catch typescript error (we are passing some prop, which we shouldn't pass.
    c) It would cause additional rerenders when the child component would implement React shouldComponentUpdate lifecycle method or use React.memo in the case of 
    d) it's harder to rewrite the code like this in the future. Currently it looks like every child component require all props from commonProps. It's more effortful to catch that only one type of subcomponent needs toolbox - DynamicDropbox.
Some people even think about overuse of spread operator as an antipattern. I would recommend this article about that: https://vhudyma-blog.eu/react-antipatterns-props-spreading/#Why-It-Should-Better-Be-Avoided
3. We may pass toolbox as an Context instead of passing it down from the widget to every component. But it's a big change to do. Can be considered as a separated PR.

widgets/common/src/inputs/InputFields.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <Help
+            description={description}
+            type={type}
+            constraints={constraints}
+            defaultValue={defaultValue}
+            dataType={dataType}
+        />
+    );
+    const required = _.isUndefined(defaultValue);
+    const booleanType = type === 'boolean';
```
jawamartus, line 43: We are checking that condition inside of InputField. I would consider moving FormField to the every *InputField component. 

Worth to notice that it's not a break of DRY rule. It's more SRP principle. The Form.Field may change from different reasons for different field. Let say that we want to make DeploymentIdInputField always required. IT would generate another condition here which would be checking also for string field.

We are checking the type in the switch inside of InputField component. And it's enough in this situation.

After above change we can get rid of this FormField private component and move:

```
     const required = _.isUndefined(defaultValue);
```
To the InputField component or somewhere else (*InputField components ?).

jawamartus: It would help to get a rid of additional level in the component tree and increase performance.

widgets/common/src/inputs/fields/NumberInputField.tsx
```diff
@@ -0,0 +1,39 @@
...
+        constraints: {
+            inRange: [number, number];
+            greaterThan: number;
+            greaterOrEqual: number;
+            lessThan: number;
+            lessOrEqual: number;
+        };
+    };
+
+export default function NumberInputField(props: NumberInputField) {
```
jawamartus, line 16: I cannot see the reason for additional component here. Do we plan to resuse it?

If yes it shouldn't be aware of constraints shape because.

IMHO the previous version without additional component and boilerplate code was OK and with better performance.
It was like:
```

        case 'float': {
            // Limit numerical values if at least one of range limitation constraints is set
            const inRange = getConstraintValue('in_range');
            const greaterThan = getConstraintValue('greater_than');
            const greaterOrEqual = getConstraintValue('greater_or_equal');
            const lessThan = getConstraintValue('less_than');
            const lessOrEqual = getConstraintValue('less_or_equal');

            const min = _.max([inRange ? inRange[0] : null, greaterThan, greaterOrEqual]);
            const max = _.min([inRange ? inRange[1] : null, lessThan, lessOrEqual]);

            return (
                <Form.Input
                    name={name}
                    value={value}
                    fluid
                    error={!!error}
                    type="number"
                    step={type === 'integer' ? 1 : 'any'}
                    min={min}
                    max={max}
                    icon={getRevertToDefaultIcon(name, value, defaultValue, onChange)}
                    onChange={onChange}
                />
            );
```

widgets/common/src/inputs/InputField.tsx
kubama: > We may take benefit of new components and use React.memo on them to limit rerenders and increase the performance of our app.

I added `React.memo` for `InputField`. Individual input field components are ATM used only there, so there would be no point for doing it for each of them.

> We should always pass only necessary props.

I created separate props object for `DynamicDropdown` based input fields.

> We may pass toolbox as an Context instead of passing it down from the widget to every component. But it's a big change to do. Can be considered as a separated PR.

This is a good idea, but as you pointed out not in scope of this PR.

widgets/common/src/inputs/InputField.tsx
kubama: This is now addressed, so I'm resolving this conversation.

widgets/common/src/inputs/fields/StringInputField.tsx
```diff
@@ -0,0 +1,23 @@
...
+            <Form.TextArea name={name} value={value} onChange={onChange} />
+            <PositionedRevertToDefaultIcon {...props} />
+        </>
+    ) : (
+        <Form.Input
+            name={name}
+            value={value}
+            fluid
+            error={error}
+            icon={<RevertToDefaultIcon {...props} />}
```
jawamartus, line 19: IIUC now we are passing toolbox prop to the RevertToDefaultIcon without a reason. It may cause hard to catch errors.

widgets/common/src/inputs/fields/StringInputField.tsx
```diff
@@ -0,0 +1,23 @@
...
+import type { ErrorAwareInputFieldProps, RevertableInputFieldProps } from './types';
+
+export default function StringInputField(props: ErrorAwareInputFieldProps & RevertableInputFieldProps) {
+    const { Form } = Stage.Basic;
+    const { name, value, onChange, error } = props;
+
+    return _.includes(value, '\n') ? (
+        <>
+            <Form.TextArea name={name} value={value} onChange={onChange} />
+            <PositionedRevertToDefaultIcon {...props} />
```
jawamartus, line 11: The same. We are passing toolbox here and other not necessary props.

widgets/common/src/inputs/fields/StringInputField.tsx
kubama: This comment is no longer valid

widgets/common/src/inputs/fields/StringInputField.tsx
kubama: This comment is no longer valid

widgets/common/src/inputs/InputField.tsx
```diff
@@ -0,0 +1,98 @@
...
+        }
+        case 'deployment_id':
+            return <DeploymentIdInputField {...commonDynamicDropdownFieldProps} />;
+        case 'blueprint_id':
+            return <BlueprintIdInputField {...commonDynamicDropdownFieldProps} />;
+        case 'capability_value':
+            return <CapabilityValueInputField {...commonDynamicDropdownFieldProps} />;
```
jawamartus, line 89: It's much harder to read the code with spread operator. IMHO

The author of the article I linked share the same opinion. Also it creates two additional variables which are not necessary.

jawamartus (commented):
I don't have more objections I aforementioned.

I really like getting rid of <div style={{position: ...}}> wrapper. Well done!

It still needs some changes to make our codebase better.

After last change the system tests should be run.

widgets/deploymentWizardButtons/src/steps/InputsStep.tsx
```diff
@@ -211,13 +212,13 @@ class InputsStepContent extends React.Component {
                                                     <InputStatus defaultValue={inputs[inputName].default} />
                                                 </Table.Cell>
                                                 <Table.Cell>
-                                                    {InputsUtils.getInputField(
-                                                        { name: inputName, ...inputs[inputName] },
-                                                        stepData[inputName],
-                                                        this.handleInputChange.bind(this),
```
jawamartus, line 217: Getting rid of bind may be risky. It's hard imagine for me how this would affect the behaviour of this part of app. You may consider checking this manually.

widgets/common/src/inputs/index.ts
```diff
@@ -1,15 +1,19 @@
 import Header from './InputsHeader';
 import HelpDescription from './InputsHelpDescription';
 import HelpIcon from './InputsHelpIcon';
+import InputField from './InputField';
 import Utils from './utils';
 import DataTypesButton from './DataTypesButton';
 import YamlFileButton from './YamlFileButton';
+import Help from './InputHelp';
```
jawamartus, line 8: For me it's little confusing that we rename the `InputHelp` component `Help`. It's harder to find the proper component without good IDE.

widgets/common/src/inputs/InputField.tsx
kubama: The approach you are pushing here is somehow in opposition to what we were accepting as a team so far. Not a long time ago we even disabled eslint rule that disallowed spread operator usage. We had a thorough discussion before proceeding this direction. I encourage you to start a new discussion as part of our weekly meetings if you think we should change our approach. For now, as it is in line with established team line, I would leave as it is. We can create a ticket to refactor such things once we agreed to change our approach.

widgets/common/src/inputs/fields/NumberInputField.tsx
kubama: > I cannot see the reason for additional component here. Do we plan to resuse it?

Possible reuse is one of the PR's goals.

> If yes it shouldn't be aware of constraints shape.

This is what's been done. Constraints shape that this component accepts is abstracted from the general inputs interface so it could be possibly reused in the context of node properties.

widgets/deploymentWizardButtons/src/steps/InputsStep.tsx
kubama: There is no risk - please see change in L151

widgets/common/src/inputs/fields/NumberInputField.tsx
jawamartus: This component shouldn't be aware of part of endpoint response `constraints`. From the same reason why React component shouldn't be aware of SQL queries. What if the shape of the response would change? It shouldn't depend on response shape.

If you want to keep it this way or consider reusing this please make sure that there is additional abstract layer between the endpoint response way and the constraints shape used by this component.

widgets/common/src/inputs/index.ts
kubama: The result of such naming is that it is going to be referenced as `Stage.Common.Inputs.Help`, although the component is still named `InputHelp`. We do this kind of stripping everywhere in `Stage.Common`.

widgets/common/src/inputs/fields/NumberInputField.tsx
kubama: As I commented previously, there is already such abstraction.

widgets/common/src/inputs/index.ts
jawamartus: Why we then import with different convention aboce?
```
import `InputField from './InputField';`
```

kubama: The term 'input' is used in two different meanings: it can either mean blueprint input, or HTML input. In case of `InputField` it is the second case. It is referenced like this: `Stage.Common.Input.InputField` where `Stage.Common.Input` refers to blueprint input while `InputField` refers to HTML input.
So we have kind of naming clash in this area.

widgets/common/src/inputs/InputFields.tsx
kubama: Even if a given solution follows one principle it doesn't automatically mean that it can't break another principle. In this case it would definitely break DRY, because certain piece of code would be repeated across all the components (with slight modification in one of them).
At the same time I don't agree it would be in line with SR principle. You just made an assumption that all `*Field` components should have responsibility of notifying of erroneous value and displaying a label. I don't agree with this assumption. At the moment these components represent just bare input fields, and those additional functionalities can be considered as decorators. In fact, a proper naming scheme for these components would be `*Input` but I decided to use `Field` suffix because of the naming clash that I mentioned in another comment.

> The Form.Field may change

If it changes then it will be the right moment to perform further refactoring :slightly_smiling_face:.

jawamartus: > (...) it would definitely break DRY, because certain piece of code would be repeated across all the components 
We have different understanding of DRY principle in our team.

Let me answer with the quote of the blog post I read some time ago.

> Dave, full of himself, claims: “You little pig! This code is not DRY!”. And you, aware of what the DRY principle is really about, answer: “But the business logic, the knowledge, is still not repeated!”."

The DRY is not about the code repetitions. It's more about the knowledge and the business logic.

The blog posts about it:
Quoted: https://thevaluable.dev/dry-principle-cost-benefit-example/
https://almostengineer.medium.com/dry-principle-of-software-development-common-mistake-15510cfea53e
Wikipedia also mention that definition: "Every piece of knowledge must have a single, unambiguous, authoritative representation within a system" (it's not every piece of code but knowledge)

Now the code is even less DRY, because we are checking type == 'boolean' twice.

kubama: OK, it would be code repetition then and not 'not DRY'. I don't think it changes anything though.

> Now the code is even less DRY, because we are checking type == 'boolean' twice.

I see nothing wrong in checking the same condition in another place at different level of component tree.

widgets/common/src/inputs/InputField.tsx
jawamartus: I don't know the history of the team discussion.

I know what are the good practices worldwide in react and I want to be a part of the team who follow them. I argued why I think that it's not good idea to use spread operator here.

What are your arguments?

widgets/common/src/inputs/InputFields.tsx
jawamartus: > I see nothing wrong in checking the same condition in another place at different level of component tree.

It is not DRY (in the correct definition of that). The logic is repeated.

> OK, it would be code repetition then and not 'not DRY'. I don't think it changes anything though.

There is nothing wrong in the code repetition here then.

widgets/common/src/inputs/InputField.tsx
kubama: I'm not going to discuss it here as it's not something specific to this PR. It's a general topic that is related to our whole code base and established team practices. Please start a new discussion on our weekly meeting if you think it is something that we should change.

widgets/common/src/inputs/InputField.tsx
```diff
@@ -0,0 +1,98 @@
...
+            const inRange = getConstraintValue('in_range');
+            const greaterThan = getConstraintValue('greater_than');
+            const greaterOrEqual = getConstraintValue('greater_or_equal');
+            const lessThan = getConstraintValue('less_than');
+            const lessOrEqual = getConstraintValue('less_or_equal');
+
+            return (
+                <NumberInputField
+                    {...commonProps}
+                    constraints={{ inRange, greaterThan, greaterOrEqual, lessThan, lessOrEqual }}
```
jawamartus, line 75: This looks good!

widgets/common/src/inputs/InputField.tsx
```diff
@@ -96,3 +96,5 @@ export default function InputField({
             return <GenericInputField {...commonProps} />;
     }
 }
+
+export default React.memo(InputField);
```
jawamartus, line 100: It doesn't prevent the *InputField components from rerendering. When the toolbox object reference would change *InputField child component would rerender. Even in the case when the component doesn't use toolbox property. Toolbox is only example property but it concerns all of them.

jawamartus: Replace asterix with Boolean, String etc.

widgets/common/src/inputs/InputFields.tsx
kubama: > It is not DRY (in the correct definition of that). The logic is repeated.

I don't agree with this statement. It's not justified to call a simple value comparison a repeated logic. It's an atomic operation.

> There is nothing wrong in the code repetition here then.

I'm sorry, but I have to point out that you are not very consistent in your comments.
In another comment you were complaining about repeated `div` wrapper that was used in some of the components. I agreed with you and removed it. Now you are suggesting an opposite thing - introducing a wrapper element that would be used across all the components, but at this time you don't see an issue with that.

jawamartus:
One system is failing. Please verify is it related to the changes or not.

widgets/common/src/inputs/InputFields.tsx
jawamartus: > In another comment you were complaining about repeated div wrapper that was used in some of the components. I agreed with you and removed it. Now you are suggesting an opposite thing - introducing a wrapper element that would be used across all the components, but at this time you don't see an issue with that.

Both situations were different. 
a) In the first case you were able to remove it. No the code is more unified. It was just redundant. And there was no additional logic there.
b) In the second case the logic is repeated. So it's not DRY. The wrapper element would be different for different *InputField. It wouldn't repeated. And we will be able to get rid of repeated logic here and make it more DRY.

widgets/common/src/inputs/InputField.tsx
kubama: > It doesn't prevent the *InputField components from rerendering.

This statement is not entirely true.
I don't think it's worth to spend any more time on optimization here, the related components are very shallow, performance improvement wouldn't even be noticeable.

jawamartus: Sure. I approve it. But please check the failing test again please.

jawamartus (approved):

widgets/common/src/inputs/InputFields.tsx
kubama: > In the first case you were able to remove it

The way the requested change was implemented is irrelevant. You are still giving contradicting suggestions.

> No the code is more unified. It was just redundant.

Yes. And now you are pushing for going the opposite direction.

> In the second case the logic is repeated. So it's not DRY. [...]  And we will be able to get rid of repeated logic here and make it more DRY.

I already replied to these statements.

> The wrapper element would be different for different *InputField.

No, it would be the same in all cases with one exception to boolean field.

kubama:
The failing system test is not related to the changes. Thus, I'm merging the PR.

widgets/common/src/inputs/InputFields.tsx
jawamartus: If you want to discuss it further let's have a call.

PR #1928, opened 2022-04-15
RD-4423: Introduced Tokens widget

NorbertAmartus:
## Description

Changes consists of providing a new Tokens widget with a corresponding test scenarios.
One think that has been done "extra", was updating `@types/react-dom` library, so that it would be possible to import `flushSync` method without the typescript warning

## Screenshots / Videos

Tokens widget
Create token
Received token
Delete token

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2540/)

## Documentation

[PR](https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2173) for the documentation update has been created.
The sychronization should take place as a part of this PR.


widgets/tokens/src/widget.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+Stage.defineWidget<never, TokensWidget.Data, TokensWidget.Configuration>({
+    id: 'tokens',
+    name: t('name'),
+    initialWidth: 12,
+    initialHeight: 16,
+    isReact: true,
+    hasReadme: true,
+    fetchUrl: '[manager]/tokens[params]',
+    // permission: Stage.GenericConfig.WIDGET_PERMISSION('tokens'),
```
NorbertAmartus, line 16: This line will be uncommented shortly, after merging another PR within [cloudify-manager-install repository](https://github.com/cloudify-cosmo/cloudify-manager-install) 😉

app/translations/en.json
```diff
@@ -1591,6 +1591,48 @@
...
+                    "lastUsed": "Last used"
+                },
+                "noTokens": "There are no Tokens available. Click \"Create\" to add tokens.",
+                "buttons": {
+                    "removeToken": "Delete token",
+                    "create": "Create"
+                }
+            },
+            "deleteModal": {
+                "content": "Are you sure you want to delete the token with ID {{tokenId}}"
```
kubama, line 1618: Question mark missing?

widgets/tokens/src/RemoveTokenButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { translationPath } from './widget.consts';
+import { tableRefreshEvent } from './TokensTable.consts';
+import { RequestStatus } from './types';
+import type { TokensWidget } from './widget.types';
+
+const {
+    Basic: { Icon, Confirm: DeleteModal },
+    Utils: { getT },
+    Hooks: { useBoolean }
+} = Stage;
```
kubama, line 11: Optional: I find nested destructuring hard to read and I personally avoid it.

widgets/tokens/src/RemoveTokenButton.tsx
```diff
@@ -0,0 +1,62 @@
+import { useEffect, useState } from 'react';
+import { translationPath } from './widget.consts';
+import { tableRefreshEvent } from './TokensTable.consts';
+import { RequestStatus } from './types';
+import type { TokensWidget } from './widget.types';
+
+const {
+    Basic: { Icon, Confirm: DeleteModal },
```
kubama, line 8: `Confirm: DeleteModal` looks quite surprising :slightly_smiling_face:

widgets/tokens/src/TextEllipsis.tsx
```diff
@@ -0,0 +1,11 @@
+import { EllipsisWrapper } from './TextEllipsis.styles';
```
kubama, line 1: `TextEllipsis` and `TextEllipsis.styles.tsx` are both so small that I would consider merging them into one.

widgets/tokens/src/TokensTableHeader/CreatedToken.styles.tsx
```diff
@@ -0,0 +1,6 @@
+import styled from 'styled-components';
```
kubama, line 1: Is it worth to create a separate file to keep such a small chunk of code?

kubama (commented):
My general comment is that I like the idea of having a pattern of separate `*.styles.tsx` and `*.types.ts` files, but I think that for simple cases where these files would contain just few lines of code it is not justified to extract them as it just brings unnecessary overhead.

widgets/tokens/src/RemoveTokenButton.tsx
NorbertAmartus: I've seen that at the [TenantsTable](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/tenants/src/TenantsTable.tsx#L134), [SecretsTable](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/secrets/src/SecretsTable.tsx#L220), [SitesTable](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/sites/src/SitesTable.tsx#L104)
So I thought that it should be done in a simillar way 😁 

But it shouldn't be a problem to modify that 😁 
So WDYT, would it be better to modify that naming? 🤔

widgets/tokens/src/TextEllipsis.tsx
NorbertAmartus: Sure!
This doesn't seem like a problem 😉

widgets/tokens/src/TokensTableHeader/CreatedToken.styles.tsx
NorbertAmartus: It's hard to determine worth of this approach 😁 
But personally I don't mind it, because when the solution is being implemented in one way (e.g extracting all of the styled components into the separate `*.styles.tsx` files), I think that it's good to follow the similar pattern in a different places as it makes the code more consistent and predictable

But as I've migrated the only component which was residing inside the `TextEllipsis.styles.tsx` file, I'll do the same here, for the consistency sake and to apply the suggestion above into our codebase 💪

app/translations/en.json
NorbertAmartus: Yup, nice catch!

widgets/tokens/src/TokensTableHeader/CreatedToken.styles.tsx
NorbertAmartus: Additionally, I've also migrated component from the `TokensTableHeader.styles.tsx` for the sake of the consistency (inside of this file there was also a single component) 😉

widgets/tokens/src/RemoveTokenButton.tsx
NorbertAmartus: I don't mind extracting it into a several declarations, so I'll apply it right away 😉

NorbertAmartus: Done 🚀

app/translations/en.json
NorbertAmartus: Questionmark has been added 🚀

widgets/tokens/src/TextEllipsis.tsx
NorbertAmartus: It has been migrated to the `TextEllipsis.tsx` 🎉

widgets/tokens/src/widget.tsx
NorbertAmartus: It has been uncommented 🥳

widgets/tokens/src/RemoveTokenButton.tsx
kubama: I wasn't aware of other appearances of such pattern. Personally I would change it, but it's rather cosmetic thing so I leave it up to you.

NorbertAmartus: In that case I'm leaving it as it is
Just because the `DeleteModal` seems to be more self explanatory than `Confirm` (without a `Modal` suffix)

widgets/tokens/src/TokensTable.tsx
```diff
@@ -0,0 +1,74 @@
...
+    return (
+        <>
+            <TokensTableHeader toolbox={toolbox} />
+            <DataTable fetchData={fetchTableData} noDataMessage={t('noTokens')}>
+                <DataTable.Column label={t('columns.token')} name={dataSortingKeys.value} />
+                <DataTable.Column label={t('columns.description')} name={dataSortingKeys.description} />
+                {shouldDisplayUsers && <DataTable.Column label={t('columns.username')} />}
+                <DataTable.Column label={t('columns.expirationDate')} name={dataSortingKeys.expirationDate} />
+                <DataTable.Column label={t('columns.lastUsed')} name={dataSortingKeys.lastUsed} />
+                <DataTable.Column label="" />
```
kubama, line 42: Since this column contains only single action icon then maybe we could set it to fixed width just enough for this icon? There would be more space for other columns then. We are doing this in some other table based widgets, for example here: widgets/filters/src/FiltersTable.tsx.

NorbertAmartus: Sure, seems like a nice adjustment, thanks for the suggestion!
I'll look on it right away 🚀

NorbertAmartus: I've applied the suggestion not only to the action column, but also to the columns containing the dates (as the width of the output dates are rather identical) 🥳

kubama (approved):
Looks nice and clean, congrats :slightly_smiling_face:.

kubama (approved):

PR #1895, opened 2022-03-25
RD-4141, RD-4138, RD-4141, RD-4146: secret_key and capabilities_value as an input for workflow parameters

jawamartus:
## Description

RD-4141, RD-4138, RD-4141, RD-4146

## Screenshots / Videos

N/A

## Checklist

- [x] My code follows the [UI Code Style](https://cloudifysource.atlassian.net/l/c/x8fq902p).
- [x] I verified that all tests and checks have passed.
- [x] I verified if that change requires any update in the [documentation](https://cloudifysource.atlassian.net/l/c/4XKtPocy).
- [x] I added proper labels to this PR.

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2365/)
[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2373/) 2022-04-07 - only test/cypress/integration/widgets/deployment_inputs_spec.ts
## Documentation

N/A

PR #1942, opened 2022-04-25
RD-4172, RD-4180, RD-4165, RD-4155: Workflow parameters - node_type, node_instance, scaling_group - tests

jawamartus:
## Description
RD-4172, RD-4180, RD-4165, RD-4155
Tests for new workflow parameters.

## Screenshots / Videos

N/A
## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2596/) - all tests
[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2595/) - single test
## Documentation

N/A


qooban:
@jawamartus I won't manage to review your PR today, will review tomorrow morning.

test/cypress/integration/widgets/workflow_parameters_spec.ts
```diff
@@ -0,0 +1,128 @@
+import type { DeploymentsViewWidgetConfiguration } from '../../../../widgets/deploymentsView/src/widget';
+
+describe('handles deployment inputs of type', () => {
```
qooban, line 3: You probably meant something like: "Execute Workflow modal handles parameters of type", am I right?

test/cypress/integration/widgets/workflow_parameters_spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    const resourcePrefix = 'workflow_input_test_';
```
qooban, line 4: To be consistent a better name would probably be `workflow_parameters_test_`.

test/cypress/integration/widgets/workflow_parameters_spec.ts
```diff
@@ -0,0 +1,128 @@
...
+        const contains = typedPrefix ? ` :contains("${typedPrefix}")` : '';
+        if (number === 0) {
+            cy.get('.menu').contains('No results found.').should('be.visible');
+        } else if (atLeast) {
+            cy.get(`.menu .item[role="option"]${contains}`).should('have.length.at.least', number);
+        } else {
+            cy.get(`.menu .item[role="option"]${contains}`).should('have.length', number);
+        }
+        cy.get('label').click();
+    };
```
qooban, line 35: Isn't the same function defined in `test/cypress/integration/widgets/deployment_inputs_spec.ts`?
I think it would be good to extract it. I'm not sure about the best place, but maybe to `test/cypress/support/asserts.ts`?

test/cypress/integration/widgets/workflow_parameters_spec.ts
```diff
@@ -0,0 +1,128 @@
...
+            const deploymentName = `${resourcePrefix}${type}_type_deployment`;
+            cy.log(`Selecting deployment: ${deploymentName}.`);
+            cy.contains('tr', deploymentName).click();
+        });
+        cy.contains('Execute workflow').click();
+        cy.contains('Script').click();
+        cy.contains('Test parameters').click();
+    };
+
+    const verifyNumberOfOptions = (number: number, atLeast = false, typedPrefix = '') => {
```
qooban, line 19: Seeing that we have two optional parameters here, maybe worth to introduce options object for `atLeast` and `typedPrefix`?

test/cypress/integration/widgets/workflow_parameters_spec.ts
```diff
@@ -0,0 +1,128 @@
...
+            customPollingTime: 10,
+            sortColumn: 'created_at',
+            sortAscending: false,
+            mapHeight: 300,
+            mapOpenByDefault: false
+        };
+        const additionalWidgetIdsToLoad: string[] = ['deploymentActionButtons', 'deploymentsViewDrilledDown'];
+
+        cy.activate('valid_trial_license')
+            .usePageMock('deploymentsView', widgetConfiguration, {
```
qooban, line 65: Why do we need `deploymentsView` and `deploymentsViewDrilledDown` widgets here?

Knowing that with `usePageMock` we already have "Resource Filter" widget available in the page I suspect that `deploymentActionButtons` should be enough.

To select deployment you can just use `setDeploymentContext` custom Cypress command.

test/cypress/integration/widgets/workflow_parameters_spec.ts
```diff
@@ -0,0 +1,128 @@
+import type { DeploymentsViewWidgetConfiguration } from '../../../../widgets/deploymentsView/src/widget';
```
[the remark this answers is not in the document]
qooban: Maybe that structure:

```
test/cypress/integration/widgets
  deploymentActionButtons
    widget_spec.ts                // test/cypress/integration/widgets/deployment_actions_spec.ts
    workflow_parameters_spec.ts   // this file
  createDeploymentButton
    widget_spec.ts                // test/cypress/integration/widgets/deployment_button_spec.ts
    deployment_inputs_spec.ts     // test/cypress/integration/widgets/deployment_inputs_spec.ts
```

would be more suitable?

test/cypress/integration/widgets/workflow_parameters_spec.ts
jawamartus: Both are valid from my perspective. However I would change it for the suggested version.

test/cypress/integration/widgets/workflow_parameters_spec.ts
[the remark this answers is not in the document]
jawamartus: The tests are cross-widget. The current structure is OK for me.

test/cypress/integration/widgets/workflow_parameters_spec.ts
jawamartus: Fixed

test/cypress/integration/widgets/workflow_parameters_spec.ts
jawamartus: Thanks for the highlighting. I forgot to removed not used part of this function. The functions are different now.

test/cypress/integration/widgets/workflow_parameters_spec.ts
jawamartus: Probably we don't. The change you requested require a few additional hours. Would you like me to spend them on that in the scope of this PR? It would put at risk 4 US from our current sprint.

test/cypress/integration/widgets/workflow_parameters_spec.ts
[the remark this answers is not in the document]
qooban: Which tests are cross-widget? :-) This test is actually using Deployment Action Buttons (Deployments View is used only for deployments filtering IIUC).

The structure is not good IMHO, `deployment_inputs_spec.ts` and `workflow_parameters_spec.ts` are violating kind of convention we tried to follow - each spec in `widgets` folder represents a test for a specific widget. It's OK to split specs, but we could do better with file/directory naming to clearly indicate the hierarchy, so that's why I proposed the different structure.

test/cypress/integration/widgets/workflow_parameters_spec.ts
qooban: If we don't need them, let's remove it.

> Would you like me to spend them on that in the scope of this PR? It would put at risk 4 US from our current sprint.

Yes. The main part of those user stories is already merged. This PR is only about the tests.

I guess you would need to remove some parts from L37 - L58 and adjust `openWorkflowParametersModal` to use `setDeploymentContext` command.

jawamartus: > Yes. The main part of those user stories is already merged. This PR is only about the tests.
Yes. Remember that tests are part of our US. We cannot close them before merging tests.

Can I make the refactoring you mention as a part of separated PR? This PR looks like valid and make our US done.

jawamartus: Personally I don't think that spending more hours (maybe day or two) is a profitable for the product. I understand your perfectionism however I prefer to use Pareto rule in my work which seems to be more profitable for the business and our team.

What additional benefits we would get from additional one or two days spent on refinement of this? Is there any bug here? Does this tests are very time consuming?

qooban: > Can I make the refactoring you mention as a part of separated PR? This PR looks like valid and make our US done.

Wouldn't it be more time consuming to create another PR, wait for another review and so on? Isn't that what you recently pointed out as a time consuming activities when describing issues with security audit fixes?

> Personally I don't think that spending more hours (maybe day or two) on refinement of this is a profitable for the product. I understand your perfectionism however I prefer to use Pareto rule in my work which seems to be more profitable for the business and our team.
> What additional benefits we would get from additional one or two days spent on refinement of this? Is there any bug here? Does this tests are very time consuming?

It's not about perfectionism, Pareto or whatever general term. Let's stick to that specific case we are discussing. I found out that you are using additional widget (Deployments View) for deployment filter. This widget is redundant here as Resource Filter is already available. You can do the same filtering without adding another widget. I don't believe it's a matter of day or two to simplify the current solution.

Correct me if I'm wrong, but my impression is that you are afraid of not closing those 4 USes before the end of the sprint, right? The sprint which normally would finish in the last week. The sprint which will start today will last for 3 days for PL team. Are we really afraid of slipping some USes from one sprint to another in such unusual situation? 

I don't believe that anyone will blame you about that as nobody did that so far in our organization. And in case anyone will, then full responsibility is on me, so please redirect such a person to me.

test/cypress/integration/widgets/workflow_parameters_spec.ts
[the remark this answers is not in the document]
qooban: > 1. The functionality and business logic I test are widget agnostic. As you may know the changes I test are present at `InputField` component. Which is not part of any widget.

That's true, but you are putting your spec test file in `widgets` folder which contains spec files testing individual widgets.

> 2. The `deployment_inputs_spec.ts` file is not a part of this ticket at all. It was approved before by our code guardian. Therefore there is no convention you are talking about.

It was not approved by code guardian. None of the code guardians were reviewing https://github.com/cloudify-cosmo/cloudify-stage/pull/1895 .

> 3. The changes you request would force me to rewrite the tests and make it dependent on specific components only. It requires a lot of time. One single run it's around one or two minutes after every change. Also it would require running it again at Jenkins.

The changes I suggest are only about directory structure changes. None of the tests need to be rewritten right now. I don't think it requires a lot of time.

> 4. As I said I before I would love to write the enzyme unit tests for this case. However I want to follow the suggested convention by our Team Leader in other discussions. Which are:
>    a) we consider Enzyme tests as deprecated
>    b) we are writing system tests for testing input parameters and workflow parameters

As far as I remember, we agreed that deployment inputs tests may require refactoring in the future and one of the options we discussed was to consider having Cypress components tests (which were not available during the time of introducing deployment input tests). But that's a different topic. New Input Parameters epic was/is good opportunity to refactor it, but if you won't find time, then it will also be OK to keep it as it is (system tests for deployment input and workflow params).

test/cypress/integration/widgets/workflow_parameters_spec.ts
jawamartus: > Correct me if I'm wrong, but my impression is that you are afraid of not closing those 4 USes before the end of the sprint, right?

You are wrong. It's very discouraging to work on things that doesn't have almost any impact on the business and doesn't provide significant value to the customers. It's like digging pits that no one needs. Not treating our team goals seriously it's also discouraging.

> It's not about perfectionism, Pareto or whatever general term. Let's stick to that specific case we are discussing. I found out that you are using additional widget (Deployments View) for deployment filter. This widget is redundant here as Resource Filter is already available. You can do the same filtering without adding another widget. I don't believe it's a matter of day or two to simplify the current solution.

It's about perfectionism. Let's stick to this specific case. You've required time consuming change which provide much much smaller benefits to the business I could do in the same time. Let me explain it wider.

The change you require needs me to:
1. For every widget I used:
- to test whether it is necessary or not and run system tests locally (it may take around 2 hours to check it on mu local computer including time for resetting cloudify-manager. It's very fluky on my machine and need some reset from time to time. )
2. It also require me to switching the context from other tickets.
3. Rewrite our tests for using different interface. ( additional 2 hours and running tests locally)
4. Push changes to the PR and run Jenkins tests (additional 2 hours)
5. Ask for rereview (up to one day).
6. Eventually spend time for another iteration, and another typo fixes and another rereview etc.

What benefits we would get from the changes?
A few seconds faster tests.

Please consider, that the bigger benefits we would get from removing the line 98 from file `deployment_inputs_spec.ts` which is very time consuming and can be done in up to two hours or even faster in the case of doing it BTW of other changes. 
This line I mean:
```
        cy.refreshPage();
```


Please note, that I don't do refreshPage in this particular test.

So, why you don't want to make me to spend my time on other more impactful things? The only answer is perfectionism.

Rationally it's better to spend 2 hours for providing bigger value (bigger test runtime optimisation). Than 8 hours for the changes which doesn't have any impact.

Please note that the test I wrote are correct. Maybe not the most efficient, but correct and providing the value to our product.

qooban: If you don't agree with my approach, let's have a meeting, there's no point in having such discussions again and again.

> What benefits we would get from the changes?

Consistency and less confusion in the future when new developers will read your code.

> Rationally it's better to spend 2 hours for providing bigger value (bigger test runtime optimisation). Than 8 hours for the changes which doesn't have any impact.

8h? Are you serious? 🙂

jawamartus: > 8h? Are you serious? 🙂

Please consider aforementioned points above.

> (...)let's have a meeting (...)

Yes, let's have a meeting.

jawamartus: > 8h? Are you serious? 🙂

If you have any idea how I can make it faster please let me know. 

"Are you serious?" It's not helpful at all. It's also not an argument.

qooban (approved):
LGTM! 👍🏻

PR #1944, opened 2022-04-25
RD-4423: Added ability to always copy the token value

NorbertAmartus:
## Description

Changes consists of passing to the `<CopyToClipboardButton />` component always a token value, not the displayed token value (which may have asterisks)

## Screenshots / Videos

N/A

## Checklist

## Tests

 

I've checked locally if the modification is working as expected.
As no tests cenario was related to the modified functionality, I didn't execute the System Tests.

## Documentation

N/A

### DIFF
--- a/widgets/tokens/src/TokensTableHeader/CreatedToken.tsx
+++ b/widgets/tokens/src/TokensTableHeader/CreatedToken.tsx
@@ -41,7 +41,7 @@ const CreatedToken = ({ token }: CreatedTokenProps) => {
                         icon={isTokenValueVisible ? 'eye slash' : 'eye'}
                         onClick={isTokenValueVisible ? hideTokenValue : showTokenValue}
                     />
-                    <CopyToClipboardButton text={displayedToken} />
+                    <CopyToClipboardButton text={token.value} />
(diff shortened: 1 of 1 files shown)


qooban (approved):
LGTM 👍🏻 

Just out of curiosity - do you know if there's an easy way to test what is the content of the clipboard eg. from Cypress?

NorbertAmartus:
@qooban 
> Just out of curiosity - do you know if there's an easy way to test what is the content of the clipboard eg. from Cypress?

If I remember correctly it coud be done with the help of the [clipboardy](https://www.npmjs.com/package/clipboardy) package or by using the `window` object inside the cypress like:

```typescript
cy.window().then(window => {
    window .navigator.clipboard.readText().then(copiedText => {
      expect(copiedText).to.eq('Something');
    });
  });
```

But just a note, that sometimes we may encounter the `Uncaught (in promise) DOMException: Document is not focused.` error with the second approach
And it's related to the fact that this functionality ONLY works when the corresponding window is being focused (e.g. you can't be in the developer console, while executing it, because it'll result in getting the error above)

PR #1946, opened 2022-04-26
RD-4650: Allow _ as a starting character for variables

jawamartus:
## Description

RD-4650

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2651/)
## Documentation

N/A


backend/templates/terraform/blueprint.ejs
```diff
@@ -22,10 +22,10 @@ node_templates:
...
       resource_config:
         source:
-          location: "<%= terraformTemplate %>"<% if (urlAuthentication) { %>
-          username: { get_secret: "<%= blueprintName %>.username" }
-          password: { get_secret: "<%= blueprintName %>.password" }<% } %>
-        source_path: "<%= resourceLocation %>"
+          location: '<%= terraformTemplate %>'<% if (urlAuthentication) { %>
+          username: { get_secret: '<%= blueprintName %>.username' }
+          password: { get_secret: '<%= blueprintName %>.password' }<% } %>
+        source_path: '<%= resourceLocation %>'
```
qooban, line 28: Why do we need this change?

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -75,7 +75,8 @@ function getDynamicTableDropdown(options: DropdownProps['options']) {
     };
 }
 
-const cloudifyResourceRegexp = /^[a-zA-Z][a-zA-Z0-9._-]*$/;
+const cloudifyBlueprintNameRegexp = /^[a-zA-Z][a-zA-Z0-9._-]*$/;
+const cloudifyResourceRegexp = /^[a-zA-Z0-9._-]*$/;
```
qooban, line 79: Few comments here:
1. I'm wondering if we are sure that for all places where `cloudifyResourceRegexp` is used loosening the regex is the correct solution. Did you check that with Michael S. or user documentation?
2. I'm not sure if the name was/is correct. It seems to be used for both Terraform resources and Cloudify resources. Can we make it more clear?


Optional:

Maybe worth to update 'validate individual form fields' test to have this kind of validation verified?

backend/templates/terraform/outputs.ejs
```diff
@@ -3,5 +3,5 @@
 <% } _%>
 <% outputs.forEach(output => { _%>
   <%= output.name %>:
-    value: { get_attribute: [ cloud_resources, outputs, <%= output.terraformOutput %>, value ] }
+    value: { get_attribute: [ cloud_resources, outputs, '<%= String(output.terraformOutput).replace("'", "''") %>', value ] }
```
qooban, line 6: Oh, do I understand it correctly that in YAML we escape single quotes by having double single quotes?

qooban (commented):
Looks like a quick fix, so the optimism from Backlog Refinement was correct! 🙂

It seems like `test/routes/Terraform.test.ts` needs some adjustments after introducing changes in the blueprint templates.

backend/templates/terraform/outputs.ejs
jawamartus: Yes. But only when we use the single quote to quote the string. That's how the yaml work.

backend/templates/terraform/blueprint.ejs
jawamartus: It's a refactoring.

jawamartus: And I made it more robust.

What if the location would include question mark? It has special meaning inside of yaml and it should be quoted. If you know any issues related to that in our backlog, let me know. This would fix them.

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useState, useRef, useEffect } from 'react';
 import type { FormEvent } from 'react';
 import type { CheckboxProps, DropdownProps } from 'semantic-ui-react';
 import { Ref } from 'semantic-ui-react';
-import _, { find, isEmpty } from 'lodash';
```
jawamartus, line 5: Ongoing refactoring

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -236,18 +238,18 @@ export default function TerraformModal({
         function validateNames(namedEntities: { name: string }[], errorPrefix: string) {
             const tNameError = Stage.Utils.composeT(tError, errorPrefix);
             let nameError = false;
-            if (find(namedEntities, variable => isEmpty(variable.name))) {
+            if (some(namedEntities, variable => isEmpty(variable.name))) {
```
jawamartus, line 241: Using more proper functions for proper purpose.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: > 1. I'm wondering if we are sure that for all places where cloudifyResourceRegexp is used loosening the regex is the correct solution. Did you check that with Michael S. or user documentation?

I checked it with our backend validation. It's acceptable by the logic on the backend side. 

> 2. I'm not sure if the name was/is correct. It seems to be used for both Terraform resources and Cloudify resources. Can we make it more clear?

Sure, I can refactor it. However keep in mind that this require me to understand the bigger context to set more perfect variable name. 

It's like writing a poem. Some poets spend whole day for one verse to make sure that the verse is almost perfect. Especially some ancient Greek poets. I like writing so I'm happy to do it.

jawamartus: >  Did you check that with Michael S. or user documentation?
Are you sure that I should engage Michael S. on this level of implementation details and explain to him all the details? Please confirm if it's required.

jawamartus: > 2. I'm not sure if the name was/is correct. It seems to be used for both Terraform resources and Cloudify resources. Can we make it more clear?

I changed the names accordingly. Let me know if you want me to elaborate more about the names I decided to use.

backend/templates/terraform/blueprint.ejs
qooban: Clear.

qooban (commented):
RD-4650 was - as your PR title says - to "Allow _ as a starting character for variables".

I see many unrelated changes, for example in the field naming convention (`name` -> `key`, `value` -> `name`, removed `default`) and there's no description about them. I'm not saying we cannot do refactoring at once, but without clear explanation it's hard to review this code, so could you please provide more explanation about all those changes not related to allowing `_` as starting character for variables?

@kubama Can I ask you take over the review? I suspect I might be quite busy today with other stuff.

jawamartus:
> I see many unrelated changes, for example in the field naming convention (name -> key, value -> name, removed default) and there's no description about them. I'm not saying we cannot do refactoring at once, but without clear explanation it's hard to review this code, so could you please provide more explanation about all those changes not related to allowing _ as starting character for variables?

Explanation.

We were using inproper and misleading names in our code. Now we are using more proper. For example - now we are using name key for the column described as "Name". Previously we were using 'value' name for the column 'name'. This mismatch led to wrong error messages.

app/translations/en.json
```diff
@@ -921,29 +921,29 @@
                     "noUsername": "Please provide username",
                     "noPassword": "Please provide password",
                     "noResourceLocation": "Please provide resource location",
-                    "invalidBlueprintName": "Please provide valid blueprint name - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
+                    "invalidBlueprintName": "Please provide valid blueprint name. It should start with a letter and contain only letters, digits and the following characters \"-\", \".\" and \"_\".",
                     "invalidTerraformTemplate": "Please provide valid Terraform template URL",
```
jawamartus, line 925: The error message was inappropriate. It's a bug. We doesn't allow the blueprint name to start from underscore. Let me know if we have a ticket for that - I will close it.

app/translations/en.json
```diff
@@ -921,29 +921,29 @@
...
                     "blueprintNameInUse": "Blueprint '{{blueprintName}}' already exists",
                     "variable": {
-                        "nameMissing": "Please provide variable name",
-                        "nameInvalid": "Please provide valid variable name - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
-                        "nameDuplicated": "Variables must be unique, duplicates are not allowed",
+                        "keyMissing": "Please provide variable key",
+                        "keyInvalid": "Please provide valid variable key. It should contain only letters, digits and the following characters \"-\", \".\" and \"_\".",
+                        "keyDuplicated": "Variable keys must be unique, duplicates are not allowed",
```
jawamartus, line 930: We were using name prefix for the first column in the table which is actually key (unique). The other column were described as name.

backend/routes/Terraform.types.ts
```diff
@@ -1,8 +1,8 @@
 export interface Variable {
+    variable: string;
     name: string;
     source: 'input' | 'secret' | 'static';
     value: string;
-    default: string;
```
jawamartus, line 5: We don't need this property. 
After the changes:
"Variable"/ "Environmental variable" -> "key"
"Name" -> "name" (before - "value")
"Value" -> "value" (before - "default")

app/translations/en.json
```diff
@@ -921,29 +921,29 @@
                     "noUsername": "Please provide username",
                     "noPassword": "Please provide password",
                     "noResourceLocation": "Please provide resource location",
-                    "invalidBlueprintName": "Please provide valid blueprint name - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
+                    "invalidBlueprintName": "Please provide valid blueprint name. It should start with a letter and contain only letters, digits and the following characters \"-\", \".\" and \"_\".",
```
kubama, line 924:
```suggestion
                    "invalidBlueprintName": "Please provide valid blueprint name. It should start with a letter and contain only letters, digits and the following characters: \"-\", \".\" and \"_\".",
```

app/translations/en.json
```diff
@@ -921,29 +921,29 @@
...
-                        "nameInvalid": "Please provide valid variable name - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
-                        "nameDuplicated": "Variables must be unique, duplicates are not allowed",
+                        "keyMissing": "Please provide variable key",
+                        "keyInvalid": "Please provide valid variable key. It should contain only letters, digits and the following characters \"-\", \".\" and \"_\".",
+                        "keyDuplicated": "Variable keys must be unique, duplicates are not allowed",
                         "sourceMissing": "Please provide variable source",
-                        "valueMissing": "Please provide variable name",
-                        "valueInvalid": "Please provide valid variable name - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed"
+                        "nameMissing": "Please provide variable name",
+                        "nameInvalid": "Please provide valid variable name. It should start with a letter and contain only letters, digits and the characters \"-\", \".\" and \"_\"."
```
kubama, line 933: IIUC the 'name' now refers to source (input/secret) name, so it shouldn't be called 'variable name'.

app/translations/en.json
```diff
@@ -921,29 +921,29 @@
...
-                    "invalidBlueprintName": "Please provide valid blueprint name - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
+                    "invalidBlueprintName": "Please provide valid blueprint name. It should start with a letter and contain only letters, digits and the following characters \"-\", \".\" and \"_\".",
                     "invalidTerraformTemplate": "Please provide valid Terraform template URL",
                     "blueprintNameInUse": "Blueprint '{{blueprintName}}' already exists",
                     "variable": {
-                        "nameMissing": "Please provide variable name",
-                        "nameInvalid": "Please provide valid variable name - only letters, digits and the characters \"-\", \".\" and \"_\" are allowed",
-                        "nameDuplicated": "Variables must be unique, duplicates are not allowed",
+                        "keyMissing": "Please provide variable key",
+                        "keyInvalid": "Please provide valid variable key. It should contain only letters, digits and the following characters \"-\", \".\" and \"_\".",
```
kubama, line 929:
```suggestion
                        "keyInvalid": "Please provide valid variable key. It should contain only letters, digits and the following characters: \"-\", \".\" and \"_\".",
```

backend/routes/Terraform.types.ts
```diff
@@ -1,8 +1,8 @@
 export interface Variable {
+    variable: string;
```
kubama, line 2: Wouldn't it be cleaner to call it `key` so it was in line with UI messages?

backend/routes/Terraform.types.ts
```diff
@@ -1,8 +1,8 @@
...
     name: string;
```
kubama, line 3: If we are already making changes in this area I would consider renaming `name` to `sourceName` so it was clear what this 'name' refers to. At the moment it's easy to think that 'name' is a variable name, which is not.

app/translations/en.json
jawamartus: > IIUC the 'name' now refers to source (input/secret) name, so it shouldn't be called 'variable name'.

That's not correct. The name refers to the "name" column.
image

backend/routes/Terraform.types.ts
jawamartus: I think differently. The column is called variable - more precisely. Key is more general term.

In our case the key might be the name or variable.

backend/routes/Terraform.types.ts
jawamartus: The `name` refer to our business logic. The business requirement is to call the column name. Not "source name" for some reasons. The variable name is currently with the column name in sync.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: @qooban What do you think about the new variable names? Can I resolve this conversation?

backend/routes/Terraform.types.ts
kubama: > In the case of Output it's different.

OK, but this interface is not used in the context of outputs.

backend/routes/Terraform.types.ts
kubama: > We don't have "source name" column.

We have, but it's titled 'Name' :slightly_smiling_face:.

> And I'm not sure that it would always refer to the source name. Are you sure about that? Do you have some knowledge I don't have?

As far as I know this is the case.

jawamartus: I trust our PMs. They decided to use Name. They also have wider picture. Feel free to send to them the proposition of your name for the column. Definitely we need more trust inside of our team.

backend/routes/Terraform.types.ts
jawamartus: That's why the name is more precise and it's variable. As you noticed we don't need to have more general name here.

Yes, the name of variable is important. However this one is good and there is no need to change it and to block the PR because of that.

It's good to discuss. However I would love to see discussion about more important things in our team according to Pareto rule. Discussing variable names and not seeing bigger issues like lack of architectural structure in our app is very discouraging for me.

jawamartus: Our time is limited, we don't have enough of the time to make everything perfect. The variable name doesn't affect customers at all. It's very very minor thing.

app/translations/en.json
jawamartus: I agree that is not perfect. We don't have a time to do everything perfect. Please notice that it's not what about is this ticket. Before the change the error messages were more confusing.

Please notice that in the case of applying of your solution (replacing variable name with source name) the user wouldn't be able to distinguish between the errors inside of environmental variables section and variables section in the TerraformModal. Now we have different message for variable name column and environmental variable name column.

Is that the reason for blocking this PR?

jawamartus: I think that we need definitely more trust in our team.

backend/routes/Terraform.types.ts
kubama: My proposal wasn't to change anything in the user interface, but change property name in the code. It has nothing to do with PMs.

kubama (approved):
There are unresolved conversations but I'm not willing to continue them.

PR #1954, opened 2022-05-02
RD-4556: Allow auto opening of the deployment modal and one blueprint upload from the marketplace

NorbertAmartus:
## Description

Changes consists of sharing information about uploading blueprints across different `BlueprintCatalog` tabs
Additionally, there has been added ability to open `Deyployments modal` on the `BlueprintActionButtons` initialization.

Other than that, some refactoring and labels migration (to the translation file) has been made 🎉 

## Screenshots / Videos

N/A

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2711/)

## Documentation

N/A


widgets/blueprintCatalog/src/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+    toolbox.getContext().setValue(Consts.CONTEXT_KEY.UPLOADING_BLUEPRINT, blueprintName);
+};
+
+const getUploadingBlueprintFromCatalogTabs = (toolbox: WidgetlessToolbox) => {
+    return toolbox.getContext().getValue(Consts.CONTEXT_KEY.UPLOADING_BLUEPRINT);
+};
+
+const isUploadingBlueprintOnAnotherCatalogTab = (toolbox: WidgetlessToolbox): boolean => {
+    return !!getUploadingBlueprintFromCatalogTabs(toolbox);
+};
```
NorbertAmartus, line 18: I tried to create names as descriptive, so they purpose would be clear (to store information about the uploading blueprint, across the blueprint catalog tabs) 

If you think that the proper comment would be beneficial, then give me a shout and I'll provide an additionall info 😉

qooban: Blueprints Catalog widget doesn't need to know anything about catalog tabs, so probably we don't need to refer to catalog tabs in function names. What do you think?

widgets/blueprintCatalog/src/widget.tsx
```diff
@@ -143,6 +145,12 @@ Stage.defineWidget<
 
     render(widget, data, _error, toolbox) {
         const { Common, Basic } = Stage;
+        const isUploadingBlueprintFromAnotherTab = Utils.isUploadingBlueprintOnAnotherCatalogTab(toolbox);
+
+        if (isUploadingBlueprintFromAnotherTab) {
+            const uploadingBlueprintName = Utils.getUploadingBlueprintFromCatalogTabs(toolbox);
+            return <UploadingMessage blueprintName={uploadingBlueprintName} />;
```
qooban, line 152: Maybe we should show `UploadingMessage` on top of the content and not instead? (similarly as we do e.g. in modal windows - blueprint upload, create deployment, etc.) 
WDYT?

widgets/blueprintCatalog/src/RepositoryList.tsx
```diff
@@ -3,23 +3,28 @@ import { without } from 'lodash';
...
+
 interface RepositoryListProps {
     data: BlueprintCatalogPayload;
     widget: Stage.Types.Widget<BlueprintCatalogWidgetConfiguration>;
     toolbox: Stage.Types.Toolbox;
     actions: Actions;
 }
 interface RepositoryListState {
-    uploadingBlueprints: string[];
+    uploadingBlueprint?: string;
```
qooban, line 27: Do you think it's necessary to keep that in state if we have that in the context already? 🤔 
(I suspect it may lead to synchronization issues)

widgets/blueprintCatalog/src/RepositoryList.tsx
```diff
@@ -137,29 +159,22 @@ export default class RepositoryList extends React.Component<RepositoryListProps,
...
-        );
+        const isUploadingBlueprint = !!uploadingBlueprint;
 
         const showNotAuthenticatedWarning = data.source === Consts.GITHUB_DATA_SOURCE && !data.isAuthenticated;
 
         const RepositoryView = widget.configuration.displayStyle === 'table' ? RepositoryTable : RepositoryCatalog;
 
+        if (isUploadingBlueprint) {
+            return <UploadingMessage blueprintName={uploadingBlueprint!} />;
+        }
```
qooban, line 176: Similar comment to one already left: Maybe worth to show `UploadingMessage` on top of the normal widget content and not instead?

widgets/blueprintCatalog/src/consts.ts
```diff
@@ -1,8 +1,12 @@
...
 export default {
+    WIDGET_ID: 'blueprintCatalog',
     GITHUB_DATA_SOURCE: 'github',
     JSON_DATA_SOURCE: 'json',
     DEFAULT_IMAGE: '/images/logo.png',
     BLUEPRINT_IMAGE_FILENAME: 'blueprint.png',
-    GITHUB_BLUEPRINT_IMAGE_URL: (user, repo) => `/github/content/${user}/${repo}/master/blueprint.png`
+    GITHUB_BLUEPRINT_IMAGE_URL: (user: string, repo: string) => `/github/content/${user}/${repo}/master/blueprint.png`,
+    CONTEXT_KEY: {
+        UPLOADING_BLUEPRINT: 'uploadingBlueprint',
```
qooban, line 9: Optional:

I'm wondering shouldn't we add some common prefixes here to indicate the source, for example something like `blueprintCatalog_uploadingBlueprint` or `uploadingBlueprintFromCatalog`. Especially that uploading blueprint from other places - Blueprint upload modal / Terraform upload modal - will not use this parameter.

widgets/blueprintCatalog/src/consts.ts
```diff
@@ -1,8 +1,12 @@
...
+        BLUEPRINT_CATALOG_ID: 'blueprintCatalogId'
```
qooban, line 10: Is it used to indicate currently selected tab in Blueprint Marketplace modal / page?

test/cypress/integration/flows/user_spec.ts
```diff
@@ -70,4 +70,18 @@ describe('User flow', () => {
...
+        const uploadBlueprintButtonSelector = 'i[title="Upload blueprint"]:not(.disabled)';
+
+        cy.contains('Create new Deployment').click();
+
+        cy.contains('Blueprint Marketplace');
+        cy.get(uploadBlueprintButtonSelector).first().click();
+        cy.contains('Uploading');
+
+        cy.get('.modal').contains('Deploy blueprint', {
+            timeout: secondsToMs(30)
```
qooban, line 84: What do you think about using `waitUntil...` Cypress command instead of defining timeout?

Something like:
```
cy.waitUntilNotEmpty(`blueprints?state=uploaded`, { search: blueprintName }); 
```
See how it is used "handle template URL authentication" test case in `test/cypress/integration/widgets/blueprints_spec.ts`.
Maybe common waiting function could be extracted for waiting for blueprint to be uploaded and placed in `test/cypress/support/blueprints.ts`. WDYT?

app/translations/en.json
```diff
@@ -993,6 +995,18 @@
                     "label": "Sort by name",
                     "description": "If set to true, then blueprints will be sorted by name."
                 }
+            },
+            "authenticationWarning": "No GitHub credentials provided! Widget may stop working after reaching unrestricted query limit (~50). Fix this by adding &github-username& and &github-password& entries to your secrets store (Secrets widget).",
+            "uploadingMessage": {
+                "uploading": "Uploading {{blueprintName}} blueprint",
```
qooban, line 1001: Maybe:
```suggestion
                "uploading": "Uploading '{{blueprintName}}' blueprint...",
```
to be inline with other parts of the application? (see for example "widgets.common.deployments.deployModal.steps" in `en.json`)

app/translations/en.json
```diff
@@ -993,6 +995,18 @@
...
+                "redirection": "After completing the upload, you'll be redirected to the dedicated blueprint page"
```
qooban, line 1002: I'm not sure about the message, maybe something like: 
```suggestion
                "redirection": "Once the blueprint is uploaded, you'll be redirected to the blueprint deployment form."
```
?

I'm also not sure about my proposal. Probably somebody from the product team should validate it.

qooban, line 1002: Looking at:

image

I'm wondering maybe we should have a little bit of separation space between `uploading` and `redirection` messages? So that we have main message on the top and some description below. WDYT?

qooban (commented):
Well done! Good refactors around Blueprints Catalog widget, including some TS migrations.

Tested the UI roughly and added some comments.

app/translations/en.json
NorbertAmartus: Sure, seems reasonable 💪

widgets/blueprintActionButtons/src/BlueprintActionButtons.tsx
```diff
@@ -2,6 +2,13 @@ import type { ComponentProps } from 'react';
 
 const t = Stage.Utils.getT('widgets.blueprintActionButtons.buttons');
 
+const MODAL_TYPE = {
+    DEPLOY: 'deploy',
+    DELETE: 'delete'
+} as const;
+
+type MODAL_TYPE = typeof MODAL_TYPE[keyof typeof MODAL_TYPE];
```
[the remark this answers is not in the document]
NorbertAmartus: Modifications has been applied 🚀

widgets/blueprintCatalog/src/RepositoryList.tsx
NorbertAmartus: I've left this on purpose, as I was assuming that the context change would not make a widget rerender, thus making component synced with the `uploadingBlueprint` state

But as I've double checked right now, it does rerender 🎉 
So I don't see a reason why we should keep the local state, in that case 🚀 

Having that said, I've removed that information from the local state

widgets/blueprintCatalog/src/consts.ts
NorbertAmartus: As I'm looking into the code it seems like `BLUEPRINT_CATALOG_ID` is beind used to determine which of the catalog items is being selected
(after clicking on any item, the whole row containing this item will have a dark background)

`Catalog view` also is implementing this functionality, but it may be a good question if this state should stay in context or if it should be migrated to be a local state of the component 🤔

widgets/blueprintCatalog/src/utils.ts
NorbertAmartus: Yeah, it seems like it 😁 
But I just wanted to give more information to the developer about the purpose of this approach

I know that it's just setting/removing some data from the context, but the developer may don't know why it's being used in that way 😁 
So I thought that putting a little more descriptive function name would make the future the development faster and the developer life easier 😄 

What do you think about that thought behind that decision?
It seems like a valid point / could we achieve the simillar effect in a different way? 🤔 
(I've tried to ommit comments as we can use functions in a different places, making sure that the purpose of they execution is being known in all of those places)

app/translations/en.json
NorbertAmartus: This approach seems to be more precise 💪 
I've just got one concern about the `you'll be redirected to the blueprint deployment form`, as the user will be redirected to a different page upon opening which he'll be able to see the mentioned modal
Once closing the modal, he'll be on a different page than he was while submitting the form

Because of that I think that there may be a room for improvement for this part of the message

But as you've suggested, I'll just contact product team to not wonder around, but just to get straight answers 🚀 

Thanks for the suggestion! 🎉 
(first part of it is really neat! 😎)

widgets/blueprintActionButtons/src/BlueprintActionButtons.tsx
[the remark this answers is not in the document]
NorbertAmartus: One note from the implementation side

I had to modify the naming of that variable, because lowercase `modalType` was used in a several places (as a function and state property) - which typescript didn't like 😄 
Because of that, we can now see in the code `availableModalType` instead of `modalType` 🚀

widgets/blueprintCatalog/src/widget.tsx
qooban: Something like:

https://user-images.githubusercontent.com/5202105/168006794-50582329-91d0-4103-bc6a-1a340a00a1b2.mp4

The content of the modal is still visible "behind" 🙂

NorbertAmartus: I've tried to implement it at first
But the problem with that approach is that the blueprints list is more likely to be scrollable 

And having that in mind, the loader was working great ... untill the user'd like to scroll it's content 😀 

I'll attach a short video which would illustrate that situation 😉   

CPT2205120844-1448x703

test/cypress/integration/flows/user_spec.ts
NorbertAmartus: Sure, seems like a nice idea! :)

I have just one question about it
I've tried not to pick a specific blueprint to be uploaded because of the following reasons:
- blueprints list may change in the future (and the specific blueprint may not be available then)
- developer may have that specific blueprint already uploaded

To resolve this issue, I'm searching for the first upload button, which is available to be clicked, as shown below:
```typescript
cy.get(uploadBlueprintButtonSelector).first().click();
```

And this approach is stopping me from waiting for a specific request to be finished
I can always try to create an uniwersal url (with the usage of asterisks 😁) which could be used in the suggested way

But on the other hand, my concerns about the inaccessibility of a specific blueprint, may not be legitemate 😄 

WDYT? 
Should I:
1. stay with the timeout?
2. implement `waitUntilNotEmpty` but with an url that could match all of blueprints upload?
3. pick one blueprint and use `waitUntilNotEmpty` specificly for it?
4. some other solution 😄

widgets/blueprintActionButtons/src/BlueprintActionButtons.tsx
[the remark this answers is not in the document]
qooban: > Btw, having this occurrence in mind, we may consider adding a proper information to our [UI Code Style guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/Cloudify+UI+Code+Style) or [Ongoing refactors](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI), depending on the future of this approach, that we may have in mind tada WDYT? thinking

I think it's already there 🙂 See "Base" section at the top of [UI Code Style](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/Cloudify+UI+Code+Style).

widgets/blueprintCatalog/src/widget.tsx
qooban: Oh, that's a pity. If there's no other way to workaround it, then leave it for now.

app/translations/en.json
qooban: > WDYT about the mentioned approaches? thinking
Cool! I like the last solution the most 👍🏻 

> WDYT about that kind of adjustment to our codebase? thinking
Oh, I have just realized that `LoadingOverlay` is taking string and not component as message, so I guess we can't easily apply any changes here? 

So... Yes, that's something for another ticket. Feel free to create it.

test/cypress/integration/flows/user_spec.ts
qooban: Hmm... Valid points, so maybe we can make kind of combination of both approaches. Cypress allows you to read text from the element - https://docs.cypress.io/faq/questions/using-cypress-faq#How-do-I-get-an-element-s-text-contents . If you read it you should be able to easily create URL to be awaited, am I correct?

widgets/blueprintCatalog/src/consts.ts
qooban: Oh, thx for explanation. So it seems more like "blueprint Id from the catalog/marketplace".
I guess it was decided to put it into the context not to allow selecting more than one blueprint across different tabs.
If it was like, let's keep it as is. 

Optional: 

If you find some time, you can refactor it to be more descriptive.

widgets/blueprintActionButtons/src/BlueprintActionButtons.tsx
```diff
@@ -2,6 +2,13 @@ import type { ComponentProps } from 'react';
 
 const t = Stage.Utils.getT('widgets.blueprintActionButtons.buttons');
 
+const availableModalType = {
+    DEPLOY: 'deploy',
+    DELETE: 'delete'
+} as const;
+
+type availableModalType = typeof availableModalType[keyof typeof availableModalType];
```
qooban, line 10: Optional:

You can also just use:
```
type ModalType = 'deploy' | 'delete';
```

and then use just strings 🙂

NorbertAmartus: After the need of renaming the shown const it may be a more optimal option 😁 
I'll modify that right away 😉 

Thanks for the suggestion! 🎉

app/translations/en.json
NorbertAmartus: >Cool! I like the last solution the most 👍🏻

Happy to hear 🎉 
In that case I'll implement it right away😉 

>Oh, I have just realized that LoadingOverlay is taking string and not component as message, so I guess we can't easily apply any changes here?

We can by passing passing a react component typed as an `any` 😁 

> So... Yes, that's something for another ticket. Feel free to create it.
I'll do just that 🎉 
In addition, I'll leave a comment to remove the `any` type once the ticket would be resolved (I'll mention this piece of code also inside of the ticket 🚀 )

NorbertAmartus: Modifications has been added 🚀

widgets/blueprintCatalog/src/consts.ts
NorbertAmartus: Sure, I'll modify it to be `SELECTED_BLUEPRINT_ID` as it reveals the intention behind that const 🤠

widgets/blueprintActionButtons/src/BlueprintActionButtons.tsx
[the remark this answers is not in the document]
NorbertAmartus: Holy guacamolle, I didn't knew earlier about the link between our code style guideliness and airbnb rules 😅 
There's always something new to be learned 😁 

Thanks for pointing that out! 🚀

widgets/blueprintCatalog/src/consts.ts
qooban: Isn't it less catalog ID, but more blueprint ID from the catalog as you've mentioned? 🤔

(we have to be careful not to mislead it with `blueprintId` which is already in the context, I guess, by e.g. selecting blueprint in Resource Filter widget)

Anyways, if it was like that it's optional change from my perspective.

NorbertAmartus: That's a good question 😅 
I was wondering about that and then I've seen [this line](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/blueprintCatalog/src/widget.tsx#L155) which was indicating that it's being named as "catalog" in that context

First thing that I've did, was to change the const name to `SELECTED_BLUEPRINT_ID`, but then I've seen this line while adjusting other parts of the application 

So WDYT?
It should be indeed a blueprint (linked code line may be somehow inconsistent with the context) or should it be catalog? 🤔 

I know that it's an optional change, but as it's just changing a const name (which I've already did 😄) I may go aswell for finishing that modification with a proper result 😁

test/cypress/integration/flows/user_spec.ts
NorbertAmartus: Oh, yeah, you're right! 🤠 

But as I'm thinking about it right now, it may lead to a cluttered code in our case 😅 
The reasoning for that is we'd need to do the following steps:
1. Search for the non blocked upload button
2. Go through the parent components till we'd find the row element (containing the specific blueprint)
3. Use that row element to find both blueprint name and upload button (once more 😢)
4. Store blueprint name
4. Click button 
5. Use suggested `waitUntilNotEmpty` with the stored blueprint name

Because of the code which seems to be cluttered (even from the description 😅) I think that it may not be the right direction
WDYT? 🤔

qooban: Hmm... If you think so, then another option you can use is to get blueprint ID from the URL. There's "GET /console/sp/blueprints/<blueprint_id>" periodically called just after blueprint upload button is clicked. So possibly you can intercept it, extract the last portion of it and then await for blueprint to be uploaded using `waitUntilNotEmpty` with the extracted blueprint ID.

In general: The concern I have is that with the `timeout` set to 30 seconds may sometimes be exceeded - depending on the network connection quality (especially that we don't know which blueprint we are uploading).

<hr />

Anyways, I don't want to block this PR, so I'm leaving it to you to decide.

widgets/blueprintCatalog/src/consts.ts
qooban: > That's a good question sweat_smile I was wondering about that and then I've seen [this line](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/blueprintCatalog/src/widget.tsx#L155) which was indicating that it's being named as "catalog" in that context

I checked the history of that name and it seems it was wrong from the beginning (introduced 5 years ago).

> So WDYT? It should be indeed a blueprint (linked code line may be somehow inconsistent with the context) or should it be catalog? thinking

I think the name of the variable can be `selectedBlueprintId` when used inside Blueprints Catalog widget, but the context parameter (as it should be unique globally ) name should indicate it (as `blueprintId` is already used), maybe something like `catalogBlueprintId`?

I'm leaving it to you to decide.

NorbertAmartus: In that case I'll go with the `selectedBlueprintId` as we are currently having `uploadingBlueprint` context property, which has no explicit connection to the Blueprints Catalog widget

Thanks for the suggestions and for the discussion 💪

NorbertAmartus: Modifications has been applied! 🎉 

In addition, I've modified the instances of the `selectedCatalogId`, so that we wouldn't worry about their correctness in the future 😎

widgets/blueprintCatalog/src/utils.ts
NorbertAmartus: Okay, now I get it 😁 
Thanks for providing such a well detailed description!

In that case I'll be more than happy to encapsulate this logic! 🥳 

Thanks once more for the explaination 🚀

NorbertAmartus: `utils.ts`  file is also containing a helper function for getting the widget translation label 

So it still may be a valid file name 😁

app/translations/en.json
NorbertAmartus: The designer is still not responding and because of that Ilan decided, that for now we can put the following message to the code:

```typescript
Uploading '<blueprint name>' blueprint...
Once the blueprint is uploaded, you'll be redirected to the blueprint deployment page.
```

So I'll do just that 😉

widgets/blueprintCatalog/src/utils.ts
NorbertAmartus: Logic has been exported in a way that was shown on our meeting 🎉

test/cypress/integration/flows/user_spec.ts
NorbertAmartus: As we've been discussing this subject on the F2F, I've decided to go with the `timeout` approach
But I've not wanted to ignore the concern about exceeding the 30s timeout timespan, so I've adjusted it to mirror the time of waiting implemented by default in `waitUntilNotEmpty` (which is currently 60s) 🥳

NorbertAmartus: And btw, thanks once more for having a discussions about the differences between those 2 approaches 🎉

widgets/blueprintCatalog/src/RepositoryList.tsx
```diff
@@ -72,15 +72,30 @@ export default class RepositoryList extends React.Component<RepositoryListProps,
...
         return toolbox.refresh(fetchParams);
     };
 
+    setUploadingBlueprint = (uploadingBlueprint: string) => {
+        const { toolbox } = this.props;
+
+        Utils.blueprintCatalogContext.setUploadingBlueprint(toolbox, uploadingBlueprint);
+    };
+
+    // TODO: Reconsider the necessity of this function wrapping
```
qooban, line 92: Do you want to leave this TODO here?

widgets/blueprintCatalog/src/UploadingMessage/UploadingMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+                (
+                    <>
+                        <MessageHeader>
+                            {t('uploading', {
+                                blueprintName
+                            })}
+                        </MessageHeader>
+                        <MessageDescription>{t('redirection')}</MessageDescription>
+                    </>
+                ) as any
```
qooban, line 24: Don't we get any errors in browser console when passing React node instead of string?

I guess we can have a ticket for updating `LoadingOverlay` to allow that and just reference a ticket here (`TODO(RD-xxxx): ...`).

qooban (dismissed):
Approved 👍🏻 

Added some comments (no blockers, they can be resolved in the next PRs/discussions).

widgets/blueprintCatalog/src/RepositoryList.tsx
NorbertAmartus: That's an eagle eye right here! 😁 
No, that one is redundant, nice catch! 😉

NorbertAmartus: TODO has been removed 🚀

widgets/blueprintCatalog/src/UploadingMessage/UploadingMessage.tsx
NorbertAmartus: We are not getting any error, because `<LoadingOverlay />` `message` prop enables us to pass down a children component (I've double checked that 🚀 )

So creating a separate ticket for that modification would be a great idea!
I'll create one right away
(a proper comment will be also provided 😉)

widgets/blueprintCatalog/src/UploadingMessage/UploadingMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+                (
+                    <>
+                        <MessageHeader>
+                            {t('uploading', {
+                                blueprintName
+                            })}
+                        </MessageHeader>
+                        <MessageDescription>{t('redirection')}</MessageDescription>
+                    </>
+                ) as any // TODO(RD-4934): Remove type assertion
```
qooban, line 24: Thanks for creating the ticket 👍🏻 

(remember to add a bit more description there 🙂)

qooban (approved):
Looks good 👍🏻

PR #1956, opened 2022-05-04
RD-2000 Enable consistent-type-imports eslint rule

kubama:
## Description
Enabled consistent-type-imports eslint rule by updating cloudify-ui-common version.

## Screenshots / Videos
N/A

## Checklist

## Tests
N/A

 

## Documentation
N/A


app/reducers/pageReducer.ts
```diff
@@ -6,19 +6,10 @@ import type { AnyAction, Reducer } from 'redux';
...
-    isWidgetsSection,
-    LayoutSection,
-    moveTab,
-    PageDefinition,
-    removeTab,
-    TabContent,
-    updateTab
-} from '../actions/page';
-import { createPagesMap, InsertPosition, PageGroup, PageMenuItem } from '../actions/pageMenu';
+import type { addTab, LayoutSection, moveTab, PageDefinition, removeTab, TabContent, updateTab } from '../actions/page';
```
qooban, line 9: This is interesting. Seems like functions returning object `as const` (e.g. `addTab`, `moveTab`) are treated as types.

app/reducers/managerReducer/licenseReducer.ts
```diff
@@ -1,7 +1,7 @@
 import type { Reducer } from 'redux';
 import * as types from '../../actions/types';
 import Auth from '../../utils/auth';
-import Consts from '../../utils/consts';
+import type Consts from '../../utils/consts';
```
qooban, line 4: This is interesting. Seems like auto-fixer is quite smart to detect in which context (JS or TS) imported file is used and then decides if to use `import` or `import type`.

qooban (approved):
Looks good! 

I'm assuming that all changes were made with auto-fixer. That's great we won't need to fix this kind of issues manually anymore 👍🏻

app/reducers/pageReducer.ts
kubama: The thing here is that those functions are used only as `typeof` operands.

app/reducers/managerReducer/licenseReducer.ts
kubama: Similarly here, it is used only as `typeof` operand.

app/reducers/pageReducer.ts
qooban: 👍🏻

app/reducers/managerReducer/licenseReducer.ts
qooban: 👍🏻

PR #1957, opened 2022-05-04
RD-4223, RD-4240, RD-4230, RD-4216: Input parameters, workflow parame…

jawamartus:
…ters, list type - tests

## Description

RD-4223, RD-4240, RD-4230, RD-4216
## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2647/)


widgets/common/src/inputs/InputField.tsx
```diff
@@ -44,7 +44,7 @@ function InputField({
     const getConstraintValue = getConstraintValueFunction(constraints);
     const validValues = getConstraintValue('valid_values');
     const componentType = type === 'list' && input.item_type ? input.item_type : type;
-    const multiple = type === 'list';
+    const multiple = !!(type === 'list' && input.item_type);
```
qooban, line 47: Seeing `type === 'list' && input.item_type` used few times in this file (L46, L47, L51), I think it would be good to extract it to a function and reuse.

widgets/common/src/inputs/fields/DeploymentIdInputField.tsx
```diff
@@ -5,25 +5,21 @@ export default function DeploymentIdInputField({
...
 }: DynamicDropdownInputFieldProps) {
     const fetchUrl = '/searches/deployments?_include=id,display_name';
 
     return (
         <DynamicDropdown
             name={name}
-            error={error}
             textFormatter={item =>
                 item.display_name && item.display_name !== item.id ? `${item.display_name} (${item.id})` : item.id
             }
```
qooban, line 17: I know we had a discussion about that. 

Wanted to double-check - are you still saying that we shouldn't use `Stage.Utils.formatDisplayName` here?

IIUC the only difference is:
```
        if (data.id === undefined) {
            log.error('id is undefined');

            return '';
        }
```
but that condition will never be met in this file as we are always fetching `id` and `display_name` (L10).

widgets/common/src/inputs/InputField.tsx
jawamartus: Done.

widgets/common/src/inputs/fields/DeploymentIdInputField.tsx
jawamartus: Done.

qooban (approved):
Looks good!

PR #1958, opened 2022-05-04
RD-4223, RD-4240, RD-4230, RD-4216: Cypress update

jawamartus:
## Description

This cypress version works faster and more stable on my local computer. It may also reduce our system tests execution time. Let's see.
## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2645/)
[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2648/) 2022-05-05


package.json
```diff
@@ -177,7 +177,7 @@
     "babel-jest": "^26.6.3",
     "babel-plugin-module-resolver": "^4.1.0",
     "chokidar": "^2.1.2",
-    "cypress": "^8.2.0",
+    "cypress": "^8.7.0",
```
qooban, line 180: Have you tried upgrading directly to the latest version of Cypress (v9)?

jawamartus: No.

qooban: Two more points:

1. For the consistency I think it'd be good to have the same Cypress version in Stage and Composer. In that case it makes sense to have separate ticket for that upgrade. We could test at once upgrade to v9 within that ticket. In case it won't be easy to do, we can always stay with the latest v8, but we'll know what we need to do in the future.

2. Apart from Plugins Catalog spec (already known issue - RD-4906), I see that Deployments View spec is failing - did you check what was the reason for that failure?

jawamartus: > (...) For the consistency I think it'd be good to have the same Cypress version in Stage and Composer (...)
I think it's not worth the efforts and cons.

jawamartus: Ad 2.
Only plugins catalog spec is failing in the second test run.

jawamartus: What is exact reason for not approving this PR?

jawamartus:
Because of the performance issues with the newer versions of Cypress it's better to stay with 8.2 for now at Jenkins. 

https://github.com/cypress-io/cypress/issues/18740

PR #1960, opened 2022-05-05
RD-4254: Modifying breakpoints

jawamartus:
## Description

RD-4254

## Screenshots / Videos

image

## Checklist

## Tests

 

2022-05-06 [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2655/)

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2653/)


qooban (commented):
1. I see just a link to the ticket in "Description" section and no additional explanation about the solution. Despite the change is quite small, it's hard to say it's the right one. Can you elaborate more on that? 
2. Is "Expected result" from the ticket description met? I guess your solution provides a bit different result. Am I correct? I'm not saying it's bad or good, but explanation should be provided, I think.
3. There are some test failures, did you check nothing is related to that change?

jawamartus:
>  1.  I see just a link to the ticket in "Description" section and no additional explanation about the solution. Despite the change is quite small, it's hard to say it's the right one. Can you elaborate more on that?

Yes.

Assumptions:
1. We don't support mobile devices. So, we prefer to have well displaying widgets on narrower 
2. Our grid system with widgets is broken by design. In the page template file we always assume that we have 12 columns always.

Solution:
To change the number of columns and breakpoints to make our broken by design widget grid system working on desktop devices with narrowed screen.



>  2. Is "Expected result" from the ticket description met? I guess your solution provides a bit different result. Am I correct? I'm not saying it's bad or good, but explanation should be provided, I think.

Yes, it's not met. If we want to have this expected solution implemented, we need to modify whole grid system used for widgets and the API we are using in our templates files for defining position of every widget. It may require a lot of work, migrations etc. This solution makes the app working more as defined in the templates (more as expected) on narrow desktop windows.

If my assumptions are not correct, let me know.

>  3. There are some test failures, did you check nothing is related to that change?

Rerun in progress.

qooban (commented):
Thanks for explanations. I'll try to dig deeper into the topic the next week. We can discuss it together on Monday in the office.

In the meantime I see still some tests failing, are they related to the change?

jawamartus:
> In the meantime I see still some tests failing, are they related to the change?

Yes, I see it too. It's very likely that this is the case.

> Thanks for explanations. I'll try to dig deeper into the topic the next week. We can discuss it together on Monday in the office.

Yes, we can talk about broken design of our Grid system related to the widgets.

Maybe I'm wrong and it's not broken by design. However I can't see the point of setting x and y, width, height in that form in our page templates if we consider to use different amount of cols and breakpoints for our app.

qooban:
> However I can't see the point of setting x and y, width, height in that form in our page templates if we consider to use different amount of cols and breakpoints for our app.

Valid point. Seems like that's something to discuss together with our PMs to make expectations more clear first. Then we should set development path for fixing the current state. I'll try to schedule a meeting with our PMs.

jawamartus:
The solution was not enough and buggy.

PR #1961, opened 2022-05-06
RD-4857: Blueprint optional description field in TerraformModal

jawamartus:
## Description

RD-4857

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2713/) 2022-05-13 10:30

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/)

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2701/) - single test blueprints_spec
## Documentation

N/A


test/cypress/integration/widgets/blueprints_spec.ts
```diff
@@ -596,6 +596,23 @@ describe('Blueprints widget', () => {
...
+            cy.typeToFieldInput('Blueprint name', 'not_existing_blueprint_dpeloyment1234');
+            typeToTextarea('Blueprint description', 'Invalid string containing non ascii Łódź');
+            setTemplateDetails(singleModuleTerraformTemplateUrl, 'local');
+
+            cy.clickButton('Create');
+            cy.contains('Errors in the form').scrollIntoView();
+            cy.contains(`Please provide valid blueprint description`).should('be.visible');
+            typeToTextarea('Blueprint description', 'VALID ASCII STRING. \n!@#$%^&*()[]?\ts');
+            cy.contains(`Please provide valid blueprint description`).should('not.be.visible');
+        });
```
kubama, line 615: 1. To fit into existing test structure I think these checks should be done in `validate individual form fields` test case.
2. There is no check if the description is actually set on the created blueprint - I think `testBlueprintGeneration` function should be extended with such check.

jawamartus: Ad 1. It's separated feature therefore I think it should be also tested test case. It would be easier to catch. Also I think that the test case `validate individual form fields ` contains enough thing right now.

Ad 2. Yes, it's always the problem how detailed the system tests should be. More detailed = more time spend for testing. That's the problem related with fundamental decision made in our team: required system tests AND optional unit tests. It's much faster to cover all cases with unit tests. On the system tests level cases need to be multiplied to cover all cases.

I will cover this one case.

jawamartus: Ad 1.
Actually there are more cases which could be classified as `validate individual form fields ` which are outside of this described section. I prefer to keep them more flat.

jawamartus: @kubama Would you advice which widget I should use to test E2E displaying description? The current one used in the test spec doesn't display description.

kubama: > Would you advice which widget I should use to test E2E displaying description? The current one used in the test spec doesn't display description.

Looks like the only way to do it is to fetch the created blueprint from the manager. Just like you implemented created secrets checking in another PR.

jawamartus: Yes sir! :) 
Done

kubama (approved):

PR #1963, opened 2022-05-09
RD-4149 Remove individual requests for templates/page groups/pages content on application bootstrap

kubama:
## Description
Removed individual requests for templates/page groups/pages content on application bootstrap

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/2679/ - the failing tests are unstable on master as well, so I consider it a pass.

 

## Documentation
N/A


test/cypress/support/commands.ts
```diff
@@ -8,19 +8,17 @@
...
 import type { GlobPattern, RouteHandler, RouteMatcherOptions } from 'cypress/types/net-stubbing';
-import type { GetCypressChainableFromCommands } from 'cloudify-ui-common/cypress/support';
-import { addCommands } from 'cloudify-ui-common/cypress/support';
-import Consts from 'app/utils/consts';
-import emptyState from 'app/reducers/managerReducer/emptyState';
-import type { ManagerData } from 'app/reducers/managerReducer';
-import type { Mode } from 'backend/serverSettings';
-
-import { secondsToMs } from './resource_commons';
+import _, { isString, noop } from 'lodash';
```
NorbertAmartus, line 21: ## Optional
What do you think about extracting functions used in the following way: `_.<functionName>` to be explicitly imported?

Because for now we have both explicit and not-explicit approaches are applied to this file
As I can see, within this file, there are 3 metods used in the non-explicit way:
- `_.castArray`
- `_.compact`
- `_.map` (which could be replaced with the native `Array.prototype.map` function, as recommended in our [UI Code Style guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/UI+Code+Style#SHOULD%3A-Use-built-in-JS-functions-instead-of-utilities-functions-from-external-packages))

WDYT?

backend/handler/templates/TemplatesHandler.ts
```diff
@@ -54,15 +54,21 @@ function getTemplates(folder: string, isCustom: boolean, filter: (fileName: stri
                 const templateFileContent = fs.readJsonSync(templateFilePath);
                 const id = pathlib.basename(templateFile, '.json');
 
-                const name = _.get(templateFileContent, 'name', id) as string;
-                const updatedBy = _.get(templateFileContent, 'updatedBy', isCustom ? '' : 'Manager') as string;
-                const updatedAt = _.get(templateFileContent, 'updatedAt', '') as string;
-                const data = {
-                    roles: _.get(templateFileContent, 'roles', []),
-                    tenants: _.get(templateFileContent, 'tenants', [])
+                const { name = id, updatedBy = custom ? '' : 'Manager', updatedAt = '', ...data } = templateFileContent;
```
NorbertAmartus, line 57: Both in `PagesHandler.ts` and `TemplatesHandler` we are assigning `Manager` value to the `updatedBy` property
And here comes the question if that would be a good moment to extract it as a const, to contain a little description about is usecase (as a variable name 😄 )

It'd give us more context in the future and additionally, would make creating a functionality based on that approach, easier to implement 🚀   

WDYT?

NorbertAmartus (commented):
Nicely done! 🎉 
It's nice to start a day with a view of dozens code lines removed! 😁 

Left 2 minor comments 😉

test/cypress/support/commands.ts
kubama: Done

backend/handler/templates/TemplatesHandler.ts
kubama: Done

NorbertAmartus (approved):
Thanks for addressing all of the comments 🎉 
Looks great!

qooban (approved):

PR #1966, opened 2022-05-10
RD-4149 Enable widgets' README.md files lazy loading

kubama:
## Description
Enabled widgets' README.md files lazy loading

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/2681/ - the failing tests are unstable on master too, so I consider it a pass

 

## Documentation
N/A


app/components/shared/widgets/Widget.tsx
```diff
@@ -91,18 +117,32 @@ class Widget<Configuration> extends Component<WidgetProps<Configuration>, Widget
...
+                    } else {
+                        readmeContent = ':-(';
+                    }
+                    this.setState({ readmeContent });
+                }
+            );
         }
-        this.setState({ readmeContent, showReadmeModal: true });
+
+        this.setState({ showReadmeModal: true });
```
qooban, line 145: Optional:

I know that README.md will be fetched quite fast in most of the cases, but maybe for proper handling we should:
1. Set `readmeContent` to show loader (loading spinner)
2. Start fetching data
3. Remove loader once resource fetching is done and set it to show proper content

app/components/shared/widgets/Widget.tsx
```diff
@@ -163,7 +203,7 @@ class Widget<Configuration> extends Component<WidgetProps<Configuration>, Widget
                     <Icon name="help circle" size={widgetIconButtonSize} link />
                 </a>
             ) : (
-                widget.definition.readme && (
+                widget.definition.hasReadme && (
```
qooban, line 206: Good catch 👍🏻

qooban (dismissed):
Nicely done 👍🏻

app/components/shared/widgets/Widget.tsx
kubama: Applied, however since `ReadmeModal` accepts only text content I used 'Loading...' label instead of spinner, so no major rework was required.

app/translations/en.json
```diff
@@ -718,7 +718,9 @@
                     "cfy_spire_name": "Cloudify Spire",
                     "mgr_premium_title": "Cloudify Premium Manager",
                     "product_name": "Cloudify"
-                }
+                },
+                "noReadme": "No README available",
+                "loading": "Loading..."
```
qooban, line 723: You can possibly use HTML code here to have the same loading spinner as in other places (something like `<div class="ui ...">...` I guess), but that probably be quite hacky ;)

app/components/shared/widgets/Widget.tsx
qooban: Thanks!

You can possibly use HTML code here to have the same loading spinner as in other places (something like <div class="ui ...">... I guess), but that probably would be quite hacky ;)

kubama: I didn't think of that :slightly_smiling_face:. But since loading indicator in this case is unlikely to be noticed by users anyway I think we can stay with the non-hack solution.

PR #1971, opened 2022-05-13
RD-1051 Update restrictions on password changing

qooban:
## Description

The main idea behind this PR is to align password changing restrictions along Cloudify UI application. For more details and requirements see RD-1051.

List of main changes:
* introduced `identityProviders` parameter inside Redux store (replaced `isLdapEnabled` parameter)
* used `identityProviders` in sidebar's user menu and User Management widget (Change Password option visibility)
* used `identityProviders` in User Groups Management widget (LDAP options visibility)

## Screenshots / Videos
N/A

## Checklist

## Tests
1. Updated tests (unit, component, e2e).
2. Full system tests run [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2719/) - all passed.
3. Another full system tests run [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2728/) - failed one test (`servers_num_spec.ts`), executed locally and it passed, so I consider it as an one-time issue not related to this PR changes.

## Documentation
I don't see a place requiring update.

### DIFF
--- a/app/utils/shared/IdpUtils.ts
+++ b/app/utils/shared/IdpUtils.ts
@@ -1,0 +1,16 @@
+import _ from 'lodash';
+import type { ManagerData } from '../../reducers/managerReducer';
+
+export default class IdpUtils {
+    static isLocal(managerData: ManagerData) {
+        return managerData.auth.identityProviders === 'local';
+    }
+
+    static isLdap(managerData: ManagerData) {
+        return managerData.auth.identityProviders.includes('ldap');
+    }
+
+    static isOkta(managerData: ManagerData) {
+        return managerData.auth.identityProviders.includes('okta');
+    }
+}
--- a/app/utils/shared/IdpUtils.ts
+++ b/app/utils/shared/IdpUtils.ts
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import type { ManagerData } from '../../reducers/managerReducer';
 
 export default class IdpUtils {
--- a/app/utils/shared/IdpUtils.ts
+++ b/app/utils/shared/IdpUtils.ts
@@ -2,7 +2,8 @@ import type { ManagerData } from '../../reducers/managerReducer';
 
 export default class IdpUtils {
     static isLocal(managerData: ManagerData) {
-        return managerData.auth.identityProviders === 'local';
+        const { identityProviders } = managerData.auth;
+        return identityProviders.length === 1 && identityProviders[0] === 'local';
     }
 
     static isLdap(managerData: ManagerData) {
--- a/backend/handler/AuthHandler.ts
+++ b/backend/handler/AuthHandler.ts
@@ -33,10 +33,6 @@ export function getToken(basicAuth: string) {
     );
 }
 
-export function getTenants(token: string) {
-    return jsonRequest('GET', '/tenants?_get_all_results=true&_include=name', getHeadersWithAuthenticationToken(token));
-}
-
 export function getUser(token: string): Promise<UserResponse> {
     return jsonRequest('GET', '/user?_get_data=true', getHeadersWithAuthenticationToken(token));
 }
--- a/backend/routes/External.ts
+++ b/backend/routes/External.ts
@@ -17,7 +17,7 @@ function pipeRequest(
     url: string,
     queryString: any
 ) {
-    logger.debug(`Piping get request to url: ${url} with query string: ${queryString}`);
+    logger.debug('Piping get request to url:', url, 'with query string:', queryString);
 
     requestAndForwardResponse(url.startsWith('//') ? `http:${url}` : url, res, { params: queryString }).catch(err =>
         res.status(500).send({ message: err.message })
--- a/test/jest/utils/Manager.test.ts
+++ b/test/jest/utils/Manager.test.ts
@@ -21,6 +21,7 @@ describe('(Utils) Manager', () => {
                     }
                 },
                 state: 'loggedIn',
+                identityProviders: 'local',
                 error: null
             },
             tenants: {
--- a/test/jest/utils/Manager.test.ts
+++ b/test/jest/utils/Manager.test.ts
@@ -11,6 +11,7 @@ describe('(Utils) Manager', () => {
         manager: {
             ...emptyState,
             auth: {
+                ...emptyState.auth,
                 username: 'admin',
                 role: 'sys_admin',
                 groupSystemRoles: {},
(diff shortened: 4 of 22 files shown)


backend/handler/AuthHandler.ts
```diff
@@ -33,10 +33,6 @@ export function getToken(basicAuth: string) {
     );
 }
 
-export function getTenants(token: string) {
```
qooban, line 36: Not related change. Removed as it is not used anywhere.

backend/routes/External.ts
```diff
@@ -17,7 +17,7 @@ function pipeRequest(
     url: string,
     queryString: any
 ) {
-    logger.debug(`Piping get request to url: ${url} with query string: ${queryString}`);
+    logger.debug('Piping get request to url:', url, 'with query string:', queryString);
```
qooban, line 20: Not related change. As it was quick to fix it, did it at once.

widgets/userManagement/src/MenuAction.tsx
```diff
@@ -1,7 +1,7 @@
 // @ts-nocheck File not migrated fully to TS
```
qooban, line 1: I have considered migration, but decided to drop it after first try as few other files should also be updated IMHO.

app/utils/shared/IdpUtils.ts
```diff
@@ -0,0 +1,16 @@
+import _ from 'lodash';
+import type { ManagerData } from '../../reducers/managerReducer';
+
+export default class IdpUtils {
+    static isLocal(managerData: ManagerData) {
+        return managerData.auth.identityProviders === 'local';
+    }
+
+    static isLdap(managerData: ManagerData) {
+        return managerData.auth.identityProviders.includes('ldap');
```
qooban, line 10: Used `includes` here as it is possible to have multiple IDPs enabled, so we could have 'ldap,okta' as a response from `GET /idp`.

NorbertAmartus:
Today I won't made it in time to review the code
I'll do it on the Monday's morning 💪

app/utils/shared/IdpUtils.ts
```diff
@@ -0,0 +1,16 @@
+import _ from 'lodash';
```
NorbertAmartus, line 1: This import seems to not be used 😉 

```suggestion

```

backend/routes/External.ts
NorbertAmartus: Just out of curiosity, what does it fixing in this example? 🤔

widgets/userGroups/src/CreateModal.tsx
```diff
@@ -1,10 +1,14 @@
-// @ts-nocheck File not migrated fully to TS
```
NorbertAmartus, line 1: 🎉

test/jest/utils/Manager.test.ts
```diff
@@ -21,6 +21,7 @@ describe('(Utils) Manager', () => {
                     }
                 },
                 state: 'loggedIn',
+                identityProviders: 'local',
```
NorbertAmartus, line 24: My IDE is showing me that the phraze `local` (in the auth context) is being used across 4 files

Maybe it'd be beneficial to extract it as a const?

As I'm looking into the code, the same string value is being used in Terraform module URL context (please see [Terraform.test.ts](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/backend/test/routes/Terraform.test.ts#L61), [blueprints_spec.ts](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/test/cypress/integration/widgets/blueprints_spec.ts#L597)), so extracting auth `local` may also give us context of it's usage, making it less confusing across the project files 🥳 

WDYT?

NorbertAmartus (commented):
Nice piece of work! 🎉 
It's always nice to see some ts migration being done 🥳 

I've left few comments related to imports and strings without context

app/utils/shared/IdpUtils.ts
qooban: I'm wondering why it was not detected by ESLint. 
Anyways, good catch, fixed.

test/jest/utils/Manager.test.ts
qooban: I was considering that when I was implementing the solution. I don't like magic strings, but since we have TypeScript and in many cases we can have type safety, I stopped being very strict about that.

I added type for `identityProviders` to provide some documentation of possible values.
It's not very strict as any kind of authentication (any string) may be introduced by plugins (IIUC information gathered from Sandy). I know it's not fully safe, so WDYT?

backend/routes/External.ts
qooban: I realized that `queryString` can be an object. In the original form it was translated to "[object Object]" so it didn't provide useful information 🙂

test/jest/utils/Manager.test.ts
NorbertAmartus: I think that in those cases, when the plugins would be updated in a mentioned way, we may get a more value by having an extracted const file - if something would change, we would be able to modify everything in one place AND store sort of "contract" which ensures us that those const values are being processed by the system 🎉 

Adding a type for `identityProvider` is a nice idea 🎉
It brings a sort of similar benefits or using a const file, which would be e.g. documenting contract between us and the core team, which ensures us that declared values should be working flawlessly 🚀 

There is just one case, which I think that for now is not being covered by the provided type
In `IdpUtils.ts` there is a `isLocal` function, which has implemented the code below:

```typescript
return identityProviders.length === 1 && identityProviders[0] === 'local';
```

And by checking the available values for the `identityProviders` it shows that it's just an array of string, which is not increasing our typing safety in a way that we'd want to 😁 
I know that any string may be passed in any time, but having that requirement in mind, we may get a similar "typescript recommendation" across other application places, which is worth to have in mind

The solution for it would be to also use const, which'd provide the `identityProvider` values used from the UI point of view 😉 

But as we can do a "drilldown clicks" to see what are the possible values of the `identityProviders`, we have an ability to avoid this issue 😁 

Having said that, I think that the const may bring some additional value, but as a type is also a solution (which is good enough 🎉 ) I'm approving this PR and leaving the decision of implementing the const to you 😉

NorbertAmartus (approved):
Looks neat! 🔥 
Thanks for addressing all of the comments! 🎉

qooban:
@kubama Are you planning to review this PR?

test/jest/utils/Manager.test.ts
qooban: Thanks for explanations. I'll re-consider adding consts when working on RD-3817.

PR #1972, opened 2022-05-13
RD-4149 Add widgets lazy loading

kubama:
## Description
Added widgets lazy loading

## Screenshots / Videos
N/A - the change is transparent to the user with exception to widget selection modal which has an added standard loader

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/2758/
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/2759/ - filter widget spec rerun after fix

 

## Documentation
N/A


NorbertAmartus:
I won't be able to review it today, but I'll be glad to do it tomorrow morning 😉

app/utils/External.ts
```diff
@@ -80,7 +79,7 @@ export default class External {
 
         log.debug(`Uploading file for url: ${url}`);
 
-        return new Promise((resolve, reject) => {
```
NorbertAmartus, line 83: Do we need to explicitly declare `<any>` here? 🤔 

If I'm not mistaken, by default `Promise` is being typed as `Promise<unknown>`, so we are getting a similar capabilities, but in more secure way

app/containers/AddWidget.ts
```diff
@@ -1,16 +1,17 @@
...
+import stageUtils from '../utils/stageUtils';
 
 const mapStateToProps = (state, ownProps) => {
     const widgetDefinitions = state.widgetDefinitions.filter(definition => {
         return (
-            stageUtils.isUserAuthorized(definition.permission, state.manager) &&
-            stageUtils.isWidgetPermitted(definition.supportedEditions, state.manager)
+            !definition.loaded ||
+            (stageUtils.isUserAuthorized(definition.permission, state.manager) &&
+                stageUtils.isWidgetPermitted(definition.supportedEditions, state.manager))
```
NorbertAmartus, line 14: This condition seems to be a little cluttered
What would you think about the suggestion below? 🤔 

```suggestion
        const isLoadingDefinition = !definition.loaded;
        const isUserAuthorized = stageUtils.isUserAuthorized(definition.permission, state.manager);
        const isWidgetPermitted = stageUtils.isUserAuthorized(definition.permission, state.manager);

        return isLoadingDefinition || (isUserAuthorized && isWidgetPermitted);
```

app/components/AddWidgetModal.tsx
```diff
@@ -365,6 +383,7 @@ function AddWidgetModal({
                         <Grid.Column width={12}>
                             <WidgetListWrapper>
                                 <WidgetList divided>
+                                    {find(widgetDefinitions, { loaded: false }) && <LoadingOverlay />}
```
NorbertAmartus, line 386: Could we use here a native `Array.prototype.find` function?
(as suggested in our [UI Code Style guideline](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/UI+Code+Style#SHOULD%3A-Use-built-in-JS-functions-instead-of-utilities-functions-from-external-packages))

```suggestion
                                    {widgetDefinitions.find(widgetDefinition => !widgetDefinition.loaded) && (
                                        <LoadingOverlay />
                                    )}
```

app/components/shared/widgets/Widget.tsx
```diff
@@ -339,7 +351,8 @@ const mapStateToProps: MapStateToProps<ReduxStateToProps<any>, WidgetOwnProps<an
 
 const mapDispatchToProps = {
     setContextValue: setValue,
-    fetchWidgetData: fetchWidgetDataThunk
+    fetchWidgetData: fetchWidgetDataThunk,
+    updateWidgetDefinition: updateWidgetDefinitionThunk
 };
 
 const connector = connect(mapStateToProps, mapDispatchToProps);
```
NorbertAmartus, line 358: ## Just a note (not strictly related to this PR)

This component has over 300 lines of code and a lot of logic inside of it
Because of that, It may be a good candidate for some extracting/refactoring exercises 😁 

And just wanted to mention that in case that we'd have time / tasks in the future, which we could use to progress on that matter 😉

app/components/AddWidgetModal.tsx
```diff
@@ -1,16 +1,20 @@
 // @ts-nocheck File not migrated fully to TS
 
-import _ from 'lodash';
-import PropTypes from 'prop-types';
+import colors from 'cloudify-ui-common/styles/_colors.scss';
+import { LoadingOverlay } from 'cloudify-ui-components';
 import i18n from 'i18next';
-import styled from 'styled-components';
+import _, { find } from 'lodash';
```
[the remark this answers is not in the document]
kubama: Done

app/utils/External.ts
kubama: Without this change I was getting an error in `widgetDefinitionsLoader.ts:147`. With this change implicit casting works there, there is no need for explicit cast.

app/containers/AddWidget.ts
kubama: It would introduce a bug :slightly_smiling_face:. But I applied the idea.

app/components/AddWidgetModal.tsx
kubama: I recently introduced a bug by applying such change (`lodash` functions are null-safe as opposed to those from `Array` and  we are not 100% TS yet). So I'd prefer not to apply such changes blindly unless I'm 100% sure they are safe.

app/components/shared/widgets/Widget.tsx
kubama: :+1:

qooban (dismissed):
In general the code looks good to me, but I'm not able to go very deep into details today.

On the other hand I don't want to block this cool initiative of speeding up system tests and the whole app, so once you have all system tests green, just re-request the review.

It would be great to have some comparison measurements before/after all the changes you did within RD-4149 (not necessarily before merging).

app/utils/StageAPI.ts
```diff
@@ -113,8 +113,9 @@ interface StageWidgetConfigurationDefinition {
 }
 export type { StageWidgetConfigurationDefinition as WidgetConfigurationDefinition };
 
-interface CommonWidgetDefinition<Params, Data, Configuration> {
+export interface CommonWidgetDefinition<Params, Data, Configuration> {
     id: string;
+    loaded: boolean;
```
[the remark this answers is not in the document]
kubama: Looks like there was already such separation but I didn't put `loaded` there - fixed.

app/components/AddWidgetModal.tsx
NorbertAmartus: Okay, seems lika a fair point 👌 
Thanks for the explanation 💪

app/actions/page.ts
```diff
@@ -1,18 +1,19 @@
-import _ from 'lodash';
-import type { ThunkAction } from 'redux-thunk';
+import _, { find, map } from 'lodash';
```
NorbertAmartus, line 1: As we are started using not explicit imports (with `_`), we may try to migrate to the explicit ones 🎉 

Lodash methods used inside this file are:
- `each`
- `compact`
- `find`
- `map`

```suggestion
import { each, compact, find, map } from 'lodash';
```

WDYT? 🤔

kubama: Done

NorbertAmartus (approved):
Nicely done! 🚀 
Thanks for addressing all of the comments! 🎉

qooban (dismissed):
Very cool addition 👍🏻 

I'll just copy here the measurements you have shared in RD-4149 comment:

```
Measured improvements in app bootstrap time (local environment):
* empty cache
  * App bootstrap before: ~8s
  * App bootstrap now: ~4.2s for loader screen to disappear plus ~0.8s for initial widgets to load
* with cached files
  * App bootstrap before: ~6.5s
  * App bootstrap now: ~3.2s for loader screen to disappear plus ~0.7s for initial widgets to load
```

These numbers are amazing! 🎉

app/utils/StageAPI.ts
```diff
@@ -226,7 +228,9 @@ declare global {
 
         // NOTE: Common items are defined in widgets/common
         /** Common widget utilities */
-        namespace Common {}
+        // eslint-disable-next-line @typescript-eslint/no-empty-interface
+        interface Common {}
+        let Common: Common;
```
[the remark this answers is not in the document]
kubama: It was required to complete `app/utils/widgetDefinitionsLoader.ts` TS migration

qooban (approved):

NorbertAmartus (approved):

PR #1975, opened 2022-05-19
RD-4403: Creating secrets on terraform modal submission - hiding secr…

jawamartus:
…ets with asterisks

## Description

RD-4403

Covering missing requirements.

## Screenshots / Videos

image

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2760/)

## Documentation

N/A


widgets/common/src/terraformModal/TerraformVariableNameInput.tsx
```diff
@@ -6,17 +6,17 @@ import { inputMaxLength } from './TerraformModal';
 
 const t = Stage.Utils.getT('widgets.blueprints.terraformModal.variablesTable');
 
-interface TerraformVariableValueInputProps extends CustomConfigurationComponentProps<string> {
+interface TerraformVariableNameInputProps extends CustomConfigurationComponentProps<string> {
     rowValues?: Variable;
 }
 
-export default function TerraformVariableValueInput({
+export default function TerraformVariableNameInput({
```
jawamartus, line 13: Refactoring

qooban: Does this component show Terraform variable name input? It didn't show Terraform variable value input, so good that you spotted it, but it seems to me that it shows an input field to provide either secret name or input name. These are Cloudify resources' names, so maybe more accurate would be `CloudifyResourceNameInput`?

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -61,6 +61,24 @@ function LengthLimitedDynamicTableInput({ name, onChange, ...rest }: CustomConfi
     );
 }
 
+interface TerraformVariableValueInputProps extends CustomConfigurationComponentProps<string> {
+    rowValues?: Variable;
+}
+
+function TerraformVariableValueInput({ name, onChange, rowValues, ...rest }: TerraformVariableValueInputProps) {
```
qooban, line 68: Seeing that `LengthLimitedDynamicTableInput` and `TerraformVariableValueInput` components are quite similar (the only difference is in `Input.type`) did you consider having one common component for it?

widgets/common/src/terraformModal/TerraformVariableNameInput.tsx
jawamartus: The column is described as name in the design. The variables and environmental variables have:
`key`
`source`
`name`
`value`

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: The components are not presentational components but more like  logic components (google for the concepts if you are not familiar with it). I cannot see duplication of knowledge here. 
What is more LengthLimitedDynamicTableInput shouldn't be aware about rowValues object.

There is plenty things to improve in this file. It's definitely to big. I decided to keep it as it is for know. At least until we don't have specific agreement where we should place side-effects and business logic and how those should be separated from UI components. 

Also I don't think that making one common component for those fields is a good practice.

jawamartus: Answering your question. Yes, I did consider this.

jawamartus:
The system tests fails are not related to the changes in this PR.

widgets/common/src/terraformModal/TerraformVariableNameInput.tsx
qooban: IMHO:
1. the component name is kind of a shortcut for component's content, so it should as much as possible describe it
(what will developer think when he's reading `TerraformVariableNameInput`? A component providing an input field for Terraform variable name)
2. the context in which a component is used is less relevant and should not decide what is the component name (what if we would like to use the same component in two different contexts?)

Also, in the list you have presented (`key`, `source`, `name`, `value`), Terraform variable name is actually a `key`.

widgets/common/src/terraformModal/TerraformModal.tsx
qooban: OK.

qooban (approved):

PR #1977, opened 2022-05-23
Rd 2498

NorbertAmartus:
## Checklist

### DIFF
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -1681,6 +1681,19 @@
                 },
                 "error": "Couldn't create a Token, please try to submit the Token data again."
             }
+        },
+        "deploymentInfo": {
+            "name": "Deployment Info",
+            "description": "Shows deployment basic information and status",
+            "configuration": {
+                "showBlueprint": "Show blueprint name",
+                "showSite": "Show site name",
+                "showCreated": "Show created date",
+                "showUpdated": "Show updated date",
+                "showCreator": "Show creator",
+                "showNodeInstances": "Show node instances status"
+            },
+            "noDeploymentSelected": "No deployment selected"
         }
     },
     "input": {
--- a/widgets/deploymentInfo/src/widget.types.ts
+++ b/widgets/deploymentInfo/src/widget.types.ts
@@ -14,18 +14,17 @@ export declare namespace DeploymentInfoWidget {
     }
 
     export interface Data {
-        // TODO: Ensure that the typing is accurate, with the core team
         deployment: {
             id: string;
             display_name?: string;
+            description: string;
             visibility: string;
-            blueprint_id?: string;
-            description?: string;
-            site_name?: string;
-            created_at?: string;
-            updated_at?: string;
-            created_by?: string;
-            isUpdated?: boolean;
+            blueprint_id: string;
+            site_name: string;
+            created_at: string;
+            updated_at: string;
+            created_by: string;
+            isUpdated: boolean;
         };
         instancesCount: number;
         instancesStates: {
--- a/widgets/deploymentInfo/src/widget.types.ts
+++ b/widgets/deploymentInfo/src/widget.types.ts
@@ -1,0 +1,25 @@
+export declare namespace DeploymentInfoWidget {
+    export interface Configuration {
+        showExpiredTokens: boolean;
+        pollingTime: number;
+    }
+
+    export interface Data {
+        deployment: {
+            id: string;
+            display_name?: string;
+            visibility: string;
+            blueprint_id?: string;
+            description?: string;
+            site_name?: string;
+            created_at?: string;
+            updated_at?: string;
+            created_by?: string;
+            isUpdated?: boolean;
+        };
+        instancesCount: number;
+        instancesStates: {
+            [key: string]: number;
+        };
+    }
+}
--- a/widgets/deploymentInfo/src/widget.types.ts
+++ b/widgets/deploymentInfo/src/widget.types.ts
@@ -1,10 +1,20 @@
 export declare namespace DeploymentInfoWidget {
+    export interface Params {
+        deployment_id?: string | string[] | null;
+    }
+
     export interface Configuration {
-        showExpiredTokens: boolean;
         pollingTime: number;
+        showBlueprint: boolean;
+        showSite: boolean;
+        showCreated: boolean;
+        showUpdated: boolean;
+        showCreator: boolean;
+        showNodeInstances: boolean;
     }
 
     export interface Data {
+        // TODO: Ensure that the typing is accurate, with the core team
         deployment: {
             id: string;
             display_name?: string;
--- a/widgets/deploymentInfo/src/DeploymentInfo.tsx
+++ b/widgets/deploymentInfo/src/DeploymentInfo.tsx
@@ -1,18 +1,24 @@
-// @ts-nocheck File not migrated fully to TS
-export default function DeploymentInfo({ data, toolbox }) {
+import type { DeploymentInfoWidget } from './widget.types';
+
+const { ErrorMessage } = Stage.Basic;
+const DeploymentDetails = Stage.Common.Deployments.Details;
+const DeploymentActions = Stage.Common.Deployments.Actions;
+
+interface DeploymentsInfoProps {
+    data: DeploymentInfoWidget.Data;
+    toolbox: Stage.Types.Toolbox;
+}
+
+export default function DeploymentInfo({ data, toolbox }: DeploymentsInfoProps) {
     const { deployment, instancesCount, instancesStates } = data;
     const [visibilityError, setVisibilityError] = React.useState('');
 
-    const { ErrorMessage } = Stage.Basic;
-    const DeploymentDetails = Stage.Common.Deployments.Details;
-
-    const setVisibility = (id, visibility) => {
-        const DeploymentActions = Stage.Common.Deployments.Actions;
+    const setVisibility = (visibility: string) => {
         const actions = new DeploymentActions(toolbox);
 
         toolbox.loading(true);
         return actions
-            .doSetVisibility(id, visibility)
+            .doSetVisibility(deployment.id, visibility)
             .then(() => toolbox.refresh())
             .catch(err => setVisibilityError(err.message))
             .finally(() => toolbox.loading(false));
@@ -26,23 +32,8 @@ export default function DeploymentInfo({ data, toolbox }) {
                 deployment={deployment}
                 instancesCount={instancesCount}
                 instancesStates={instancesStates}
-                onSetVisibility={visibility => setVisibility(deployment.id, visibility)}
-                toolbox={toolbox}
+                onSetVisibility={setVisibility}
             />
         </div>
     );
 }
-
-DeploymentInfo.propTypes = {
-    data: PropTypes.shape({
-        deployment: PropTypes.shape({
-            id: PropTypes.string.isRequired
-        }).isRequired,
-        instancesCount: PropTypes.number.isRequired,
-        instancesStates: PropTypes.objectOf(PropTypes.number).isRequired
-    }).isRequired,
-    toolbox: PropTypes.shape({
-        loading: PropTypes.func,
-        refresh: PropTypes.func
-    }).isRequired
-};
--- a/widgets/deploymentInfo/src/consts.ts
+++ b/widgets/deploymentInfo/src/consts.ts
@@ -1,0 +1,3 @@
+export default {
+    WIDGET_ID: 'deploymentInfo'
+};
--- a/widgets/deploymentInfo/src/consts.ts
+++ b/widgets/deploymentInfo/src/consts.ts
@@ -1,3 +1,6 @@
+const WIDGET_ID = 'deploymentInfo';
+
 export default {
-    WIDGET_ID: 'deploymentInfo'
+    WIDGET_ID,
+    WIDGET_TRANSLATION_PATH: `widgets.${WIDGET_ID}`
 };
--- a/widgets/deploymentInfo/src/utils.ts
+++ b/widgets/deploymentInfo/src/utils.ts
@@ -1,0 +1,11 @@
+import Consts from './consts';
+
+const widgetTranslation = Stage.Utils.getT(Consts.WIDGET_TRANSLATION_PATH);
+
+// TODO: Add to utils a dedicated method for the purpose of getting the translation method for the specific widget
+
+const getWidgetTranslation = (translationPath = '') => Stage.Utils.composeT(widgetTranslation, translationPath);
+
+export default {
+    getWidgetTranslation
+};
--- a/widgets/deploymentInfo/src/widget.tsx
+++ b/widgets/deploymentInfo/src/widget.tsx
@@ -1,55 +1,58 @@
 import type { DeploymentInfoWidget } from './widget.types';
 import DeploymentInfo from './DeploymentInfo';
 import Consts from './consts';
+import Utils from './utils';
+
+const t = Utils.getWidgetTranslation();
 
 Stage.defineWidget<DeploymentInfoWidget.Params, DeploymentInfoWidget.Data, DeploymentInfoWidget.Configuration>({
     id: Consts.WIDGET_ID,
-    name: 'Deployment Info',
-    description: 'Shows deployment basic information and status',
+    name: t('name'),
+    description: t('description'),
     initialWidth: 16,
     initialHeight: 7,
     isReact: true,
     hasReadme: true,
     showHeader: false,
     showBorder: false,
-    permission: Stage.GenericConfig.WIDGET_PERMISSION('deploymentInfo'),
+    permission: Stage.GenericConfig.WIDGET_PERMISSION(Consts.WIDGET_ID),
     categories: [Stage.GenericConfig.CATEGORY.DEPLOYMENTS, Stage.GenericConfig.CATEGORY.CHARTS_AND_STATISTICS],
 
     initialConfiguration: [
         Stage.GenericConfig.POLLING_TIME_CONFIG(10),
         {
             id: 'showBlueprint',
-            name: 'Show blueprint name',
+            name: t('configuration.showBlueprint'),
             default: true,
             type: Stage.Basic.GenericField.BOOLEAN_TYPE
         },
         {
             id: 'showSite',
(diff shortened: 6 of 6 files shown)


app/translations/en.json
```diff
@@ -497,8 +497,12 @@
         "fetchError": "Error fetching widget data",
         "loadError": "Cannot load widget {{widgetName}}. It might not be installed in your env. Please contact administrator.",
         "loading": "Loading...",
-        "renderError": "Cannot render widget. Check browser console for details.",
-        "unexpectedError": "An unexpected error occurred",
+        "unexpectedError": "Cannot render widget. Check browser console for details.",
```
qooban, line 500: I see `unexpectedError` and `widget.unexpectedError` and they have different texts. Is it intentional?

app/translations/en.json
```diff
@@ -497,8 +497,12 @@
...
+        "fetchingError": "An unexpected error occurred",
+        "renderError": {
```
qooban, line 502: Why we need `unexpectedError` saying "Cannot render widget" and `renderError` saying "Could not render widget"?

1. Let's clarify if it is needed.
2. Let's use the same form - either "Cannot" or "Could not" - in both cases.

app/translations/en.json
```diff
@@ -497,8 +497,12 @@
         "fetchError": "Error fetching widget data",
         "loadError": "Cannot load widget {{widgetName}}. It might not be installed in your env. Please contact administrator.",
         "loading": "Loading...",
-        "renderError": "Cannot render widget. Check browser console for details.",
-        "unexpectedError": "An unexpected error occurred",
+        "unexpectedError": "Cannot render widget. Check browser console for details.",
+        "fetchingError": "An unexpected error occurred",
```
qooban, line 501: If it's fetching error shouldn't we say that in the message instead of "unexpected errror"?

app/components/shared/widgets/Widget.tsx
```diff
@@ -301,7 +301,7 @@ class Widget<Configuration> extends Component<WidgetProps<Configuration>, Widget
                         <ErrorMessage
                             autoHide={false}
                             error={i18n.t(
-                                'widget.renderError',
+                                'widget.unexpectedError',
                                 'Cannot render widget. Check browser console for details.'
```
qooban, line 305: Default text can be removed not to mislead developers.

It was temporarily introduced, IMHO we don't need it.

app/components/shared/index.ts
```diff
@@ -35,5 +36,6 @@ export {
     ProductLogo,
     SemanticIconDropdown,
     VerticallyAlignedCell,
+    ErrorPopup,
```
qooban, line 39: Let's put it after `DynamicTable` and before `ExecutionStatus`.

app/components/shared/ErrorPopup.tsx
```diff
@@ -0,0 +1,63 @@
...
+        &::before {
+            background-color: inherit;
+            box-shadow: var(--box-shadow-position) 0 0 ${errorMessageBorderColor};
+        }
+    }
+`;
+
+const HeaderWrapper = styled.div`
+    /* NOTE: In some cases the "x" icon is overflowing the header content. */
+    /* Note: Because of the '.ui.popup>.header' selector (coming from semanti-ui-react) specificy, !important flag seems like a valid option here */
```
qooban, line 33:
```suggestion
    /* Note: Because of the '.ui.popup>.header' selector (coming from semantic-ui-react) specificity, !important flag seems like a valid option here */
```

app/components/WidgetDynamicContent/ErrorMessageWithPopup.tsx
```diff
@@ -0,0 +1,47 @@
...
+                    <StyledMessage error>
+                        <Ellipsis content={messageContent} />
+                    </StyledMessage>
+                </div>
+            }
+        />
+    );
+};
+
+export default ErrorMessageWithPopup;
```
qooban, line 47: I suspect that the name may be misleading. We have `ErrorMessage` component and this one does not use it.
Error message is static (`'${widgetName}' widget`), so maybe it could be named a bit different to indicate the purpose - showing widget placeholder with an error message in the popup? Am I correct?

app/components/WidgetDynamicContent/Ellipsis.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+interface EllipsisProps {
+    content: string;
+}
+
+const Ellipsis = ({ content }: EllipsisProps) => {
+    return <StyledWrapper title={content}>{content}</StyledWrapper>;
+};
+
+export default Ellipsis;
```
qooban, line 18: The name may be a bit misleading. Actual it's not just ellipsis (which is multi-dot = "...") it's a text with ellipsis, so maybe `TextWithEllipsis`?

app/components/WidgetDynamicContent/WidgetDynamicContent.tsx
```diff
@@ -239,13 +241,17 @@ export default class WidgetDynamicContent extends Component {
 
     renderReact() {
         const { data, widget } = this.props;
+        const showErrorWithPopup = !widget.definition.showBorder;
```
qooban, line 244: Wouldn't it be more accurate to say:
```suggestion
        const showErrorInPopup = !widget.definition.showBorder;
```
?

app/components/WidgetDynamicContent/WidgetDynamicContent.tsx
```diff
@@ -254,13 +260,17 @@ export default class WidgetDynamicContent extends Component {
...
-                    <ErrorMessage
-                        error={i18n.t('widget.detailedRenderError', `Error rendering widget: {{errorDetails}}`, {
-                            errorDetails: e.message
-                        })}
-                        autoHide
+                    <WidgetErrorMessage
+                        widgetName={widget.definition.name}
+                        showErrorWithPopup={showErrorWithPopup}
+                        header={i18n.t('widget.renderError.title', 'Could not render widget')}
+                        content={i18n.t('widget.renderError.content', 'For more details see the browser console')}
```
qooban, line 273: Let's not use default values in `i18n.t` calls. It's misleading and hard to maintain.

app/components/WidgetDynamicContent/WidgetDynamicContent.tsx
```diff
@@ -239,13 +241,17 @@ export default class WidgetDynamicContent extends Component {
...
+
             return (
-                <ErrorMessage
-                    error={data.error}
-                    header={i18n.t('widget.unexpectedError', 'An unexpected error occurred')}
-                    autoHide
+                <WidgetErrorMessage
+                    widgetName={widget.definition.name}
+                    showErrorWithPopup={showErrorWithPopup}
+                    header={i18n.t('widget.fetchingError', 'Could not fetch the widget data')}
```
qooban, line 253: Let's not use default values in i18n.t calls. It's misleading and hard to maintain.

app/components/WidgetDynamicContent/Ellipsis.tsx
qooban, line 18: Isn't there similar component already in Tokens widget? Can we reuse it?

qooban (commented):
In general - important and well done change! This PR is quite big, so I did first quick review round and didn't get very deep into details and I didn't test the UI, not to prolong waiting for review. Most of my comments relates to naming and translations. 

Minor, general comment: I'd like us to have common way of naming directories. IMHO we should be consistent in that matter and as we didn't name directories starting with big letters in other places of the application (except `GettingStartedModal` and `ContactDetailsModal`) I don't think we should do that. Probably we should clarify that in our code style. WDYT?

Thanks for additional TS migrations and adding translation support in few places.

> In addition, while refactoring components, it seems like I've also delivered the scope of [RD-3351](https://cloudifysource.atlassian.net/browse/RD-3351)

Migrating `widgets/deploymentWizardButtons/src/WizardButton.tsx` can be seen as part of it, but I don't think it is fully covered. That user story requires deeper discussion with PMs do decide what to do with that widget. In general it's outdated, so simply migrating one file to TS does not solve it.

app/translations/en.json
NorbertAmartus: Label has been removed 🎉

app/translations/en.json
NorbertAmartus: [It was fixed](https://github.com/cloudify-cosmo/cloudify-stage/pull/1977/commits/0d5f4f2f62a9a201d51a016123235fc9c3c1996c) as a part of resolving [this comment](https://github.com/cloudify-cosmo/cloudify-stage/pull/1977/files#r884482920) 🚀

app/components/WidgetDynamicContent/Ellipsis.tsx
NorbertAmartus: Sounds good to me 😉 
I'll implement it right away

NorbertAmartus: Component has been renamed 🎉

NorbertAmartus: Yeah, [there is](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/tokens/src/TextEllipsis.tsx)
But we can't just extract it to a specific place and reuse it, as the [one from the Tokens widget](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/tokens/src/TextEllipsis.tsx) needs to have a specified `max-width`

And as it'd require some structural changes, I'd prefer to leave it as a part of [RD-2255](https://cloudifysource.atlassian.net/browse/RD-2255) - I'll write a comment about the existence of this component (Some time ago I've put there [a comment](https://cloudifysource.atlassian.net/browse/RD-2255?focusedCommentId=75994) about the one from the [Tokens widget](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/tokens/src/TextEllipsis.tsx))

app/components/WidgetDynamicContent/ErrorMessageWithPopup.tsx
NorbertAmartus: Yeah, you're correct!
Hmm in that case we may think about adjusting it 🎉 

WDYT about: `WidgetErrorPlaceholder`, because it's a placeholder which would appear only when an error would occur?

Alternatively I've got `WidgetPlaceholderWithError` on my mind 😁

app/components/WidgetDynamicContent/WidgetDynamicContent.tsx
NorbertAmartus: They've been removed 🎉

app/components/shared/index.ts
NorbertAmartus: It has been done 🎸

app/components/WidgetDynamicContent/WidgetDynamicContent.tsx
qooban: Hmm... Wait, didn't we agree to show all errors in popup when we have borderless widgets and inside widget box just show "<widget-id> widget"?

app/components/shared/widgets/Widget.tsx
NorbertAmartus: It has been removed 🎉

widgets/deploymentInfo/src/widget.types.ts
```diff
@@ -0,0 +1,37 @@
...
+        deployment: {
+            id: string;
+            display_name?: string;
+            description: string;
+            visibility: string;
+            blueprint_id: string;
+            site_name: string;
+            created_at: string;
+            updated_at: string;
+            created_by: string;
```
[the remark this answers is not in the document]
NorbertAmartus: From what I was able to find, some parts of it are being also declared [here](https://github.com/cloudify-cosmo/cloudify-stage/blob/71e35843b689526439f54434bdaf435d8e84d743/widgets/common/src/deploymentsView/types.ts#L19-L38)

But there are not 1:1
There are some differences between them (even when it comes to typing)

So, as it would require from us more work towards merging those interfaces, rather than just sharing them, I think that we may omit this procedure 😁 
WDYT?

PS: I've checked the types above with Sandy, so they should be up to date 🎉

app/translations/en.json
NorbertAmartus: While cleaning the labels (due to the other comments) one of the labels value has been modified, so the issue may not appear anymore

Let me know if you think that we should still do some actions in this area 😉

app/components/WidgetDynamicContent/WidgetDynamicContent.tsx
NorbertAmartus: Oh yeah, you're right 😁 
In that case it's 100% valid 🎉 

I'll rename it here and in the `WidgetErrorMessage` component aswell 💪

NorbertAmartus: It has been renamed 🎉

app/components/WidgetDynamicContent/ErrorMessageWithPopup.tsx
qooban: `WidgetPlaceholderWithError` sounds good, maybe we can have even `WidgetPlaceholderWithErrorPopup`, but I'm leaving it to you to decide which on to choose..

app/translations/en.json
qooban: I still see in L499:
```
        "unexpectedError": "Cannot render widget. Check browser console for details.",
```
and in L501:
```
        "renderError": {
            "title": "Could not render widget",
            "content": "For more details see the browser console"
        },
```
so my previous comment seems still valid:

> 1. Let's clarify if it is needed.
> 2. Let's use the same form - either "Cannot" or "Could not" - in both cases.

app/components/shared/widgets/Widget.tsx
```diff
@@ -297,15 +293,7 @@ class Widget<Configuration> extends Component<WidgetProps<Configuration>, Widget
...
-                    {hasError && (
-                        <ErrorMessage
-                            autoHide={false}
-                            error={i18n.t(
-                                'widget.renderError',
-                                'Cannot render widget. Check browser console for details.'
-                            )}
-                        />
-                    )}
+                    {hasError && <ErrorMessage autoHide={false} error={i18n.t('widget.unexpectedError')} />}
```
qooban, line 296: Shouldn't we have here:
```suggestion
                    {hasError && <ErrorMessage autoHide={false} header={i18n.t('widget.renderError.title')} error={i18n.t('widget.renderError.content')} />}
```
? 
That way we may probably remove `widget.unexpectedError` key, am I right?

app/components/WidgetDynamicContent/TextEllipsis.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+interface TextEllipsisProps {
+    content: string;
+}
+
+const TextEllipsis = ({ content }: TextEllipsisProps) => {
+    return <StyledWrapper title={content}>{content}</StyledWrapper>;
+};
+
+export default TextEllipsis;
```
qooban, line 18: I'll continue here previous discussion:

> Yeah, [there is](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/tokens/src/TextEllipsis.tsx)
> But we can't just extract it to a specific place and reuse it, as the [one from the Tokens widget](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/tokens/src/TextEllipsis.tsx) needs to have a specified max-width

> And as it'd require some structural changes, I'd prefer to leave it as a part of [RD-2255](https://cloudifysource.atlassian.net/browse/RD-2255) - I'll write [a comment](https://cloudifysource.atlassian.net/browse/RD-2255?focusedCommentId=75994) about the existence of this component
> (Some time ago I've put there a comment about the one from the [Tokens widget](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/tokens/src/TextEllipsis.tsx))

OK. We have to be aware that this way we are intentionally introduce code duplication (= tech-debt), so it should be clearly stated in RD-2255 that it should be fixed there.

widgets/deploymentInfo/src/widget.types.ts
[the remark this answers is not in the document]
qooban: > From what I was able to find, some parts of it are being also declared [here](https://github.com/cloudify-cosmo/cloudify-stage/blob/71e35843b689526439f54434bdaf435d8e84d743/widgets/common/src/deploymentsView/types.ts#L19-L38)

Yes, this is the place which I meant in my comment.

OK, we can omit this, but as this kind of duplications may be hard to find later, could you create a ticket for deduplication? It can be added to RD-4541 epic. As an implementation suggestion - I think the best would be to have Deployments API resource defined as TS interface in one place (e.g. `widgets/common/deployments`), then in `widgets/common/src/deploymentsView/types.ts`  and `widgets/deploymentInfo/src/widget.types.ts` just reuse relevant parts of that interface. 

Should be quick to implement, so may also be a good candidate for tech-debt item example for Patrycja.

app/translations/en.json
NorbertAmartus: After a short slack discussion I'll implement it right away! 🚀

NorbertAmartus: Done 🎉

app/components/shared/widgets/Widget.tsx
NorbertAmartus: Yup
Necessary changes are coming right away 🚀

NorbertAmartus: Changes has been provided 🎉

widgets/deploymentInfo/src/widget.types.ts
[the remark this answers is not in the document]
NorbertAmartus: Seems to me like a great idea 💪 
I'll create a corresponding ticket shortly after merging this PR 😉

app/components/WidgetDynamicContent/TextEllipsis.tsx
NorbertAmartus: Okay, I'll add a proper note to the ticket about that 👌

qooban (approved):
Didn't manage to test it in the UI, but looking at the code it's good.

Important improvement from my perspective. Thanks for all the efforts 👍🏻

PR #1982, opened 2022-05-24
RD-4997, RD-4904 BlueprintCatalog buttons UI improvements

PatrycjaKubiczek:
## Description

- changed the structure of BlueprintCatalog component
- removed which seems to be unnecessary styles from widget.css file

## Screenshots / Videos

Before: 
86fd3fec-50ad-4900-bfa8-ae61c471bc50
26acd23e-ed3a-4063-9799-f3dc0da3021c

After:
Zrzut ekranu 2022-05-24 152628

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2806/)

## Documentation

N/A


widgets/blueprints/src/BlueprintsCatalog.tsx
```diff
@@ -218,7 +218,7 @@ export default function BlueprintsCatalog({
                 searchable
                 noDataMessage={noDataMessage}
             >
-                <Grid>{blueprintsRows}</Grid>
+                <Grid stackable>{blueprintsRows}</Grid>
```
qooban, line 221: Just a note - from my perspective it's fine to use stackable version of `Grid` component, but I'd suggest to double check that with Ilan or Alex.

widgets/blueprints/src/BlueprintsCatalog.tsx
```diff
@@ -135,54 +135,54 @@ export default function BlueprintsCatalog({
...
-                                    onClick={event => {
-                                        event.stopPropagation();
-                                        onCreateDeployment(item);
-                                    }}
-                                />
-
-                                {!manager.isCommunityEdition() && widget.configuration.showComposerOptions && (
+                    {BlueprintActions.isCompleted(item) && (
+                        <Grid style={{ marginTop: 'auto', paddingTop: '1rem' }}>
+                            <Grid.Row className="centered">
```
qooban, line 140: Can we use here:
```suggestion
                            <Grid.Row centered>
```
? If so, then it's a bit cleaner solution IMHO.

widgets/blueprints/src/BlueprintsCatalog.tsx
```diff
@@ -135,54 +135,54 @@ export default function BlueprintsCatalog({
...
+                                <Grid.Column className="centered aligned">
```
qooban, line 141: Similar comment - can we use here:
```suggestion
                                <Grid.Column centered>
```
? 

I'm not sure about `aligned` class. Seems to me like it is added only when `textAlign` prop is provided. Do we need it?

qooban (commented):
Looks good in general 👍🏻 Added some comments.

I tested the UI and looks much better than it was. Found small issue and I'm wondering would it be possible to fix it. See "Edit a copy..." button in this video:

https://user-images.githubusercontent.com/5202105/170218838-33a34529-ad04-49e2-9065-cf8d8ab05899.mp4

Regarding "Tests" section - I don't expect any issues, but to be 100% sure all is fine, let's trigger system tests. For speedy check, I think we can limit check to `blueprints_spec` only. @NorbertAmartus I guess you can help @PatrycjaKubiczek with it.

widgets/blueprints/src/BlueprintsCatalog.tsx
PatrycjaKubiczek: @qooban you're 100% correct, good catch, I'm gonna change it.

widgets/blueprints/src/BlueprintsCatalog.tsx
PatrycjaKubiczek: this is a bit tricky because if I want center buttons and I'm using "centered" on rows, Semantic UI gives Column "text-align: left" by default:
Zrzut ekranu 2022-05-25 111605 however I might found a bit cleaner solution without className something like this:

``
```suggestion
   <Grid.Row centered>
     <Grid.Column textAlign="center">
```

qooban: Looks good 👍🏻  Thx for checking that.

NorbertAmartus:
I won't be able to do a review today
I'll do it tomorrow morning 🎸

widgets/blueprints/src/widget.css
```diff
@@ -2,24 +2,30 @@
...
-
 .uploadBlueprintButton {
     z-index: 5;
     position: absolute !important;
     right: 12px;
 }
+
+.blueprintCatalog .button {
+    margin: 0.2rem;
+}
```
NorbertAmartus, line 19: With this approach we may be creating side effects in the future ... without our knowledge 😅 

What do I mean by that?
`.button` className is being automatically added to the `<Button />` component

Because of that, we may create some unexpected styling, while adding any button to the `BlueprintCatalog` component 😁 
(especially, when we wouldn't see any declaration inside the mentioned component, which would imply some sort of styling)

So, what do you think about providing more specific `className` to the blueprint catalog buttons, just to be more explicit and precise (both in component and in this stylesheet)? 🤔

PatrycjaKubiczek: good suggestion, thanks! added actionButtons to avoid side effects in the future :)

widgets/blueprints/src/BlueprintsCatalog.tsx
PatrycjaKubiczek: @qooban I messaged Ilan and waiting for their approval :)

PatrycjaKubiczek:
> Looks good in general 👍🏻 Added some comments.
> 
> I tested the UI and looks much better than it was. Found small issue and I'm wondering would it be possible to fix it. See "Edit a copy..." button in this video:
> 
>  Peek.2022-05-25.10-24.mp4 
> Regarding "Tests" section - I don't expect any issues, but to be 100% sure all is fine, let's trigger system tests. For speedy check, I think we can limit check to `blueprints_spec` only. @NorbertAmartus I guess you can help @PatrycjaKubiczek with it.

@qooban after some time... but I finally managed to get the green badge on tests on my changes! :) 🚀

NorbertAmartus (approved):
Looks good to me! 🎸 
Thanks for adressing the comments! 🎉

widgets/blueprints/src/BlueprintsCatalog.tsx
PatrycjaKubiczek: @qooban I've got approval from Ilan :)

qooban: Great to hear that 👍🏻

qooban (approved):
Thanks for addressing all comments. LGTM 👍🏻

PR #1983, opened 2022-05-27
RD-4909 Add migration for `BlueprintAdditions` table removal

qooban:
## Description

Created migration script which moves `BlueprintAdditions` table content into files:
  * for each row image is extracted (either directly from `image` field or fetched from the network from URL: `imageUrl`)
  * then it is saved as PNG to `userData/blueprint-icons/<blueprint-id>/icon.png`

I used [sharp](https://sharp.pixelplumbing.com/) library for image manipulation (considered also [jimp](https://github.com/oliver-moran/jimp) and [gm](https://github.com/aheckmann/gm) packages, but `sharp` looked the best - well maintained and most popular). All supported image types are listed at [here](https://sharp.pixelplumbing.com/#formats).

Next step is moving extracted `icon.png` to blueprint resources directories in Cloudify Manager. This will be done within RD-5061. 

Additional, not related changes:
  * upgraded `package-lock.json` file to v2 (NPM >= v7)
  * removed adding `ResourcesModel` to DB module

Snyk reported an issue with `sharp` library, but I don't think at this point there's better alternative. It's not reported by NPM audit and it's not reported as issue in `sharp` library, yet. I think we don't have to be worried about it right now.

## Screenshots / Videos
N/A

## Checklist

## Tests
Added DB migration test with additional verification.
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2837/) (01/06/2022) - failed, but on known issues, so I'm considering it as PASSED

## Documentation
Updated internal documentation.


backend/jest.config.js
```diff
@@ -18,7 +18,7 @@ try {
 
 module.exports = {
     collectCoverage: true,
-    collectCoverageFrom: [...changedFiles, 'config.ts', '!migrations/**', '!jest.config.js', '!migration.ts'],
+    collectCoverageFrom: [...changedFiles, 'config.ts', '!db/**', '!migrations/**', '!jest.config.js', '!migration.ts'],
```
qooban, line 21: I can be more specific here and exclude only `BlueprintAdditionsModel.ts`, `ResourcesModel.ts`, `ResourceTypes.ts` (which are now used only by migrations), but as we are mocking DB in all tests I don't think we are losing much by excluding this directory from coverage collection.

kubama: > as we are mocking DB in all tests I don't think we are losing much by excluding this directory from coverage collection

I don't think I understand this comment, actually this change effectively includes `db` dir in coverage calculation, right?

backend/migrations/20220527182401-6_4-RemoveBlueprintAdditionsTable.ts
```diff
@@ -0,0 +1,44 @@
...
+
+            logger.info(`Migrating icon for '${blueprintId}' blueprint...`);
+            try {
+                let imageBuffer: Buffer = image;
+                if (imageUrl) {
+                    logger.info(`Fetching icon from URL: ${imageUrl}...`);
+                    imageBuffer = (await axios.get(imageUrl, { responseType: 'arraybuffer' })).data;
+                }
+                await fs.promises.mkdir(iconPath, { recursive: true });
+                await sharp(imageBuffer).toFile(path.join(iconPath, iconFilename));
```
kubama, line 30: So we need this specialized lib only to save buffer content to file? But is this operation anyhow specific to an image? Isn't it something generic that can be performed for any kind of binary data? If yes, do we need this lib at all?

backend/test/migration.test.ts
```diff
@@ -56,6 +62,7 @@ describe('Migration script', () => {
 
                 throw e;
             }
+            expect(verifyMigration()).toEqual(true);
```
kubama, line 65: I think the test would be better if `verifyMigration` was called directly here and actual expectations were done inside it individually, in case of test failure it would be possible to spot the failing assertion immediately.

backend/migrations/20220527182401-6_4-RemoveBlueprintAdditionsTable.ts
qooban: It actually do a bit more, as we are not sure what is the format of the stored image, it allows to get the image buffer (all supported image formats are listed at [here](https://sharp.pixelplumbing.com/#formats)) and saves it as `icon.png` file (behind the scenes it converts the image to PNG).

backend/test/migration.test.ts
qooban: Agree, I had the same intention originally, but I failed to do so (due to some errors). I'll try to put more effort to fix it.

backend/jest.config.js
qooban: Hmm... Isn't `!` (exclamation mark) working as a negation here? Similarly to `migrations` directory, `jest.config.js` and `migration.ts` files - it's not included in coverage calculation, or am I missing something?

backend/test/migration.test.ts
qooban: Had to turn off some ESLint checks, but I managed to address the comment.

backend/migrations/20220527182401-6_4-RemoveBlueprintAdditionsTable.ts
kubama: I see, thanks for explanation.

backend/jest.config.js
kubama: Of course, my fault :slightly_smiling_face:.

kubama (approved):

PR #1990, opened 2022-06-01
RD-4844: Update all marketplaces to 6.4

PatrycjaKubiczek:
## Description

[RD-4844](https://cloudifysource.atlassian.net/browse/RD-4844)

- update URLs to point to 6.4.
- change the default number of tabs from 6 to 13 in `blueprintMarketplaceTabsConfig.ts`
- add custom style to `Tabs.tsx` to display all tabs inside modal

## Screenshots / Videos

PR-3
PR-2
PR-4

important note - currently, on 3 tabs (Docker, Ansible & Other) we could have a problem with access so it will looks like this, but it will be fixed:  
Zrzut ekranu 2022-06-01 134706

## Checklist

## Tests

 

https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2839/ - there are fails, but it seems like it doesn't come from the changes, the same fails are from the NightlyUI

## Documentation

n/a


qooban:
@PatrycjaKubiczek I won't be able to review it before my vacation, so I have removed myself and added @kubama 🙂 Good luck!

NorbertAmartus:
@kubama 
> I think having a scrollbar for tabs looks quite awkward. Did you consider extending the modal's width so the scrollbar wasn't needed, at lest in cases when horizontal screen resolution would allow it?

I just did a quick check on how it'd look on 1280px width screen (our minimum supported width):
image

It still fits, but barely 😅 
So having in mind that marketplace is lately being frequently updated, the universal solution may be a good option

Another thing, which may be against a pixel-adjusted solution, is that the `Modal` component max-width may be adjusted with the usage of the following prop:
```typescript
size?: 'mini' | 'tiny' | 'small' | 'large' | 'fullscreen'
``` 

So, as showned at the screenshot above, we'd need to go with the `fullscreen` option, which may not be ideal for all of the screen sizes

Another option would be to override `<Modal />` styles using `className` property with a very specific selectors / `!important` flag, which seems to not be an ideal solution 😅 

WDYT?
Maybe it should be discussed with Ilan? 🤔

widgets/common/src/blueprintMarketplace/blueprintMarketplaceTabsConfig.ts
```diff
@@ -1,7 +1,7 @@
 import { times } from 'lodash';
 
 const t = Stage.Utils.getT(`widgets.blueprints.configuration.marketplaceTabs`);
-const defaultNumberOfTabs = 6;
+const defaultNumberOfTabs = 13;
```
NorbertAmartus, line 4: There is [a neat option](https://www.i18next.com/translation-function/objects-and-arrays) to extract an `object`/`array` from the `i18n` 

We could try to use it here, just so we wouldn't need to remember about adjusting this number in the future 💪 

```suggestion
import { times, size } from 'lodash';

const t = Stage.Utils.getT(`widgets.blueprints.configuration.marketplaceTabs`);
const defaultNumberOfTabs = size(t('columns.defaultValues', { returnObjects: true }));
```

NorbertAmartus (changes requested):

PatrycjaKubiczek:
@NorbertAmartus I'm afraid it won't fit on 1280px resolution with fullscreen class on modal
tabs
I think I needed to add "Getting Started" column as well? 🤔 

beside it won't look good on larger screens  :( 
tabs-3

the option "large" doesn't fit neither
tabs-3

I will try some other options

kubama:
If we can't fit it then I would say we should consider a different approach. I think scroll is not the best option here, I would rather consider something like this:
image

PatrycjaKubiczek:
@kubama I tried quickly reproduce this solution in our product. I'm wondering if this 
tabs-work
it would probably also need extra work on adjusting editMode

I also tried some different approach - create a wrapper around tabs, imo it doesn't look perfect as well... :( but it's kinda bulletproof on a different resolutions? WDYT @NorbertAmartus @kubama ?
tabs-different

kubama:
@PatrycjaKubiczek Ah, yes, I didn't consider edit mode. Changes there would be definitely out of scope of this ticket.

I think your second suggestion (with second row of the tabs) is still better than having a scroll.

kubama:
> do you think that we should ping PM's about that kind of `<Tabs />` adjustment, so that they'd be aware of it?

Sure, lets let them now.

widgets/common/src/blueprintMarketplace/blueprintMarketplaceTabsConfig.ts
PatrycjaKubiczek: done 👌

PatrycjaKubiczek:
I update the CSS of tabs a bit, so it looks better but it would still look like a tab
Zrzut ekranu 2022-06-06 103607
Zrzut ekranu 2022-06-06 103623

NorbertAmartus:
I've just had a call with Ilan about the attached results and he said, that for the time being, it's prefect 🚀 
Just wanted to notify about it, as it's being consider a right direction 🎉

NorbertAmartus (dismissed):
Looks great 💪 
Thanks for addressing all of the comments and for putting extra work on adjusting `Tabs` styling! 🎸

kubama (approved):

NorbertAmartus (approved):
Looks good to me 🎉

PR #1993, opened 2022-06-06
RD-4327 Show tasks graph even when its name does not match workflow name

kubama:
## Description
Show tasks graph even when its name does not match workflow name.

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/2863/ - none of the failures seem to be related to the changes in the PR, so I consider it a pass

 

## Documentation
N/A

### DIFF
--- a/widgets/executions/src/backend.ts
+++ b/widgets/executions/src/backend.ts
@@ -184,7 +184,7 @@ export default r => {
             // Connecting all the operations into a graph
             // *IMPORTANT NOTE* - Retrying tasks depend on their previous failed task
             const edges: any[] = [];
-            _.map(operationsList, task => {
+            _.each(operationsList, task => {
                 if (task.parameters.current_retries > 0) {
                     allSubgraphs[task.id].labels[0].retry = task.parameters.current_retries;
                 }
@@ -192,7 +192,7 @@ export default r => {
                     allSubgraphs[task.id].width = 270;
                     allSubgraphs[task.id].height = 40;
                 }
-                _.map(operationsList, dependantTask => {
+                _.each(operationsList, dependantTask => {
                     const edge = {
                         id: '',
                         sources: [],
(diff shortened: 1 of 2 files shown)


widgets/executions/src/backend.ts
```diff
@@ -183,7 +183,7 @@ export default r => {
         const constructDependencies = (operationsList, allSubgraphs) => {
```
jawamartus, line 183: What do you think about refactoring it a little bit and to split it at two separated functions?

1. Returning altered allSubgraphs.
2. Returning edges without modifying allSubgraphs (using Array.reduce/_.reduce to do that without declaring edges variable). (with the following name for example: `function getEdges() ...` )

Some comments modification might be needed.

widgets/executions/src/backend.ts
```diff
@@ -410,15 +419,18 @@ export default r => {
             const params = { ...req.query };
             const headers = _.pick(req.headers, 'cookie', 'tenant');
 
-            const operationsList = [];
             helper.Manager.doGet(tasksGraphsFetchUrl, { params, headers })
```
jawamartus, line 422: Async/await might be use here as a syntactic sugar to make code more flat and easier to think about.

jawamartus (commented):
Looks good!

I left a few comments.

widgets/executions/src/backend.ts
kubama: I tried to do this but when I introduce a `catch` clause I'm getting the following mysterious error from vm2:
`SyntaxError: Use of internal vm2 state variable`
No clue what's wrong and how to fix it. I think such issues is another argument for using regular backend for internal widgets instead of widget backend feature. Anyway, I must leave it as it is for now.

NorbertAmartus (approved):

widgets/executions/src/backend.ts
jawamartus: *code comments

jawamartus (dismissed):

widgets/executions/src/backend.ts
```diff
@@ -183,7 +183,7 @@ export default r => {
         const constructDependencies = (operationsList, allSubgraphs) => {
             // Connecting all the operations into a graph
             // *IMPORTANT NOTE* - Retrying tasks depend on their previous failed task
-            allSubgraphs.edges = [];
+            const edges: any[] = [];
             _.map(operationsList, task => {
```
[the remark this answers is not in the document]
kubama: Done

PR #2004, opened 2022-06-13
RD-4629: Realignment of the form fields

NorbertAmartus:
## Description

Changes consists of:
- modifying form fields labels
- modifying the structure of the `formFields` object and extracting `ContactDetailsFormField` component to be able to easily adjust order and the allignment of the form fields
- adding explicit imports for the `lodash` functions

### Before
Before

### After
After

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2924/)

## Documentation

N/A


app/translations/en.json
```diff
@@ -30,26 +30,26 @@
         "form": {
             "fields": {
                 "firstName": {
-                    "label": "First name",
+                    "label": "First Name",
```
kubama, line 33: Why the change to uppercase 'N'?

NorbertAmartus: Please take a look at the screenshot attached to [RD-4629](https://cloudifysource.atlassian.net/browse/RD-4629)
Here we can see, that form contains a field with the `First Name` label, not `First name`

And as the ticket is about alligning labels / fields layout accordingly to the provided screenshot, I've put here an uppercase 'N'

kubama: I see no reason why it should be in upper case. Besides, why 'Phone **n**umber' but 'First **N**ame'? Personally I would consult it with ticket submitter, but since it's a minor thing and it's clearly stated in the ticket I have no other option than to approve :slightly_smiling_face:.

PR #2012, opened 2022-06-20
RD-4442: Add a link to get premium license from the License page

PatrycjaKubiczek:
## Description

[RD-4442](https://cloudifysource.atlassian.net/browse/RD-4442)
- create a CTA button "Get a license" on License Page

## Screenshots / Videos

Before
image-20220331-101410

After
Zrzut ekranu 2022-06-27 115142

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2952/)

after modification license_page.spec https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/2994/
## Documentation

n/a


app/components/LicensePage.tsx
```diff
@@ -278,6 +280,17 @@ export default class LicensePage extends Component {
                         </Grid.Column>
 
                         <Grid.Column textAlign="right" verticalAlign="middle">
+                            {status === 'no_license' && (
+                                <Button
+                                    content={i18n.t('licenseManagement.getLicense', 'Get a license')}
```
kubama, line 285: Looks like we now have a number of calls like `i18n.t('licenseManagement....')` so it would be beneficial to use `getT` from stage utils.

kubama, line 285: I think the default value argument is not necessary here.

kubama (commented):
I think the corresponding cy spec should be extended to cover the added functionality

app/components/LicensePage.tsx
```diff
@@ -278,6 +280,17 @@ export default class LicensePage extends Component {
                         </Grid.Column>
 
                         <Grid.Column textAlign="right" verticalAlign="middle">
+                            {status === 'no_license' && (
+                                <Button
```
qooban, line 249: Optional:

I know that Alex didn't respond in RD-4442 on my question about the mockup, but what I had in mind is something like:
image

For other cases - license expired / license active - shown at https://docs.cloudify.co/staging/dev/install_maintain/installation/manager-license/ we have a button in that place, so maybe it would be suitable for "Get a license" as well? What do you think?

app/components/LicensePage.tsx
```diff
@@ -5,23 +5,21 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
```
qooban, line 5: Optional:

Re. L1 - as we are trying to proceed with TS migration, could you check how difficult would it be to migrate this file to TS? It's fine to keep the file as is if the effort is big, but maybe there's a chance to migrate it at once.

app/components/LicensePage.tsx
PatrycjaKubiczek: okay, I think you're right :) it seems better to keep a consistent design across all the cases. I also noticed that our doc about the expired case is a bit outdated since the picture in the [docs](https://docs.cloudify.co/staging/dev/install_maintain/installation/manager-license/) shows [red theme version](https://docs.cloudify.co/staging/dev/images/ui/license/expired-license.png) but we have a brown version :D 
Zrzut ekranu 2022-06-24 102326

qooban: Good catch 🙂 Would it be possible for you to update the screenshot within this ticket or shall we have separate ticket? (it won't require syncing docs repo and stage repo, so it's just image update)

PatrycjaKubiczek: sure I will do it

app/components/LicensePage.tsx
```diff
@@ -37,35 +51,57 @@ LicenseSwitchButton.propTypes = {
...
+    status
+}: DescriptionMessageProps) {
+    const { redirectToPage } = StageUtils.Url;
+    const StyledMessageContent = styled(Message.Content)`
+        &&&& {
+            display: flex;
+            justify-content: center;
+            align-items: center;
+        }
+    `;
```
kubama, line 75: This should be defined on the top level, not as part of render function. Otherwise we are creating a new component on each render.

kubama:
> I think the corresponding cy spec should be extended to cover the added functionality

Looks like this was not yet addressed.

PatrycjaKubiczek:
> I think the corresponding cy spec should be extended to cover the added functionality

@kubama ah correct! Sorry, I missed this comment, I will change the cy spec

app/components/LicensePage.tsx
```diff
@@ -1,27 +1,41 @@
...
-import { Button, Form, FullScreenSegment, Grid, Header, Icon, Message, MessageContainer } from './basic';
+import StageUtils from '../utils/stageUtils';
+import { Button, FullScreenSegment, Grid, Header, Icon, Message, MessageContainer } from './basic';
 import CurrentLicense from './license/CurrentLicense';
 import EulaLink from './license/EulaLink';
 import UploadLicense from './license/UploadLicense';
 import SplashLoadingScreen from '../utils/SplashLoadingScreen';
+import type Manager from '../utils/Manager';
+
+type LicenseStatus = 'no_license' | 'expired_license' | 'active_license';
```
qooban, line 19: I think the same type is already defined in `app/reducers/managerReducer/licenseReducer.ts`. Can we use it here?

app/components/LicensePage.tsx
```diff
@@ -205,8 +235,10 @@ export default class LicensePage extends Component {
         this.setState({ error: null });
     }
 
-    onLicenseEdit(proxy, field) {
-        this.setState(Form.fieldNameValue(field));
+    onLicenseEdit(_proxy: any, field: Field) {
```
qooban, line 238: I guess we can use here cool TS feature to get type from a function argument:
```suggestion
    onLicenseEdit(_proxy: any, field: Parameters<typeof Stage.Basic.Form.fieldNameValue>[0]) {
```
and avoid defining `Field` type.

Ref.: https://stackoverflow.com/questions/51851677/how-to-get-argument-types-from-function-in-typescript

app/components/LicensePage.tsx
```diff
@@ -236,7 +268,7 @@ export default class LicensePage extends Component {
 
         return (
             <FullScreenSegment>
-                <HeaderBar>
+                <HeaderBar className={undefined}>
```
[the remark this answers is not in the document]
PatrycjaKubiczek: className is required on HeaderBar :( Zrzut ekranu 2022-06-24 133133

app/components/LicensePage.tsx
PatrycjaKubiczek: applied adjustment: https://github.com/cloudify-cosmo/cloudify-stage/pull/2012/commits/3f8a17a693b9233c677d5c097e00bca99f5e93a5

app/components/LicensePage.tsx
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2012/commits/3f8a17a693b9233c677d5c097e00bca99f5e93a5

app/components/LicensePage.tsx
```diff
@@ -1,27 +1,41 @@
...
+    name: string;
+    value: unknown;
+    type: string;
+    checked?: string;
+}
+
+interface LicenseSwitchButtonProps {
+    color: ButtonProps['color'];
+    isEditLicenseActive: boolean;
+    onClick: () => void;
```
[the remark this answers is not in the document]
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2012/commits/3f8a17a693b9233c677d5c097e00bca99f5e93a5

app/components/LicensePage.tsx
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2012/commits/3f8a17a693b9233c677d5c097e00bca99f5e93a5

app/components/LicensePage.tsx
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2012/commits/79e8cb8dfdc1c9ca8485a6e98aec5d783fba7d06

PatrycjaKubiczek:
@kubama @qooban do you have any suggestions for the structure of yaml file for an inactive license? I think I need that to change cy spec to test cases with no active license

qooban:
> @kubama @qooban do you have any suggestions for the structure of yaml file for an inactive license? I think I need that to change cy spec to test cases with no active license

@PatrycjaKubiczek To test "no active license" case I suggest to remove the license before the test. Quickly checked Clodify CLI/API and seems like you should be able to do so with `DELETE /api/v3.1/license`, so something like `cy.cfyRequest('/license', 'DELETE')` should prepare your test. 

Alternatively you can probably (didn't test it) mock the response from `GET /api/v3.1/license` to give you empty `items` field:
```
{"metadata": {"pagination": {"total": 0, "size": 1000, "offset": 0}, "filtered": null}, "items": []}
```

app/components/LicensePage.tsx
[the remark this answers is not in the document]
PatrycjaKubiczek: is it okay if I keep `className={undefined}`?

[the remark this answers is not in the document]
qooban: Yeah, it's minor thing, so we can keep it. 


Optional: 

Alternatively - **only if you find time to check the root cause** - you can omit adding `className` prop and add `@ts-ignore` annotation mentioning the reason in a description.

test/cypress/integration/license_spec.ts
```diff
@@ -122,5 +122,10 @@ describe('License Management', () => {
                 uploadLicense(license.file).then(() => verifyError(license.error));
             });
         });
+
+        it('shows inactive license error', () => {
```
qooban, line 126: Optional:

Maybe to be aligned with the other test case names
```suggestion
        it('handles inactive license error', () => {
```

test/cypress/integration/license_spec.ts
```diff
@@ -122,5 +122,10 @@ describe('License Management', () => {
...
+            cy.cfyRequest('/license', 'DELETE');
```
qooban, line 127: I'm afraid about test robustness. The flow seems like that we visit the license page in `beforeEach` and then we remove the license. Visiting page triggers `GET /license` call. 

If you are sure that `DELETE /license` will always be called before `GET /license`, then all fine, but let's be sure.

qooban (approved):
Thanks for addressing all the comments 👍🏻 
I've added two small ones about newly added system test.

test/cypress/integration/license_spec.ts
```diff
@@ -122,5 +122,10 @@ describe('License Management', () => {
...
+            cy.contains('Get a license');
```
kubama, line 128: I think we can add `.should('be.visible')` to be more precise and have an explicit assertion.

kubama (dismissed):

test/cypress/integration/license_spec.ts
PatrycjaKubiczek: I run the test a couple of times in local env and in Jenkins and it always passed, also checked the Network tab in Cypress 
Zrzut ekranu 2022-06-27 145809 
so it looks like `license` is called after tenants and users GET calls, as they're part of promises in here: https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/components/AuthRoutes.tsx

test/cypress/integration/license_spec.ts
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2012/commits/be9b0be74edf3262450eef15a03b5f8921357834

qooban (approved):
Looks good!

PR #2014, opened 2022-06-22
RD-5066 Use new Marketplace API for plugins catalog

qooban:
## Description
This PR changes default JSON path for plugins catalog. As Marketplace API for plugins catalog produces slightly different JSON response, some adjustments were necessary. See RD-5066 comments for more explanation.

Related refactorings:
* Migrated some files to TypeScript
* Extracted common utility functions for plugins catalog
* Extracted common types for plugin resource and plugins catalog

Solved some system tests issues:
* RD-4906
* RD-5101
* RD-5072 (only partially - plugin issue)
* RD-5100 (only partially - plugin issue)

## Screenshots / Videos
New set of plugins in the catalog:
image

## Checklist

## Tests
1. Updated Cypress tests.
2. Triggered [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2957/) - fails on known issues only, so I'm considering it as PASSED
3. Triggered [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2963/) after small refactoring - fails on known issues only, so I'm considering it as PASSED
4. Triggered [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2965/) after code review adjustments - in progress

## Documentation
No changes needed.


app/utils/shared/PluginUtils.ts
```diff
@@ -0,0 +1,19 @@
+import { find, last } from 'lodash';
+import type { PluginCatalogEntry } from '../../../backend/routes/Plugins.types';
+
+export default class PluginUtils {
+    static getYamlUrl(plugin: Pick<PluginCatalogEntry, 'yaml_urls'>) {
+        const latestDslVersion = 'cloudify_dsl_1_4';
+        const latestDslYamlUrl = find(plugin.yaml_urls, yaml => yaml.dsl_version === latestDslVersion)?.url;
+        const fallbackYamlUrl = last(plugin.yaml_urls)!.url;
```
qooban, line 8: This logic may change in the future. Currently (v6.4+) we are supposed to use `plugin.yaml` associated with `cloudify_dsl_1_4` version. There may be a change in the Marketplace API which may allow us to simplify this logic (no need to check DSL version).

app/components/GettingStartedModal/plugins/model.ts
```diff
@@ -1,31 +1,10 @@
...
-    distribution: string;
-    // eslint-disable-next-line camelcase
-    package_name: string;
-    // eslint-disable-next-line camelcase
-    package_version: string;
-};
-
+export type CatalogPluginResponse = PluginCatalogEntry;
+export type ManagerPluginResponse = PluginEntry;
 export type ManagerPluginsResponse = Stage.Types.PaginatedResponse<ManagerPluginResponse>;
```
kubama, line 10: Does it make sense to keep this file at all?

app/utils/shared/PluginUtils.ts
```diff
@@ -0,0 +1,19 @@
+import { find, last } from 'lodash';
+import type { PluginCatalogEntry } from '../../../backend/routes/Plugins.types';
+
+export default class PluginUtils {
```
kubama, line 4: What's the reason for wrapping `class`? Wouldn't it be simpler to directly export utility functions?

app/utils/shared/PluginUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+    static getYamlUrl(plugin: Pick<PluginCatalogEntry, 'yaml_urls'>) {
+        const latestDslVersion = 'cloudify_dsl_1_4';
+        const latestDslYamlUrl = find(plugin.yaml_urls, yaml => yaml.dsl_version === latestDslVersion)?.url;
```
kubama, line 7:
```suggestion
        const latestDslYamlUrl = find(plugin.yaml_urls, { dsl_version: latestDslVersion})?.url;
```

backend/routes/Plugins.types.ts
```diff
@@ -0,0 +1,30 @@
...
+    arch: string;
+    file_size: string;
+    release: string;
+    url: string;
+}
+
+export interface PluginYaml {
+    dsl_version: string;
+    url: string;
+}
```
kubama, line 13: Wouldn't it be more accurate to name these types `PluginWagonUrl` and `PluginYamlUrl`? With current naming this is quite surprising:
```
    wagon_urls: PluginWagon[];
    yaml_urls: PluginYaml[];
```

app/utils/stageUtils.ts
```diff
@@ -26,6 +27,8 @@ export default class StageUtils {
 
     static Json = JsonUtils;
 
+    static Plugin = PluginUtils;
```
kubama, line 30: I'm not sure if this is the best place for this. I would rather make it part of `Stage.Shared`. The reason is that these 'utils' are specific to a narrow part of the domain, while most others (for example `Json`, `Time` or `Url`) are indeed generic utils. I would say this also applies to existing `Execution` utils.

app/components/GettingStartedModal/plugins/model.ts
qooban: I was considering removal, but I didn't want to make too many changes at once.
Some types (e.g. `URLString`) are used not only inside of `app/components/GettingStartedModal/plugins`, so I think that bigger refactoring is needed here.

backend/routes/Plugins.types.ts
qooban: Yep, changed.

app/utils/shared/PluginUtils.ts
qooban: Applied.

app/utils/shared/PluginUtils.ts
qooban: I considered that, but decided to use the same approach as in `ExecutionUtils`. 
Probably you are right it'll be simpler with direct exports.

app/utils/stageUtils.ts
qooban: `Stage.Shared` contains only (at least right now) React components shared between framework and widgets, so I'm not sure it's better place. The idea when introducing it (IIRC) was to distinguish between components used only internally and not exposed officially (`Shared`) and components which are exposed officially (`Basic`).

`Stage.Utils` contains different utilities. I'm not sure we should be worried about Cloudify domain specificity as we have all nicely wrapped (`Stage.Utils.Plugin`, `Stage.Utils.Execution`, ...). We do the same with `Stage.Common`. Maybe the naming or the description about each group should be better.

What do you think?

kubama: It's a valid reasoning, I think we can leave it as is.

app/utils/shared/PluginUtils.ts
qooban: Changed to direct exports.

kubama (approved):

PR #2015, opened 2022-06-22
RD-3666, RD-5188: Use context based error messages

jawamartus:
## Description

RD-3666, RD-5188

## Screenshots / Videos

image

## Checklist

## Tests

 

new blueprint spec: [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2995/)

blueprint spec: [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/2978/)


jawamartus:
System tests verified. ESLint fix applied. Waiting for Code Review.

package.json
```diff
@@ -58,7 +58,7 @@
         "blueimp-md5": "^2.7.0",
         "cloudify-blueprint-topology": "^2.5.5",
         "cloudify-ui-common": "^7.2.0",
-        "cloudify-ui-components": "^2.14.0",
+        "cloudify-ui-components": "^2.15.0",
```
jawamartus, line 61: To make the CI/CD tests to be passing I need to merge this PR first: https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2234 

I will do it just before merging this PR. Otherwise the change in doc repository would block other contributors.

CC: @PatrycjaKubiczek

widgets/common/src/terraformModal/TerraformModalTableAccordion.tsx
```diff
@@ -11,20 +11,23 @@ export interface TerraformModalTableAccordionProps<T extends any[]> {
     title: string;
     columns: (Stage.Types.WidgetConfigurationDefinition & { id: keyof T[number] })[];
     toolbox: Stage.Types.WidgetlessToolbox;
+    idPrefix: string;
```
jawamartus, line 14: I think I would remove it and replace with some more elegant solution with another PR/iteration to this US.

Also I plan to place toolbox (one line above) in the context instead of passing it down through all components.

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -614,26 +639,31 @@ export default function TerraformModal({ onHide, toolbox }: { onHide: () => void
...
             <Modal.Header>
                 <TerraformLogo src={terraformLogo} size="mini" inline /> {t('header')}
             </Modal.Header>
 
             <Modal.Content>
-                <Form errors={errors} scrollToError onErrorsDismiss={clearErrors}>
-                    <Form.Field label={t(`blueprintName`)} required error={errors.blueprint}>
-                        <Form.Input value={blueprintName} onChange={setBlueprintName}>
+                <Form scrollToError>
+                    <Form.Field label={t(`blueprintName`)} required>
```
PatrycjaKubiczek, line 651: Shouldn't we add   `error={getFieldError('blueprintName')}` or some className to `Form.Field` as well? I think it's required have red text in input title while error occurs.

see example of design in [docs](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2351104024/Terraform+Wrappers#Input-field-errors): 
7dc6e44c-b00c-4a5d-9114-8a419c7cb05b

jawamartus: Good highlight!

Actually I realized that in some places we don't need additional Form.Field component.
And I removed it.

Why we don't need it?
It was twice in the component tree.
I suppose that on the beginning Form.Input wasn't generating own Form.Field subcomponent. Somehow the implementation of Form.Input had changed and now we have Form.Field twice in some place.

@qooban We might consider it as ongoing refactoring - getting rid of doubled Form.Field component in component tree wherever it's possible.

Red labels:
image

After the change that's how components tree looks like - we have Form.Field once:
image

jawamartus: BTW
About component tree. It looks like too much nested. I guess we can get rid of half of the intermediate components. For example many times we are using Context.Provider and Context.Consumer. We might consider getting rid of those Context data which might be places inside of redux store AND using data from redux instead.

qooban (commented):
Will review tomorrow.

widgets/common/src/terraformModal/TerraformModal.css
```diff
@@ -0,0 +1,7 @@
+#terraformModal .accordion .content .ui.table td {
+    vertical-align: top;
+}
+
+#terraformModal .error.field > .label {
+    z-index: 1;
+}
```
qooban, line 7: Showing error on top of the error seems to be used only for DynamicDropdown input fields in Variables, Env Variables, Outputs sections.

image

Wouldn't it look nicer and be consistent with other places (e.g. Terraform module folder shown in the screenshot) when we show it under the input and push down the other content?

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -305,122 +332,124 @@ export default function TerraformModal({ onHide, toolbox }: { onHide: () => void
...
-        clearErrors();
+        cleanFormErrors();
 
-        const formErrors: Record<string, string> = {};
+        let formErrors = false;
 
         function validateBlueprintName() {
             if (!blueprintName) {
-                formErrors.blueprint = tError('noBlueprintName');
+                formErrors = true;
```
qooban, line 341: Maybe instead of setting `formErrors` in many lines we could extend `useFormErrors` hook to provide a function checking if there's any error in the form? That would allow removing all `formErrors = true;` statements. WDYT?

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -614,30 +639,36 @@ export default function TerraformModal({ onHide, toolbox }: { onHide: () => void
     }
 
     return (
-        <Modal open onClose={onHide}>
+        <Modal open onClose={onHide} id="terraformModal">
```
qooban, line 642: Do we need `id`to be set here?

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -614,30 +639,36 @@ export default function TerraformModal({ onHide, toolbox }: { onHide: () => void
...
-                    </Form.Field>
-                    <Form.Field label={t(`blueprintDescription`)} error={errors.blueprintDescription}>
-                        <Form.TextArea
-                            name="blueprintDescription"
-                            value={blueprintDescription}
-                            onChange={setBlueprintDescription}
-                            rows={5}
-                        />
-                    </Form.Field>
+                <Form scrollToError>
```
qooban, line 650: Does `scrollToError` work here? 🤔 
Checked it quickly and it seems it doesn't.

widgets/common/src/terraformModal/TerraformModalTableAccordion.tsx
qooban: > I think I would remove it and replace with some more elegant solution with another PR/iteration to this US.

Just out of curiosity - in which direction are you planning to go with removing it?

qooban (commented):
Good piece of work! Added some comment.

I did also basic UI tests and in general I like how the errors are displayed. 

Found some issues worth to address:
1. There's no general information about the errors in the form (AFAIR we agreed to keep it). When user provides Blueprint name and there's another error related in the form in collapsed section, the user does not know why after clicking "Create" button nothing happens.
2. Similarly to what was spotted in Create Deployment modal (RD-5188), I think we should open accordion sections automatically if they contain erroneous fields.
3. On opening Terraform modal errors are shown (when previously it was closed with error in the form). I think they should be cleared every time we open the modal as we do not persist modal data (on each close user agree to "close the modal and discard provided Terraform data").

app/translations/en.json
```diff
@@ -969,23 +969,23 @@
                     "invalidBlueprintDescription": "Please provide valid blueprint description. It should contain only ASCII characters.",
                     "invalidTerraformTemplate": "Please provide valid Terraform template URL",
                     "blueprintNameInUse": "Blueprint '{{blueprintName}}' already exists",
-                    "variable": {
+                    "variables": {
```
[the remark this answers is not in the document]
jawamartus: To make it aligned with accordion sections identifiers in the code.

widgets/common/src/terraformModal/TerraformModal.css
jawamartus: Is it required in the scope of this ticket?

I wouldn't consider it as a bug.

It might required a lot of effort to achieve consistency here and I can't see the profits for the users worth this tremendous efforts.

If you know how to easily achieve it let me know.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: It might be hard to achieve with immutable redux store. Do you have any idea how can I achieve it?

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: >     1. There's no general information about the errors in the form (AFAIR we agreed to keep it). When user provides Blueprint name and there's another error related in the form in collapsed section, the user does not know why after clicking "Create" button nothing happens.

I removed it. We have `scrollToError`, so the form doesn't need additional message on the top. Soon I will push the fix for not working `scrollToError`. Thanks for spotting it.

> 
>     2. Similarly to what was spotted in Create Deployment modal ([RD-5188](https://cloudifysource.atlassian.net/browse/RD-5188)), I think we should open accordion sections automatically if they contain erroneous fields.
> 

It's different case. The user doesn't open the modal with fields in accordion which might be invalidated by default. Therefore it doesn't improve UX in any way in this case. If you have additional concerns about this case let's have a call.

>     3. On opening Terraform modal errors are shown (when previously it was closed with error in the form). I think they should be cleared every time we open the modal as we do not persist modal data (on each close user agree to "close the modal and discard provided Terraform data").

Fixed. Thanks for spotting it.

widgets/common/src/terraformModal/TerraformModalTableAccordion.tsx
jawamartus: > > I think I would remove it and replace with some more elegant solution with another PR/iteration to this US.
> 
> Just out of curiosity - in which direction are you planning to go with removing it?

I didn't decide yet. If I would decide I would show you in the next PR.

app/components/shared/DynamicTable.tsx
```diff
@@ -52,23 +52,21 @@ const DynamicTable: FunctionComponent<DynamicTableProps> = ({ name, value = [],
...
-                                        />
-                                    </Table.Cell>
-                                );
-                            })}
+                            .map(({ id, label, width, ...columnRest }) => (
+                                <Table.Cell key={id} width={width}>
+                                    <GenericField
+                                        label=""
+                                        key={id}
+                                        index={index}
```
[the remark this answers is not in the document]
jawamartus: Yes. As you can see I passed index as a prop here.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: I have found some idea how to make it more clean. However I wouldn't plan to do it in the scope of this PR.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: Yes.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: > Does scrollToError work here? 🤔
> Checked it quickly and it seems it doesn't.

Yes. I created separated ticket for that and I created a PR.
Ticket: https://cloudifysource.atlassian.net/browse/RD-5267
PR: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/138

jawamartus:
It's blocked by: https://cloudifysource.atlassian.net/browse/RD-5267 . However I'm open for any additional comments to make future process of getting approval faster.

widgets/common/src/terraformModal/TerraformModal.tsx
qooban: > @qooban We might consider it as ongoing refactoring - getting rid of doubled Form.Field component in component tree wherever it's possible.

@jawamartus Sounds reasonable. Could you add a point to UI weekly (I've just created new section for the next week meeting) to notify others and then we'll update ongoing refactors page?

> About component tree. It looks like too much nested. I guess we can get rid of half of the intermediate components. For example many times we are using Context.Provider and Context.Consumer. We might consider getting rid of those Context data which might be places inside of redux store AND using data from redux instead.

IIUC we are not passing these components explicitly to the tree. They come from using Redux `connect` function, so we already use the data from the Redux store. 

See as an example this component tree (generated on Dashboard page):
image
we can see `Context.Provider` near `Home`, `SystemServicesStatus`, `SystemStatusHeader` and `SystemStatusIcon` components. All of them are using `connect`.

Let me know if I'm wrong and you have an idea how to make the tree less nested.

app/translations/en.json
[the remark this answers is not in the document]
qooban: OK.

widgets/common/src/terraformModal/TerraformModal.tsx
qooban: Thanks for addressing some UI/UX issues.

Reg.
>>  2. Similarly to what was spotted in Create Deployment modal ([RD-5188](https://cloudifysource.atlassian.net/browse/RD-5188)), I think we should open accordion sections automatically if they contain erroneous fields.
> It's different case. The user doesn't open the modal with fields in accordion which might be invalidated by default. Therefore it doesn't improve UX in any way in this case. If you have additional concerns about this case let's have a call.

I think it is very similar case. User can expand/collapse any section anytime. See no errors displayed in the video below:

https://user-images.githubusercontent.com/5202105/175547991-2689bb4c-b947-4c8d-ae13-5090abe4fcbd.mp4

User needs to open the accordion by himself, so if possible I would open accordions with erroneous field automatically.

If hard to achieve, then of course let's leave it as is, it's not a blocking issue from my perspective, but I think that @MichaelShn should be notified.

widgets/common/src/terraformModal/TerraformModal.tsx
qooban: Another idea I guess is creating simple function inside `handleSubmit` and use it instead of `setFieldError`.
```
function setFormErrors(fieldName: string, message?: string) {
    setFieldError(fieldName, message);
    formErrors = true;
}
```
WDYT?

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: As I said It's not the same case.

> User can expand/collapse any section anytime. See no errors displayed in the video below:

You are right. But the probability that the user would end up in this scenario in Terraform Modal is dramatically smaller.

In the deployment modal the accordions might contain fields which are required before collapsing them. It's mean that It's higly probable that the user might be not aware about existence of fields in accordion. Also the accordion is closed by default.

In contrast the case of terraform modal the user the user is required to do two actions to encounter the error:
1. Need to collapse the accordion and add some input/output manually (therefore he is conscious about them).
2. He is required to fold up the accordion.
To encounter the mentioned scenario.
Which is IMHO not highly probable that the user would do that.

However if you still consider it as a very similar case AND it's blocking the approval I can fix it in the scope of this PR.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: I think that another not pure function with side effect would shadow the code more and make it harder to maintain in the future.

qooban: Shadow the code? It's a simple wrapper function which allows to do 2 operations in one function call 🙂 
Harder to maintain? There are 17 calls like `formErrors = true;`. When adding new error check you need to remember to add that line as well. With single function you don't need to know that, you just call `setFormErrors` and that's it, so IMHO it's easier to maintain.

Anyways, I'm not willing to continue that discussion. 
You decide if you leave it as is or change it.

widgets/common/src/terraformModal/TerraformModal.tsx
qooban: > However if you still consider it as a very similar case AND it's blocking the approval I can fix it in the scope of this PR.

As I said in the previous comment:

> (...) it's not a blocking issue from my perspective, but I think that @MichaelShn should be notified.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: Yes, indeed. It's much harder to understand the code when you are using a not pure function in another not pure function. Also the function would be redundant if I would get rid of formErrors variable at all which is quite easy to do by removing side effects from all validation related functions. However as I stated previously it wouldn't be in the scope of this PR.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/139 The related PR is in code review.

widgets/common/src/terraformModal/TerraformModal.css
jawamartus: Unified. In the css file. The more sophisticated solution require much bigger changes and a lot of testing in other repository.

widgets/common/src/terraformModal/TerraformModal.tsx
jawamartus: Fixed.

widgets/common/src/terraformModal/TerraformModal.css
qooban: Thanks. I'm resolving the discussion.

qooban (commented):
Thanks for addressing UI/UX issues.

I don't have more comments and I see no blockers. There's just `cloudify-ui-components-2.15.0.tgz` to be removed and `package.json` to be updated.

jawamartus:
> I don't have more comments and I see no blockers. There's just cloudify-ui-components-2.15.0.tgz to be removed and package.json to be updated.

package.json was up to date.

Now `cloudify-ui-components-2.15.0.tgz` has been removed.

qooban (approved):

PR #2025, opened 2022-06-29
RD-2255: Long resource names - UI enhancements

jawamartus:
## Checklist


widgets/tokens/src/TokensTable.tsx
```diff
@@ -44,32 +43,26 @@ const TokensTable = ({ data, toolbox }: TokensTableProps) => {
...
-                                <TextEllipsis content={Time.formatTimestamp(dataItem.last_used)} />
-                            </DataTable.Data>
-                            <DataTable.Data>
-                                <RemoveTokenButton token={dataItem} toolbox={toolbox} />
-                            </DataTable.Data>
-                        </DataTable.Row>
-                    );
-                })}
+                        )}
+                        <DataTable.Data>{Time.formatTimestamp(dataItem.expiration_date)}</DataTable.Data>
```
jawamartus, line 59: Date doesn't need Ellipsis wrapper. Removed.

widgets/tokens/src/TokensTable.tsx
```diff
@@ -44,32 +43,26 @@ const TokensTable = ({ data, toolbox }: TokensTableProps) => {
                 />
                 <DataTable.Column label={t('columns.lastUsed')} name={dataSortingKeys.lastUsed} width="156px" />
                 <DataTable.Column label="" width="48px" />
-                {data?.items?.map(dataItem => {
-                    return (
```
jawamartus, line 48: Return statement redundant. Shortened.

widgets/common/src/deploymentsView/table/columns.tsx
```diff
@@ -39,44 +39,40 @@ const partialDeploymentsViewColumnDefinitions: Record<
...
+            return <DeploymentStatusIcon status={deploymentStatus} />;
         },
         // NOTE: do not show the column label
         label: ''
     },
     id: {
         width: '20px',
-        render(deployment) {
-            // NOTE: Rule reports false positives (ref.: https://github.com/yannickcr/eslint-plugin-react/issues/3022)
-            // eslint-disable-next-line react/destructuring-assignment
```
jawamartus, line 54: Fixed. Now the code is in accordance with eslint rules.

qooban:
Adding @NorbertAmartus to reviewers list as he was mainly working on ellipsis-related stuff in stage.

app/components/TextEllipsis.tsx
```diff
@@ -0,0 +1,44 @@
...
+     * undefined or number of allowed line
+     */
+    multiline?: number;
+}
+
+const StyledDiv = styled.div<TextEllipsisProps>`
+    overflow: hidden;
+    text-overflow: ellipsis;
+    ${props =>
+        props.multiline
```
jawamartus, line 25: For multiline we are using -webkit-line-clamp css attribute.

app/components/WidgetDynamicContent/TextEllipsis.tsx
```diff
@@ -1,18 +0,0 @@
-import React from 'react';
```
jawamartus, line 1: removed cause not needed.

widgets/tokens/src/TextEllipsis.tsx
```diff
@@ -1,18 +0,0 @@
-import styled from 'styled-components';
```
jawamartus, line 1: Removed cause no needed. Combined into one reusable component with richer functionality.

NorbertAmartus:
I'll make a review tomorrow as today I won't made it in time

jawamartus:
System tests looks like failing because of other reasons.

qooban (commented):
I like the changes in general. Did not test the UI. Added few comments.

I like how you divided RD-2255 into subtasks 👍🏻 BTW are filters and users widget cases (listed in RD-5261) covered by this PR?

app/components/TextEllipsis.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * text value of max-width style property
+     */
+    maxWidth?: string;
+
+    /**
+     * undefined or number of allowed line
+     */
+    multiline?: number;
```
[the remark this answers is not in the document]
jawamartus: I prefer to have multiline prop here without default value. The comments are clear enough what values are acceptable. IMHO having required numeral prop here it's not enough informative that the behaviour is completely different for values 1 and 2.

app/components/TextEllipsis.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import styled from 'styled-components';
+
+interface TextEllipsisProps {
+    /**
+     * text content
+     */
+    children?: string | null;
```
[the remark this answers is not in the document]
jawamartus: It's not enough to allow undefined value here. It has been provide to keep backward-compatibility with one use case for one of previous two TextEllipsis component.

app/components/TextEllipsis.tsx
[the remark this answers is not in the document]
NorbertAmartus: I agree with @qooban, name `multiline` implies for me that the prop could be used as a `boolean`, which cannot happen.

So maybe a more descriptive names as:
- `linesCount` (suggested by @qooban )
- `displayedLinesCount`
- `maxDisplayedLines`
- `clampedLinesCount`

would be more adequate here 🤔

app/components/TextEllipsis.tsx
```diff
@@ -0,0 +1,44 @@
...
+    multiline?: number;
+}
+
+const StyledDiv = styled.div<TextEllipsisProps>`
+    overflow: hidden;
+    text-overflow: ellipsis;
+    ${props =>
+        props.multiline
+            ? `
+        display: -webkit-box;
```
NorbertAmartus: Is this rule required? 
Till this point, functionality was working fine without it 🤔

app/components/basic/index.ts
```diff
@@ -171,6 +172,7 @@ export {
     Step,
     Tab,
     Table,
+    TextEllipsis,
```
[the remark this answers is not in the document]
jawamartus: Moved to Shared components

app/components/Breadcrumbs.tsx
```diff
@@ -47,7 +47,7 @@ export default function Breadcrumbs({ isEditMode, onPageNameChange, onPageSelect
                 {historyPageList.map((page, pageIndex) => (
                     <React.Fragment key={page.id}>
                         <Breadcrumb.Section link onClick={() => onPageSelected(page, pagesList, pageIndex)}>
-                            {page.name}
+                            <TextEllipsis maxWidth="300px">{page.name}</TextEllipsis>
```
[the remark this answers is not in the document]
jawamartus: It was defined as absolute value in the existing TextEllipsis components. I followed existing pattern.

app/components/TextEllipsis.tsx
```diff
@@ -0,0 +1,44 @@
...
+        -webkit-line-clamp: ${props.multiline};
+        -webkit-box-orient: vertical;`
+            : `white-space: nowrap;`}
+    ${props =>
+        props.maxWidth &&
+        `
+        max-width: ${props.maxWidth}
+    `}
+`;
```
NorbertAmartus: What do you think about using`css` [helper function](https://styled-components.com/docs/api#css) for applying conditional css rules?

IMHO it makes a code more readable (especially looking at the first condition 😁)

```typescript
const StyledDiv = styled.div<TextEllipsisProps>`
    overflow: hidden;
    text-overflow: ellipsis;
    ${props =>
        props.multiline
            ? css`
                  display: -webkit-box;
                  -webkit-line-clamp: ${props.multiline};
                  -webkit-box-orient: vertical;
              `
            : css`
                  white-space: nowrap;
              `}
    ${props =>
        props.maxWidth &&
        css`
            max-width: ${props.maxWidth};
        `}
`;
```

widgets/tokens/src/TextEllipsis.tsx
```diff
@@ -1,18 +0,0 @@
...
-
-export const EllipsisWrapper = styled.div`
-    max-width: 300px;
```
jawamartus, line 4: @qooban The absolute value was present here before. However newly created component allow easy refactoring in the future without modifying the component itself.

app/components/TextEllipsis.tsx
```diff
@@ -0,0 +1,44 @@
...
+     */
+    maxWidth?: string;
+
+    /**
+     * undefined or number of allowed line
+     */
+    multiline?: number;
+}
+
+const StyledDiv = styled.div<TextEllipsisProps>`
```
NorbertAmartus: WDYT about renaming it to `StyledTextEllipsis` or `TextEllipsisWrapper`?
As it would give us more information about it purpose, than just `StyledDiv`?

May suggestions may not be ideal, but the general idea is to indicate that this component is responsible for applying the css ellipsis functionality

app/components/Breadcrumbs.tsx
```diff
@@ -61,6 +63,12 @@ export default function Breadcrumbs({ isEditMode, onPageNameChange, onPageSelect
                         enabled={isEditMode}
                         onChange={newName => onPageNameChange(editablePage, newName)}
                         inputSize="mini"
+                        style={{
+                            overflow: 'hidden',
+                            textOverflow: 'ellipsis',
+                            whiteSpace: 'nowrap',
+                            maxWidth: '450px'
+                        }}
```
NorbertAmartus, line 71: Can we use here `<TextEllipsis />` component? 🤔

app/components/Breadcrumbs.tsx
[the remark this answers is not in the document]
qooban: Got it.

app/components/basic/index.ts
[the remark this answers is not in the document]
qooban: OK, thanks.

app/components/TextEllipsis.tsx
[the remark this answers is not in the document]
qooban: OK, then let's leave it here.

widgets/tokens/src/TextEllipsis.tsx
qooban: Sure 👍🏻

app/components/TextEllipsis.tsx
jawamartus: It's local name. The StyledDiv is enough here.

app/components/TextEllipsis.tsx
jawamartus: css is not helper function for applying conditional css rules.

> A helper function to generate CSS from a template literal with interpolations. You need to use this if you return a template literal with functions inside an interpolation due to how tagged template literals work in JavaScript.

We don't use interpolation here. Do we?

Also, the text with css function is more indented and more complex. Not following KISS principle.

app/components/TextEllipsis.tsx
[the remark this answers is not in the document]
jawamartus: > Hmm... With the current prop definition it's also not clear that there's different behavioUr for values 1 and 2. Maybe we should state in the prop description that it is supposed to be used with values starting from 2? 🤔

It is. Because the it's optional field without default value.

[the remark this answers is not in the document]
jawamartus: > Minor addition to the current discussion - I think the description should be modified a bit:
> 
>     the last word should be plural ("number of allowed lines").
>    we don't need to write "undefined or " (other props - maxWidth and children - can also be undefined, so I guess it's natural to assume it can be undefined just by looking at the type definition)

Enhancements added.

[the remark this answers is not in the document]
jawamartus: > would be more adequate here

What is your argument?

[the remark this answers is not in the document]
jawamartus: By argument I mean an array of propositions, one of which - the conclusion - is supposed to follow from the others - the premises.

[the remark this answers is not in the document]
qooban: @jawamartus I guess that one argument for name change mentioned by @NorbertAmartus is that:

> name `multiline` implies for me that the prop could be used as a boolean

which is valid point IMHO.

app/components/TextEllipsis.tsx
jawamartus: Yes.
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp

jawamartus: IIRC We didn't have multiline with ellipsis previously implemented.

qooban (approved):
From my perspective it looks fine, so @jawamartus @NorbertAmartus let's finalize open discussions and get it done 🙂

app/components/TextEllipsis.tsx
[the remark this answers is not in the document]
qooban: @jawamartus Take a look at that video:

https://user-images.githubusercontent.com/5202105/176629349-d364842d-2869-48cd-a7de-c57222d260f6.mp4

Majority of occurrences of `multiline: ` string turns out to be somehow associated with boolean values 🙂 so it's not unusual that when looking at `multiline` parameter the first idea popping up in mind is that it might most probably be a boolean.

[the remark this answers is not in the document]
jawamartus: > Majority of occurrences of multiline: string turns out to be somehow associated with boolean values 🙂 so it's not unusual that when looking at multiline parameter the first idea popping up in mind is that it might most probably be a boolean.

It doesn't prove anything. In different cases the multiline argument is boolean type. It doesn't imply that we cannot use it as numeral type. Also inductive reasoning is risky here

Both names have some props and cons. Using multiline as optional prop gives information that setting it would change behaviour somehow. Also it saves us from making another property - one for switching between multiline mode and another for defining a number of lines. It's common practice in the programming to combine use cases like that into one variable/prop which is numerical and optional ( or ZERO imply false in some programming languages).

`clampedLinesCount` and other proposed names doesn't provide information that the mode (rendered result) would be much different when the value is different than one.
**WHAT IS IMPORTANTANT**, the proposed solution doesn't give the opportunity to use `-webkit-line-clamp` with value 1. It limits the component flexibility and functionality.

Keep in mind that not all browsers support multiline with ellipsis. Take a look at caniuse for `-webkit-line-clamp`. It would be good to clearly give the possibility for the user to enable it or disable.

Using the name like `clampedLinesCount` doesn't look for the component user that he can disable this multiline mode for the component.

Therefore I consider multiline in this case as a more informative and appropriate prop name.

app/components/TextEllipsis.tsx
NorbertAmartus: Thanks for the confirmation and explaination 💪

app/components/Breadcrumbs.tsx
NorbertAmartus: Why?

app/components/TextEllipsis.tsx
NorbertAmartus: Yeah, but in this example the indentation is making the code more readable as the rules are being format with a similar structure

Looking at the following piece of code:
```typescript
        props.multiline
            ? `
        display: -webkit-box;
        -webkit-line-clamp: ${props.multiline};
        -webkit-box-orient: vertical;`
            : `white-space: nowrap;`}
```

We can see that the css rules used in both condition are being displayed differently (indendation is different, backticks (`) are being placed differently)

With the usage of `css` helper function we can omit that behavior, making it more readable

app/components/Breadcrumbs.tsx
jawamartus: Because we are using EditableLabel component which is not part of cloudify-stage repository and which is not flexible enough - it doesn't allow to pass the wrapping component.

app/components/TextEllipsis.tsx
jawamartus: > We can see that the css rules used in both condition are being displayed differently (indendation is different, backticks (`) are being placed differently)

I think it's because of prettier.

> With the usage of css helper function we can omit that behavior, making it more readable

> With the usage of css helper function we can omit that behavior, making it more readable

1. IIUC We don't need css to do it.
2. I think that the literal template containing two functions is less simple (KISS) than one containing one function.

app/components/Breadcrumbs.tsx
jawamartus: Do you see any possibility how we can use <TextEllipsis /> component without modifying other repositories?

PR #2026, opened 2022-06-29
RD-5071 Use filename from Content-Disposition header when downloading blueprint archives

qooban:
## Description
Download Blueprint button from Blueprint Action Buttons widget provides blueprint download functionality. The downloaded file was named always "<blueprint-id>.zip", so we assumed that the format of the file is always `zip`. That was not valid as there can be another archive formats used to store the blueprint archive (`tar`, `tar.gz`, `tar.bz2`).

This PR improves `External` class to support setting the filename from `Content-Disposition` header. It is provided in the response from `GET /blueprints/<blueprint-id>/archive` API call. It has the following format:
```
content-disposition: attachment; filename=test-blueprint.zip
```

During the development I spotted a problem in the core - RD-5284. Until it is fixed, we will still encounter issues with blueprint download (not only in the UI, but in CLI as well).

## Screenshots / Videos
N/A

## Checklist

## Tests
1. Updated Blueprint Action Buttons widget spec.
2. Full system tests run - [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3008/) - fails on known issues only, so I'm considering it as PASSED

## Documentation
N/A


app/utils/External.ts
```diff
@@ -32,6 +31,22 @@ function getContentType(type?: string) {
...
 
+function getFilenameFromHeaders(headers: Headers, fallbackFilename: string) {
+    let filename = fallbackFilename;
+    const contentDispositionHeader = headers.get('content-disposition');
+
+    if (contentDispositionHeader && contentDispositionHeader.indexOf('attachment') !== -1) {
+        // NOTE: Regex taken from https://stackoverflow.com/questions/23054475/javascript-regex-for-extracting-filename-from-content-disposition-header
+        const filenameRegex = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/;
+        const matches = filenameRegex.exec(contentDispositionHeader);
+        if (matches != null && matches[1]) {
```
kubama, line 42:
```suggestion
        if (matches?.[1]) {
```

app/utils/External.ts
```diff
@@ -32,6 +31,22 @@ function getContentType(type?: string) {
     return { 'content-type': type || 'application/json' };
 }
 
+function getFilenameFromHeaders(headers: Headers, fallbackFilename: string) {
+    let filename = fallbackFilename;
+    const contentDispositionHeader = headers.get('content-disposition');
+
+    if (contentDispositionHeader && contentDispositionHeader.indexOf('attachment') !== -1) {
+        // NOTE: Regex taken from https://stackoverflow.com/questions/23054475/javascript-regex-for-extracting-filename-from-content-disposition-header
+        const filenameRegex = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/;
```
kubama, line 40: I think this regexp can be moved to top level to prevent recreating it on each function call

app/utils/External.ts
```diff
@@ -32,6 +31,22 @@ function getContentType(type?: string) {
     return { 'content-type': type || 'application/json' };
 }
 
+function getFilenameFromHeaders(headers: Headers, fallbackFilename: string) {
+    let filename = fallbackFilename;
+    const contentDispositionHeader = headers.get('content-disposition');
+
+    if (contentDispositionHeader && contentDispositionHeader.indexOf('attachment') !== -1) {
```
kubama, line 38:
```suggestion
    if (contentDispositionHeader?.indexOf('attachment') >= 0) {
```

app/utils/External.ts
qooban: Done.

app/utils/External.ts
```diff
@@ -32,6 +31,22 @@ function getContentType(type?: string) {
     return { 'content-type': type || 'application/json' };
 }
 
+// NOTE: Regex taken from https://stackoverflow.com/questions/23054475/javascript-regex-for-extracting-filename-from-content-disposition-header
+const filenameRegex = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/;
+function getFilenameFromHeaders(headers: Headers, fallbackFilename: string) {
+    let filename = fallbackFilename;
+    const contentDispositionHeader = headers.get('content-disposition');
+
+    if (contentDispositionHeader && contentDispositionHeader?.indexOf('attachment') >= 0) {
```
qooban, line 40: @kubama Applied your suggestion partially. We cannot omit `contentDispositionHeader &&` as TS is complaing: "Object is possibly 'undefined'.".

kubama (approved):

app/utils/External.ts
```diff
@@ -32,6 +31,22 @@ function getContentType(type?: string) {
...
+        const matches = filenameRegex.exec(contentDispositionHeader);
+        if (matches?.[1]) {
+            filename = matches[1].replace(/['"]/g, '');
+        }
+    }
+
+    return filename;
+}
```
NorbertAmartus, line 48: WDYT about removing the let `filename` variable in order of having a `return` statements?

It may make the code more predictable and understandable 💪 

```suggestion
function getFilenameFromHeaders(headers: Headers, fallbackFilename: string) {
    const contentDispositionHeader = headers.get('content-disposition');

    if (contentDispositionHeader && contentDispositionHeader?.indexOf('attachment') >= 0) {
        const matches = filenameRegex.exec(contentDispositionHeader);
        if (matches?.[1]) {
            const filename = matches[1].replace(/['"]/g, '');
            return filename;
        }
    }

    return fallbackFilename;
}
```

qooban: OK, done.

NorbertAmartus (approved):

kubama (approved):

app/utils/External.ts
```diff
@@ -32,6 +31,22 @@ function getContentType(type?: string) {
     return { 'content-type': type || 'application/json' };
 }
 
+// NOTE: Regex taken from https://stackoverflow.com/questions/23054475/javascript-regex-for-extracting-filename-from-content-disposition-header
+const filenameRegex = /filename[^;=\n]*=((['"]).*?\2|[^;\n]*)/;
```
jawamartus, line 35: Making It case-sensitive might cause some problems in the future.

qooban: Could you elaborate more on that? Which cases you have in mind?

jawamartus: `filename` might be `Filename` or `FiLeNaMe`. IIUC all of them are valid. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#as_a_header_for_a_multipart_body

However if it would work there is very likely that it would be always filename with lowercases

qooban: Thx, noted.

PR #2033, opened 2022-07-01
RD-4254 Set minimal width for the main layout

qooban:
## Description

According to recently settled requirement: "Minimal supported screen width: 1280px" and decision to use horizontal scrollbar for smaller screens, I have adjusted minimal width for `div.main` element.

In addition to the main change I did the following:
* Used `ReactGridLayout` component with fixed number of columns (no need to use responsive grid with minimal width set)
* Fixed content of blueprint and plugin marketplace modals to always use full width of the modal (using `maximized` does not work, RD-5023 should make it possible to use `maximized` in layout definition JSON)

## Screenshots / Videos
https://user-images.githubusercontent.com/5202105/176841106-23ab05a9-b5cb-4ca9-a8e6-7bf1c0d2535d.mp4

## Checklist

## Tests
1. I did many manual tests - didn't spot any issues
2. [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3024/) - PASSED (one test failed - known issue - RD-5020)

## Documentation
I don't think we need any update right now.
There's RD-4998 to document supported screen size, which seems enough.

### DIFF
--- a/app/components/layout/Grid.tsx
+++ b/app/components/layout/Grid.tsx
@@ -1,13 +1,10 @@
 import _ from 'lodash';
-import PropTypes from 'prop-types';
-import type { CSSProperties, PropsWithChildren, ReactNode } from 'react';
+import type { CSSProperties, FunctionComponent, ReactNode } from 'react';
 import React from 'react';
 import type { ReactGridLayoutProps } from 'react-grid-layout';
-import { Responsive } from 'react-grid-layout';
+import ReactGridLayout from 'react-grid-layout';
 import { useWidthObserver } from '../../utils/hooks';
 import GridItem from './GridItem';
-
-const ReactGridLayout = Responsive;
 
 interface GridProps {
     isEditMode: boolean;
@@ -15,7 +12,7 @@ interface GridProps {
     style?: CSSProperties;
 }
 
-export default function Grid({ children, isEditMode, onGridDataChange, style }: PropsWithChildren<GridProps>) {
+const Grid: FunctionComponent<GridProps> = ({ children, isEditMode, onGridDataChange, style }) => {
     const [wrapperRef, getWidth] = useWidthObserver();
 
     const saveChangedItems: ReactGridLayoutProps['onLayoutChange'] = layout => {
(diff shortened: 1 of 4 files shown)


kubama (approved):

app/components/layout/Grid.tsx
```diff
@@ -66,8 +63,7 @@ export default function Grid({ children, isEditMode, onGridDataChange, style }:
             {width && (
                 <ReactGridLayout
                     className={['layout', isEditMode && 'isEditMode'].join(' ')}
-                    breakpoints={{ lg: 1000, md: 800, sm: 640, xs: 320, xxs: 0 }}
-                    cols={{ lg: 12, md: 10, sm: 8, xs: 6, xxs: 2 }}
```
NorbertAmartus, line 70: That removal looks like a dream 🎉

app/styles/style.scss
```diff
@@ -21,6 +21,7 @@ a:hover {
 }
 
 .main {
+  min-width: 1100px;
```
NorbertAmartus, line 24: Why is it set to `1100px` when the requirement is to support `1280px`? 🤔

qooban: We say that we support widths >= 1280px, but it doesn't mean that 1279px won't work 🙂 
Our app was still looking good for widths lower than `1280px`, so I thought that we can have some margin here and show scrollbar only for widths lower than `1100px`. 

If you have a better suggestion for the actual number, I'm open to discuss and change it.

NorbertAmartus: Oh, I see 😁 
In that case I may suggest to stay with the `1280px` min width restriction, as:
- It was a decision agreed on during the conversation with Illan
- We can still go with a lower width and some pages still would look quite decent, so I don't think that it may be a proper metric to take into the account
- Supporting only required width would make us sure, that the people which are not fully familar with the mentioned agreement (like new team members / team members that have forgot about it 😅) wouldn't spend time on adjusting the application view to the lowest **declared** application width
- In long term it may bring confusion as the documentation and the code wouldn't be consistent with eachother 

In conclusion, I totally agree that application may be looking good on a screen widths < `1280px`, but as we've agreed on not supporting those widths, it may be wise to not care about them, as it may require from us more work than it is required and expected.

WDYT? 🤔

qooban: The requirement is "Minimal supported screen width: 1280px", so it is about **screen** width and not about the browser width or viewport width or `div.main` element width. Setting `min-width: 1280px;` would actually brake our agreement IMHO, so I wouldn't mix those values.

The requirement itself is going to be documented as part of RD-4998 as stated in the description. 

If you still think we should change it, I see the following options:
* a comment in `styles.scss`, but I'm not sure we need it here (as it will be in the user docs).
* have `min-width: (1280 - $browserWindowMargin)px` or something similar

NorbertAmartus: Oh okay, in that case I'm fine with the change 😁 
Thanks for the clarification 🎉

NorbertAmartus (approved):

PR #2044, opened 2022-07-06
RD-5330 Finalize TS migration of Cypress system tests

qooban:
## Description
I've noticed that we are close to finalize TS migration of Cypress integration specs, so created to RD-5330 to finalize it. In total migrated 10 files in this PR.

As `cluster_status` related specs required more typing changes outside of the integration tests, I have created another PR for it - #2045 which extends this one.

## Screenshots / Videos
N/A

## Checklist
- [x] My code follows the [UI Code Style](https://cloudifysource.atlassian.net/l/c/x8fq902p).
- [x] I verified that all tests and checks have passed.
- [x] I verified if that change requires any update in the [documentation](https://cloudifysource.atlassian.net/l/c/4XKtPocy).
- [x] I added proper labels to this PR.

## Tests
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3083/) - Getting Started spec failed, so I'm considering it as PASSED (as I didn't change this spec as well as I didn't change any Cypress support commands)

## Documentation
N/A

PR #2045, opened 2022-07-07
RD-5330 Migrate Cluster Status components to TypeScript

qooban:
## Description
This PR extends https://github.com/cloudify-cosmo/cloudify-stage/pull/2044 with TypeScript migration for Cluster Status related files (in total: 7 files).

## Screenshots / Videos
N/A

## Checklist

## Tests
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3093/) - PASSED (Deployments widget spec failed, but it's a known issue - RD-5338)

## Documentation
N/A


NorbertAmartus:
I won't be able to do the review today
I'll do it tomorrow morning 💪

app/components/shared/cluster/ClusterServicesOverview.tsx
```diff
@@ -1,20 +1,40 @@
...
+}
+const ClusterServicesOverview: FunctionComponent<ClusterServicesOverviewProps> = ({
+    services = _.mapValues(clusterServiceEnum, () => ({
+        status: clusterServiceStatusEnum.Unknown as ClusterServiceStatus,
+        is_external: false
+    })),
+    clickable = false,
+    isFetching = false,
+    fetchingError = '',
+    header = null
```
NorbertAmartus, line 31: Looking at the the use cases of those props, it seems like they don't need default value assignment 🎉 

```suggestion
    clickable,
    isFetching,
    fetchingError,
    header
```

NorbertAmartus (commented):
Looks great! 🎉 
After merging both of your PRs our TS migration progress may hit 100% 😁 

I've just left one suggestion

app/components/shared/cluster/ClusterServicesOverview.tsx
qooban: Let me understand it better. Do you mean that e.g. if we don't have `clickable` passed it will be `undefined` and in the place where it is used L61:
```
{clickable && isSystemHealthPagePresent ? (
```
it will work the same as when `clickable` === `false`?

If so, then right, we can have it also `undefined` and it'll work, but are we losing anything by being a bit more descriptive here? 🤔

NorbertAmartus: Yeah, that was one of the reasons behind the idea 😁 

I don't think that there are some outstanding benefits which we could get due to romoving those default props assignments
But as they're not necessary (in that particular case) I thought that ommiting them would simplify code a little 🎉 

But other than that, I don't think that there is much of an additional benefit to it  😅 
Because of that, I'm leaving a decision up to you and as it seems to be a not required change, I'm approving this PR 🎉

NorbertAmartus (approved):

app/components/shared/cluster/ClusterServicesOverview.tsx
qooban: Thanks for clarification. In that case, I'll leave it as is and keep the original functionality.

PR #2046, opened 2022-07-07
RD-5199 mark only one blueprint as main

PatrycjaKubiczek:
## Description

RD-5199
- only one blueprint will have "Main" label
- I tried to migrate file to TS, any suggestions highly appreciated

## Screenshots / Videos

n/a

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3074/)

## Documentation

n/a


qooban:
@PatrycjaKubiczek I'm sorry I won't manage to review this PR today, I'll take it as the first one tomorrow morning.

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -58,11 +92,12 @@ export default function BlueprintSources({ data, toolbox, widget }) {
 
     const { CancelButton, NodesTree, Message, Label, Modal, HighlightText, ErrorMessage, Icon } = Stage.Basic;
 
-    const loop = (blueprintId, timestamp, items) => {
+    const loop = (blueprintId: string, timestamp: number, items: ItemProps[]) => {
         return items.map(item => {
             const key = `${blueprintId}/file/${timestamp}/${item.key}`;
             if (item.children) {
                 return (
+                    // @ts-ignore Missing the following properties from type 'TreeNodeProps': style, onSelect, icon, switcherIconts
```
NorbertAmartus, line 100: Issue is appearing due to the fact that the `NodeTree.Node` component is not yet migrated to TS

Because of that, we may just leave an information about it
With that approach we could omit listing all of the `NodeTree.Node` props 🎸 

WDYT about adjusting it (and other comments) to look in the following way:
```
@ts-ignore NodesTree.Node is not migrated to typescript
```

? 🤔

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -1,15 +1,46 @@
...
+
+interface BlueprintSourcesProps {
+    data: {
+        blueprintId: string;
+        blueprintTree: BlueprintTree;
+        importedBlueprintIds: string[];
+        importedBlueprintTrees: {
+            timestamp: number;
+            children: any;
+        };
```
qooban, line 20: I think we should define it using `BlueprintTree`:
```suggestion
        importedBlueprintTrees: BlueprintTree[];
```

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -1,15 +1,46 @@
...
+}
+
+interface ItemProps {
+    children: any;
+    key: string;
+    title: string;
+    isDir: boolean;
+}
+
+type HighlightTextProps = 'bash' | 'javascript' | 'json' | 'python' | 'yaml';
```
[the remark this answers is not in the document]
qooban: Agree with @NorbertAmartus , `FileType` seems more suitable here. 

Also we can somehow refer it to `HighlightText` component - and specifically it's `language` prop. Unfortunately it doesn't export props, but still we can provide proper reference I guess with use of `ComponentProps` generic type. 

It should be possible to have something like:

```suggestion
type FileType = ComponentProps<typeof HighlightText>['language'];
```
WDYT? (you need to have `ComponentProps` type imported from `react` and `HighlightText` imported from our basic components)

widgets/blueprintSources/src/BlueprintSources.tsx
qooban: Agree with @NorbertAmartus . In other files we use such convention, so it's good idea to adjust `@ts-ignore` comments in this file.

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -76,8 +111,9 @@ export default function BlueprintSources({ data, toolbox, widget }) {
                     </NodesTree.Node>
                 );
             }
+            const mainYamlFilePath = `${data.blueprintTree.children[0].key}/${data.yamlFileName}`;
```
qooban, line 114: I'm wondering if the main blueprint YAML file will always be in the root directory 🤔 

Probably in majority of cases it is true, so we should be good with that, at least for now.

qooban, line 114: I guess `data.blueprintTree.children[0].key` is the name of the root folder in the blueprint archive, right?
I think it'll make it a bit more clear if we have a const for that. WDYT about something like:
```suggestion
            const blueprintRootDirectory = data.blueprintTree.children[0].key;
            const mainYamlFilePath = `${blueprintRootDirectory}/${data.yamlFileName}`;
```
?

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -1,15 +1,46 @@
...
+interface ItemProps {
+    children: any;
+    key: string;
+    title: string;
+    isDir: boolean;
+}
+
+type HighlightTextProps = 'bash' | 'javascript' | 'json' | 'python' | 'yaml';
+
+export default function BlueprintSources({ data, toolbox, widget }: BlueprintSourcesProps) {
```
qooban, line 36: Optional:

Seeing that we use `data.*` prefixes in many lines, we could maybe destructure it at the top of the function and use directly object properties?

```suggestion
export default function BlueprintSources({ data, toolbox, widget }: BlueprintSourcesProps) {
    const { blueprintId, blueprintTree, importedBlueprintIds, importedBlueprintTrees, yamlFileName } = data;
```

It's optional as I didn't check if we won't have shadowing issues.

qooban (commented):
Tested UI - issue fixed, that's great 👍🏻 
Added some comments related to TS and code style.

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -1,15 +1,46 @@
...
+            children: any;
+        };
+        yamlFileName: string;
+    };
+    toolbox: Stage.Types.Toolbox;
+    widget: Stage.Types.Widget;
+}
+
+interface ItemProps {
+    children: any;
```
qooban, line 28: Similarly to https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/files#r916482566 you can possibly have here:
```suggestion
    children?: ItemProps[];
```

qooban (commented):
Added one more comment 🙂

widgets/blueprintSources/src/BlueprintSources.tsx
PatrycjaKubiczek: yeah, I was wondering the same, but tbh I didn't know how could I verify that :(

widgets/blueprintSources/src/BlueprintSources.tsx
PatrycjaKubiczek: I was wondering if we could create a ticket for migrating this component to TS? It looks rather small, but it could resolve the issue with ts-ignore in cloudify-stage and could improve migration check on ui-components :D

widgets/blueprintSources/src/BlueprintSources.tsx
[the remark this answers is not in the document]
PatrycjaKubiczek: thanks for the tip @qooban about ComponentProps I wasn't aware we could do that :) always happy to learn something new! addressed it here: https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/50f940b7f1625ce6830044350f7ee29d768834cd

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -1,15 +1,46 @@
...
+    key: string;
+    title: string;
+    isDir: boolean;
+}
```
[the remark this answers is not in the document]
PatrycjaKubiczek: correct, that name makes much more sense :) applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/50f940b7f1625ce6830044350f7ee29d768834cd

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -48,7 +82,7 @@ export default function BlueprintSources({ data, toolbox, widget }) {
                 setType(fileType);
                 clearError();
             })
-            .catch(err => {
+            .catch((err: { message: string }) => {
```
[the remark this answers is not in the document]
PatrycjaKubiczek: most of time I've seen typing `err` with `any` so I didn't know we have global `Error` type accessible :D applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/50f940b7f1625ce6830044350f7ee29d768834cd

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -19,7 +50,10 @@ export default function BlueprintSources({ data, toolbox, widget }) {
         resetType();
     }, [data]);
 
-    function selectFile(selectedKeys, info) {
+    function selectFile(
+        selectedKeys: string[],
+        info: { node: { props: { children: any; title: { props: { children: React.SetStateAction<string>[] } } } } }
```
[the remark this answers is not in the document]
PatrycjaKubiczek: I extracted info o seperate interface to avoid any, is that ok? https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/50f940b7f1625ce6830044350f7ee29d768834cd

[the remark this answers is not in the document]
NorbertAmartus: Yeah, the implementation looks good to me :)
Thanks for applying one of the suggested approaches 🎉

widgets/blueprintSources/src/BlueprintSources.tsx
NorbertAmartus: Till this point we were not making tickets for migrating single components into typescript, as we are trying to do it during our development work (@qooban pleasea correct me if I'm wrong 😁)

But leaving a trace behind it may be a good idea, as we could not only take action upon it, but we could also group migration tasks (which are related to the same package / application part) to do complete them at the same time 🎉 

It could be a beneficial strategy, especially when some parts of the system don't look like they will be touched by the developer anytime soon 😅 

Additionally those tasks could make a good candidate for a new UI team member tasks 🚀 

It may be worth noticing that with having that approach in mind, one day we could search for all not migrated components, which are implemented with the usage of the mentioned `@ts-ignore <component name> is not migrated to typescript` comment, to create corresponding tasks.


@qooban WDYT about this subject? 🤔

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -107,10 +153,11 @@ export default function BlueprintSources({ data, toolbox, widget }) {
                 <SplitterLayout
                     primaryIndex={0}
                     percentage
-                    secondaryInitialSize={widget.configuration.contentPaneWidth}
+                    secondaryInitialSize={widget.configuration.contentPaneWidth as number}
                 >
                     <div>
                         <NodesTree className="nodes-tree" showLine selectable defaultExpandAll onSelect={selectFile}>
+                            {/* @ts-ignore Missing the following properties from type 'TreeNodeProps': style, onSelect, icon, switcherIconts */}
```
NorbertAmartus, line 160: As discussed [here](https://github.com/cloudify-cosmo/cloudify-stage/pull/2046#discussion_r915911542)

```suggestion
                            {/* @ts-ignore NodesTree.Node is not migrated to typescript */}
```

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -1,15 +1,46 @@
-// @ts-nocheck File not migrated fully to TS
 import SplitterLayout from 'react-splitter-layout';
 import Actions from './actions';
 
-export default function BlueprintSources({ data, toolbox, widget }) {
+interface BlueprintTree {
+    children: any;
```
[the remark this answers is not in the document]
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/50f940b7f1625ce6830044350f7ee29d768834cd

widgets/blueprintSources/src/BlueprintSources.tsx
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/50f940b7f1625ce6830044350f7ee29d768834cd

widgets/blueprintSources/src/BlueprintSources.tsx
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/50f940b7f1625ce6830044350f7ee29d768834cd

widgets/blueprintSources/src/BlueprintSources.tsx
PatrycjaKubiczek: addressed it here: https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/3d2833b9ddcbdc0958510f7563fb53a463d6e86b

widgets/blueprintSources/src/BlueprintSources.tsx
qooban: Hm... Maybe let's leave it then unchanged. Thanks for checking.

widgets/blueprintSources/src/BlueprintSources.tsx
qooban: @NorbertAmartus Actually I was thinking about that last week 🙂 My observation is that TS migration in `cloudify-ui-components` is a bit slow and I get to a conclusion that we may speed it up only by having dedicated tasks and trying to fit it into our sprints, so...

I have just created RD-5339 epic to collect such issues. 
@PatrycjaKubiczek Please add there migration task for `NodeTree`.
I'll add a point to UI weekly to discuss it further with the whole team.

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -1,15 +1,57 @@
-// @ts-nocheck File not migrated fully to TS
+import type { HighlightText } from 'cloudify-ui-components';
+import type { ComponentProps } from 'react';
 import SplitterLayout from 'react-splitter-layout';
 import Actions from './actions';
 
-export default function BlueprintSources({ data, toolbox, widget }) {
+interface NodeTreeItem {
+    children: NodeTreeItem[];
```
qooban, line 7: I think we should make it optional as leafs of that node tree does not have `children` prop.

image

widgets/blueprintSources/src/BlueprintSources.tsx
```diff
@@ -1,15 +1,57 @@
...
+    key: string;
+    title: string;
+    isDir: boolean;
+}
+
+type FileType = ComponentProps<typeof HighlightText>['language'];
```
qooban, line 13: Hmm... As in L1 we are importing a type: 
```
import type { HighlightText } from 'cloudify-ui-components';
```
do we need to have `typeof` here? 🤔

widgets/blueprintSources/src/BlueprintSources.tsx
PatrycjaKubiczek: @qooban created the corresponding ticket: https://cloudifysource.atlassian.net/browse/RD-5341 💪

widgets/blueprintSources/src/BlueprintSources.tsx
qooban: If `typeof` is needed here, then:
```
import type { HighlightText } from 'cloudify-ui-components';
``` 
doesn't work as type import 🤔 There's no `HighlightText` type in 'cloudify-ui-components'.

I suggest to move L101:
```
    const { CancelButton, NodesTree, Message, Label, Modal, HighlightText, ErrorMessage, Icon } = Stage.Basic;
```
before L13:
```
type FileType = ComponentProps<typeof HighlightText>['language'];
```
and remove L1:
```
import type { HighlightText } from 'cloudify-ui-components';
``` 

That should make it a bit more clear.

qooban (commented):
Looks good. I commented on just one thing.

widgets/blueprintSources/src/BlueprintSources.tsx
PatrycjaKubiczek: okay, got it :) applied here: https://github.com/cloudify-cosmo/cloudify-stage/pull/2046/commits/394b2e817eee9c7c6a302ef06ec02571ddfae3c3 actually, I moved all the imports to have more concise structure

qooban: Thanks. Looks clearer 👍🏻

NorbertAmartus (approved):
LGTM 🎉 
Thanks for addressing all of the comments 💪

qooban (approved):

PR #2051, opened 2022-07-11
RD-4926 fix autofill bug in Firefox and migrate LoginPage to TS

PatrycjaKubiczek:
## Description

RD-4926
- fix autofill colors in FF
- migrate files to TS

## Screenshots / Videos

Before: 
Screen Shot 2022-05-11 at 11 26 54

After:
Zrzut ekranu 2022-07-07 133138

## Checklist

## Tests

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3084/)

 

## Documentation

n/a


app/components/LoginPage.tsx
```diff
@@ -23,44 +65,44 @@ export default class LoginPage extends Component {
...
-    handleInputChange = (proxy, field) => {
+    handleInputChange = (_proxy: any, field: Parameters<typeof Stage.Basic.Form.fieldNameValue>[0]) => {
         const fieldNameValue = Form.fieldNameValue(field);
         this.setState({ ...fieldNameValue, errors: {} });
     };
 
     render() {
         const { errors, password, username } = this.state;
-        const { isLoggingIn, isSamlEnabled, loginError } = this.props;
+        const { isLoggingIn, isSamlEnabled, loginError = null } = this.props;
```
PatrycjaKubiczek, line 100: I'm wondering if we had default props should I transfer them as well while migrating files to TS? 🤔

app/components/LoginPage.tsx
```diff
@@ -23,44 +65,44 @@ export default class LoginPage extends Component {
...
 
         const query = parse(location.search);
-        const redirect = query.redirect || null;
+        const redirect = (query.redirect as string) || null;
 
         return onLogin(username, password, redirect);
     };
 
-    handleInputChange = (proxy, field) => {
+    handleInputChange = (_proxy: any, field: Parameters<typeof Stage.Basic.Form.fieldNameValue>[0]) => {
```
qooban, line 93: As `Form` should be available here, we can use shorter form:
```suggestion
    handleInputChange = (_proxy: any, field: Parameters<typeof Form.fieldNameValue>[0]) => {
```

app/components/LoginPage.tsx
```diff
@@ -23,44 +65,44 @@ export default class LoginPage extends Component {
...
-        const { isSamlEnabled, location, onLogin, samlSsoUrl } = this.props;
-        const errors = {};
+        const { isSamlEnabled, location, onLogin, samlSsoUrl = '' } = this.props;
+        const errors: Errors = {};
 
         if (isSamlEnabled) {
-            // eslint-disable-next-line scanjs-rules/assign_to_location
-            window.location = samlSsoUrl;
+            // eslint-disable-next-line scanjs-rules/assign_to_href
+            window.location.href = samlSsoUrl;
```
qooban, line 73: Are we sure it works the same and we are not introducing any regression here? 🤔

app/containers/LoginPage.ts
```diff
@@ -22,7 +22,7 @@ const mapStateToProps = (state: ReduxState) => {
 
 const mapDispatchToProps = (dispatch: Dispatch) => {
     return {
-        onLogin: (username: string, password: string, redirect: string) => {
```
qooban, line 25: General:

Just FYI - as part of Ongoing Refactors we are also trying to move out from `containers`. Ref.: https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Moving-containers-to-components

In case of this PR I think it might be too much to implement it, it may shadow more the core change of this PR. But I'm leaving the decision to you 🙂

app/components/LoginPage.tsx
qooban: Yeah, I guess so. That would make the migration to not contain functional changes.

app/components/LoginPage.tsx
PatrycjaKubiczek: @qooban `window.location` and `window.location.href` are different. First one is an object and second returns a string. However according to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/location):

> Though Window.location is a read-only Location object, you can also assign a string to it. **This means that you can work with location as if it were a string in most cases: location = 'http://www.example.com' is a synonym of location.href = 'http://www.example.com'.**

 so I think in our case it will act the same, because we're assigning URL to redirect a page. In TS `window.location` will return a type error because location is an object with typing like this `(string | Location) & Location`. That's why I change it `window.location.href`. 
 
 The other way I could solve this issue is type assertion: `window.location = samlSsoUrl as (string | Location) & Location;`

app/components/LoginPage.tsx
```diff
@@ -1,17 +1,59 @@
...
+
+type Redirect = string | null;
+
+const StyledInput = styled(Input)`
+    &&&&&& input:autofill,
+    input:autofill:focus,
+    input:-webkit-autofill:hover,
+    input:-webkit-autofill:focus,
+    input:-webkit-autofill:active {
+        -webkit-box-shadow: 0 0 0 100px transparent !important;
```
NorbertAmartus, line 47: As the [styled components](https://styled-components.com/docs/basics#motivation) are adding vendor prefixes (like -webkit-`) by default, it is required to add this particular css rule? 🤔

app/components/LoginPage.tsx
```diff
@@ -1,17 +1,59 @@
...
+
+type Errors = {
+    username?: string;
+    password?: string;
+} | null;
+
+type Redirect = string | null;
+
+const StyledInput = styled(Input)`
+    &&&&&& input:autofill,
```
NorbertAmartus, line 42: As we are using `!important` flags for the css rules, do we need to bump the selector specificity? 🤔

NorbertAmartus:
### Comment about the PR title
It may be beneficial to use `Firefox` instead of `FF` as this abbreviation (out of context) may not be clear for everybody

app/components/LoginPage.tsx
PatrycjaKubiczek: I think we do because these selectors have already `!important` 😅 
Zrzut ekranu 2022-07-11 151228

app/components/LoginPage.tsx
PatrycjaKubiczek: good catch, I will modify it

PatrycjaKubiczek: applied changes: https://github.com/cloudify-cosmo/cloudify-stage/pull/2051/commits/493167abb3829b339a00e83d4690942cd2630ee2

app/containers/LoginPage.ts
PatrycjaKubiczek: I was curious I tried to move it out of `containers` :) https://github.com/cloudify-cosmo/cloudify-stage/pull/2051/commits/493167abb3829b339a00e83d4690942cd2630ee2 looks like everything still works as intended

app/components/LoginPage.tsx
qooban: Thanks for explanation. Tested locally and it seems to work properly, so... All fine 🙂

qooban (dismissed):
LGTM 👍🏻 

Nicely done with all additional refactors 👍🏻

app/components/LoginPage.tsx
NorbertAmartus: Ou okay 😄
In that case let's fight fire ... with fire! 🔥  

Thanks for the explaination! 🎉

app/components/LoginPage.tsx
```diff
@@ -1,17 +1,62 @@
...
+type Errors = {
+    username?: string;
+    password?: string;
+} | null;
+
+const StyledInput = styled(Input)`
+    &&&&&& input:autofill,
+    input:autofill:focus,
+    input:autofill:hover,
+    input:autofill:focus,
```
NorbertAmartus, line 49: After refactoring, some code duplication has sneak into the selectors 😁 
additionalcode

```suggestion
    input:autofill:focus,
    input:autofill:hover,
```

NorbertAmartus (approved):
LGTM 🎸 

Thanks for addressing all of the comments 🚀 
(and for the opportunity to create an occasional meme 😁)

PR #2068, opened 2022-07-15
RD-5365 Fix Topology widget data loading

qooban:
## Description
As described in RD-5365 the topology view didn't load on second time open.
The reason for that problem was the fact that topology data was set (`topology.setTopology` calls in [widgets/topology/src/Topology.tsx](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/topology/src/Topology.tsx)) only in case of blueprint data update. On `Topology` component mount the view was not initialized properly.

Small refactorings added: 
1. fixed `propTypes` definition
2. simplified nodes difference checking function

NOTE: This widget is hard to maintain (it's using non-React library for rendering the view) and the code is very messy. I considered TS migration, but I decided to drop it as a time-consuming and risky before the release.

## Screenshots / Videos

https://user-images.githubusercontent.com/5202105/179209766-a429fd00-3377-4e7d-854e-a8326754b004.mp4

## Checklist

## Tests
1. Topology spec passes locally
2. [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3148/) - PASSED

## Documentation
N/A

### DIFF
--- a/widgets/topology/src/Topology.tsx
+++ b/widgets/topology/src/Topology.tsx
@@ -80,6 +80,7 @@ export default class Topology extends React.Component<TopologyProps, TopologySta
         toolbox.getEventBus().on('blueprints:refresh', toolbox.refresh, this);
         toolbox.getEventBus().on('deployments:refresh', toolbox.refresh, this);
         this.startTopology();
+        this.updateTopology();
     }
 
     shouldComponentUpdate(nextProps: TopologyProps, nextState: TopologyState) {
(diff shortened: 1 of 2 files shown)


widgets/topology/src/Topology.tsx
```diff
@@ -97,34 +99,18 @@ export default class Topology extends React.Component<TopologyProps, TopologySta
...
-                this.topology.setTopology(this.topologyData, layout);
-                if (isFirstTimeLoading) this.topology.setScale(_.get(layout, 'scaleInfo'));
-            } else {
-                this.topology.refreshTopologyDeploymentStatus(this.topologyData);
-            }
-        }
+            expandedTerraformNodes !== prevState.expandedTerraformNodes;
+
+        if (topologyRestartAndUpdateRequired) this.startTopology();
+        if (topologyRestartAndUpdateRequired || topologyUpdateRequired) this.updateTopology();
```
NorbertAmartus, line 113:
```suggestion
        const topologyRestartRequired =
            configuration !== prevProps.configuration ||
            blueprintId !== prevProps.blueprintId ||
            !_.isEqual(data.icons, prevProps.data.icons);

        const topologyUpdateRequired =
            topologyRestartRequired ||
            !_.isEqual(data.blueprintDeploymentData, prevProps.data.blueprintDeploymentData) ||
            expandedDeployments !== prevState.expandedDeployments ||
            expandedTerraformNodes !== prevState.expandedTerraformNodes;

        if (topologyRestartRequired) this.startTopology();
        if (topologyUpdateRequired) this.updateTopology();
```
 
With that approach we could correlate one variable with a corresponding functionality (restart is for restarting and updating is for updating 😁), making the function more readable and easier to understand 🎉 

(I know that by adding extra condition to one of the variables, it makes the 'if`ology' bigger, but as it already had a few conditions, I think that we shouldn't be worried about it)

WDYT? 🤔

qooban: Of course, good idea. Applied.

NorbertAmartus (approved):
LGTM 💪

PR #2079, opened 2022-07-20
RD-5299:  search by deployment name in Deployments Widget

PatrycjaKubiczek:
## Description

RD-5299 
- add additional `searchByDeploymentName` to `widget.tsx` in Deployments Widget to enable searching by `id` and by `display_name`
- migrate `DeploymentsList` to TS
- modify `FirstJourneyButtons` behaviour on search - now it will display buttons only if there's no deployments
- modify the `noDeployments` message 
before: 
_"There are no Deployments available. Click \"Create deployment\" to add deployments."_ 
after :
_'No results found. Please, try a different search."._ 
Previous message was misleading since there's no "create deployment" button available
- modify the corresponding cypress test

## Screenshots / Videos

Before:

https://user-images.githubusercontent.com/17828621/179978551-beeccedc-d66f-4d9f-96c6-0589b0309604.mp4

After:

https://user-images.githubusercontent.com/17828621/179987800-35a876c5-f9b0-44fa-a0a0-d0498c047fd0.mp4

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3180/)
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3205/) (25.07.2022)
## Documentation

n/a


test/cypress/integration/widgets/deployments_spec.ts
```diff
@@ -39,9 +39,16 @@ describe('Deployments widget', () => {
...
+    describe('should be present in Deployments page', () => {
+        it('should be able search by depyloyment ID', () => {
+            cy.searchInDeploymentsWidget(deploymentId);
+            cy.get('.deploymentSegment h3 [aria-label="Deployment name"]').should('have.text', deploymentName);
+        });
+
+        it('should be able by deployment name', () => {
+            cy.searchInDeploymentsWidget(deploymentName);
+            cy.get('.deploymentSegment h3 [aria-label="Deployment name"]').should('have.text', deploymentName);
+        });
```
qooban, line 51: I suggest to adjust the naming a bit.
```suggestion
    describe('should be present in Deployments page and', () => {
        it('should allow to search by depyloyment ID', () => {
            cy.searchInDeploymentsWidget(deploymentId);
            cy.get('.deploymentSegment h3 [aria-label="Deployment name"]').should('have.text', deploymentName);
        });

        it('should allow to search by deployment name', () => {
            cy.searchInDeploymentsWidget(deploymentName);
            cy.get('.deploymentSegment h3 [aria-label="Deployment name"]').should('have.text', deploymentName);
        });
```

WDYT?

Optional:

Seeing the same 2 lines, maybe also common function for 
```
cy.get('.deploymentSegment h3 [aria-label="Deployment name"]').should('have.text', deploymentName);
```
can be extracted. It's optional IMO.

test/cypress/integration/widgets/deployments_spec.ts
```diff
@@ -88,25 +95,8 @@ describe('Deployments widget', () => {
...
-
-            const mockDeploymentsResponse = (mockedResponse: any) =>
-                cy.interceptSp('GET', '/deployments*', mockedResponse);
-
-            it('should display showFirstUserJourneyButtons view when there are not installed deployments', () => {
-                const mockedResponse = getMockedResponse([]);
-                mockDeploymentsResponse(mockedResponse);
-
+            it('should display showFirstUserJourneyButtons view when there are no deployments', () => {
+                cy.deleteDeployments('', true);
```
qooban, line 99: Hmm... That may be an overkill if we have many deployments created in the Cloudify Manager.
Can we user mocked response as before? 🤔

widgets/deployments/src/widget.tsx
```diff
@@ -1,6 +1,7 @@
 // @ts-nocheck File not migrated fully to TS
 
 import { get, isEmpty, isEqual } from 'lodash';
+import { searchAlsoByDeploymentName } from '../../common/src/actions/SearchActions';
```
qooban, line 4: We shouldn't import directly from `common/src` (optimization reasons - bundle size). Can we use `Stage.Common` instead? Either we can add it inside `SearchActions` class and access it through `Stage.Common.Actions.Search` or find a better place in `Stage.Common...` for it.

widgets/deployments/src/widget.tsx
```diff
@@ -145,13 +149,11 @@ Stage.defineWidget({
         const {
             configuration: { showFirstUserJourneyButtons }
         } = widget;
-        const shouldShowFirstUserJourneyButtons = showFirstUserJourneyButtons && isEmpty(data?.items);
-
```
qooban, line 149: I think we can keep it here, but extend it to check if `_search` was set.

You should be able to get it from the local scope:
image

Check: `data.metadata.gridParams._search`.
Probably changes in L142 and L156 won't be needed then.

widgets/deployments/src/DeploymentsList.tsx
```diff
@@ -1,11 +1,53 @@
...
+        };
+        pagination: {
+            offset: number;
+            size: number;
+            total: number;
+        };
+    };
+    showFirstUserJourneyButtons: boolean;
+    total: number;
+}
```
qooban, line 27: I think you can use `PaginatedResponse` generic type (accessible from widget using `Stage.Types.PaginatedResponse`) here instead of defining `items` and `metadata`.

widgets/deployments/src/DeploymentsTable.tsx
```diff
@@ -16,6 +16,7 @@ export default function DeploymentsTable({
     onSelectDeployment,
     onWorkflowAction,
     onSetVisibility,
+    onError,
```
qooban, line 19: Is it used anywhere? 
If so, then probably we should add it also to `DeploymentsSegment`? 🤔 
If not, maybe we shouldn't pass it down in `DeploymentsList` component do `DeploymentsView` (`DeploymentsSegment` or `DeploymentsTable`) at all?

widgets/deployments/src/DeploymentsList.tsx
```diff
@@ -47,41 +89,42 @@ export default class DeploymentsList extends React.Component {
             .finally(() => toolbox.loading(false));
     };
 
-    setError = errorMessage => {
+    setError = (errorMessage: Error) => {
```
qooban, line 92: Just out of curiosity - are we sure it's generic `Error`?

app/translations/en.json
```diff
@@ -1625,7 +1625,7 @@
                 }
             },
             "list": {
-                "noDeployments": "There are no Deployments available. Click \"Create deployment\" to add deployments."
+                "noDeployments": "No results found. Please, try a different search."
```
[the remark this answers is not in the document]
PatrycjaKubiczek: Good point. Do you think only "No results found." will be better?

NorbertAmartus (commented):
I'll wait for the current comments to be resolved, as they may come with a large amount of changes, that could resolve issues which I may have in mind 😁 

Besides that. I agree with most of the suggestions/questions raised within the current comments 🚀

app/translations/en.json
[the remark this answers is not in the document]
PatrycjaKubiczek: I remove second sentence: https://github.com/cloudify-cosmo/cloudify-stage/pull/2079/commits/40a4652300c209b1746a3947495475d338092e48

test/cypress/integration/widgets/deployments_spec.ts
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2079/commits/40a4652300c209b1746a3947495475d338092e48

widgets/deployments/src/DeploymentsList.tsx
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2079/commits/40a4652300c209b1746a3947495475d338092e48

widgets/deployments/src/DeploymentsTable.tsx
PatrycjaKubiczek: yeah, I don't think we use it anywhere, so I agree - it's better to remove it :) https://github.com/cloudify-cosmo/cloudify-stage/pull/2079/commits/40a4652300c209b1746a3947495475d338092e48

widgets/deployments/src/widget.tsx
PatrycjaKubiczek: added to SearchActions: https://github.com/cloudify-cosmo/cloudify-stage/pull/2079/commits/40a4652300c209b1746a3947495475d338092e48

test/cypress/integration/widgets/deployments_spec.ts
PatrycjaKubiczek: reverted to mocked response: https://github.com/cloudify-cosmo/cloudify-stage/pull/2079/commits/40a4652300c209b1746a3947495475d338092e48

widgets/deployments/src/widget.tsx
PatrycjaKubiczek: I tried to only use `data.metadata.gridParams._search` as an additional condition in  `shouldShowFirstUserJourneyButtons` but it didn't work properly - so I decided to leave this search value in data and it works as expected. 
WDYT?

widgets/deployments/src/DeploymentsList.tsx
PatrycjaKubiczek: I checked it once again and I think error in ErrorMessage could be a string, array, or object, but in that particular case, it should be just a regular string

test/cypress/integration/widgets/deployments_spec.ts
```diff
@@ -39,9 +41,16 @@ describe('Deployments widget', () => {
             .mockLogin();
     });
 
-    it('should be present in Deployments page', () => {
-        cy.searchInDeploymentsWidget(deploymentId);
-        cy.get('.deploymentSegment h3 [aria-label="Deployment name"]').should('have.text', deploymentName);
+    describe('should be present in Deployments page and', () => {
```
qooban, line 44: Optional:

I just spotted that the name can be misleading as we don't have "Deployments" page 🙂 
It's optional to fix it.

test/cypress/integration/widgets/deployments_spec.ts
```diff
@@ -77,6 +86,7 @@ describe('Deployments widget', () => {
                 cy.get('.icon').should('be.visible');
                 cy.get('.label').should('be.visible');
             });
+            cy.getSearchInput().clear();
```
qooban, line 89: I guess it was added for other (next) tests to have clean search input, right?

It's not the best to make that kind of inter-test dependencies (as failure before that line will trigger failure of the next test as well), but IMO we can live with it.

For a reference - it is somehow related to https://docs.cypress.io/guides/references/best-practices#Using-after-or-afterEach-hooks and https://docs.cypress.io/guides/references/best-practices#Having-tests-rely-on-the-state-of-previous-tests.

test/cypress/integration/widgets/deployments_spec.ts
```diff
@@ -103,7 +113,7 @@ describe('Deployments widget', () => {
             const mockDeploymentsResponse = (mockedResponse: any) =>
                 cy.interceptSp('GET', '/deployments*', mockedResponse);
 
-            it('should display showFirstUserJourneyButtons view when there are not installed deployments', () => {
+            it('should display showFirstUserJourneyButtons view when there are no deployments', () => {
```
qooban, line 116: Good catch 👍🏻

test/cypress/integration/widgets/deployments_spec.ts
```diff
@@ -121,7 +131,7 @@ describe('Deployments widget', () => {
                 cy.contains('Blueprint Marketplace').should('be.visible');
             });
 
-            it("should hide showFirstUserJourneyButtons view when there's at least one deployment installed", () => {
+            it("should hide showFirstUserJourneyButtons view when there's at least one deployment", () => {
```
qooban, line 134: Good catch 👍🏻

widgets/deployments/src/widget.tsx
```diff
@@ -145,11 +149,12 @@ Stage.defineWidget({
...
         if (isEmpty(data)) {
             return <Loading />;
         }
+        // eslint-disable-next-line no-underscore-dangle
```
qooban, line 155: Do we need this line? 🤔

widgets/deployments/src/widget.tsx
qooban: OK 👍🏻

widgets/deployments/src/widget.tsx
```diff
@@ -145,11 +149,12 @@ Stage.defineWidget({
...
+        const searchValue = data?.searchValue;
+        const shouldShowFirstUserJourneyButtons = showFirstUserJourneyButtons && isEmpty(data?.items) && !searchValue;
```
NorbertAmartus, line 157: ### Optional

WDYT about adjusting this line to firstly check if the string is being parsed to the `truthy` value and then to go through the array?

That order of code execution should be more performant 💪 
Maybe in a small way, but still 🎉 

```suggestion
        const shouldShowFirstUserJourneyButtons = showFirstUserJourneyButtons && !searchValue && isEmpty(data?.items);
```

Because it's a minor improvement, I'm marking this suggestion as optional 😁

NorbertAmartus (approved):
Looks amazing!

Thanks for putting so much work into both TS migration and feature itself 🎉 
This code could have been read like a book before a good night sleep 😁 

Left one minor comment, which is optional
That's why with all the pleasure, I'm approving this PR 

great job

test/cypress/integration/widgets/deployments_spec.ts
PatrycjaKubiczek: implemented: https://github.com/cloudify-cosmo/cloudify-stage/pull/2079/commits/b4a7a4cb07f5ea18a40ccf93982c68b1c922760d

widgets/deployments/src/widget.tsx
PatrycjaKubiczek: we don't 😅 I removed it https://github.com/cloudify-cosmo/cloudify-stage/pull/2079/commits/b4a7a4cb07f5ea18a40ccf93982c68b1c922760d

test/cypress/integration/widgets/deployments_spec.ts
PatrycjaKubiczek: @qooban yes, exactly, that was my intention.

NorbertAmartus (approved):

PR #2092, opened 2022-07-26
RD-5399: Fixed variables & outputs detection on zip files

NorbertAmartus:
## Description

Changes consists of modifying current terraform mechanism for detecting modules, variables and outputs across both zip and git files.

As I had to analyze currently existing mechanisms, I've found out that some of them can be simplified and reused.
And thus, those adjustments were made 🎉 

I've lefted comments containing explanation of some made decisions, to keep everyone on the same page 💪 

Other than that, I've tried to
- increase the typing security
- modify variable names, so that their context would be reflected more accurately

## Screenshots / Videos

N/A

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3273/)

## Documentation

N/A

## Changes should be a part of `6.4.0` release?
Yes, corresponding PR with cherrypick will be created 🚀


backend/routes/Terraform.ts
```diff
@@ -147,32 +146,32 @@ const getModuleListForGitUrl = async (url: string, authHeader?: string) => {
...
 const getTfFileBufferListFromGitRepositoryUrl = async (url: string, resourceLocation: string, authHeader?: string) => {
     const repositoryPath = getUniqNotExistingTemporaryDirectory();
-    const files: string[] = [];
+    const files: Buffer[] = [];
 
     await cloneGitRepo(repositoryPath, url, authHeader);
 
-    await directoryTree(repositoryPath, directoryTreeOptions, (_file, filePath) => {
-        const isInRootDirectory = !filePath.includes('/');
-        if (isInRootDirectory || filePath.indexOf(`${resourceLocation}/main.tf`) > -1) {
```
NorbertAmartus, line 156: ### Note
This line is making `main.tf` file to not be scanned in some scenarios
I think it's a residue from the previous implementation of the scanning mechanism (for git files we were scanning only files with names: `variables.ts` and `outputs.ts` - which was an incorrect mechanism)

As stated in [RD-5399](https://cloudifysource.atlassian.net/browse/RD-5399?focusedCommentId=77021) all `*.tf` files should be scanned.
Because of that we shouldn't treat `main.tf` file in any different way than the other files.

Having that in mind, I've removed this condition 💪

backend/routes/Terraform.ts
```diff
@@ -147,32 +146,19 @@ const getModuleListForGitUrl = async (url: string, authHeader?: string) => {
...
 const getTfFileBufferListFromGitRepositoryUrl = async (url: string, resourceLocation: string, authHeader?: string) => {
     const repositoryPath = getUniqNotExistingTemporaryDirectory();
-    const files: string[] = [];
+    const terraformModulePath = path.join(repositoryPath, resourceLocation);
+    const files: Buffer[] = [];
 
     await cloneGitRepo(repositoryPath, url, authHeader);
 
-    await directoryTree(repositoryPath, directoryTreeOptions, (_file, filePath) => {
-        const isInRootDirectory = !filePath.includes('/');
```
NorbertAmartus, line 155: ### Note
`directoryTree` can be provided with the path to the directory, which it should traverse
Having that in mind, we can combine `repositoryPath` and `resourceLocation` to have a certainty that the traversed files are the ones that we've been looking for 🎉 

Having that in mind, I've removed the highlighted condition to simplify the code 💪

widgets/common/src/terraformModal/TerraformModal.tsx
```diff
@@ -289,45 +289,42 @@ export default function TerraformModal({ onHide, toolbox }: { onHide: () => void
...
-                    source: 'input',
-                    value: '',
-                    duplicated: false
-                }));
-
-                setOutputsDeferred(outputsTmp);
-                setVariablesDeferred(variablesTmp);
-                unsetTemplateModulesLoading();
-            }
+    useEffect(() => {
```
NorbertAmartus, line 292: ### Note
This function needs a refactoring but as there is a dedicated [task for refactoring the whole component](https://cloudifysource.atlassian.net/browse/RD-5332) I decided to not do it within the scope of this PR

qooban: Optional:

If necessary you can add a point in RD-5332 about it.

backend/routes/Terraform.ts
```diff
@@ -48,8 +46,8 @@ type ResourcesRequest = Request<
     }
 >;
 
-const getTerraformFilePaths = (pathPrefix: string): string[] => {
-    return terraformFilesToScan.map(fileName => path.join(pathPrefix, fileName));
+const isTerraformFilePath = (filePath: string, directoryPath: string): boolean => {
+    return !!filePath.match(`^(${escapeRegExp(directoryPath)})[/]?[^/]+\\.tf$`);
```
qooban, line 50: I'm wondering

1. Is `directoryPath` necessary here? 
2. Can't we just check file extension here the same like we do in L101?

backend/routes/Terraform.ts
```diff
@@ -92,18 +90,20 @@ const cloneGitRepo = async (repositoryPath: string, url: string, authHeader?: st
...
-                    files.some(
-                        file => file.type === 'file' && file.path.match(`^${escapeRegExp(directory)}[^/]+\\.tf$`)
-                    )
-                )
-                .map(directory => trimEnd(directory, '/'))
+                .filter(file => file.type === 'file' && file.path.endsWith('.tf'))
+                .map(file => {
+                    const parentPath = path.dirname(file.path);
+                    return parentPath;
+                })
```
qooban, line 105: Optional:

Parent path sounds a bit odd to me. It's directory path extracted from the full path, right?

I would then simplify it to:
```suggestion
                .map(file => path.dirname(file.path)
```

or you can probably make it even shorter by importing `dirname` from `path` and having:

```suggestion
                .map(file => dirname(file.path)
```

backend/routes/Terraform.ts
```diff
@@ -289,10 +276,13 @@ router.post('/resources', async (req: ResourcesRequest, res) => {
...
 async function getTerraformFileBufferListFromZip(zipBuffer: Buffer, resourceLocation: string) {
     const resourceLocationTrimmed = trimStart(resourceLocation, '/');
-    const acceptableFilePaths = getTerraformFilePaths(resourceLocationTrimmed);
 
     const files = await decompress(zipBuffer);
-    return files.filter(file => acceptableFilePaths.includes(`${file.path}`)).map(file => file?.data);
+    return files
+        .filter(file => {
+            return file.type === 'file' && isTerraformFilePath(file.path, resourceLocationTrimmed);
+        })
```
qooban, line 284: I think you can safely shorten it to:
```suggestion
        .filter(file => file.type === 'file' && isTerraformFilePath(file.path, resourceLocationTrimmed))
```

qooban (commented):
Added some comments.

Two more questions:
1. I spotted the following problem when doing quick tests:

https://user-images.githubusercontent.com/5202105/181733607-9842c0bd-233e-4e48-b9bb-52c3cd62b074.mp4

After clearing URL or File input I see loading spinner and Terraform module details section gets blocked.

Could you check where it comes from? I checked on `staging` environment and there's no such issue.

2. Have you considered adding a test comparing variables & outputs detection mechanism using Git repository and ZIP URL to that repository? It's optional IMHO, but if it's quick to add, I would try it.

backend/routes/Terraform.ts
NorbertAmartus: Yeah, here it is necessary.
While decompressing zip files, we are getting all files/directories put inside that zip file (without any nested structure, we ar ejust getting a plain array of those items 😁)

When we'd want to get outputs/variables, we could get provided with the following zip file structure:
```
  > module
      variables.ts
      outputs.ts
      > nested_module
          nested_variables.ts
          nested_outputs.ts
```

When the user would like to get all variables/outputs from the `module` we should scan ONLY files which are being nested directly in that directory, as `nested_variables.ts` and `nested_outputs.ts` are related to the other module, named `nested_module`.

Because of that it is necessary to check not only if files are having a `.ts` extension, but also if they are inside a specific directory.

backend/routes/Terraform.ts
NorbertAmartus: > It's directory path extracted from the full path, right?
Yeah 😁 

The first approach seems good to me, I'll implement it right away :)

NorbertAmartus:
> Have you considered adding a test comparing variables & outputs detection mechanism using Git repository and ZIP URL to that repository? It's optional IMHO, but if it's quick to add, I would try it.

To be honest, I did not think about it 😅 
I think that it could be beneficial to have a same similar tests for both mechanisms 🎉 
But I think that it may take some time as we still may encounter some differences, which should be discussed with one of the PMs
And as for now I wouldn't like to prelong this PR, so having a dedicated ticket for that could be a valid option here 🚀

NorbertAmartus:
> I spotted the following problem when doing quick tests

Nicely spotted! 
I've found a couse of that issue and have [fixed it](https://github.com/cloudify-cosmo/cloudify-stage/pull/2092/commits/76c1c2e79591e2f663b2f78fa3be1353fd053d77) 💪

widgets/common/src/terraformModal/TerraformModal.tsx
NorbertAmartus: It may not be necessary, but I've left a comment [there](https://cloudifysource.atlassian.net/browse/RD-5332), so that it wouldn't be forgotten 🎉

backend/routes/Terraform.ts
qooban: Got it 👍🏻

qooban (approved):
Looks good 👍🏻

PR #2093, opened 2022-07-26
RD-4936 Add first login credentials hint to the Login page

qooban:
## Description

The main goal of this PR was to provide  first login hint popup in the Login page. It should be displayed only during first time login. I'm checking if `UserApps` table is empty to determine if user is trying to log in for the first time.

List of all other changes:
* Added `showFirstLoginHint` to the user configuration as it needs to be turned off for SaaS environment (further work is to be done as part of RD-5443)
* Refactored `GET /config` endpoint allowing to use proper types generated directly from config JSON files
* Improved TypeScript support for configuration related parts of the application code

## Screenshots / Videos

image

## Checklist

## Tests
1. [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3246/) - failed due to RD-5460
2. [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3253/) - two specs failed (`login_spec` and `user_configuration_spec`)
3. [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3259/) - only `login_spec` and `user_configuration_spec`

## Documentation
Added description for `showFirstLoginHint` parameter in internal configuration. 
I don't think we need any update in the user documentation.


backend/config.ts
```diff
@@ -54,13 +58,12 @@ export function getConfig(mode?) {
...
 }
 
-export function getClientConfig(mode) {
+export function getClientConfig(mode: Mode) {
     const config = getConfig(mode);
 
     // For client only get from app config the relevant part (and not send passwords and shit)
     return {
         app: {
-            initialTemplate: config.app.initialTemplate,
```
qooban, line 63: I found that it was not used defined in our app config and I didn't find any uses in the application code, so I think we can safely remove it.

backend/config.ts
```diff
@@ -10,14 +10,18 @@ import app from '../conf/app.json';
 import logging from '../conf/logging.json';
```
qooban, line 10: There's something more to do here to remove `@ts-nocheck` as dynamic `require` used in this file don't work perfect with TypeScript.

backend/routes/Auth.ts
```diff
@@ -104,4 +106,10 @@ router.get('/RBAC', authenticateWithCookie, (req, res) => {
         });
 });
 
+router.get('/is-first-login', (_req, res, next) => {
```
kubama, line 109: I think `first-login` suits better for a resource name:
```suggestion
router.get('/first-login', (_req, res, next) => {
```

kubama, line 109: Did you consider including this flag as part of `config` endpoint response?

test/cypress/support/commands.ts
```diff
@@ -213,8 +213,8 @@ const commands = {
             }
         });
 
-        cy.get('.form > :nth-child(1) > .ui > input').clear().type(username);
-        cy.get('.form > :nth-child(2) > .ui > input').clear().type(password);
+        cy.get('.form > div > :nth-child(1) > .ui > input').clear().type(username);
+        cy.get('.form > div > :nth-child(2) > .ui > input').clear().type(password);
```
kubama, line 217: Can't we just use `get('input').eq(0)` and `get('input').eq(1)`? It would simply read to 'first input field' and 'second input field'.

qooban: Sure, good point. Changed.

backend/routes/Auth.ts
qooban: OK, changed.

kubama (dismissed):

backend/routes/Auth.ts
qooban: I did originally when I didn't have all requirements clarified. 

Ended up with dedicated endpoint, but as I have end-to-end understanding right now I can take a deeper look and check it again.

backend/db/Connection.ts
```diff
@@ -1,12 +1,16 @@
...
 
 import BlueprintUserDataModel from './models/BlueprintUserDataModel';
 import UserAppsModel from './models/UserAppsModel';
 import WidgetBackendsModel from './models/WidgetBackendsModel';
 
-const dbModule = getDbModule(getConfig().app.db, loggerFactory, [
+// NOTE: app.db.options.dialectOptions.ssl is defined as boolean in the main configuration file,
+// but manager installer overrides it, so that app.options.dialectOptions.ssl has object structure
+// as defined in `DialectOptions` type in `cloudify-ui-common/backend/db`
+const dbModule = getDbModule(getConfig().app.db as unknown as DbConfig, loggerFactory, [
```
NorbertAmartus, line 13: Woah, first time seeing chained type assertion <takes a photo>, it may require some sort of celebration! 🥳 

Nicely thought! 🎉

backend/config.ts
```diff
@@ -54,13 +56,12 @@ export function getConfig(mode?) {
...
 }
 
-export function getClientConfig(mode) {
+export function getClientConfig(mode: Mode): ClientConfig {
     const config = getConfig(mode);
 
     // For client only get from app config the relevant part (and not send passwords and shit)
     return {
         app: {
-            initialTemplate: config.app.initialTemplate,
```
NorbertAmartus, line 63: We don't need to return this parameter anymore?
Just wanted to make sure, as I don't see its replacement across the PR

backend/routes/Config.types.ts
```diff
@@ -0,0 +1,32 @@
...
+import type logging from '../../conf/logging.json';
+import type dbOptions from '../../conf/db.options.json';
+import type userConfig from '../../conf/userConfig.json';
+import type manager from '../../conf/manager.json';
+import type { Mode } from '../serverSettings';
+
+export type UserConfig = typeof userConfig;
+
+export interface Config {
+    app: typeof app & typeof root & typeof logging & { db: { options: typeof dbOptions } } & typeof userConfig;
```
NorbertAmartus, line 13: WDYT about extracting the `app` type to make the interface look more readable?

```suggestion
type App = typeof app & typeof root & typeof logging & { db: { options: typeof dbOptions } } & typeof userConfig;

export interface Config {
    app: App;
```

It's a minor change but it can always bring us some value in the future 🎉

backend/routes/Auth.ts
qooban: Tried to include the flag in `config` response. 

IMHO it increases complexity of `config` endpoint (either `backend/routes/Auth.ts` or `backend/config.ts` and the whole typing) as we are right now not using async code (and with DB access we would have to), so I would stay with the new endpoint.

backend/config.ts
qooban: Yes, as mentioned in https://github.com/cloudify-cosmo/cloudify-stage/pull/2093#discussion_r929850458 it is not used, so I believe we can safely remove it.

backend/routes/Config.types.ts
qooban: Yes, makes sense, adjusted it and reused in `ClientConfig` interface.

kubama (approved):

qooban:
Triggered system tests as in the previous one there were some failures (hopefully coming from RD-5452 only).

NorbertAmartus (approved):
Thanks for adressing the comments 🎉 

Looks great 🚀

qooban:
There's system test failure coming from RD-5460, so I'll wait for fix (from DevOps possibly), then re-run tests and decide on further steps.

NorbertAmartus (approved):

PR #2115, opened 2022-08-02
RD-1103 Deduplicate some consts

kubama:
## Description
Deduplicated some consts

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/3318/ - pass, failed tests are failing on `master` too

 

## Documentation
N/A

### DIFF
--- a/backend/utils.ts
+++ b/backend/utils.ts
@@ -52,8 +52,3 @@ export function getHeadersWithAuthenticationTokenFromRequest(req: Request, heade
 export function getHeadersWithAuthenticationToken(token: string, headers: AxiosRequestHeaders = {}) {
     return { ...headers, ...getAuthenticationTokenHeader(token) };
 }
-
-export function isYamlFile(filename: string) {
-    const lowercaseFilename = filename.toLowerCase();
-    return lowercaseFilename.endsWith('.yaml') || lowercaseFilename.endsWith('.yml');
-}
--- a/backend/utils.ts
+++ b/backend/utils.ts
@@ -1,6 +1,6 @@
-import pathlib from 'path';
-import type { Request } from 'express';
 import type { AxiosRequestHeaders } from 'axios';
+import type { Request } from 'express';
+import pathlib from 'path';
 import * as Consts from './consts';
 import { TOKEN_COOKIE_NAME } from './consts';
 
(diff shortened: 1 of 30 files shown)


NorbertAmartus (approved):
LGTM 🎉

backend/utils.ts
```diff
@@ -1,6 +1,6 @@
-import pathlib from 'path';
-import type { Request } from 'express';
 import type { AxiosRequestHeaders } from 'axios';
+import type { Request } from 'express';
+import pathlib from 'path';
```
qooban, line 3: Minor: 

Out of curiosity - seeing few such changes recently in you PRs, I suspect that comes from some specific IDE configuration, am I right? 🤔 Probably we can align our configurations to have common way of defining imports.

widgets/common/src/Consts.ts
```diff
@@ -1,11 +1,14 @@
 import type { MapOptions } from 'leaflet';
+import appConsts from '../../../app/utils/consts';
```
qooban, line 2: General:

This is somehow violating our rule not to import files from the outside of the widgets folder and use `Stage` global instead. I guess for consts is not a big problem as the `common.js` bundle size increase would not be big, but maybe before we combine app framework and built-in widgets (you suggested that topic for research) we should stick to the rules? 🤔

qooban (commented):
I think it would be good to document it somewhere that we have a common approach for sharing code between backend and frontend. 

We could add there the following items:
`backend/consts.ts` - shared consts
`backend/sharedUtils.ts` - shared utility functions
`backend/routes/*.types.ts` - shared types
Correct me if my assumptions are wrong.

Please also add a point on UI weekly meeting agenda to spread that knowledge across the whole UI team.

backend/utils.ts
kubama: I'm not sure, I think it's just "Optimize imports" option in WebStorm, however the imports sorting behavior seems to be present since I upgraded - I'm using version 2021.3.2 at the moment.

widgets/common/src/Consts.ts
kubama: > We could add there the following items: ...

The items are listed in `tsconfig.json`, so I just referenced this file in the README.
 
> Please also add a point on UI weekly meeting agenda to spread that knowledge across the whole UI team.

Done.

kubama: > maybe before we combine app framework and built-in widgets (you suggested that topic for research) we should stick to the rules?

I'm not sure... At the moment the only place for consts on widgets side is `Stage.Common.Consts`. If we want to get away from direct import then we have to introduce something like `Stage.Consts`. This way would have two 'consts' objects in `Stage` namespace. A bit messy as for me.

qooban: OK, let's keep it as is.

qooban (approved):

PR #2132, opened 2022-08-11
RD-5389 Disable Bulk action button when there's no deployment

PatrycjaKubiczek:
## Description

RD-5389
- pass additional parameter to verify if bulk action buttons should be disabled or not

## Screenshots / Videos

After:
Zrzut ekranu 2022-08-11 151516

## Checklist

## Tests

 

deployments_view_spec.ts: [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3441/)

## Documentation

n/a


widgets/common/src/deploymentsView/header/index.tsx
```diff
@@ -81,7 +84,12 @@ const DeploymentsViewHeader: FunctionComponent<DeploymentsViewHeaderProps> = ({
                     style={{ marginRight: 0 }}
                 />
             )}
-            <Dropdown button text={headerT('bulkActions.button')} style={{ marginLeft: '0.25em' }}>
+            <Dropdown
+                button
+                text={headerT('bulkActions.button')}
+                style={{ marginLeft: '0.25em' }}
+                disabled={deployments === undefined || deployments.length === 0}
```
Vorbert-Kruk, line 91: `deployments` are being typed in the following way:
```typescript
    deployments: Deployment[];
```

That type does not allow `deployments` to be `undefined`
If that condition is somehow achievable, maybe we have to align the typing to be:

```typescript
    deployments?: Deployment[];
```

Having that condition in mind (deployments may / may not be undefined), we can modify the code in the following ways:

### Deployments MAY be `undefined`

```suggestion
                disabled={!deployments?.length}
```

### Deployments MAY NOT be `undefined`

```suggestion
                disabled={!deployments.length}
```

or 


```suggestion
                disabled={deployments.length === 0}
```
--- 
WDYT about that proposal? 🤔

PatrycjaKubiczek: you're correct about undefined 🤦 I've chosen option number two - it fulfills its task and well... it's the shortest ;) https://github.com/cloudify-cosmo/cloudify-stage/pull/2132/commits/55e90b1a6ef979351eeb503677921876de397093

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -184,6 +184,7 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
                     toolbox={toolbox}
                     onFilterChange={setUserFilterRules}
                     filterRules={finalFilterRules}
+                    deployments={deploymentsResult.data.items}
```
qooban, line 187: Maybe to be more precise and not add unused data to `DeploymentsViewHeader` we could just have 
`deployOnAllowed` or `deployOnEnabled`? Similarly to `mapOpen` prop? WDYT?

qooban (commented):
Tested UI and it works according to the expected result from the ticket 👍🏻 

... but looking at the description of this feature (Deploy On) at https://docs.cloudify.co/staging/dev/working_with/console/widgets/deploymentsview/ I'm wondering are we actually doing the right change, but that's for Ilan, so I'll write my concerns in the ticket.

widgets/common/src/deploymentsView/DeploymentsView.tsx
PatrycjaKubiczek: applied: https://github.com/cloudify-cosmo/cloudify-stage/pull/2132/commits/9b6fdd1629523c3bd83663064bbee04a396994ab

widgets/common/src/deploymentsView/DeploymentsView.tsx
```diff
@@ -174,6 +174,7 @@ export const DeploymentsView: FunctionComponent<DeploymentsViewProps> = ({
 
     const mapOpen = !!(toolboxContext.getValue(mapOpenContextKey) as boolean | undefined);
     const toggleMap = () => toolboxContext.setValue(mapOpenContextKey, !mapOpen);
+    const deployOnEnabled = !!deploymentsResult.data.items.length;
```
Vorbert-Kruk, line 177: This flag is being used only for disabling bulk actions, so WDYT about renaming it to be `disableBulkActions`?

Name `deployOnEnabled` may suggest relationship with different kind of functionalities, which is not the case in this example 😉 

```suggestion
    const deployOnEnabled = !deploymentsResult.data.items.length;
```

(if you'd like to apply the suggestion, other parts of the code should be updated aswell, to use new prop name)

PatrycjaKubiczek: good suggestion, added it here: https://github.com/cloudify-cosmo/cloudify-stage/pull/2132/commits/cc2f53dfe9ddb5fccd183f5464a7eaaead52abd0

qooban (approved):

widgets/common/src/deploymentsView/DeploymentsView.tsx
qooban: Good idea 👍🏻

Vorbert-Kruk (approved):
Looks great! 
Thanks for addressing the comments 🎉

PR #2143, opened 2022-08-23
RD-3480: Introducing icon font

Vorbert-Kruk:
## Description

Changes are consisting of:
- Creating custom icons font (further refered to as `cloudify-icons`) and connecting it to the application
- Creating documentation for the process of working with `cloudify-icons`
- Automate the process of updating `cloudify-icons` by providing a dedicated CLI command and `fontello` configuration file 🎉 

## Screenshots / Videos

N/A

## Checklist

## Tests

 

N/A

## Documentation

Package documentation has within this PR 🎉


.size-limit
```diff
@@ -1,7 +1,7 @@
 [
     {
         "name": "Main JS bundle",
-        "limit": "315 KB",
+        "limit": "355 KB",
```
Vorbert-Kruk, line 4: Size has been increased due to the import of multiple fonts (which have been generated by `fontello`)

semantic-ui/README.md
```diff
@@ -0,0 +1,56 @@
+## Custom icons font
```
Vorbert-Kruk, line 1: ### Note
I'm not sure if that's a good place for that documentation 🤔     
Maybe it would be more beneficial to relocate it to a different project directory (possibly merge with other documentation) or to a dedicated confluence page 🤔

package.json
```diff
@@ -126,6 +127,7 @@
         "cross-env": "^7.0.3",
         "css-loader": "^3.2.0",
         "file-loader": "^1.1.11",
+        "fontello-cli": "^0.6.2",
```
qooban, line 130: As in `devDependencies` we have dependencies required to build the package (see L108), can we put it to `optionalDependencies`?

semantic-ui/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+Custom icon font is being used for storing custom icons.  
+It enables us to use newly created icons within the application framework
+
+### Using custom icons
+
+We can defer the usage of custom icons, depending on the used component
```
qooban, line 8: Is `defer` suitable in this context? 
I suspect you wanted to mention that there are different ways of using custom icons, right?

semantic-ui/README.md
qooban: I think it's OK for the time being to have it somewhere inside `semantic-ui` directory.
Maybe even more accurate would be to put it inside `semantic-ui/fonts` directory?

fontello.config.json
```diff
@@ -0,0 +1,80 @@
+{
```
qooban, line 1: Maybe that configuration file should be located closer to the generated fonts (in `semantic-ui/fonts`)? 🤔 WDYT?

app/styles/fonts.scss
```diff
@@ -10,3 +10,15 @@
...
 }
+
+@font-face {
+    font-family: 'cloudify-icons';
+    src: url('../../semantic-ui/fonts/cloudify-icons/cloudify-icons.eot?75470793');
+    src: url('../../semantic-ui/fonts/cloudify-icons/cloudify-icons.eot?75470793#iefix') format('embedded-opentype'),
+        url('../../semantic-ui/fonts/cloudify-icons/cloudify-icons.woff2?75470793') format('woff2'),
+        url('../../semantic-ui/fonts/cloudify-icons/cloudify-icons.woff?75470793') format('woff'),
+        url('../../semantic-ui/fonts/cloudify-icons/cloudify-icons.ttf?75470793') format('truetype'),
+        url('../../semantic-ui/fonts/cloudify-icons/cloudify-icons.svg?75470793#cloudify-icons') format('svg');
```
qooban, line 21: I suspect that bundle size increase of 40kB comes from the fact that we include multiple formats here.

From Webpack:
```
modules by path ./semantic-ui/fonts/cloudify-icons/ 44.6 KiB
  ./semantic-ui/fonts/cloudify-icons/cloudify-icons.eot?75470793 12.2 KiB [built] [code generated]
  ./semantic-ui/fonts/cloudify-icons/cloudify-icons.woff2?75470793 5.99 KiB [built] [code generated]
  ./semantic-ui/fonts/cloudify-icons/cloudify-icons.woff?75470793 7.1 KiB [built] [code generated]
  ./semantic-ui/fonts/cloudify-icons/cloudify-icons.ttf?75470793 11.9 KiB [built] [code generated]
  ./semantic-ui/fonts/cloudify-icons/cloudify-icons.svg?75470793 7.4 KiB [built] [code generated]
```

As we provide only TTF for Roboto, do we actually need to provide all these formats for custom icons font? 🤔 Maybe we can limit it only to TTF as well?

qooban (commented):
Nicely done!

Added few comments with questions, I guess none of them should be difficult to resolve.

package.json
Vorbert-Kruk: Yeah, it seems like a good idea 🎉

Vorbert-Kruk: Done 🎉

semantic-ui/README.md
Vorbert-Kruk: Yeah, you're right 😁 
I'll adjust it, thanks for having an eagle eye! 🦅

Vorbert-Kruk: I've rewritten this sentence in a more simple way 🚀

fontello.config.json
Vorbert-Kruk: I'm not certain about it, as generated bundle is consisting of:
- font files
- css files
- config file

If we'd get (and need to only adjust) config and font files, it could be a nice adjustment 🎉 
But because of the generated css, I think that it may not be the best idea
 

WDYT about moving this file to the `semantic-ui` directory instead? 🤔 
(All of the adjustments, after updating icons font, are being made in that place 🚀 )

app/styles/fonts.scss
Vorbert-Kruk: Sure, we can test it out 🚀 
If something would seem to be broken, we'll just import the rest of the fonts 💪

Vorbert-Kruk: After removing imports for fonts other than TTF, everything seems to be working as expected 🎉 

Because of that, I've:
- Removed those imports
- Removed unused fonts from `semantic-ui/fonts/cloudify-icons` directory
- Updated documentation

semantic-ui/README.md
Vorbert-Kruk: I'm not certain about the `semantic-ui/fonts` directory due to the [reasoning stated here](https://github.com/cloudify-cosmo/cloudify-stage/pull/2143#discussion_r955757954)

So if the current directory seems to be ok, I'd leave it here 🎉

fontello.config.json
qooban: > WDYT about moving this file to the `semantic-ui` directory instead? 🤔 

Yeah, make sense. 

Side note: Maybe in the future we can change the name of that directory to be something like `theme` instead of `semantic-ui`.

app/styles/fonts.scss
qooban: Great 👍🏻

semantic-ui/README.md
```diff
@@ -0,0 +1,74 @@
...
+    Example of passing icon name as a `className`
+
+    ```typescript
+    <Button
+        content="Example"
+        icon={{
+            className: "cloudify-dashboard"
+        }}
+    >
+    ```
```
qooban, line 54: General:

For now I think it's perfectly fine that we have all ways described. In the future we should probably decide which way we prefer and as we discussed probably have some wrappers.

fontello.config.json
Vorbert-Kruk: Yeah, I was wondering about the same thing 😁 

`theme` name may not be as accurate, as it can contain `themes` directory
But it surely is the right direction to modify the `semantic-ui` name 💪

semantic-ui/README.md
Vorbert-Kruk: Seems like a good idea to me 💪 

I think that during the theming implementation, we'll gain more understanding on which direction should we go and what types of actions should we take in order of making the process as easy and as undestandable as it may be🚀

fontello.config.json
Vorbert-Kruk: Config file has been relocated 🎉

qooban (approved):
Thanks for all adjustments. LGTM 👍🏻

PR #2145, opened 2022-08-24
RD-261 Remove usages of deprecated React APIs

kubama:
## Description
Removed usages of deprecated React APIs

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/3496/

 

## Documentation
N/A


app/components/Home.tsx
```diff
@@ -20,25 +20,19 @@ import Page from './Page';
 import SideBar, { collapsedSidebarWidth, expandedSidebarWidth } from './sidebar/SideBar';
 
 class Home extends Component {
-    // TODO: Context handling should not be here. Currently necessary to use deprecated methods.
-    // eslint-disable-next-line camelcase
-    UNSAFE_componentWillMount() {
+    componentDidMount() {
```
qooban, line 23: Are you sure drilldown context handling works properly after these changes?

I checked quickly and I'm seeing that:

https://user-images.githubusercontent.com/5202105/186853311-fc9fc635-5a46-463d-ab71-ddce629069d4.mp4

You can use `Nested-Example` blueprint from **Blueprint Marketplace > Getting Started** to reproduce it.

app/components/Home.tsx
```diff
@@ -20,25 +20,19 @@ import Page from './Page';
 import SideBar, { collapsedSidebarWidth, expandedSidebarWidth } from './sidebar/SideBar';
```
qooban, line 20: Maybe worth to migrate to TS at once?

qooban (commented):
Added 2 comments.

Turned on Strict Mode in local environment and saw some warnings (about using `findDOMNode` and legacy context API). Are all of them coming from Semantic UI React and we can just wait for v3 or there are still places which we should fix on our side?

app/components/Home.tsx
kubama: Done

app/components/Home.tsx
kubama: I rechecked locally and it's working just fine for me. This is also covered by system tests which are passing. Could you please re-validate?

kubama:
> Are all of them coming from Semantic UI React and we can just wait for v3 or there are still places which we should fix on our side?

I identified 2 more dependencies that have the same issue:
1. `react-visibility-sensor` - https://github.com/joshwnj/react-visibility-sensor/issues/141
The project seems to be dead, no fix can be expected.
2. `react-sortable-hoc` - https://github.com/clauderic/react-sortable-hoc/issues/677
Same as above.

I think we should have separate subtasks for those, however as fixes involve replacing these with actively maintained equivalents then these subtasks can be quite time consuming, so I'm not sure if I should include them or they should be planned separately.

app/components/Home.tsx
qooban: OK, I'll recheck and let you know.

qooban: Rechecked and it works fine. Sorry for misleading. There was probably an issue in my local environment.

qooban (approved):
Thanks for additional effort on TS migration. LGTM 👍🏻 

> I identified 2 more dependencies that have the same issue:
> 
> 1. `react-visibility-sensor` - [findDOMNode will be deprecated joshwnj/react-visibility-sensor#141](https://github.com/joshwnj/react-visibility-sensor/issues/141)
>    The project seems to be dead, no fix can be expected.
> 2. `react-sortable-hoc` - [React's StrictMode Warnings clauderic/react-sortable-hoc#677](https://github.com/clauderic/react-sortable-hoc/issues/677)
>    Same as above.
> 
> I think we should have separate subtasks for those, however as fixes involve replacing these with actively maintained equivalents then these subtasks can be quite time consuming, so I'm not sure if I should include them or they should be planned separately.

Yes, we should have seperate subtasks for them.

What do you think about creating similar ticket to RD-261 containing 3 subtasks - 1. RD-263, 2. for `react-visibility-sensor` replacement, 3. for `react-sortable-hoc` replacement? If that's fine, could you create it?

That will allow us to close RD-261 with what we have and continue this activity in the future (not in this sprint) with a new ticket.

Vorbert-Kruk (approved):
LGTM 🎉

kubama:
> What do you think about creating similar ticket to [RD-261](https://cloudifysource.atlassian.net/browse/RD-261) containing 3 subtasks - 1. [RD-263](https://cloudifysource.atlassian.net/browse/RD-263), 2. for `react-visibility-sensor` replacement, 3. for `react-sortable-hoc` replacement? If that's fine, could you create it?

Created RD-5688.

PR #2150, opened 2022-08-26
RD-3481: Updated sidebar icons

Vorbert-Kruk:
## Description

Changes consisting of modifying icons across the sidebar (according to [RD-3481](https://cloudifysource.atlassian.net/browse/RD-3481) guideline)

## Checklist

## Tests

 

N/A

## Documentation

N/A

### DIFF
--- a/app/components/sidebar/TenantSelection.tsx
+++ b/app/components/sidebar/TenantSelection.tsx
@@ -43,7 +43,7 @@ const TenantSelection: FunctionComponent = () => {
 
     return (
         <SideBarDropdownItem
-            icon="user circle"
+            icon="cloudify-user-cog"
             label={selectedTenant || t('noTenants')}
             onClose={clearSearch}
             onOpen={refreshTenants}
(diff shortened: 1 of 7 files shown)


app/components/sidebar/TenantSelection.tsx
```diff
@@ -43,7 +43,7 @@ const TenantSelection: FunctionComponent = () => {
 
     return (
         <SideBarDropdownItem
-            icon="user circle"
+            icon="cloudify-user-cog"
```
qooban, line 46: Optional:

WDYT about adding some more type-checking and code assistance for the developers?
I don't suggest any specific solution. Rough ideas: some generic functions or consts that will allow to get icons from `cloudify-icons` font or type extending `SemanticICONS` as we discussed.

We don't need to do it right now, so if you think it'd better to do it later, let's maybe have a dedicated ticket for it? (to simplify getting icons and improve icons type-checking)

qooban (approved):

app/components/sidebar/TenantSelection.tsx
Vorbert-Kruk: I think that it may be a great adjustment 🎉 
But also I think that it may be beneficial to create a separate ticket for it, as current development related to the theming changes seems kind of messy

I'll take care of it 💪 
And in addition, I'll try to investigate if there are some `fontello`  integrations which could help us in automating this process 🎉

PR #2154, opened 2022-08-30
RD-3483 Change Page Styling

PatrycjaKubiczek:
## Description

RD-3483 
- remove divider
- add Roboto Medium to fonts
- change breadcrumb's divider icon
- enhance design

## Screenshots / Videos

Zrzut ekranu 2022-08-30 121236

## Checklist

## Tests

 

n/a

## Documentation

n/a


app/components/Page.tsx
```diff
@@ -80,23 +84,25 @@ class Page extends Component<PageProps, never> {
...
+                    <Breadcrumbs
+                        pagesList={pagesList}
+                        onPageNameChange={onPageNameChange}
+                        isEditMode={isEditMode}
+                        onPageSelected={onPageSelected}
                     />
-                </div>
-                <div className="ui divider" />
+
+                    {(page.description || isEditMode) && (
```
Vorbert-Kruk, line 95: WDYT about extracting this condition as a separate `const`?

E.g.:
```typescript
const showPageDescription = page.description || isEditMode;
```

It could make the Page's insides more clear and readable  💪

app/components/Page.tsx
```diff
@@ -80,23 +84,25 @@ class Page extends Component<PageProps, never> {
...
-                />
-                <div>
-                    <EditableLabel
-                        value={page.description}
-                        placeholder={i18n.t('page.description', 'Page description')}
-                        className="pageDescription"
-                        enabled={isEditMode}
-                        onChange={newDesc => onPageDescriptionChange(page.id, newDesc)}
-                        inputSize="mini"
+                <StyledWrapper>
```
Vorbert-Kruk, line 87: ### Optional

We have already `StyledContainer` and `StyledWrapper`, which seems to be not that kind of different (when it comes to the meaning of those names 😁)

Maybe we could modify this name to be more meaningful / adjusted to purpose of this wrapper?

It's an open question, as I don't have a lot of ideas about the replacement 😅 
That's why I've marked it as optional 💪

semantic-ui/site/collections/breadcrumb.overrides
```diff
@@ -1,3 +1,19 @@
...
+    .page {
+        .pageDescription {
+            font-size: 14px;
+            padding: 0;
+            color: @grey;
+            font-weight: normal;
+            opacity: 1
+        }
+    }
+}
```
Vorbert-Kruk, line 15: As the purpose of files with `.overrides` extension is to globally style components, we should not include here styling related to a specific case scenario
This kind of styling should be done on the component level

From what I've found out, the `pageDescription` class name is only related to one instance of the `EditableLabel` (which is placed inside the `Page.tsx` component)

So we should be able to make a specific styling applied only to that component instance, not globally to the `Breadcrumbs` component 🎉 

Just to give you an additional context, styling related to the `pageDescription` is being stored [in this stylesheet](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/styles/style.scss#L44)
We can try to modify it **or** drop it in order to create a dedicated styled component 🎉

Vorbert-Kruk:
I've got some more comments, regarding differences between mockups and application presentation     

   
## Link inside the breadcrumbs is being colored differently, than on mockups

### Current branch
image
### Mockup 
image

### More context
It seems that current branch link is being colored with the usage of the following selector
```css
.ui.breadcrumb a {
    color: #4183C4;
}
```

Expected color by the mockups is `#65adff` which is available under `@blue` variable in theming files 💪 

## Divider seems to not be styled as shown in the mockups

### Current branch
image

### Mockup
image

### More context
Divider is being styled with the usage of `.ui.breadcrumb .divider` selector, so we should be able to style it within `breadcrumbs.overrides` file 💪 t

app/components/Page.tsx
PatrycjaKubiczek: @Vorbert-Kruk WDYT about changing the name to PageHeader? we have StyledContianer, we page PageContent so PageHeader seems like a good fit

app/components/Page.tsx
PatrycjaKubiczek: good suggestion - applied it here: https://github.com/cloudify-cosmo/cloudify-stage/pull/2154/commits/2c8eb4c4a69ae37511e28f6c8d3126a827478b56

semantic-ui/site/collections/breadcrumb.overrides
PatrycjaKubiczek: change styles placement: https://github.com/cloudify-cosmo/cloudify-stage/pull/2154/commits/da82c98119650e0cb26e8a6eea3f874fe33e0198

semantic-ui/site/elements/label.overrides
```diff
@@ -1,3 +1,7 @@
 /*******************************
          Site Overrides
 *******************************/
+
+.ui.label.pageDescription {
```
qooban, line 5: I'm not sure if we should add this kind of app-specific classes (`pageDescription`) to the SUIR components overrides 🤔  Maybe in general stylesheet or even better pass it to `EditableLabel` component? 🤔

qooban (commented):
Checked UI - looks good in general. Added just two comments.

app/components/Breadcrumbs.tsx
```diff
@@ -51,24 +59,18 @@ export default function Breadcrumbs({ isEditMode, onPageNameChange, onPageSelect
                         <Breadcrumb.Section link onClick={() => onPageSelected(page, pagesList, pageIndex)}>
                             <TextEllipsis maxWidth="300px">{page.name}</TextEllipsis>
                         </Breadcrumb.Section>
-                        <Breadcrumb.Divider key={`d_${page.id}`} icon="right angle" />
+                        <Breadcrumb.Divider key={`d_${page.id}`} />
                     </React.Fragment>
                 ))}
                 {editablePage && (
-                    <EditableLabel
+                    <StyledEditableLabel
```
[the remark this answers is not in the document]
Vorbert-Kruk: It's worth to mention that the same thing is appearing on the [staging](https://staging.trn.cloudify.co/console/login) 😁 
So there is a big chance that the issue has been there for a while

semantic-ui/site/elements/label.overrides
Vorbert-Kruk: Yea it may not be ideal, but we've got an issue related to reusing color from the newly themed pallete

To avoid putting this class selectors we can do 2 things:
- style the specific class with the hardcoded hex color
- reflect currently implemented colors (implemented in `site.overrides` file) in some sort of **JS** or **SCSS** variable

WDYT about it? 🤔 

And in addition, @PatrycjaKubiczek if you'd have any other suggestions feel free to mention them 💪

app/components/Page.tsx
Vorbert-Kruk: @PatrycjaKubiczek 
`StyledPageHeader` seems like a very nice solution 🎉 
Nicely thought 🎸

semantic-ui/site/collections/breadcrumb.overrides
```diff
@@ -1,3 +1,18 @@
 /*******************************
          Site Overrides
 *******************************/
+
+.ui.breadcrumb {
+    a {
+        color: @blue;
```
Vorbert-Kruk, line 7: WDYT about adding this declaration also to the `site.variables`?    

I can see that in the UI `.ui.breadcrumb a` selector is responsible for stating the color of a links inside of it

image

But as `.ui.breadcrumb a` was providing the same text color for the links as were `a` selector, maybe it would be beneficial to align it?

WDYT? 🤔

Vorbert-Kruk (commented):
Looks good 💪 
Added some comments and contributed to the discussions started by Kuba 🎉

semantic-ui/site/elements/label.overrides
PatrycjaKubiczek: @qooban yes, @Vorbert-Kruk is correct, I did it because of @grey variable which is used as part of SUIR changes - ofc I'm open to another solution, but I don't know if others mentioned by Norbert are better :( 
WDYT?

app/components/Page.tsx
PatrycjaKubiczek: renamed it to StyledPageHeader 👍 https://github.com/cloudify-cosmo/cloudify-stage/pull/2154/commits/4031e36448f3f10a8a9a8c8db3711d33351fec8d

app/components/Breadcrumbs.tsx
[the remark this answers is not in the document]
PatrycjaKubiczek: fix the issue with EditableLabel in Breadcrumbs and also improve behavior of Page Description's input in editMode (there was a difference between spacing so it made the layout jump a bit):
https://user-images.githubusercontent.com/17828621/188082460-a3cf9217-98dc-4400-aa5b-31e20317d08b.mp4

semantic-ui/site/collections/breadcrumb.overrides
PatrycjaKubiczek: @Vorbert-Kruk do you mean simply:

```
a {
  color: @blue;
}
```
to add to `site.variables` besides `breadcrumb.overrides`?

app/components/Breadcrumbs.tsx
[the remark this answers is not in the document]
qooban: > It's worth to mention that the same thing is appearing on the [staging](https://staging.trn.cloudify.co/console/login) grin
> So there is a big chance that the issue has been there for a while

Indeed, I wasn't aware. In such cases we can omit fixing it, but as Patrycja managed to fix it at once, then great 🙂

semantic-ui/site/collections/breadcrumb.overrides
qooban: I think your idea @PatrycjaKubiczek is good as I believe we should use the same blue color for all links, but let's wait for @Vorbert-Kruk opinion.

qooban (approved):
All fine from my perspective 👍🏻 

Let's wait for @Vorbert-Kruk .

Side comment - @Vorbert-Kruk @PatrycjaKubiczek - maybe worth to rebase [RD-3478-theme-changes](https://github.com/cloudify-cosmo/cloudify-stage/tree/RD-3478-theme-changes) branch from time to time in order to:
1. have fully working CI 
2. be up to date with the latest changes

semantic-ui/site/elements/label.overrides
qooban: Oh, missed that discussion. I guess we may need to access colors from the newly themed pallete in the future, so maybe it'd be worth to expose them through some sort of variables (second bullet in @Vorbert-Kruk comment).

Probably something like we have already here:
https://github.com/cloudify-cosmo/cloudify-ui-common/blob/40f666d98c4bd0e3ea1da3bf5c67321a2516de5b/styles/_colors.scss

I guess for the time being we could have something similar in `cloudify-stage` and in the future we could unify it.

semantic-ui/site/collections/breadcrumb.overrides
Vorbert-Kruk: @PatrycjaKubiczek, @qooban Yeah I was thinking about something like this (but in `site.overrides`, not `site.variables`) :)

But as you've mentioned `site.variables` file, [I've found out that there is](https://github.com/Semantic-Org/Semantic-UI-LESS/blob/master/themes/default/globals/site.variables#L141) a `@linkColor` variable, which we may use to adjust all link colors 🎉 

So we should be able to just provide the following declaration
```typescript
@linkColor: @blue;
```

to the `site.variables` file 🎸 

WDYT? 😉

Vorbert-Kruk:
@qooban

> Side comment - maybe worth to rebase [RD-3478-theme-changes](https://github.com/cloudify-cosmo/cloudify-stage/tree/RD-3478-theme-changes) branch from time to time

Yeah, it seems like a good idea 🚀 
I'll allign master and theming branches shortly 🎸 

In the meantime I'll try to come up with a plan/rules, which could help us with making an alignment as frequently as it would be beneficial 🎉 
Thanks for pointing that out!

semantic-ui/site/collections/breadcrumb.overrides
qooban: Makes sense 👍🏻

semantic-ui/site/elements/label.overrides
Vorbert-Kruk: Seems good to me 💪 
Having that approach in mind, we'd need to just align styling in [styles.scss](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/styles/style.scss#L44), so it'd be using freshly exposed color 🎸

PatrycjaKubiczek: created additional file for new colors from Semantic UI - in the future we could add it to Common repo or adjust the original colors.

semantic-ui/site/collections/breadcrumb.overrides
PatrycjaKubiczek: added `@linkColor: @blue;` and removed colors from `.ui.breadcrumb a`: 
https://github.com/cloudify-cosmo/cloudify-stage/pull/2154/commits/9ace9dbfb34b7e79705f0b1e30441aab5c02b9ad

app/styles/colors.scss
```diff
@@ -0,0 +1,15 @@
...
+$suir-grey: #5f6368;
+$suir-black: #050229;
+
+:export {
+    suirRed: $suir-red;
+    suirYellow: $suir-yellow;
+    suirBlue: $suir-blue;
+    suirGrey: $suir-grey;
+    suirBlack: $suir-black;
+}
```
Vorbert-Kruk, line 15: Nice idea about exporting variables, so that they may be accessed even from the component level

But as we are not using them anywhere, I think that we can safely remove this declaration
(keeping in mind, that in the future we may use this approach, if we'd have a usecase for it 🎉 )

```suggestion

```

app/styles/colors.scss
```diff
@@ -0,0 +1,15 @@
+/*--- Semantic UI Site Colors  ---*/
+
+$suir-red: #ef5350;
+$suir-yellow: #fee584;
+$suir-blue: #65adff;
+$suir-grey: #5f6368;
+$suir-black: #050229;
```
Vorbert-Kruk, line 7: WDYT about renaming colors to have a `theme` or `theme-color` prefix?
Currently those names are a little bit inaccurate, as those colors are not coming out of SUIR but are customized via theme, to fit our needs

PatrycjaKubiczek: @Vorbert-Kruk hmm so if there are customized theme colors, maybe we should use `cloudify-` prefix? 🤔

Vorbert-Kruk: As not a long time ago, we've implemented a new, custom font named `cloudify-icons`, I think that it's indeed a valid option! 🎉 

@qooban WDYT about it?
(I just wanted to ask about it, as we've had some conversations about prefixing different assets with `cloudify-` prefix 😉 )

qooban: Yeah, sounds good to me 👍🏻 

In the future (far, far in the future I think ;-) ) I think we'll need to merge `app/styles/colors.scss` with https://github.com/cloudify-cosmo/cloudify-ui-common/blob/master/styles/_colors.scss , so I suggest to write down somewhere all this kind of intermediate solutions.

PatrycjaKubiczek: adjusted colors names: https://github.com/cloudify-cosmo/cloudify-stage/pull/2154/commits/405dff23675982307b5cbf0090d860dc54a496c7

app/styles/colors.scss
PatrycjaKubiczek: okay, sounds good, removed (for now at least :)) https://github.com/cloudify-cosmo/cloudify-stage/pull/2154/commits/405dff23675982307b5cbf0090d860dc54a496c7

Vorbert-Kruk (approved):

PR #2160, opened 2022-09-02
RD-5605 Add a description to secrets properties

PatrycjaKubiczek:
## Description

RD-5605
- added label to Form Field
- added test
- mockup description since we don't have a description set yet

PS. could we create another ticket to fix the placement of Popup? I think it should be above (?) icon. Currently is a bit off to the right. 

## Screenshots / Videos

Zrzut ekranu 2022-09-02 131101

## Checklist

## Tests

 

https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3526/

## Documentation

n/a

### DIFF
--- a/app/components/GettingStartedModal/steps/SecretsStep.tsx
+++ b/app/components/GettingStartedModal/steps/SecretsStep.tsx
@@ -4,15 +4,12 @@ import { Form } from '../../basic';
 import { useInputs } from '../../../utils/hooks';
 
 import type { GettingStartedSecretsData, GettingStartedSchemaItem } from '../model';
-import StageUtils from '../../../utils/stageUtils';
 
 type Props = {
     selectedEnvironment: GettingStartedSchemaItem;
     typedSecrets?: GettingStartedSecretsData;
     onChange: (typedSecrets: GettingStartedSecretsData) => void;
 };
-
-const t = StageUtils.getT('gettingStartedModal');
 
 const SecretsStep = ({ selectedEnvironment, typedSecrets, onChange }: Props) => {
     const defaultSecretInputs: Record<string, any> = useMemo(
@@ -28,8 +25,6 @@ const SecretsStep = ({ selectedEnvironment, typedSecrets, onChange }: Props) => 
     );
 
     const [secretInputs, setSecretInputs, resetSecretInputs] = useInputs(typedSecrets || defaultSecretInputs);
-    const secretDescription = t('secretsDescription');
-    const hasSecretDescrtiption = secretDescription;
 
     useEffect(() => resetSecretInputs(), [typedSecrets]);
     useEffect(() => {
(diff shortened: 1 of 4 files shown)


app/components/GettingStartedModal/steps/SecretsStep.tsx
```diff
@@ -38,11 +43,10 @@ const SecretsStep = ({ selectedEnvironment, typedSecrets, onChange }: Props) =>
                     onChange(secretInputs);
                 };
                 return (
-                    <Form.Field key={name}>
+                    <Form.Field key={name} label={label} help={hasSecretDescrtiption ? secretDescription : undefined}>
```
qooban, line 46: I think we shouldn't add "Secret description" to all secrets. We just need to have the framework ready to support new property.

Regarding change in that line, I'm wondering if we change L41 to:
```
            {selectedEnvironment.secrets.map(({ name, label, type, description }) => {
```
could we have here just:
```suggestion
                    <Form.Field key={name} label={label} help={description}>
```
? 🤔

app/components/GettingStartedModal/steps/SecretsStep.tsx
```diff
@@ -25,6 +28,8 @@ const SecretsStep = ({ selectedEnvironment, typedSecrets, onChange }: Props) =>
     );
 
     const [secretInputs, setSecretInputs, resetSecretInputs] = useInputs(typedSecrets || defaultSecretInputs);
+    const secretDescription = t('secretsDescription');
+    const hasSecretDescrtiption = secretDescription;
```
qooban, line 32: Do we need it? 

1. We don't have secret descriptions yet defined, and the goal of RD-5605 was to add support for `description` property. No need to define default value for all secrets, I think.

2. `hasSecretDescription` variable name suggests that this is a boolean containing information if certain secret has description defined. The value is a string "Secret description".

qooban (commented):
On the visual side - I think all is fine.

On the implementation side - I think we don't need to add any default help descriptions right now, we should just add support for new property.

Reg.
> PS. could we create another ticket to fix the placement of Popup? I think it should be above (?) icon. Currently is a bit off to the right.

Yes, you're right. It looks like generic problem with `PopupHelp` component. Feel free to create a ticket for that.

app/components/GettingStartedModal/steps/SecretsStep.tsx
PatrycjaKubiczek: https://github.com/cloudify-cosmo/cloudify-stage/pull/2160/commits/53662860fd59f312c20c62992405bb3b63c24b0f applied here :)

app/components/GettingStartedModal/steps/SecretsStep.tsx
PatrycjaKubiczek: thanks for the clarification, I think I misunderstood the ticket's goal a bit :) implemented changes

qooban (commented):
Looks good 🙂 

Just tests need to be fixed (wrote some ideas on Slack).

PatrycjaKubiczek:
@qooban so I decided to remove the test for now, as we would need to mock up the new secret. I think it'll be very easy to test after adding description - and as it is planned to add description anyway in the future, I'll leave it at it is right now, if you don't mind :)

qooban (approved):

PR #2165, opened 2022-09-07
RD-5606 Add a description to secrets modal

PatrycjaKubiczek:
## Description

RD-5606 (in ticket's description there is information about page, but we have only modal with secrets step in gettingStartedModal)
- add Message component to Modal Content
- add secretsPageDescription

## Screenshots / Videos

Zrzut ekranu 2022-09-07 103556

## Checklist

## Tests

 

https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3540/

## Documentation

n/a


Vorbert-Kruk (approved):
LGTM 🎸 

nicely-done

Vorbert-Kruk:
@PatrycjaKubiczek 
> in ticket's description there is information about page, but we have only modal with secrets step in gettingStartedModal
Just for more context:   

Management is usually referring to something that they can see and it's taking most of the screen size, as a page 😅 
It's good to be aware of it as it may bring less confusion over time 😁 

But remember that you can always raise a question about it, if something would seems to be unclear 🎉  
It's good to be aware, but sometimes it's even better to be certain! 😁

app/components/GettingStartedModal/ModalContent.tsx
```diff
@@ -45,12 +45,19 @@ const ModalContent = ({
...
+    const { Message } = Stage.Basic;
+
     return (
         <Modal.Content style={{ minHeight: 220, flexDirection: 'column' }}>
             {stepName === StepName.Welcome && <WelcomeStep />}
             {stepName === StepName.Environments && (
                 <EnvironmentsStep schema={schema} onChange={onEnvironmentsStepChange} />
             )}
+            {stepName === StepName.Secrets && secretsStepPageDescription && (
+                <Message>{secretsStepPageDescription}</Message>
```
qooban, line 59: How HTML links provided as part of `secretsPageDescription` text field are rendered with that solution?
If I'm not mistaken with such solution it'll be raw text, am I correct?

I think we should add support to that as in RD-5606 Alex provided such description:
> For example, for AWS, GCP, Azure I can guide a user to go to a certain link to create the cloud credentials.

PatrycjaKubiczek: you're 100% correct, added HTML support: https://github.com/cloudify-cosmo/cloudify-stage/pull/2165/commits/06670f40e9e018958f2ab3e32b7e02e766ad0e3c

qooban (approved):
Thanks for update, that's exactly it 👍🏻 

As we all know using `dangerouslySetInnerHTML` is not the best option, but for this purpose I think we should be fine.

PR #2167, opened 2022-09-08
RD-5894 Remove Deployment Wizard Buttons widget

qooban:
## Description
As described in RD-5894 this widget was not developed for a long time and it was decided by Alex to remove it in v7.

This change will solve also RD-5857.

## Screenshots / Videos
N/A

## Checklist

## Tests
CI check should be enough, so no need to run Stage system tests.

## Documentation
Widget documentation page removal: https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2323


test/jest/resources/rbac.ts
```diff
@@ -146,7 +146,6 @@ export default {
         administrators: ['sys_admin', 'manager'],
         cluster_start_or_join: ['sys_admin'],
         manager_rest_config_get: ['sys_admin'],
-        widget_deploymentWizardButtons: ['sys_admin', 'manager', 'user'],
```
Vorbert-Kruk, line 149: As we are removing permissions related to the widget, we should also address it in [cloudify-manager-install](https://github.com/cloudify-cosmo/cloudify-manager-install/blob/master/cfy_manager/components/restservice/config/authorization.conf#L918)

Vorbert-Kruk (approved):
LGTM 💪 
Checked out a repo and it seems that there are no residues after this removal 🚀 
It's good to see that we're not only focusing on adding some code, but also on removing code that we don't need 😄 

Lefted one comment about an additional PR which creation may be required

test/jest/resources/rbac.ts
qooban: Great point. Will create PR to `cloudify-manager-install` right away 👍🏻

PR #2171, opened 2022-09-12
RD-6005 RD-2873 Use multipart form data for blueprint upload params

geokala:
The backend is being changed to do this to avoid silliness around encoding query params.


geokala:
This is part of the following batch of PRs:
https://github.com/cloudify-cosmo/cloudify-common/pull/1144
https://github.com/cloudify-cosmo/cloudify-manager/pull/3826
https://github.com/cloudify-cosmo/cloudify-cli/pull/1439
https://github.com/cloudify-cosmo/cloudify-stage/pull/2171

As noted in several places, it's to avoid us having to do silly things with encoding of query params (and thus make the new snapshots not require more ugliness).

qooban:
Let's check it with Jenkins. I updated this branch with latest master and triggered system tests - [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3588/)

qooban:
@geokala Some of the blueprint upload tests failed, so I guess we'll need someone from the UI team to provide the solution. Do we have dedicated task for that, so we can add it to the next sprint?

Vorbert-Kruk:
@qooban, I've managed to get this thing going! 🎉 
Locally everything is passing as expected adjusted the PR and locally

Vorbert-Kruk:
One more thing 😁 
[System Tests](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3619/) has just passed and it seems that the blueprint issues are indeed not appearing anymore 🥳

test/cypress/support/blueprints.ts
```diff
@@ -16,24 +16,53 @@ interface UploadBlueprintOptions {
...
+    return cy
+        .fixture(filePath, 'binary')
+        .then(binary => Cypress.Blob.binaryStringToBlob(binary))
+        .then(fileContent => {
+            const formData = new FormData();
+            formData.append('blueprint_archive', fileContent);
+            formData.append('params', JSON.stringify(parameters));
+
+            return cy.doXhrPutRequest(`/blueprints/${blueprintId}`, formData, timeout);
+        });
```
Vorbert-Kruk, line 40: ### NOTE
As we are having an eslint rule, which disallows assigning a value to a cypress command results, I've went with `.then` approach, to minimize the usage of `// eslint-disable-next-line`

test/cypress/support/blueprints.ts
```diff
@@ -16,24 +16,53 @@ interface UploadBlueprintOptions {
     timeout?: number;
 }
 
+const uploadBlueprint = (blueprintId: string, parameters: Record<string, any>, timeout?: number) => {
```
qooban, line 19: We can improve `parameters` type here, in L28 and in L49 as there's limited set of allowed query params for `PUT /blueprints`.

See: https://docs.cloudify.co/api/staging/dev/#upload-blueprint

test/cypress/support/blueprints.ts
```diff
@@ -16,24 +16,53 @@ interface UploadBlueprintOptions {
...
+    parameters: Record<string, any>,
+    timeout?: number
+) => {
+    return cy
+        .fixture(filePath, 'binary')
+        .then(binary => Cypress.Blob.binaryStringToBlob(binary))
+        .then(fileContent => {
+            const formData = new FormData();
+            formData.append('blueprint_archive', fileContent);
+            formData.append('params', JSON.stringify(parameters));
```
qooban, line 37: Probably we can extract common function for creating form data for `PUT /blueprints` calls (L20-L21 and L35-37).

As similar code is in `widgets/common/src/blueprints/BlueprintActions.ts`, probably the right place for it would be in source code and then we can use it in Cypress test.

backend/package-lock.json
```diff
@@ -23381,7 +23381,7 @@
             "resolved": "https://registry.npmjs.org/axios/-/axios-0.27.2.tgz",
             "integrity": "sha512-t+yRIyySRTp/wua5xEr+z1q60QmLq8ABsS5O9Me1AsE5dfKqgnCFzwiCZZ/cGNd1lq4/7akDWMxdhVlucjmnOQ==",
             "requires": {
-                "follow-redirects": "^1.15.2",
+                "follow-redirects": "^1.14.9",
```
qooban, line 23384: I guess that was changed by mistake? 🤔 Let's revert it.

It's important to have `follow-redirects` in version >=1.15.2.
See RD-5965 or RD-5663 for more details.

app/utils/External.ts
```diff
@@ -165,8 +168,6 @@ export default class External {
                         const reader = new FileReader();
                         const zip = new JSZip();
 
-                        xhr.setRequestHeader('content-type', 'application/octet-stream');
-
```
qooban, line 169: Taking into account note from L247-249 browser will adjust content-type not only for multipart data, but also when we are sending file as payload? (so it'll still be `application/octet-stream` even though not explicitly provided)

app/utils/External.ts
```diff
@@ -141,7 +145,6 @@ export default class External {
                         log.error('xhr upload error', e, xhr.responseText);
 
                         reject({ message: StageUtils.resolveMessage(response.message) });
-                        return;
```
qooban, line 144: Without it won't we call reject first and then resolve (L154)? 🤔

qooban (approved):
@Vorbert-Kruk Thanks for your effort on this change. I've added some comments, but IMHO it looks good enough to merge it and as you mentioned you can address my comments on Monday.

@geokala Are you planning also to adjust `POST /plugins` as well? I see it share similar approach - request body for file payload and have query params as well. Maybe worth to transform it to multipart form data as well?

backend/package-lock.json
Vorbert-Kruk: Addressed it in [6d29396](https://github.com/cloudify-cosmo/cloudify-stage/pull/2198/commits/6d293967d30fcdee7ce6cb8fd50451194d777788)

test/cypress/support/blueprints.ts
Vorbert-Kruk: Good idea!
Implemented it right away, as [693f8ed](https://github.com/cloudify-cosmo/cloudify-stage/pull/2198/commits/693f8ed658948e0ce8defa374f45580d46f60613)

test/cypress/support/blueprints.ts
Vorbert-Kruk: Good idea! 🎸 
Done in [84a8487](https://github.com/cloudify-cosmo/cloudify-stage/pull/2198/commits/84a848785b75f0efd952094e31f73a12572c8a7b) 🎉

app/utils/External.ts
Vorbert-Kruk: Yeah, you're right
I've removed it by accident (during development I've had put many return statements and as we can see ... I've also removed to much of them 😅)

Issue has been addressed in [2145c57](https://github.com/cloudify-cosmo/cloudify-stage/pull/2198/commits/2145c57c37e15d6b0deb4c3f87150381a53fdf8d)

app/utils/External.ts
Vorbert-Kruk: I'll answer on #2198, as there may be some room for a discussion 😁

PR #2173, opened 2022-09-12
RD-5665 Add configuration option for maximum manager forwarded request size

kubama:
## Description
Added configuration option for maximum manager forwarded request size

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/3561 - failures look unrelated to the change and are occurring on other branches as well - I consider it a pass

 

## Documentation
Updated README.md


qooban (approved):
Adding configuration option for maximum body size looks good 👍🏻 

But is that fixing the issue described in RD-5665? Was there a problem with default value for `limit`? 🤔 
(on https://expressjs.com/en/api.html#express.raw I see the default for `express.raw` is `100kb`, so probably that's not the case as with such limit most of the snapshots won't work)

PatrycjaKubiczek (approved):

kubama:
> But is that fixing the issue described in [RD-5665](https://cloudifysource.atlassian.net/browse/RD-5665)?

As far as I tested - yes.

> Was there a problem with default value for `limit`? thinking (on https://expressjs.com/en/api.html#express.raw I see the default for `express.raw` is `100kb`, so probably that's not the case as with such limit most of the snapshots won't work)

I wasn't able to upload 500MB snapshot without increasing this option - it looks like it has to be part of the solution.

PR #2177, opened 2022-09-15
RD-5965 Bump `axios` and `follow-redirects` packages in backend to make Stage DB migration more robust

qooban:
## Description

A proper fix for app-crashing issue (https://github.com/follow-redirects/follow-redirects/issues/211 and RD-5663) was implemented in `follow-redirects` package and new version was released. Bumped `axios` package to get the latest `follow-redirects`.

Noticed that axios v1 is coming, so created RD-5969.

## Screenshots / Videos
N/A

## Checklist

## Tests
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3569/) - failed due to known issues (RD-5918, RD-5921, RD-5922), so I'm considering it as PASSED

## Documentation
N/A

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -97,9 +97,6 @@
         "sequelize": "^6.17.0",
         "sequelize-cli": "^6.4.1"
     },
-    "overrides": {
-        "follow-redirects": "^1.15.2"
-    },
     "engines": {
         "node": "16.x"
     },
--- a/backend/package.json
+++ b/backend/package.json
@@ -32,7 +32,7 @@
         "@evops/hcl-terraform-parser": "^1.0.0",
         "@tsconfig/node14": "^1.0.1",
         "archiver": "^5.3.0",
-        "axios": "^0.26.0",
+        "axios": "^0.27.2",
         "body": "^5.1.0",
         "cloudify-ui-common": "^7.3.1",
         "cookie": "^0.4.2",
@@ -94,6 +94,9 @@
         "nodemon": "^2.0.15",
         "sequelize": "^6.17.0",
         "sequelize-cli": "^6.4.1"
+    },
+    "overrides": {
+        "follow-redirects": "^1.15.2"
     },
     "engines": {
         "node": "16.x"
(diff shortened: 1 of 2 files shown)


backend/package.json
```diff
@@ -97,6 +97,9 @@
         "sequelize": "^6.17.0",
         "sequelize-cli": "^6.4.1"
     },
+    "overrides": {
+        "follow-redirects": "^1.15.2"
```
kubama, line 101: Would be good to somehow document the reason for this override - once it is no longer needed it should be removed.

kubama (approved):

kubama:
I had a deeper look... please see https://github.com/cloudify-cosmo/cloudify-stage/pull/2178#issuecomment-1249016141 as I suspect it may also apply here. I think `override` may not be needed.

backend/package.json
qooban: Adding comments in `package.json` is problematic, but as I mentioned in the description:

> ... I have created [RD-5969](https://cloudifysource.atlassian.net/browse/RD-5969) and added a note to remove overrides once we move to axios v1.

qooban:
> I had a deeper look... please see [#2178 (comment)](https://github.com/cloudify-cosmo/cloudify-stage/pull/2178#issuecomment-1249016141) as I suspect it may also apply here. I think `override` may not be needed.

Hmm... That was my original approach, but unfortunately `npm` did not pick up the latest version of `follow-redirects`, so I had to add `override`, but... I removed `override` right now and seems like it just works 🙂

kubama (approved):

backend/package.json
qooban: After recent changes (https://github.com/cloudify-cosmo/cloudify-stage/pull/2177/commits/f7d7b83e2d2cd4712cadf6067c557d392c17b889) removed that note from [RD-5969](https://cloudifysource.atlassian.net/browse/RD-5969).

PR #2178, opened 2022-09-15
RD-5965 Bump `follow-redirects` package in cloudify-stage backend

qooban:
## Description
It's similar change to #2177 , but with only `follow-redirects` version bump.

## Screenshots / Videos
N/A

## Checklist

## Tests
Tested manually in real production environment.

## Documentation
N/A

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -26,6 +26,7 @@
                 "express": "^4.14.0",
                 "express-static-gzip": "^0.3.2",
                 "flat": "^4.1.1",
+                "follow-redirects": "^1.15.2",
                 "fs-extra": "^3.0.1",
                 "js-beautify": "1.6.14",
                 "js-yaml": "^3.13.1",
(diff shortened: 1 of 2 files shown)


kubama:
Hmm... I'm not sure this is the way to go. `axios` uses `follow-redirects@^1.14.x`, so `1.15.2` fits. If so, isn't it sufficient to perform just an update to bump version in `package-lock`?

qooban:
> Hmm... I'm not sure this is the way to go. `axios` uses `follow-redirects@^1.14.x`, so `1.15.2` fits. If so, isn't it sufficient to perform just an update to bump version in `package-lock`?

As mentioned in https://github.com/cloudify-cosmo/cloudify-stage/pull/2178 I was not able to bump it without explicitly specifying the expected version, but... Seems like the following:
```
npm install --save-dev follow-redirects@latest
npm remove follow-redirects
```
is doing the trick 🙂

kubama (approved):

PR #2185, opened 2022-09-22
RD-4255 Made dynamic dropdown reset search query on focus

Blewin:
## Description
<! -- Describe your changes to help reviewers understand the details. -->

## Checklist

- [] I followed the [UI Testing Guidelines](https://cloudifysource.atlassian.net/l/cp/udCVfvrx) and verified that all tests/checks have passed.

## Tests

 
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3587/) known issues


widgets/common/src/components/DynamicDropdown.tsx
```diff
@@ -307,6 +307,10 @@ export default function DynamicDropdown({
                     setSearchQuery(data.searchQuery);
                     if (isFunction(onSearchChange)) onSearchChange(event, data);
                 }}
+                onFocus={(event, data) => {
```
Blewin, line 310: I had mixed feelings about making some common funciton for this and the above, since data override would make it problematic

Blewin: any comments?

qooban: > I had mixed feelings about making some common funciton for this and the above, since data override would make it problematic

Agree with that. But let's leave that for [RD-2591](https://cloudifysource.atlassian.net/browse/RD-2591). I see that I added there a point about `onSearchChange`, so it's already part of the considered refactoring.

Getting back to the original issue:

1. Adding `onFocus` for all cases seems to brake behavior in `LabelsList` component, so we need to limit it. 

2. I did some research and found that in the original component - [Dropdown](https://react.semantic-ui.com/modules/dropdown/), search query clearing is done on blur (ref. [Dropdown.js#L535](https://github.com/Semantic-Org/Semantic-UI-React/blob/0e1a55ae671bccf9310cf96f974977f6b6f21bca/src/modules/Dropdown/Dropdown.js#L535)).
I think we can do something similar and instead of taking action `onFocus` clear search query in `onBlur`.

3. There's only one case in our application where we explicitly define `onSearchChange` prop for `DynamicDropdown`. It's in [CommonDropdown.tsx#L82](https://github.com/cloudify-cosmo/cloudify-stage/blob/c7ab13234a2c5edbefdf57e9b193c4c6d5c5767c/widgets/common/src/labels/CommonDropdown.tsx#L82) and as I see in that case (adding/modyfing deployment labels) we should not clear search query.

4. Taking all previous points into account, maybe instead of overridding `onFocus` we could do:
```
onBlur={(event, data) => {
    if (!isFunction(onSearchChange)) setSearchQuery('');
    if (isFunction(rest.onBlur)) rest.onBlur(event, data);
}}
```
(of course `onBlur` can be extracted from `rest` spread) 

5. I think we can have `clearable={false}` in [GenericDeployModal#L671](https://github.com/cloudify-cosmo/cloudify-stage/blob/c7ab13234a2c5edbefdf57e9b193c4c6d5c5767c/widgets/common/src/deployModal/GenericDeployModal.tsx#L671) as the value for this dropdown is required, so no need to confuse users ;-) 

After all - once Jenkins is operational again - we need to trigger system tests to see if there are other dropdown cases which we didn't take into account.

qooban (commented):
Added comment.

qooban:
@Blewin Let's check the change with system tests:
1. Go to [Stage-UI-System-Test job](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/)
2. Click "Run with parameters"
3. Set this PR's branch name (`RD-4255`) in `STAGE_BRANCH`
4. Build!

I usually embed link with status in "Tests" section. You can find the link on "Embeddable Build Status" page, accessible from specific build page.

qooban (approved):

PR #2186, opened 2022-09-23
RD-5788 Migrate User Management widget to TypeScript

qooban:
## Description
Migrated User Management widget to TypeScript and added full translation support.

## Screenshots / Videos
N/A

## Checklist

## Tests
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3585/) - failing on known issues only, so... PASSED

## Documentation
N/A


widgets/userManagement/src/IsAdminCheckbox.tsx
```diff
@@ -0,0 +1,71 @@
+import type { FunctionComponent } from 'react';
```
qooban, line 1: Extracted from `UsersTable` component.

widgets/userManagement/src/CreateModal.tsx
```diff
@@ -1,24 +1,38 @@
...
-
+import type { FunctionComponent } from 'react';
 import Actions from './actions';
-import { mapTenantsToRoles } from '../../common/src/tenants/utils';
+import type { CancelablePromise } from '../../../app/utils/types';
+import type { RolesPickerProps } from '../../common/src/roles/RolesPicker';
+import type { TenantItem, TenantsDropdownProps } from '../../common/src/tenants/TenantsDropdown';
+import getWidgetT from './getWidgetT';
+
+const modalT = (key: string) => getWidgetT()(`createModal.${key}`);
```
kubama, line 8: I think to be in line with common approach this function should be called `tModal`

widgets/userManagement/src/MenuAction.tsx
```diff
@@ -1,30 +1,33 @@
...
+    EDIT_TENANTS_ACTION = 'EDIT_TENANTS_ACTION',
+    EDIT_GROUPS_ACTION = 'EDIT_GROUPS_ACTION',
+    DELETE_ACTION = 'DELETE_ACTION',
+    DEACTIVATE_ACTION = 'DEACTIVATE_ACTION',
+    ACTIVATE_ACTION = 'ACTIVATE_ACTION',
+    SET_DEFAULT_USER_ROLE_ACTION = 'SET_DEFAULT_USER_ROLE_ACTION',
+    SET_ADMIN_USER_ROLE_ACTION = 'SET_ADMIN_USER_ROLE_ACTION',
+    ENABLE_GETTING_STARTED_MODAL_ACTION = 'ENABLE_GETTING_STARTED_MODAL_ACTION',
+    DISABLE_GETTING_STARTED_MODAL_ACTION = 'DISABLE_GETTING_STARTED_MODAL_ACTION'
+}
```
kubama, line 19: Do we need values to be set if they are all equal to the keys?

widgets/userManagement/src/TenantModal.tsx
```diff
@@ -1,16 +1,30 @@
...
 const RolesPicker = Stage.Common.Roles.Picker;
 const { getDefaultRoleName } = Stage.Common.Roles.Utils;
 
-export default function TenantModal({ onHide, open, user, toolbox, tenants }) {
+export interface TenantModalProps {
+    onHide: () => void;
+    open: boolean;
+    tenants: {
+        items: { name: string }[];
+    };
```
kubama, line 18: Maybe we could simplify it to `tenants: string[]`?

widgets/userManagement/src/actions.ts
```diff
@@ -1,31 +1,42 @@
-// @ts-nocheck File not migrated fully to TS
+import type { SystemRole } from '../../common/src/roles/types';
+import type { Toolbox } from '../../../app/utils/StageAPI';
+import type { RolesAssignment } from '../../common/src/tenants/utils';
+
+export type ResourceWithNameResponse = Stage.Types.PaginatedResponse<{ name: string }>;
```
kubama, line 5:
```suggestion
export type NamedResourceResponse = Stage.Types.PaginatedResponse<{ name: string }>;
```

widgets/userManagement/src/widget.tsx
```diff
@@ -1,11 +1,25 @@
-// @ts-nocheck File not migrated fully to TS
-
 import UsersTable from './UsersTable';
+import type { User, UserManagementWidget } from './widget.types';
+import getWidgetT from './getWidgetT';
+
+export interface ExtendedUser extends User {
```
kubama, line 5: How about:
```suggestion
export interface UserViewItem extends User {
```

widgets/userManagement/src/CreateModal.tsx
qooban: That's right. This is what we have in [UI Translation Files Guidelines](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/1875083446/UI+Translation+Files+Guidelines). Fixed and at once fixed other places (found inconsistency in Deployment View widget).

widgets/userManagement/src/MenuAction.tsx
qooban: Yes, with the current implementation we need it as it is passed to `name` prop of `Menu.Item` component, which is expected to be string.

widgets/userManagement/src/actions.ts
qooban: OK, changed.

widgets/userManagement/src/widget.tsx
qooban: Sure, done.

widgets/userManagement/src/TenantModal.tsx
qooban: Thought about it originally, but deferred to focus on migration.
I'll check it out. Possibly other places can be adjusted as well.

qooban: Done 🙂

widgets/userManagement/src/MenuAction.tsx
kubama:
```
enum Enum {
    A
}
let nameOfA = Enum[Enum.A]; // "A"
```
Won't it work to obtain the string value?

qooban: It will work to obtain the string value, but won't it make the final solution more complex than it is now?

`invokeAction` in `UsersTable` and `onSelectAction` in `MenuAction` components would need to be changed to expect `string` I guess and we would need mapping function (something like you shared) used in all places instead.

kubama: I'm not sure I understand. Why `invokeAction` in `UsersTable` would need to be changed to accept `string`?

By the way, I think the common approach is to name `enum`s in singular form - I suggest `MenuAction` instead of `MenuActions` as a single enum entry represents a single action, not multiple actions.

qooban: > I'm not sure I understand. Why invokeAction in UsersTable would need to be changed to accept string?

OK, I think I found a way to omit that. Let me know WDYT about https://github.com/cloudify-cosmo/cloudify-stage/pull/2186/commits/35a55d909ed8abcf02084c3a13eaa93253b01240 

> By the way, I think the common approach is to name enums in singular form - I suggest MenuAction instead of MenuActions as a single enum entry represents a single action, not multiple actions.

Changed to `Action` as `MenuActions` is already used by the component.

qooban: Actually, when I look at it right now - the component itself should probably be named `ActionsMenu` instead of `MenuActions`. Then we can have `MenuAction` enum. WDYT?

qooban: I decided to refactor component naming. Let me know WDYT about https://github.com/cloudify-cosmo/cloudify-stage/pull/2186/commits/35a55d909ed8abcf02084c3a13eaa93253b01240 & https://github.com/cloudify-cosmo/cloudify-stage/pull/2186/commits/d7896b15b119a512b8ec5e0bc15a6e12d96ea14c.

kubama: Looks perfect :+1:

kubama (approved):

PR #2197, opened 2022-10-03
RD-5450 new design of Blueprint marketplace

PatrycjaKubiczek:
## Description

RD-5450
zeplin mockups: https://app.zeplin.io/project/612e8cb57788ff10babc7900/screen/62d54f617ecb5513598e4c2e
- changed the design to match mockups

Important notes: 
- I didn't change the buttons and texts colors because I wasn't sure if this should be part of this ticket 🤔 or it should be part of theming
- Also, I'm wondering what color should be on the button on the hover (because as I understand correctly the blue button should appear on the hover on the whole segment, not only on the button). The mockups don't cover this state. 
- Another issue with the icons: currently, we could have a disabled state of "Upload" button which is slightly lighter than active. So if I match the colors of icons to the current design it looks like all the icons and buttons are disabled:
Zrzut ekranu 2022-09-29 125132 (1)

## Screenshots / Videos

Before:
Zrzut ekranu 2022-10-03 134627

After:
Zrzut ekranu 2022-10-03 133554

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/3628/) failed because of known issue with User spec

## Documentation

I'll update the docs as separate PR.

### DIFF
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -990,8 +990,8 @@
             "noDataMessage": "There are no Blueprints available in catalog. Check widget's configuration.",
             "catalog": {
                 "properties": {
-                    "created": "Created",
-                    "updated": "Updated"
+                    "created": "Created:",
+                    "updated": "Updated: "
                 }
             }
         },
@@ -1503,7 +1503,6 @@
                         "note": "Admin users have full permissions to all tenants on the manager."
                     }
                 }
-
             },
             "editGroupsModalHeader": "Edit user groups for {{username}}",
             "editTenantsModalHeader": "Edit tenants for {{username}}",
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -992,7 +992,7 @@
             "catalog": {
                 "properties": {
                     "created": "Created:",
-                    "updated": "Updated: "
+                    "updated": "Updated:"
                 }
             }
         },
--- a/app/translations/en.json
+++ b/app/translations/en.json
@@ -942,6 +942,7 @@
             "name": "Blueprints Catalog",
             "description": "Shows blueprints catalog",
             "actions": {
+                "openBlueprintRepository": "Open blueprint repository",
                 "openDocumentation": "Open documentation",
                 "uploadBlueprint": "Upload blueprint"
             },
--- a/widgets/blueprintCatalog/src/RepositoryTable.tsx
+++ b/widgets/blueprintCatalog/src/RepositoryTable.tsx
@@ -35,22 +35,22 @@ const RepositoryTable: FunctionComponent<RepositoryViewProps> = ({
             noDataMessage={noDataMessage}
         >
             <DataTable.Column
-                label={t('name')}
+                label={t('configuration.fieldsToShow.items.name')}
                 width="25%"
                 show={fieldsToShow.includes(t('configuration.fieldsToShow.items.name'))}
             />
             <DataTable.Column
-                label={t('description')}
+                label={t('configuration.fieldsToShow.items.description')}
                 width="40%"
                 show={fieldsToShow.includes(t('configuration.fieldsToShow.items.description'))}
             />
             <DataTable.Column
-                label={t('created')}
+                label={t('configuration.fieldsToShow.items.created')}
                 width="12%"
                 show={fieldsToShow.includes(t('configuration.fieldsToShow.items.created'))}
             />
             <DataTable.Column
-                label={t('updated')}
+                label={t('configuration.fieldsToShow.items.updated')}
                 width="12%"
                 show={fieldsToShow.includes(t('configuration.fieldsToShow.items.updated'))}
             />
@@ -92,7 +92,7 @@ const RepositoryTable: FunctionComponent<RepositoryViewProps> = ({
                                 name="github"
                                 bordered
                                 onClick={() => Stage.Utils.Url.redirectToPage(html_url)}
-                                title={html_url}
+                                title={t('actions.openBlueprintRepository')}
                             />
                             <Icon
                                 name={isReadmeLoading ? 'spinner' : 'info'}
--- a/widgets/blueprintCatalog/src/RepositoryTable.tsx
+++ b/widgets/blueprintCatalog/src/RepositoryTable.tsx
@@ -7,7 +7,7 @@ import type { RepositoryViewProps } from './types';
 import ExternalBlueprintImage from './ExternalBlueprintImage';
 
 const { DataTable, Icon } = Stage.Basic;
-const t = Utils.getWidgetTranslation('configuration.fieldsToShow.items');
+const t = Utils.getWidgetTranslation();
 
 const RepositoryTable: FunctionComponent<RepositoryViewProps> = ({
     fetchData = noop,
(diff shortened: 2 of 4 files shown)


app/translations/en.json
```diff
@@ -991,8 +991,8 @@
             "noDataMessage": "There are no Blueprints available in catalog. Check widget's configuration.",
             "catalog": {
                 "properties": {
-                    "created": "Created",
-                    "updated": "Updated"
+                    "created": "Created:",
+                    "updated": "Updated: "
```
qooban, line 995: Minor thing - I suspect we don't need space here:
```suggestion
                    "updated": "Updated:"
```

qooban (commented):
Looks good IMHO.

Regarding your concerns:

> I didn't change the buttons and texts colors because I wasn't sure if this should be part of this ticket thinking or it should be part of theming

That's good. This should be part of theming epic.

> Also, I'm wondering what color should be on the button on the hover (because as I understand correctly the blue button should appear on the hover on the whole segment, not only on the button). The mockups don't cover this state.

I would leave it as is.

> Another issue with the icons: currently, we could have a disabled state of "Upload" button which is slightly lighter than active. So if I match the colors of icons to the current design it looks like all the icons and buttons are disabled:

I would leave it as is.

Some UI related comments:
1. I personally don't see a point of selecting the tile right now, but we can leave that functionality as is.
2. I would add tooltips for GitHub icon and (i) icon. WDYT?
3. In Table view action icons have tooltips (except GitHub) and there is an issue with translation - not necessarily in this PR, but worth to fix. Should be quick - the problem is in [widgets/blueprintCatalog/src/RepositoryTable.tsx](https://github.com/cloudify-cosmo/cloudify-stage/blob/a81e017f4426f7a99dc3f9a2bbc3de04987808ca/widgets/blueprintCatalog/src/RepositoryTable.tsx)

Vorbert-Kruk (commented):
> I didn't change the buttons and texts colors because I wasn't sure if this should be part of this ticket 🤔 or it should be part of theming

When it comes to the colors, I think that they'll be adjusted during a theming
But we should be able to modify the `color` prop of the `<Button />` to show a default or a blue button 💪 

With that approach, we wouldn't need to worry about recoloring buttons (and their different states) within the CSS 🎉 

> Also, I'm wondering what color should be on the button on the hover (because as I understand correctly the blue button should appear on the hover on the whole segment, not only on the button). The mockups don't cover this state.
It may be worth to discuss it with Illan if those kind of questions are arising
It's better to be certain than to be wondering 😁 

If I may add some input from my side:
From what do I understand, button should colored as blue, when the user would hover over a specific card (e.g `Deploy Openshift` on [mockups](https://app.zeplin.io/project/612e8cb57788ff10babc7900/screen/62d54f617ecb5513598e4c2e) )
We can achieve this by manipulating the props of the `<Button />` component, so that it would be either `default` (as it is a grey color 😁) or `blue`
If we'd go with that approach, we'd have a different `hover`, `focus`, `active` presentation for the button by default 😉 

**Implementation notes**
I can see that making button modifications purely in CSS may be hard to develop and maintain (especially lookin at the approaching theming changes - which would require some adjustments 😅)
Because of that, we can go with more JS related solution
Here's an example:

```jsx
const RepositoryCatalog = () => {
    const [isHovered, setIsHovered, unsetIsHovered] = useBoolean();
    const buttonColor = isHovered ? 'blue' : undefined;

    <div onMouseEnter={setIsHovered} onMouseLeave={unsetIsHovered}>
        <Button color={buttonColor}>Example</Button>
    </div>;
};
```

It may not look ideal, but it resolves the problem 🎉 
Let me know WDYT, both @PatrycjaKubiczek and @qooban 😉 

> Another issue with the icons: currently, we could have a disabled state of "Upload" button which is slightly lighter than active. So if I match the colors of icons to the current design it looks like all the icons and buttons are disabled:

That's a good candidate for a discussion with Illan 😁 

IMO, we should be good to go with the changes provided within mockups
As you've mentioned, there are not ideal, as there is an issue with disabled button state, but while the current direction is to make user hover over the cards (to see a more colorful cards) the problem won't be as prominent, as upon the user hover, he'll be able to distinguish disabled `<Button />` from the clickable icons 🎉 

Because of that I think that we should be good to go with a changes provided within the mockups 🎸 
But once more, discussion with Illan could be very beneficial here 💪 

Let me know WDYT @PatrycjaKubiczek and @qooban 😉

app/translations/en.json
PatrycjaKubiczek: fixed: https://github.com/cloudify-cosmo/cloudify-stage/pull/2197/commits/78e3f858954df8f13e5240da997fab051143ca12

widgets/blueprintCatalog/src/RepositoryTable.tsx
```diff
@@ -7,7 +7,7 @@ import type { RepositoryViewProps } from './types';
 import ExternalBlueprintImage from './ExternalBlueprintImage';
 
 const { DataTable, Icon } = Stage.Basic;
-const t = Utils.getWidgetTranslation('configuration.fieldsToShow.items');
+const t = Utils.getWidgetTranslation();
```
qooban, line 10: Seems like there's an issue with translation in table headers right now:

image

widgets/blueprintCatalog/src/RepositoryTable.tsx
```diff
@@ -72,7 +88,12 @@ const RepositoryTable: FunctionComponent<RepositoryViewProps> = ({
                         <DataTable.Data>{created_at}</DataTable.Data>
                         <DataTable.Data>{updated_at}</DataTable.Data>
                         <DataTable.Data className="center aligned rowActions">
-                            <Icon name="github" bordered onClick={() => Stage.Utils.Url.redirectToPage(html_url)} />
+                            <Icon
+                                name="github"
+                                bordered
+                                onClick={() => Stage.Utils.Url.redirectToPage(html_url)}
+                                title={html_url}
```
qooban, line 95: Instead of URL maybe let's have something like "Open blueprint repository" (similarly to "Open documentation" used for (i) icon)? 

In both - Table and Catalog - views we should have the same.

qooban (commented):
Thanks for additional work. Added two comments.

Regarding @Vorbert-Kruk 's comment - I actually don't understand if you are suggesting to do more changes within this PR (e.g. change CSS-like styling into JS-like) or it's more like a sharing ideas what we can do in the future 🙂 I'm OK with both approaches, but as you guys have better understanding around styling, I'm leaving it to you to decide. 
IMHO it looks good enough to merge it soon (just translation to be fixed from my point of view). And you can discuss details with Ilan, if needed.

widgets/blueprintCatalog/src/RepositoryTable.tsx
PatrycjaKubiczek: changed title: https://github.com/cloudify-cosmo/cloudify-stage/pull/2197/commits/53995f578ef0e6e464fc9abe0acd9f8f85cf7f33

widgets/blueprintCatalog/src/RepositoryTable.tsx
PatrycjaKubiczek: fix issue with labels: https://github.com/cloudify-cosmo/cloudify-stage/pull/2197/commits/53995f578ef0e6e464fc9abe0acd9f8f85cf7f33

Vorbert-Kruk (approved):
More details related [this comment](https://github.com/cloudify-cosmo/cloudify-stage/pull/2197#pullrequestreview-1131004222) were discussed offline

qooban (approved):
Checked UI and it looks good from my perspective 👍🏻

PatrycjaKubiczek:
@qooban & @Vorbert-Kruk I reach out to Ilan and ask if should change the color to blue or leave at it is and I got approval on current look: 
Zrzut ekranu 2022-10-06 091614
So I'll merge this PR if you're okay with that

Vorbert-Kruk:
> @qooban & @Vorbert-Kruk I reach out to Ilan and ask if should change the color to blue or leave at it is and I got approval on current look: Zrzut ekranu 2022-10-06 091614 So I'll merge this PR if you're okay with that

Sure, it's fine with me 🎉 
Thanks for getting the confirmation! :)

PR #2202, opened 2022-10-06
RD-5748 marketplace as a catalog bug

PatrycjaKubiczek:
## Description

RD-5748

## Checklist

### DIFF
--- a/widgets/blueprintCatalog/src/RepositoryCatalog.tsx
+++ b/widgets/blueprintCatalog/src/RepositoryCatalog.tsx
@@ -81,6 +81,9 @@ const RepositoryCatalog: FunctionComponent<RepositoryViewProps> = ({
     onReadme,
     widget
 }) => {
+    const showUpdatedColumn = widget.configuration.fieldsToShow.includes(t('configuration.fieldsToShow.items.updated'));
+    const showCreatedColumn = widget.configuration.fieldsToShow.includes(t('configuration.fieldsToShow.items.created'));
+
     const catalogItems = data.items.map(item => {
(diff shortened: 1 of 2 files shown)


widgets/blueprintCatalog/src/RepositoryCatalog.tsx
```diff
@@ -117,21 +120,25 @@ const RepositoryCatalog: FunctionComponent<RepositoryViewProps> = ({
                             </Grid.Column>
                         </Grid.Row>
```
qooban, line 122: We have also `Name` and `Description` fields. There should also be a way to show/hide them by using `fieldsToShow` parameter.

PatrycjaKubiczek:
found one bug: Zrzut ekranu 2022-10-07 110943

fix it by adding `flex-wrap: nowrap`
Zrzut ekranu 2022-10-07 111130

qooban (commented):
Change looks good 👍🏻 

However I did quick UI tests and after resetting templates I don't see "Created" and "Updated" fields in the tiles:
image
and looking at mockups - it should be there. Am I correct?

I guess it's easy to fix - just remove:
```
"fieldsToShow": ["Name", "Description"]
```
from [templates/pages/blueprintMarketplace.json](https://github.com/cloudify-cosmo/cloudify-stage/blob/76e28bb2625c4ef38a87df1d469b6aebe931ed48/templates/pages/blueprintMarketplace.json) file, so our framework will pick up the defaults which is to show all fields - Name, Description, Created and Updated.

qooban (approved):

PR #2219, opened 2022-10-17
RD-3002 Replace `scanjs`

kubama:
## Description
Replaced `scanjs`

## Screenshots / Videos
N/A

## Checklist

## Tests
CI verification

 

## Documentation
N/A


app/components/LoginPage.tsx
```diff
@@ -85,7 +85,7 @@ class LoginPage extends Component<LoginPageProps, LoginPageState> {
         const errors: Errors = {};
 
         if (isSamlEnabled) {
-            // eslint-disable-next-line scanjs-rules/assign_to_href
+            // eslint-disable-next-line xss/no-location-href-assign
             window.location.href = samlSsoUrl;
```
qooban, line 89: According to https://github.com/Rantanen/eslint-plugin-xss/blob/master/docs/rules/no-location-href-assign.md the following is the solution
```
window.location.href = escape(samlSsoUrl);
```
where `escape` should be a function to sanitize user input.

Maybe worth to implement it at once as `samlSsoUrl` can be provided by user (in configuration)?

kubama: I think it would be impossible to define escape logic for this case - I assume the SSO URL could point to any domain and we shouldn't anyhow restrict it.

app/components/LicensePage.tsx
```diff
@@ -1,21 +1,21 @@
...
+import type Manager from '../utils/Manager';
+import SplashLoadingScreen from '../utils/SplashLoadingScreen';
 import StageUtils from '../utils/stageUtils';
+import Banner from './banner/Banner';
 import { Button, FullScreenSegment, Grid, Header, Icon, Message, MessageContainer } from './basic';
 import CurrentLicense from './license/CurrentLicense';
 import EulaLink from './license/EulaLink';
 import UploadLicense from './license/UploadLicense';
-import SplashLoadingScreen from '../utils/SplashLoadingScreen';
-import type Manager from '../utils/Manager';
```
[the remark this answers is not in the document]
kubama: No, reverted.

.eslintrc.json
```diff
@@ -10,7 +10,9 @@
                 "import/no-extraneous-dependencies": ["error", { "devDependencies": true }],
                 "no-console": "off",
                 "security/detect-child-process": "off",
-                "no-process-exit": "off"
+                "no-process-exit": "off",
+                "xss/no-mixed-html": "off",
+                "security-node/detect-crlf": "off"
```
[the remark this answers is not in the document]
kubama: Same as here: https://github.com/cloudify-cosmo/cloudify-blueprint-topology/pull/229#discussion_r998116501

app/components/LoginPage.tsx
qooban: I agree that we can't detect problematic domains, but... 

XSS is about different case. According to https://stackoverflow.com/questions/24078332/is-it-secure-to-use-window-location-href-directly-without-validation/24089350#24089350:

> XSS issues arising from the use of javascript: and vbscript: URIs.

so I would block at least these two and basically allow `http://` and `https://` prefixes. WDYT?

.eslintrc.json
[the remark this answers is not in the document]
qooban: OK.

qooban (commented):
Replied in discussion.

app/components/LoginPage.tsx
kubama: Still, this URL is not supplied by the application end-user, but someone who has access to configuration files. Such person is very likely to have a full access to the application installation, so constraining a single configuration option doesn't seem to increase security at all. I think it's the customer responsibility to ensure configuration access security, and once it is ensured there is just no need for further constraints.

qooban: Yeah, OK, I'm not insisting to push it. 

I think still it may be helpful more, but in other cases.
Don't have any particular in mind right now, so OK, let's leave it.

qooban (approved):

qooban:
I forced Snyk to give green light on "License check for cloudify-cosmo/cloudify-stage".

PR #2222, opened 2022-10-19
RD-5719 RD-1701 RD-2083 Fix types and usage of DataTable component

Blewin:
## Description

one bug found 
## Checklist

## Documentation

not needed


widgets/plugins/src/PluginsTable.tsx
```diff
@@ -142,7 +142,7 @@ const PluginsTable: FunctionComponent<PluginsTableProps> = ({ data, toolbox, wid
                             key={item.id}
                             selected={item.isSelected}
                             onClick={() => selectPlugin(item)}
-                            onMouseOver={setHoveredPlugin}
+                            onMouseOver={setHoveredPlugin as any} // TODO: How does it work? :(, probably a bug
```
Blewin, line 145: the whole "hoveredPlugin" seems to be a noop

qooban:
@Blewin As @PatrycjaKubiczek finished his changes I think you should be able to continue this PR (probably you can merge here #2226 .

PatrycjaKubiczek:
@Blewin @qooban just for clarification - should I merge my PR in Stage? or will @Blewin will include my changes in his PR?

qooban:
@PatrycjaKubiczek Let's not combine them, but merge it separately, so you will handle this PR and @Blewin will handle his PR.

widgets/plugins/src/PluginsTable.tsx
```diff
@@ -2,6 +2,7 @@ import type { FunctionComponent } from 'react';
 import { useState } from 'react';
 import type { Visibility } from '../../common/src/types';
 import MarketplaceModal from './MarketplaceModal';
+import type { DataTableConfiguration } from '../../../tsc-dist/app/utils/GenericConfig';
```
qooban, line 5: Oops 🙂 

```suggestion
import type { DataTableConfiguration } from '../../../app/utils/GenericConfig';
```

app/utils/GenericConfig.ts
```diff
@@ -85,7 +85,7 @@ export interface SortColumnConfiguration {
 }
 
 export interface SortAscendingConfiguration {
-    sortAscending: string;
+    sortAscending: boolean;
```
qooban, line 88: Good catch 👍🏻

qooban (commented):
Looks mostly good. Added just one comment in the code.

Let's fix also PR title 🙂

PatrycjaKubiczek:
do we still need this ts-ignore on https://github.com/cloudify-cosmo/cloudify-stage/blob/1bab9e18ca3ad8080d9bef6ed95b8eb189561218/app/components/templateManagement/pageGroups/PageGroups.tsx#L111 ? 🤔

Blewin:
> do we still need this ts-ignore on

good catch! somebody used ts-ignore instead of ts-expect-error and that is why i missed it

widgets/plugins/src/PluginsTable.tsx
qooban: Seeing how `hoveredPlugin` is used (L149), shouldn't we have here:
```suggestion
                            onMouseOver={_event => setHoveredPlugin(item.id)}
```
? 🤔

qooban (commented):
See my comment in the discussion you opened - https://github.com/cloudify-cosmo/cloudify-stage/pull/2222#discussion_r999376019.

Also please address https://github.com/cloudify-cosmo/cloudify-stage/pull/2222#pullrequestreview-1170368097: 

> Let's fix also PR title 🙂

Blewin:
what is the correct pr title?

widgets/plugins/src/PluginsTable.tsx
Blewin: I'd avoid functional changes that don't seem to fix any bugs in a refactor.

Could you verify that whole useState of hoveredPlugin is an elaborate noop, so we can just delete it?

Blewin:
The end of the title is in body, and merging would glue it back together ;)

PatrycjaKubiczek (approved):

widgets/plugins/src/PluginsTable.tsx
qooban: I don't have enough time to get too deep, but from what I saw it seems like it needs a fix and I'm almost sure that the fix I provided would do the trick (but as always - I might be wrong 🙂).

As I really want to close this PR soon please do the following:
1. Create a ticket in Jira (type: Task, label: UI, stage) to investigate and provide the solution for `hovered*` states (the problem is not only here in this file, but also in other files - [widgets/deployments/src/DeploymentsSegment.tsx](https://github.com/cloudify-cosmo/cloudify-stage/blob/38c0689708f4254b3ce294f5a21c0e09bbc52f4d/widgets/deployments/src/DeploymentsSegment.tsx), [widgets/deployments/src/DeploymentsTable.tsx](https://github.com/cloudify-cosmo/cloudify-stage/blob/38c0689708f4254b3ce294f5a21c0e09bbc52f4d/widgets/deployments/src/DeploymentsTable.tsx)
2. Send me new ticket ID, so I'll try to plan it somehow
3. Add `@ts-expect-error` or `@ts-ignore` comment in all 3 places with e.g. the following description: `TODO(RD-xxxx) Set hovered state properly` or something similar. The most important is to provide related ticket ID in the brackets, so anyone can get more details about the problem by reading Jira ticket.

qooban:
> The end of the title is in body, and merging would glue it back together ;)

Thanks for letting me know. I wasn't aware 🙂 

Replied in the last outstanding discussion.

widgets/plugins/src/PluginsTable.tsx
Blewin: https://cloudifysource.atlassian.net/browse/RD-6366

sorry if i am too stingy about this, I'm just very strict about not changing functinality in refactor.

qooban: Thanks for creating the ticket. It would be better if you could add more description there (if you or someone else look at this ticket in the future he/she might not understand what was the idea). Also you mentioned there `WidgetTable.tsx`, but there's no such file in `cloudify-stage`, so please fix that. In p. 3 I asked to add TODO in all 3 places. I see only one file changed, but... OK, as this PR is taking too long let it be like that.

qooban (approved):

PR #2226, opened 2022-10-24
RD-5717 adjustment after NodesTree & HighlightText TS  migration

PatrycjaKubiczek:
## Description

linked PR: https://github.com/cloudify-cosmo/cloudify-ui-components/pull/174
after merging this PR in components issues appearing in CI should be resolved

## Screenshots / Videos

after updating RC-tree lib, there was tiny CSS changes so I need to update the "Main" label styles so it will remain in the center

Before
Zrzut ekranu 2022-11-04 092005
After:
Zrzut ekranu 2022-11-04 091922

## Checklist

- [x] My code follows the [UI Code Style](https://cloudifysource.atlassian.net/l/c/x8fq902p).
- [ ] I followed the [UI Testing Guidelines](https://cloudifysource.atlassian.net/l/cp/udCVfvrx) and verified that all tests/checks have passed.
- [x] I verified if that change requires any update in the [documentation](https://cloudifysource.atlassian.net/l/c/4XKtPocy).
- [x] I added proper labels to this PR.

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/3791/)

## Documentation

n/a

PR #2234, opened 2022-10-26
RD-6075 Start reusing toolbox related types exported by `cloudify-ui-components`

kubama:
## Description
Started reusing toolbox related types exported by `cloudify-ui-components`.

To validate the approach taken and set a direction for future components reuse I applied the toolbox types to agents widget's components - `StageToolbox` references were dropped and replaced with references to types exported by `cloudify-ui-components`. The top level `AgentsTable` component still relays on the monolith `toolbox` helper object - the component is heavily dependent on Stage API, so it should be refactored before it could be eventually reused.

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/job/Stage-UI-System-Test/3794/

 

## Documentation
N/A


widgets/common/src/deployments/DeploymentActions.ts
```diff
@@ -12,31 +13,31 @@ export interface WorkflowOptions {
 }
 
 export default class DeploymentActions {
-    constructor(private toolbox: Stage.Types.Toolbox) {}
+    constructor(private manager: Manager) {}
```
qooban, line 16: I like that direction of narrowing `*Actions` classes constructor argument to only HTTP client - makes perfect sense 👍🏻 

I'm wondering if there are any `*Actions` classes which actually uses anything else that HTTP client methods?

widgets/common/src/executions/ExecutionActions.ts
```diff
@@ -1,21 +1,23 @@
+import type { Manager } from 'cloudify-ui-components';
```
qooban, line 1: This imports looks a bit odd to me (as we are importing here `Manager` type from a components packages). 

Maybe we'll need to adjust mindset to it or there's a better solution? (maybe https://github.com/cloudify-cosmo/cloudify-ui-components/pull/175#discussion_r1007785622 could improve it)

app/utils/Toolbox.ts
```diff
@@ -59,6 +60,11 @@ class Toolbox implements Stage.Types.Toolbox {
         );
     };
 
+    getDrilldownHandler(): DrilldownHandler {
+        return (pageId: string, drilldownContext: Record<string, any>, drilldownPageTitle: string) =>
+            this.drillDown(this.getWidget(), pageId, drilldownContext, drilldownPageTitle);
+    }
```
qooban, line 66: Removing the need of passing `widget` is a good idea 👍🏻 We can reuse this method and maybe end up changing visibility of `drillDown` method to private.

app/utils/External.ts
```diff
@@ -49,7 +39,7 @@ function getFilenameFromHeaders(headers: Headers, fallbackFilename: string) {
     return fallbackFilename;
 }
 
-export default class External {
+export default class External implements Http {
```
qooban, line 42: I'm wondering about `External-Internal-Manager` class hierarchy naming changes.

Right now the names are a bit mysterious. I don't have great proposal, but I think we should make the name easier to understand. Maybe by adding `HttpClient` suffix to all classes? 🤔 

It's out of the scope of RD-6075, but maybe enhancement for RD-6121.

widgets/common/src/executions/ExecutionActions.ts
kubama: Please see https://github.com/cloudify-cosmo/cloudify-ui-components/pull/175#discussion_r1007972009.
With that change the import will be from `cloudify-ui-components/toolbox`.

widgets/common/src/deployments/DeploymentActions.ts
kubama: I haven't investigated. But if there are then we can always have two (or more) constructor argument to supply only the needed subset of current toolbox functionality.

app/utils/Toolbox.ts
kubama: Yes, I think it should be the long-term goal.

app/utils/External.ts
kubama: Indeed, looks like a perfect fit for [RD-6121](https://cloudifysource.atlassian.net/browse/RD-6121).

qooban:
Will review on Monday.

CI is failing. I suspect that's related to #2226 (actually lack of it).

qooban (approved):
Looks good 👍🏻

PR #2244, opened 2022-10-31
RD-5280 Deduplicate plugin wagon name creation logic

leon-good-life:
## Description
Jira ticket: RD-5280:

> There are 2 places in Stage code where we construct distribution full name (distro name + distro release) to select wagon files when uploading plugins: 
> - Plugins Catalog widget 
> - Getting Started Modal (part of the app framework)
>
> It could be extracted and re-used across the app framework and widgets.

There is  also a comment in the Jira ticket:
> @ Leon Gilyadov Please check with @ Jakub Madej if there’s any case in Composer app where we use similar constructs to select wagon files when uploading plugins. If there’s something in Composer as well, then it would make sense to extract the code even higher (to cloudify-ui-common-*).

## Screenshots / Videos

N/A
`wagonUrl` is used to fetch data.
Here is a related screenshot of Plugins Catalog widget, after the changes in this PR:
Screen Shot 2022-10-31 at 18 28 31

## Checklist

## Tests

all specs:
https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/3772/pipeline

 

## Documentation
N/A

### DIFF
--- a/app/components/GettingStartedModal/common/managerHooks.ts
+++ b/app/components/GettingStartedModal/common/managerHooks.ts
@@ -2,16 +2,9 @@ import { useMemo } from 'react';
 import { useSelector } from 'react-redux';
 
 import Internal from '../../../utils/Internal';
-import type Manager from '../../../utils/Manager';
 import useManager from '../../../utils/hooks/useManager';
 
 import type { ReduxState } from '../../../reducers';
-
-const getCurrentDistribution = (manager: Manager) => {
-    const currentDistributionName = manager.getDistributionName().trim();
-    const currentDistributionRelease = manager.getDistributionRelease().trim();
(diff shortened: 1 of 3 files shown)


leon-good-life:
Searching globally in `cloudify-ui-components` for `dist`, `distro` `distribution` `wagon` do not give relevant results. So I conclude that there is no `distribution` in `cloudify-ui-components`.

kubama:
> There is also a comment in the Jira ticket:
> 
> > @ Leon Gilyadov Please check with @ Jakub Madej if there’s any case in Composer app where we use similar constructs to select wagon files when uploading plugins. If there’s something in Composer as well, then it would make sense to extract the code even higher (to cloudify-ui-common-*).

I don't think there is this kind of logic present in composer, so this PR looks sufficient.

kubama (approved):

PR #2267, opened 2022-11-18
RD-6189 PP sidebar adjustment in editMode

PatrycjaKubiczek:
## Description

RD-6189
- add css style to change color on hover on buttons in sidebar in editMode
- add styles to change position of removeIcon so it will be aligned 
- also found a bug with removeIcon in subitem - this icon was slighty outside the menu (see screenshot) so I fixed that as well

## Screenshots / Videos

Before: 
Zrzut ekranu 2022-11-18 120503

Zrzut ekranu 2022-11-18 120423

After (I disabled `hidden` properties just for screenshots so we could see more removeIcons)
Zrzut ekranu 2022-11-18 120245

## Checklist

## Documentation

n/a

### DIFF
--- a/app/components/sidebar/PagesList.tsx
+++ b/app/components/sidebar/PagesList.tsx
@@ -49,14 +49,19 @@ function consumeEvent(event: React.MouseEvent) {
 }
 
 const RemoveIcon = styled(Icon)`
-    position: relative;
-    top: 3px;
-    right: 3px;
+    position: absolute;
+    top: 15px;
+    right: 0;
     visibility: hidden;
     float: right;
-
+    line-height: 8px !important;
+    font-size: 10px !important;
     ${SideBarItemWrapper}:hover & {
         visibility: visible;
+    }
+    &.subitem {
+        top: 0;
+        right: 10px;
     }
 `;
 
@@ -283,6 +288,7 @@ const PagesList: FunctionComponent<PagesListProps> = ({ pageId, expandedGroupIds
                                     consumeEvent(event);
                                     onItemRemoved(pageMenuItem);
                                 }}
+                                className={subItem ? 'subitem' : ''}
                             />
                         )}
                     </>
--- a/app/components/sidebar/PagesList.tsx
+++ b/app/components/sidebar/PagesList.tsx
@@ -6,7 +6,7 @@ import { closestCenter, DndContext, PointerSensor, useSensor, useSensors } from 
 import { SortableContext, verticalListSortingStrategy } from '@dnd-kit/sortable';
 import { restrictToVerticalAxis } from '@dnd-kit/modifiers';
 import { useDispatch, useSelector } from 'react-redux';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 
 import type { SemanticICONS } from 'semantic-ui-react';
 import { EditableLabel, Icon } from '../basic';
@@ -47,14 +47,14 @@ function consumeEvent(event: React.MouseEvent) {
     event.stopPropagation();
     event.preventDefault();
 }
+interface RemoveIconProps {
+    isSubItem: boolean;
+}
 
-const RemoveIcon = styled(Icon).attrs(({ isSubItem }) => ({
-    right: isSubItem ? '10px' : '2px',
-    top: isSubItem ? '0' : '15px'
-}))`
+const RemoveIcon = styled(Icon)<RemoveIconProps>`
     position: absolute;
-    top: ${props => props.top};
-    right: ${props => props.right};
+    top: 15px;
+    right: 2px;
     visibility: hidden;
     float: right;
     line-height: 8px !important;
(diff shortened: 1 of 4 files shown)


qooban (approved):

app/components/sidebar/SideBarItem.tsx
```diff
@@ -67,7 +67,14 @@ const SideBarItem: FunctionComponent<SideBarItemProps> = ({
...
                             rotated={expanded ? undefined : 'counterclockwise'}
-                            style={{ position: 'absolute', right: 12, margin: 0 }}
+                            style={{
+                                position: 'absolute',
+                                right: 12,
+                                margin: 0,
+                                height: 13,
+                                width: 15,
+                                fontSize: '14px'
+                            }}
```
Vorbert-Kruk, line 77: Lots of custom css properties are being passed here
As advised on our [UI Code Style guideline](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/435388492/UI+Code+Style#MUST%3A-Use-CSS-in-JS-over-SCSS-files) we could use `styled-components` to make the styling layer separated and the view more readable 

I've checked and `<Icon />` has a prop `className`, so the usage of `styled-components` should be fully achievable 🎉 

WDYT? 🤔

app/components/sidebar/PagesList.tsx
```diff
@@ -283,6 +288,7 @@ const PagesList: FunctionComponent<PagesListProps> = ({ pageId, expandedGroupIds
                                     consumeEvent(event);
                                     onItemRemoved(pageMenuItem);
                                 }}
+                                className={subItem ? 'subitem' : ''}
```
Vorbert-Kruk, line 291: As `RemoveIcon` is being created with the usage of `styled-components` we can pass custom props to it - [reference](https://styled-components.com/docs/api#styledcomponent)

Because of that, using this functionality seems like a more fitting solution than adding custom `className` just for the same purpose

We could declare a boolean prop `isSubItem` which could be passed to the `RemoveIcon`
Depending on the `isSubItem` flag we could provide the same styling that we are doing right now for `.subitem` 

WDYT? 😉

PatrycjaKubiczek: oh, I haven't though about it, implemented in this commit: https://github.com/cloudify-cosmo/cloudify-stage/pull/2267/commits/3bf91cfe0b13b1c4ff7cf86fcb736e0ac9e61aff is this something what you mean? anyway great idea, thanks for sharing!

Vorbert-Kruk: It's almost the idea, that I had in mind 😁 

I was thinking of something like

```typescript
import styled, { css } from 'styled-components';

interface RemoveIconProps {
    isSubtItem: boolean;
}

const RemoveIcon = styled(Icon)<RemoveIconProps>`
    position: absolute;
    top: 15px;
    right: 0;
    visibility: hidden;
    float: right;
    line-height: 8px !important;
    font-size: 10px !important;
    ${SideBarItemWrapper}:hover & {
        visibility: visible;
    }

    ${({ isSubItem }) =>
        isSubItem &&
        css`
            right: 2px;
            top: 0;
        `}
`;
```

It's a recommended way of using props from `styled-components` which enables you to use custom props and manipulate `css` based on them 🎉
It differs from the example that you've shown in the sense that with that approach you're able to do the multiple `if` conditions inside of the `styled-component` css - which are more readable IMO 
So e.g. if we'd have to implement another prop `isRunaLikingThisItem` and modify `right` and `top` css properties accordingly, we could do just the following:

```typescript
import styled, { css } from 'styled-components';

interface RemoveIconProps {
    isSubtItem: boolean;
    isRunaLikingThisItem: boolean;
}

const RemoveIcon = styled(Icon)<RemoveIconProps>`
    position: absolute;
    top: 15px;
    right: 0;
    visibility: hidden;
    float: right;
    line-height: 8px !important;
    font-size: 10px !important;
    ${SideBarItemWrapper}:hover & {
        visibility: visible;
    }

    ${({ isSubItem }) =>
        isSubItem &&
        css`
            right: 2px;
            top: 0;
        `}

    ${({ isRunaLikingThisItem }) =>
        isRunaLikingThisItem &&
        css`
            right: 13px;
            top: 21px;
        `}
`;
```

Additionally, we are getting a benefit of type safety from typescript 😁

Vorbert-Kruk: To be honest, I'd never thought about using it in this kind of way 🤯 
I'm not telling this in a negative sense, but more like in amazed way! 🥳 

wow-amaze

If I wouldn't have a propose, which would have some additional benefits, I'd definitelly try to go with this one 😉 

Greatly thought! :)

PatrycjaKubiczek: so we both learn something new here :D 🥳

Vorbert-Kruk: Yeah 😁 
Win-win scenario 🥳

Vorbert-Kruk (approved):
Looks good!
Thanks for both addressing comments and sharing some new technical ideas  🎉

PR #2287, opened 2022-11-28
RD-6176 Align drilldown buttons

Vorbert-Kruk:
## Description

Changes consists of updating `button.overrides.less` file to apply the same `margin-top` property button icons 🎉 

### Before
Before

### After
After

Difference is almost not visible 😄, but pixel-wise everything is now perfectly aligned 🎉 

## Checklist

## Tests

 

N/A

## Documentation

N/A

### DIFF
--- a/semantic-ui/site/elements/button.overrides.less
+++ b/semantic-ui/site/elements/button.overrides.less
@@ -35,7 +35,7 @@
         position: relative;
         font-size: @relative17px;
         top: @additionalIconOffset;
-        margin-top: @additionalIconOffset * -2;
+        margin-top: @additionalIconOffset * -2 !important;
     }
 
     &.disabled,


semantic-ui/site/elements/button.overrides.less
```diff
@@ -35,7 +35,7 @@
         position: relative;
         font-size: @relative17px;
         top: @additionalIconOffset;
-        margin-top: @additionalIconOffset * -2;
+        margin-top: @additionalIconOffset * -2 !important;
```
Vorbert-Kruk, line 38: ### Note

I've put here `!important` flag, as this rule has been overridden by some semantic selector for `.right.icon`
As there may be some other cases like that (with a more complicated, therefor stronger, selectors), I've decided to go with using `!important` flag, rather than going with the following direction:

```typescript
&,
&.right,
&.left {
    margin-top: @additionalIconOffset * -2; 
}
```

qooban (approved):

PR #2289, opened 2022-11-29
RD-5753 Migrating app/utils files to TS - Part 1

Vorbert-Kruk:
## Description

Changes consists of migrating the following files:
- `app/utils/External.ts` 
- `app/utils/Manager.ts`

In the upcoming PR the rest of not migrated `app/utils/` files will be migrated 🎉 

## Screenshots / Videos

N/A

## Checklist

## Tests

 

N/A

## Documentation

N/A

### DIFF
--- a/app/utils/Manager.ts
+++ b/app/utils/Manager.ts
@@ -7,6 +7,17 @@ import StageUtils from './stageUtils';
 import Consts from './consts';
 import type { PaginatedResponse } from '../../backend/types';
 
+const emptyPaginatedResponse: PaginatedResponse<any> = {
+    items: [],
+    metadata: {
+        pagination: {
+            offset: 0,
+            size: 0,
+            total: 0
+        }
+    }
+};
+
 export default class Manager extends Internal {
     getCurrentUsername() {
         return this.managerData?.auth?.username ?? null;
@@ -53,16 +64,7 @@ export default class Manager extends Internal {
     doFetchFull<ResponseBodyItem>(
         fetcher: (params: QueryStringParams) => Promise<PaginatedResponse<ResponseBodyItem>>,
         params: QueryStringParams = {},
-        fullData: PaginatedResponse<ResponseBodyItem> = {
-            items: [],
-            metadata: {
-                pagination: {
-                    offset: 0,
-                    size: 0,
-                    total: 0
-                }
-            }
-        },
+        fullData: PaginatedResponse<ResponseBodyItem> = emptyPaginatedResponse,
         size = 0
     ): Promise<PaginatedResponse<ResponseBodyItem>> {
         const fetchParams = {
--- a/app/utils/Manager.ts
+++ b/app/utils/Manager.ts
@@ -1,15 +1,11 @@
 import { filter, get, concat } from 'lodash';
+
+import type { QueryStringParams } from '../../backend/sharedUtils';
 import { getUrlWithQueryString } from '../../backend/sharedUtils';
 import Internal from './Internal';
 import StageUtils from './stageUtils';
 import Consts from './consts';
-
-type RequestParams = Record<string, any>;
-type RequestBody = Record<string, any>;
-type Fetcher = (parameters: RequestParams) => Promise<any>;
-
-type BuildActualUrl = Internal['buildActualUrl'];
-type DoFetchFull<T> = (fetcher: Fetcher, params?: RequestParams, data?: { items: any[] }, size?: number) => Promise<T>;
+import type { PaginatedResponse } from '../../backend/types';
 
 export default class Manager extends Internal {
     getCurrentUsername() {
@@ -36,8 +32,8 @@ export default class Manager extends Internal {
         return this.managerData?.version?.edition === Consts.EDITION.COMMUNITY;
     }
 
-    getManagerUrl: BuildActualUrl = (url, data?) => {
-        return this.buildActualUrl(url, data);
+    getManagerUrl: typeof this['buildActualUrl'] = (url, params) => {
+        return this.buildActualUrl(url, params);
     };
 
     getSelectedTenant() {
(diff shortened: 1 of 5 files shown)


app/utils/Manager.ts
```diff
@@ -33,52 +43,59 @@ export default class Manager extends Internal {
...
             return fullData;
         });
     }
 
-    doPostFull(url, params, body) {
+    doPostFull<ResponseBody, RequestQueryParams extends QueryStringParams>(
+        url: string,
+        body: ResponseBody,
+        params?: RequestQueryParams
+    ) {
```
qooban, line 94: I see arguments order change, so no need to update any other places? 🤔

app/utils/Manager.ts
```diff
@@ -1,12 +1,22 @@
...
+    metadata: {
+        pagination: {
+            offset: 0,
+            size: 0,
+            total: 0
+        }
+    }
+};
 
 export default class Manager extends Internal {
```
qooban, line 21: Documentation:

I see inconsistency between `app/utils/Manager.ts` and https://docs.cloudify.co/staging/dev/developer/writing_widgets/widget-apis/#getmanager, but as I'll be doing more changes here, I'll take it as part of RD-6121.

app/utils/Manager.ts
Vorbert-Kruk: Yeah, you're right 😅 
I've found one place needed an update - which has ben provided within 0da49adee992d10314c25310baf7325e6d8ceda3 🎉

app/utils/Manager.ts
Vorbert-Kruk: Ou, that's good to know!
Till now I didn't know that we are documenting some of our inner classes, like e.g. `Manager`, `External` - new day, new knowledge portion 🥳 

Just out of curiosity, you're referring to the list of `Available methods for getting Cloudify Manager properties`?
Or do you see some inconsistency in other description, related to the `Manager`? 🤔

app/utils/Manager.ts
```diff
@@ -33,52 +43,58 @@ export default class Manager extends Internal {
...
                 return this.doFetchFull(fetcher, fetchParams, fullData, cumulativeSize);
             }
+
             return fullData;
         });
     }
 
-    doPostFull(url, params, body) {
-        return this.doFetchFull(currentParams => this.doPost(url, { params: currentParams, body }), params);
+    doPostFull<ResponseBodyItem>(url: string, body: Record<string, any>, params?: QueryStringParams) {
```
qooban, line 90: Optional:

As:
```
    doPost<ResponseBody = any, RequestBody = any, RequestQueryParams extends QueryParams = QueryParams>(
        url: string,
        requestOptions?: RequestOptions<RequestBody, RequestQueryParams>
    ) {
        return this.ajaxCall<ResponseBody, RequestBody, RequestQueryParams>(url, 'post', requestOptions);
    }
```
I assume that `body` (the second argument of `doPostFull`) don't have to be an object (`Record<string, any>`).

app/utils/Manager.ts
```diff
@@ -33,52 +43,58 @@ export default class Manager extends Internal {
...
+        const path = `/sp${getUrlWithQueryString(url, params)}`;
         return StageUtils.Url.url(path);
-    }
-
-    doFetchFull(fetcher, params = {}, fullData = { items: [] }, size = 0) {
+    };
+
+    doFetchFull<ResponseBodyItem>(
+        fetcher: (params: QueryStringParams) => Promise<PaginatedResponse<ResponseBodyItem>>,
+        params: QueryStringParams = {},
```
qooban, line 66: Optional:

You could probably extend that generic type to get second one and be something like `doGet` in `External` class:

```suggestion
    doFetchFull<ResponseBodyItem, RequestQueryParams extends QueryParams = QueryParams>(
        fetcher: (params: RequestQueryParams) => Promise<PaginatedResponse<ResponseBodyItem>>,
        params: RequestQueryParams = {},
```

Similar approach can be taken to `doPostFull`.

I didn't test it. 

It's definitely not a blocker.

app/utils/Manager.ts
qooban: > Till now I didn't know that we are documenting some of our inner classes, like e.g. `Manager`, `External` - new day, new knowledge portion

Yeah, that might be something to improve. I mean - I'd like to make it clearer in the code - what parts are exposed and what parts are not.

> Just out of curiosity, you're referring to the list of Available methods for getting Cloudify Manager properties?

Yes, exactly to that part 👍🏻

qooban (approved):

app/utils/Manager.ts
Vorbert-Kruk: Thanks for the explanation 🎉

app/utils/Manager.ts
Vorbert-Kruk: As for now I resigned from the idea of passing a type for `params` (92f574de16f56908de28bce85331c503a1a0f3b5) I wouldn't push it within those changes, to keep things consistent 💪 

But idea seems to be interesting 🎉

app/utils/Manager.ts
Vorbert-Kruk: Oh yeah, you're right - I'll adjust it in the upcoming PR 💪

widgets/common/src/actions/SearchActions.ts
```diff
@@ -23,8 +25,11 @@ export default class SearchActions {
             .doPost(`/searches/${resourceName}`, { params, body: { filter_rules: filterRules } });
     }
 
-    private doListAll(resourceName: ResourceName, filterRules: FilterRule[], params?: Params) {
-        return this.toolbox.getManager().doPostFull(`/searches/${resourceName}`, params, { filter_rules: filterRules });
+    private doListAll<ResponseBody>(resourceName: ResourceName, filterRules: FilterRule[], params?: Params) {
+        return this.toolbox.getManager().doPostFull<ResponseBody>(`/searches/${resourceName}`, {
+            filter_rules: filterRules,
+            params
+        });
```
kubama, line 32: I suspect this change introduces a bug - before the change `params` were passed to `doPostFull` as the value for `params` argument, now they are part of the object that is passed as the value for `body` argument.

app/utils/Manager.ts
```diff
@@ -33,52 +43,58 @@ export default class Manager extends Internal {
...
+        fullData: PaginatedResponse<ResponseBodyItem> = emptyPaginatedResponse,
```
kubama, line 67: Looks like this change introduces a bug.
Before the change the `fullData` argument was assigned a new object instance on each invocation without the argument value provided.
After the change the value assigned is always the same object for each invocation. This makes `emptyPaginatedResponse` to accumulate responses from different requests.

PR #2314, opened 2022-12-07
RD-6342 Declare Dockerfiles for building Stage images

tehasdf:
There's two Dockerfiles, because Stage consists of a "backend" part and a "frontend" part.

- declare the two dockerfiles
- make sure to re-use cached layers as much as possible, hence the two-step copy + install in them
- some formalities: dockerignore, a note in the readme
- the config files rendered in the entrypoint are a bit rough around the edges, but for right now, this is what we need. We'll most likely want to add more there in the future (eg. postgres ssl, logging config). But that depends on how we progress with the rest of the system.

For right now, this can be used with the [docker compose file](https://github.com/cloudify-cosmo/cloudify-manager/blob/master/docker-compose.yml) but in the future, the deployment is of course going to change.


tehasdf:
Used in cloudify-cosmo/cloudify-manager#3963

qooban:
I'll take a look tomorrow.

Dockerfile.backend
```diff
@@ -0,0 +1,48 @@
+# Stage backend
```
qooban, line 1: Shouldn't we have that file in the `backend` directory? 🤔

Dockerfile
```diff
@@ -0,0 +1,39 @@
+# Stage frontend
```
qooban, line 1: I'm wondering do we actually need two containers - if backend container is using part of frontend files anyway? 🤔 
Is there real benefit from having two containers for stage app?

packaging/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env -S bash -eux
```
qooban, line 1: General:

I'm wondering how that docker images will be tested?
The docker files uses specific directory structure, NPM scripts, etc., so it would be easy to brake these images. Will that be on your side in the manager?

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /home/stage
+
+# For now, we only need the static/ subdirectory. There are other artifacts
+# available too, such as /app/dist/appData, but they are currently
+# not required.
+COPY --from=builder /app/dist/static /home/stage/static
+
+# -f = nodaemon
+# -p 3000 = listen on port 3000, all interfaces
+CMD ["busybox", "httpd", "-f", "-p", "3000"]
```
qooban, line 39: FYI Composer backend is using port `3000`.
Probably it's not a problem as they're separate containers.

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+FROM busybox
+
+RUN adduser -D stage
+USER stage
+WORKDIR /home/stage
+
+# For now, we only need the static/ subdirectory. There are other artifacts
+# available too, such as /app/dist/appData, but they are currently
+# not required.
```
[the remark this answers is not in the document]
tehasdf: indeed they are not, and they are part of the backend container, yes. Stage backend serves them, and it even checks up front, that these are available in the backend ([in here](https://github.com/cloudify-cosmo/cloudify-stage/blob/4b58dc5e5611e2a1eaa50172293631f5ce0a2fdf/backend/server.ts#L19))

.dockerignore
```diff
@@ -0,0 +1,27 @@
...
+coverage-*
+conf/me.json
+/dist
+/tsc-dist
+userData
+node_modules
+test/cypress/screenshots
+test/cypress/videos
+test/cypress/results
+test/cypress/downloads
```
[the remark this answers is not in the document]
tehasdf: `test` is already in here. Those are listed only because they are listed in `.gitignore`, and I wanted to copy it directly. But OK, I'll remove those entries from here.

packaging/entrypoint.sh
tehasdf: The other PR that I have linked.

tehasdf: And test suite is an interesting question indeed; eventually we'll be able to run just regular integration-tests on the "split" manager.
Of course, stage isn't even tested in those!

Dockerfile
tehasdf: There isn't much of a benefit, there just is a requirement - I can't serve both static files and the backend from a single container.
When you make the backend also serve the static files (which IMO would be a bad idea anyway), then we can make this a single container.

packaging/entrypoint.sh
qooban: As discussed in Slack - we'll need for sure some tests (at least the same level as with RPM testing).

qooban (approved):

Dockerfile
```diff
@@ -0,0 +1,39 @@
+# Stage frontend
+# This describes the stage frontend image, which serves the frontend
+# static files, packaged by webpack.
+# The build is two-stage, where we build the artifacts in a node container,
+# and then only serve them using a simple busybox http server.
+# Build this with e.g.: `docker build . -t stage_frontend`
+
+FROM node:16 as builder
```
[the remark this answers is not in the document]
qooban: Updated.

PR #2315, opened 2022-12-07
RD-6363 Add support for deleting Secret Provider

PatrycjaKubiczek:
## Description

RD-6363
- Add support for deleting Secret Provider

## Screenshots / Videos

https://user-images.githubusercontent.com/17828621/206483827-44077003-651a-4dba-ade2-2f9740883293.mp4

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4102/)

## Documentation

n/a


PatrycjaKubiczek:
@qooban I requested for your review to speed up the progress with this ticket, but as I mentioned - tests are not fully working I will fix this on Monday :)

test/cypress/integration/widgets/secretProviders_spec.ts
```diff
@@ -4,27 +4,27 @@ describe('Secret Providers widget', () => {
         pollingTime: 3,
```
qooban, line 4: Minor: Seems like the filename should be changed to `secret_providers_spec.ts` to be in line with other files in this directory 🙂

test/cypress/support/secret_providers.ts
```diff
@@ -0,0 +1,37 @@
...
+        // NOTE: necessary for extending the Cypress API
+        // eslint-disable-next-line @typescript-eslint/no-empty-interface
+        export interface Chainable extends GetCypressChainableFromCommands<typeof commands> {}
+    }
+}
+export interface SecretProvider {
+    name: string;
+    type: string;
+    visibility: string;
+}
```
qooban, line 15: I think we already have a type for it, right? (in the widget)
(potentially we can use `Partial` generic type to get just part of it)

test/cypress/support/secret_providers.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const commands = {
+    createSecretProvider: (secretProvider: SecretProvider) => {
+        return cy.cfyRequest('/secrets-providers', 'PUT', null, {
+            name: secretProvider.name,
+            type: secretProvider.type,
+            visibility: secretProvider.visibility
+        });
```
qooban, line 23: WDYT about a bit shorter version: 
```suggestion
        const { name, type, visibility } = secretProvider;
        return cy.cfyRequest('/secrets-providers', 'PUT', null, { name, type, visibility });
```
? 🙂

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+        toolbox
+            .getManager()
+            .doDelete(`/secrets-providers/${secretProvider.id}`)
+            .then(() => {
+                setDeletingStatus(RequestStatus.SUBMITTED);
+                toolbox.getEventBus().trigger(tableRefreshEvent);
+            })
+            .catch(() => {
+                setDeletingStatus(RequestStatus.ERROR);
+            });
```
qooban, line 36: I'm not sure we need `RequestStatus`.

What do you think about:
```suggestion
            .then(() => toolbox.getEventBus().trigger(tableRefreshEvent))
            .catch(() => setDeletingStatus(RequestStatus.ERROR))
            .finally(hideModal);
```
?

Will it work the same or I'm missing something? 🤔

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
```diff
@@ -0,0 +1,55 @@
+import { useEffect, useState } from 'react';
+import { tableRefreshEvent } from './SecretProvidersTable.consts';
+import { RequestStatus } from './types';
+import type { SecretProvidersWidget } from './widget.types';
+
+const { Icon, Confirm: DeleteModal } = Stage.Basic;
+
+const { useBoolean } = Stage.Hooks;
+
+const t = Stage.Utils.getT(`widgets.secretProviders`);
```
qooban, line 10: We'll be adjusting that according to recent discussions started by Bartosz L.
Unfortunately we didn't have time during UI weekly for that, so I'm not saying it should be done at once, but what you may consider it for this or the next PRs: Create `translateSecretProviders` function in one of the main widget files (maybe `widget.tsx`?) and then reuse across all other widget files. Sometimes it might be worth to use `Stage.Utils.composeT(translateSecretProviders, <inner-key-path>)`, but probably not in case of this widget (at least not yet).

test/cypress/integration/widgets/secretProviders_spec.ts
```diff
@@ -4,27 +4,27 @@ describe('Secret Providers widget', () => {
...
     });
 
     it('should allow to list secret providers', () => {
-        cy.contains('Secret Provider 1').should('be.visible');
-        cy.contains('Secret Provider 2').should('be.visible');
+        cy.contains('Secret_Provider_1').should('be.visible');
+    });
+
+    it('should allow to delete secret providers', () => {
+        cy.contains('Secret_Provider_1').parent().find('.trash').click();
```
qooban, line 26: I suspect that instead of `find('.thrash')` you cshould be able to use the title you created, am I correct? 🤔 
That would be more inline with Cypress best practices to avoid using class selectors.

Also - maybe worth to introduce a function to get secret provider row at this point? 🤔
Probably something like:
```
function getSecretProviderRow(name: string) {
    return cy.contains(name).parent();
}
```
It's just something to consider.

test/cypress/integration/widgets/secretProviders_spec.ts
```diff
@@ -4,27 +4,27 @@ describe('Secret Providers widget', () => {
...
-                        offset: 0
-                    }
-                }
-            });
+            cy.activate('valid_trial_license')
+                .createSecretProvider({
+                    name: secretProviders[0].name,
+                    type: secretProviders[0].type,
+                    visibility: secretProviders[0].visibility
+                })
```
qooban, line 15: Optional:

```suggestion
            const { name, type, visibility } = secretProviders[0];
            cy.activate('valid_trial_license')
                .createSecretProvider({ name, type, visibility })
```

qooban (commented):
Thanks for providing that draft 👍🏻 

Added some comments, mostly minor, code style related and some hints for the future.

test/cypress/integration/widgets/secretProviders_spec.ts
PatrycjaKubiczek: correct 😅  changed the file name: https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/03cc36aea9e9248dec115582e85c90ab6d486bf9

test/cypress/integration/widgets/secretProviders_spec.ts
PatrycjaKubiczek: update the test spec: https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/03cc36aea9e9248dec115582e85c90ab6d486bf9#diff-c598bfbd81ac382bbe428a6b21304f538447760406575ab7abde4b39045e31a0 to implement required changes

test/cypress/integration/widgets/secret_providers_spec.ts
```diff
@@ -0,0 +1,29 @@
+describe('Secret Providers widget', () => {
+    const widgetId = 'secretProviders';
+    const widgetConfiguration = {
+        pollingTime: 3,
+        pageSize: 0 // NOTE: Setting page size to 0 to list all secret providers and be able to find the one created in test
+    };
+    function getSecretProviderRow(name: string) {
+        return cy.contains(name).parent();
+    }
```
Vorbert-Kruk, line 9: ### Optional
We could move the variables & function above the `describe` to increase test scenarios readability

```suggestion
const widgetId = 'secretProviders';
const widgetConfiguration = {
    pollingTime: 3,
    pageSize: 0 // NOTE: Setting page size to 0 to list all secret providers and be able to find the one created in test
};
function getSecretProviderRow(name: string) {
    return cy.contains(name).parent();
}

describe('Secret Providers widget', () => {
```

WDYT? 🤔

test/cypress/integration/widgets/secret_providers_spec.ts
```diff
@@ -0,0 +1,29 @@
...
+        });
+    });
+
+    it('should allow to list secret providers', () => {
+        cy.contains('Secret_Provider_1').should('be.visible');
+    });
+
+    it('should allow to delete secret providers', () => {
+        getSecretProviderRow('Secret_Provider_1').find('i[title="Delete Secret Provider"]').click();
+        cy.contains('Yes').click();
```
Vorbert-Kruk, line 26: ### Optional
WDYT about using

```suggestion
        cy.clickButton('Yes');
``` 

To be more explicit and consistent with other test scenarios? 🤔

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+            .catch(() => {
+                setDeletingStatus(RequestStatus.ERROR);
+            });
+    };
+
+    useEffect(() => {
+        if (deletingStatus === RequestStatus.SUBMITTED) {
+            hideModal();
+        }
+    }, [deletingStatus]);
```
Vorbert-Kruk, line 40: If we are using `deletingStatus` just to show/hide modal, there is not need for actually using it, see the example below 😉 

```suggestion
    const deleteModalContent = translateSecretProviders('deleteModal.content', {
        secretProviderId: secretProvider.id
    });

    const removeSecretProvider = () => {
        setDeletingStatus(RequestStatus.SUBMITTING);

        toolbox
            .getManager()
            .doDelete(`/secrets-providers/${secretProvider.id}`)
            .then(() => {
                hideModal();
                toolbox.getEventBus().trigger(tableRefreshEvent);
            })
    };
```

One would say that we also are removing error handling, but till now we were setting `RequestStatus.ERROR` flag, once an error would appear, but we weren't handling it anyhow 😁 
So the provided example has the same amount of anti-faulty mechanisms as the proposed solution 😉

widgets/secretProviders/src/SecretProvidersTable.consts.ts
```diff
@@ -1,9 +1,14 @@
...
     name: 'name',
     type: 'type',
     createdAt: 'created_at',
     updatedAt: 'updated_at'
 } as const;
+const translationSecretProvider = Stage.Utils.getT('widgets.secretProviders');
+
+export const translateSecretProviders = (key: string, params?: Record<string, any>) => {
+    return translationSecretProvider(key, params);
+};
```
Vorbert-Kruk, line 14: File in which we are storing those changes is called `SecretProvidersTable.consts.ts`
As changes are consisting of an additional method, I'd rather call it a util function than a const 😁 

Because of that I'd create a separate `SecretProvidersTable.utils.ts` file, where I'd migrate those changes 🎉 

WDYT? 🤔

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
Vorbert-Kruk: If the modification would seem okay for you, we should also remove `./types.ts` file, as its content wouldn't be useful anymore 💪

widgets/secretProviders/src/SecretProvidersTable.tsx
```diff
@@ -29,20 +29,36 @@ const SecretProvidersTable = ({ configuration, data, toolbox }: SecretProvidersT
         <>
             <DataTable
                 fetchData={fetchTableData}
-                noDataMessage={t('noSecretProviders')}
+                noDataMessage={translateSecretProviders('table.noSecretProviders')}
```
Vorbert-Kruk, line 32: We are repeating ourselves a lot in this file, by setting `table.` prefix in every translation

WDYT about taking advantage of composition, which we could use in the following way:

1. Creating a composed function
```typescript
const translation = Stage.Utils.composeT(translateSecretProviders, 'table');
```

2. Using it the same way, as previous translation function, but without `table` prefix 
```typescript
noDataMessage={translation('noSecretProviders')}
```

? 🤔

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+        setDeletingStatus(RequestStatus.SUBMITTING);
+
+        toolbox
+            .getManager()
+            .doDelete(`/secrets-providers/${secretProvider.id}`)
+            .then(() => {
+                toolbox.getEventBus().trigger(tableRefreshEvent);
+            })
+            .catch(() => {
+                setDeletingStatus(RequestStatus.ERROR);
```
qooban, line 32: Is that status used anywhere? 

Also, to handle the error properly we should show it somehow.

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
qooban, line 40: Is it working? I don't see a place where we set `deletingStatus` to `RequestStatus.SUBMITTED`.

I think it would be simpler to add `hideModal()` just after `toolbox.getEventBus().trigger(tableRefreshEvent);` and remove this `useEffect`.

widgets/secretProviders/src/SecretProvidersTable.consts.ts
Vorbert-Kruk, line 14: As we are redeclaring and passing again parameters to the `translationSecretProvider` function, could we use it directly then?

```suggestion
export const translateSecretProviders = Stage.Utils.getT('widgets.secretProviders');
```

Or do you see some advantages of creating additional wrapper for that function? 🤔

widgets/secretProviders/src/SecretProvidersTable.tsx
```diff
@@ -29,20 +29,36 @@ const SecretProvidersTable = ({ configuration, data, toolbox }: SecretProvidersT
...
+                    label={translateSecretProviders('table.columns.dateCreated')}
+                    name={dataSortingKeys.createdAt}
+                    width="156px"
+                />
+                <DataTable.Column
+                    label={translateSecretProviders('table.columns.dateUpdated')}
+                    name={dataSortingKeys.updatedAt}
+                    width="156px"
+                />
                 <DataTable.Column label="" width="10%" />
```
qooban, line 62: Is it good to have a mix of `%` and `px`? 🤔 

If we are using fixed widths (`px`), then I would go with fixed width in actions column (to fit icons and nothing more).

widgets/secretProviders/src/SecretProvidersTable.consts.ts
qooban, line 14: I'm not sure about `translationSecretProvider` name (as a function it would be better to use verb in the name).

Can't we actually do it this way:
```suggestion
export const translateSecretProviders = Stage.Utils.getT('widgets.secretProviders');
```
?

Vorbert-Kruk:
@qooban I guess we have a similar point of view CR wise 😄

test/cypress/integration/widgets/secret_providers_spec.ts
PatrycjaKubiczek: sounds good :)

test/cypress/integration/widgets/secret_providers_spec.ts
PatrycjaKubiczek: I remember that this kind of command it's not recommended by Cypress: https://docs.cypress.io/api/cypress-api/custom-commands#Best-Practices 
there is exactly quote like that: 😆 
```
Don't do things like:
 cy.clickButton(selector)
```

so do you think I should change it anyway? 🤔 I have seen it clickButton was used multiple times across the stage, however, if we follow Cypress's best practices it's not recommended

test/cypress/integration/widgets/secret_providers_spec.ts
PatrycjaKubiczek: Zrzut ekranu 2022-12-13 100005
@Vorbert-Kruk actually after putting const before describe I'm getting conflict with another const - so WDYT about putting consts inside `describe` but function getSecretProviderRow outside it? 🤔

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
PatrycjaKubiczek: implemented adjustments: https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/dc55e4bf73feaf22389713b03d31ac4fd97d5ed3 :)

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
PatrycjaKubiczek: made adjustments according to Norbert's and your suggestions to simplify the structure and add a simple catch with `error` https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/abe6402f21e0a3ed15468ac372f967b429147618 & https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/abe6402f21e0a3ed15468ac372f967b429147618

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
PatrycjaKubiczek: modified code as I mentioned earlier :) https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/dc55e4bf73feaf22389713b03d31ac4fd97d5ed3

widgets/secretProviders/src/SecretProvidersTable.consts.ts
PatrycjaKubiczek: https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/19fb95240d6966f41d10dbbf9343f6c737a5ecde done

widgets/secretProviders/src/SecretProvidersTable.tsx
PatrycjaKubiczek: good idea, changed it 🙌 https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/3d42996b8017bc2c69ea5f010b6a82d9063221bc

widgets/secretProviders/src/SecretProvidersTable.consts.ts
PatrycjaKubiczek: as I've already replied to @Vorbert-Kruk comment (https://github.com/cloudify-cosmo/cloudify-stage/pull/2315#discussion_r1047056174) I changed the code https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/3d42996b8017bc2c69ea5f010b6a82d9063221bc

widgets/secretProviders/src/SecretProvidersTable.tsx
PatrycjaKubiczek: removed fixed width from all columns (https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/d2fe2e85789a5dfb4fc3dd7bad5e0879354a9119) Do you think I stay with % on the action column? % is always a bit more responsive

qooban:
I'll review it tomorrow.

test/cypress/integration/widgets/secret_providers_spec.ts
Vorbert-Kruk: @qooban looking at the point made by Patrycja, it may be worth to consider going through custom commands and migrate away from those which are not meeting best practice guidelines from cypress (like `clickButton`)

Of course it should be discussed, but as we are touching this subject right now, just wanted to share this thought 😉

widgets/secretProviders/src/SecretProvidersTable.consts.ts
Vorbert-Kruk: Everyday we learn something new 😁

test/cypress/integration/widgets/secret_providers_spec.ts
Vorbert-Kruk: Oh, it's kind of strange, as no other `const widgetId = ...` is being declared within the same file 🤯 
As I can see, inside `fibonacciSequenceWidget.tsx` there is that declaration ... but it seems kind of strange that those files are overlaping with each other

I can see, that the solution is to add `export {}` at the end of the file - we won't go in that way, but just wanted to share example of resolving this issue ([more details about the issue](https://stackoverflow.com/questions/35758584/cannot-redeclare-block-scoped-variable))

As it's quite counter intuitive, I'll try to mention it on one of ours weekly 💪 

Having that said, we can go back to the original question:
> so WDYT about putting consts inside describe but function getSecretProviderRow outside it?   

and answer it with: yeah, let's leave it as it is 🥳 
I've seen that you've put method in the upper scope, which is a partial upgrade that's always great to see! 💪

Vorbert-Kruk: CC @qooban, just so you'd be aware about this strange issue 😁
One of the solutions is to modify TS compiler, to compile code to `ES6`, which also resolves the issue that we've encountered not a long time ago (arrow functions inside classes)

Vorbert-Kruk (approved):
LGTM 💪 
Thanks for putting additional effort to address all CR comments (even those optional!) 🎉 

In the future we can have in mind implementing some anti-faulty mechanisms (related to the removal of a secret provider), but for now it looks quite promising and totally okay 🚀 

All discussions are considered to be resolved by me, but some of them I didn't resolve just to increase the visibility of the comments

test/cypress/integration/widgets/secret_providers_spec.ts
qooban: Good points, I'm OK with leaving it as is right now and I'll add a point to UI Weekly about verification of our custom commands against Cypress best practices. @PatrycjaKubiczek I think we may consider that idea when discussing long-term goals 🙂 

Let me share my point of view related to statements from https://docs.cypress.io/api/cypress-api/custom-commands#2-Don-t-overcomplicate-things

> There's no reason to add this level of complexity when you're only wrapping a couple commands.

That's debatable IMHO. I'm not sure it applies to every single case. Writing always something like `cy.contains('button', 'Yes').click()` may not be an overkill, but still it is possible to shorten it and manage it from one place.

> This first custom command is wrapping cy.get(selector).click(). 

We define our custom command as: 
```
clickButton: (buttonLabel: string) => cy.contains('button', buttonLabel).click()
```
so it's not taking CSS selector (in which case I agree with Cypress best practices that it wouldn't be good idea), but button label, so it's not exactly the same case.

> Going down this route would lead to creating dozens or even hundreds of custom commands to cover every possible combination of element interactions. It's completely unnecessary.

I don't see that problem in our code, we didn't went down this route and we didn't create "dozens or even hundreds of custom commands to cover every possible combination of element interactions".

widgets/secretProviders/src/SecretProvidersTable.tsx
```diff
@@ -52,9 +55,9 @@ const SecretProvidersTable = ({ configuration, data, toolbox }: SecretProvidersT
 
                         <DataTable.Data>{Time.formatTimestamp(secretProvider.created_at)}</DataTable.Data>
                         <DataTable.Data>{Time.formatTimestamp(secretProvider.updated_at)}</DataTable.Data>
-                        <DataTable.Data>
-                            <Icon name="edit" title={t('buttons.removeSecretProvider')} />
-                            <Icon name="trash" title={t('buttons.updateSecretProvider')} />
+                        <DataTable.Data textAlign="center">
+                            <Icon name="edit" title={translation('buttons.updateSecretProvider')} />
+                            <RemoveSecretProviderButton secretProvider={secretProvider} toolbox={toolbox} />
```
qooban, line 60: I went over ticket description and there is "Validation" section, please take a look.

I agree that Michael S. could write it better, but IMHO we can deduce from it that we should either:
1. Disable delete icon when there are secrets defined using certain secret provider
2. Change confirmation modal description to disallow secret provider deletion

I would go with the first option. If you wish, you can also talk directly with Michael S.

Implementation can be done in another PR if that's bigger change, but it should still be part of RD-6363.

widgets/secretProviders/src/SecretProvidersTable.tsx
```diff
@@ -1,10 +1,13 @@
...
+import RemoveSecretProviderButton from './RemoveSecretProviderButton';
 import { dataSortingKeys, tableRefreshEvent } from './SecretProvidersTable.consts';
+import { translateSecretProviders } from './SecretProvidersTable.utils';
 import type { SecretProvidersWidget } from './widget.types';
 
 const { DataTable, Icon, Button } = Stage.Basic;
 const { Time } = Stage.Utils;
-const t = Stage.Utils.getT(`widgets.secretProviders.table`);
+
+const translation = Stage.Utils.composeT(translateSecretProviders, 'table');
```
qooban, line 10: Let's not use `translation` (noun) for a function name (which should be a verb in majority of cases). IMHO it can be `translateTable` or something like that.

widgets/secretProviders/src/SecretProvidersTable.tsx
qooban: OK. 

Regarding actions column, we have it fixed in some widgets (e.g. Filters widget).
I'm OK with both solutions, so I'm leaving it to you to decide (either leave as is or change to `px`).

widgets/secretProviders/src/SecretProvidersTable.utils.ts
```diff
@@ -0,0 +1 @@
+export const translateSecretProviders = Stage.Utils.getT('widgets.secretProviders');
```
qooban, line 1: Having this file named `SecretProvidersTable.utils.ts` sounds not fully accurate to me.

I assume that `translateSecretProviders` may be used across all widget files (you'll probably use it in create and update modals), so shouldn't it be `widget.utils.ts`? 🤔 The same applies to `SecretProvidersTable.consts.ts`.

It's not critical to address in this PR, but IMHO worth to take closer look at some point during Secret Providers widget development.

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+        toolbox
+            .getManager()
+            .doDelete(`/secrets-providers/${secretProvider.id}`)
+            .then(() => {
+                toolbox.getEventBus().trigger(tableRefreshEvent);
+                hideModal();
+            })
+            .catch(err => {
+                log.error(err);
+            });
```
qooban, line 30: We are just not closing the modal on error and saying nothing to the user. Hm... I'm not sure that's the way to go.
Can be part of separate PR if you want to deliver first part of RD-6363, but IMHO it should be covered.

test/cypress/integration/widgets/secret_providers_spec.ts
qooban: Thanks for letting me know @Vorbert-Kruk. Could you start discussion about it in UI team during the next UI weekly meeting? I'm pretty sure we have enough topics for this week (but feel free to add it at the end of agenda, and we'll move it to the next week once new section is created).

widgets/secretProviders/src/SecretProvidersTable.utils.ts
Vorbert-Kruk: Oh yeah, valid point
It would be more accurate 💪

widgets/secretProviders/src/SecretProvidersTable.tsx
PatrycjaKubiczek: changed to `translateTable` https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/7f95ad2a02c92e0b668ef893f77e63fb0b2e8ce4

widgets/secretProviders/src/SecretProvidersTable.utils.ts
PatrycjaKubiczek: changed file names to widget https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/7f95ad2a02c92e0b668ef893f77e63fb0b2e8ce4

widgets/secretProviders/src/SecretProvidersTable.tsx
PatrycjaKubiczek: ok, leaving as it is 👍

widgets/secretProviders/src/SecretProvidersTable.tsx
PatrycjaKubiczek: as I already said on slack I'm leaning towards 2nd option - https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/e7267ed76c0a568024f8e455e4178a8e9d7f81db ofc we can have further discussion about it in meantime I did implement this solution https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/e7267ed76c0a568024f8e455e4178a8e9d7f81db tell WDYT about it

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
PatrycjaKubiczek: hmm there was no information about error handling (displaying an error to users), so I assumed it was necessary at this point, but I implement a simple solution https://github.com/cloudify-cosmo/cloudify-stage/pull/2315/commits/e7267ed76c0a568024f8e455e4178a8e9d7f81db WDYT?

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    const WarningSecretProviderModal = (
+        <Modal open={isWarningModalOpen} onClose={closeWarningModal}>
+            <Modal.Header style={{ border: 'none' }}> {warningModalContent}</Modal.Header>
+            {error && <Modal.Content>{ErrorMsg}</Modal.Content>}
+            <Modal.Actions>
+                <Button content={translateSecretProviders('warningModal.buttons.cancel')} onClick={closeWarningModal} />
+            </Modal.Actions>
+        </Modal>
+    );
```
qooban, line 82: IIRC by having it here we are recreating all 3 components on each rerender of `RemoveSecretProviderButton`.
That maybe won't be big problem in this specific case, but IIRC it's considered as an anti-pattern.

widgets/secretProviders/src/SecretProvidersTable.tsx
qooban: I'm OK with it.

qooban (approved):
Thanks for all the efforts on making this PR better.

I left one comment. I think it can be addressed in another PR, not to prolong this one.

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
qooban: Yep, I think it's OK.

widgets/secretProviders/src/RemoveSecretProviderButton.tsx
PatrycjaKubiczek: okay, I'll improve that in next PR :)

PR #2321, opened 2022-12-09
RD-6162 Merged `widgets/common` into main application bundle

kubama:
## Description
Merged `widgets/common` into main application bundle

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/4078/ - the tests are starting and running, failures due to environment issues.

Component tests are failing on first spec - application bundle now takes slightly longer to build, thus the timeout. I consider it an unacceptable CI slowness - it takes 195s to build the bundle on CI while only 25s locally - almost 8x slower. I consider https://cloudifysource.atlassian.net/browse/RD-6549 a blocker to this ticket.

 

## Documentation
N/A


app/widgets/common/components/parameter/ParameterValue.tsx
```diff
@@ -1,6 +1,12 @@
 // @ts-nocheck File not migrated fully to TS
+import React from 'react';
 import { types } from 'cloudify-ui-common-frontend';
 import highlighterStyles from 'react-syntax-highlighter/dist/esm/styles/hljs/idea';
+import PropTypes from 'prop-types';
+import { CopyToClipboardButton, HighlightText } from '../../../../components/basic';
+import Url from '../../../../utils/shared/UrlUtils';
+import Json from '../../../../utils/shared/JsonUtils';
+import { AnyData } from '../../../../utils/props';
```
qooban, line 9: General:

As we'll have more nested directory structure, I think we could improve import paths configuration to allow 'app' prefix instead of having multiple '../'. What do you think? Probably not for this PR (as it's already big). Maybe worth to have dedicated ticket for that?

app/widgets/common/deploymentsView/styles.scss
```diff
@@ -1,5 +1,5 @@
 @use 'sass:math';
-@import '~cloudify-ui-common-frontend/styles/colors';
+@import '../../../../node_modules/cloudify-ui-common-frontend/styles/colors';
```
qooban, line 2: Do we need this change? 🤔

app/widgets/common/map/MarkerIcon.ts
```diff
@@ -1,11 +1,11 @@
+import { CONTEXT_PATH } from '../../../../backend/consts';
```
qooban, line 1: Probably we don't need to "ask" backend for that, and can take it from `app/utils/consts.ts`.

test/cypress/integration/widgets/deployment_actions_spec.ts
```diff
@@ -1,5 +1,5 @@
-import type { LabelInputType } from '../../../../widgets/common/src/labels/types';
-import { isLabelModifiable } from '../../../../widgets/common/src/labels/common';
+import type { LabelInputType } from '../../../../app/widgets/common/labels/types';
+import { isLabelModifiable } from '../../../../app/widgets/common/labels/common';
```
qooban, line 2: General:

Seeing that in `test/cypress/components/widgets/filters/RulesForm_spec.tsx` we use shorter paths in imports, can we have:
```suggestion
import type { LabelInputType } from 'app/widgets/common/labels/types';
import { isLabelModifiable } from 'app/widgets/common/labels/common';
```
here?

Applies to all other changes in `test/cypress/integration`.

test/jest/widgets/common/map/site.test.ts
```diff
@@ -5,7 +5,7 @@ Stage.Utils = StageUtils;
 
 // NOTE: use `require` to lazily load the module after `Stage.Utils` is set
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const { isSiteWithPosition } = require('../../../../../widgets/common/src/map/site');
+const { isSiteWithPosition } = require('../../../../../app/widgets/common/map/site');
```
qooban, line 8:
```suggestion
const { isSiteWithPosition } = require('widgets/common/map/site');
```

.size-limit
```diff
@@ -1,19 +1,19 @@
...
         "name": "Vendor JS bundle",
-        "limit": "5.08 MB",
+        "limit": "5.15 MB",
         "gzip": false,
         "path": "dist/static/js/vendor.bundle.js"
     },
     {
         "name": "Widgets",
-        "limit": "3.00 MB",
+        "limit": "2.62 MB",
```
qooban, line 16: So all in all we have a decrease in the whole application size? 🤔 
(main - ~170kB increase, vendor - ~70kB increase, widgets - ~380kB decrease)

If so, maybe we can also adjust L26?

qooban (commented):
Great work 👍🏻 I like the direction. 

General:

I think that as we are moving out from using `Stage` global object in built-in widgets, we would need stronger verification for it in tests. That'll ensure that we don't brake custom widget development. Can we make it part of RD-6346 epic? (either within already existing ticket or a new one)

app/widgets/common/deploymentsView/styles.scss
kubama: Reverted, looks like WS changed this while moving the file.

app/widgets/common/blueprints/UploadBlueprintForm.tsx
```diff
@@ -1,10 +1,14 @@
...
 import Consts from '../Consts';
-import ToolboxPropType from '../props/Toolbox';
+import ToolboxPropType from '../../../utils/props/Toolbox';
 import BlueprintActions from './BlueprintActions';
 import UploadBlueprintBasicForm from './UploadBlueprintBasicForm';
+import StageUtils from '../../../utils/stageUtils';
+import { Form } from '../../../components/basic';
 
-const t = Stage.Utils.getT('widgets.common.blueprintUpload.inputs');
+const t = StageUtils.getT('widgets.common.blueprintUpload.inputs');
```
PatrycjaKubiczek, line 11: just out of curiosity - what is the reason for changing Stage.Utils into StageUtils?

.size-limit
kubama: Done

app/widgets/common/blueprints/UploadBlueprintForm.tsx
kubama: To use explicitly imported `StageUtils` instead of implicit `Stage` global.

PatrycjaKubiczek: I've seen Stage.Utils/Stage.Basic etc multiple times - WDYT about adding this rule as a part of ongoing refactors/ui code style? maybe having one consistent way would be beneficial?

PatrycjaKubiczek (approved):

test/cypress/integration/widgets/deployment_actions_spec.ts
kubama: Yes, done.

app/widgets/common/blueprints/UploadBlueprintForm.tsx
kubama: Actually as for now this depends on the context - for widgets source files that are still being built as separate bundles the correct way is to use `Stage` global. For files being built into main application bundle (the `app` dir) the only possible way is to use explicit import, trying to reference `Stage` global will result in runtime error. Eventually the `Stage` global access should be dropped when built-in widgets are merged into main application bundle. This is in scope of RD-6346.

kubama:
@bartoszkosciug made pipeline changes that solved the timeout issue. Re-approval needed.

PatrycjaKubiczek (approved):

PR #2348, opened 2022-12-19
RD-6341 Allow absolute paths for backend type imports

kubama:
## Description
Allowed absolute paths for backend type imports

## Screenshots / Videos
N/A

## Checklist

## Tests
CI verification only

 

## Documentation
N/A


app/components/LicensePage.tsx
```diff
@@ -1,13 +1,13 @@
-import _ from 'lodash';
```
Vorbert-Kruk, line 1: Nice catch 💪

app/components/LicensePage.tsx
```diff
@@ -1,13 +1,13 @@
...
-import React, { Component } from 'react';
 import type { ChangeEvent } from 'react';
+import React, { Component } from 'react';
```
Vorbert-Kruk, line 2:
```suggestion
import React from 'react';
import type { Component } from 'react';
```

app/utils/shared/UrlUtils.ts
```diff
@@ -1,6 +1,6 @@
 import _ from 'lodash';
-import Const from '../consts';
 import { getUrlWithQueryString } from '../../../backend/sharedUtils';
```
Vorbert-Kruk, line 2: Did we omitted this line or is troublesome to also modify this path? 🤔 
```suggestion
import { getUrlWithQueryString } from 'backend/sharedUtils';
```

Vorbert-Kruk (commented):
I did a quick investigation upon searching for possible leftovers and it seems that there are 7 more occurrences of relative `backend` imports (on 1 I've already commented on 😁)

Could we also modify those paths to be absolutes or would it be troublesome? 🤔 

image

PatrycjaKubiczek (approved):

kubama:
> Could we also modify this path to be absolutes or would it be troublesome? thinking

Good catch, I somehow missed it. Updated.

Vorbert-Kruk (approved):
LGTM 💪

app/components/LicensePage.tsx
kubama: I had to revert this change - `Component` is used as a value so it broke CI.

Vorbert-Kruk: Oh, got it,
Sorry in that case for the incorrect comment

Vorbert-Kruk (approved):

PR #2358, opened 2023-01-04
RD-4899 Introduce Deploy On functionality on environments page

Vorbert-Kruk:
## Description

Changes consists of passing additional props to [GenericDeployModal](https://github.com/cloudify-cosmo/cloudify-stage/blob/90633bb404a5695c3513d0295bbd5578c0bf9539/app/widgets/common/deployModal/GenericDeployModal.tsx), which enables us to deploy blueprint on a predefined deployment - for now this functionality has been used only on the `environments` page 🎉 

Other than that some refactors were made in area of:
- Redundant `propTypes` removal
- Providing missing translations

### Before
Dropdown item

### After
Dropdown item
Modal

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4252/)
Failing due to the redirection made after making deployment

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4256/)
Passing successfully after providing [a fix](https://github.com/cloudify-cosmo/cloudify-stage/pull/2358/commits/c6d3803b4b51b0f0d655b997f9f2a06c457f2337)

## Documentation

N/A


app/widgets/common/deployModal/GenericDeployModal.tsx
```diff
@@ -265,20 +275,25 @@ class GenericDeployModal extends React.Component<GenericDeployModalProps, Generi
...
         const { installWorkflow } = this.state;
         this.setState({
             baseInstallWorkflowParams: installWorkflow.parameters,
             userInstallWorkflowParams: mapValues(installWorkflow.parameters, parameterData =>
                 getInputFieldInitialValue(parameterData.default, parameterData.type)
-            )
+            ),
+            deploymentId: generateDeploymentIdOnMount ? StageUtils.uuid() : GenericDeployModal.initialState.deploymentId
         });
     }
```
Vorbert-Kruk, line 292: ### Note
This functionality, along with the `generateDeploymentIdOnMount` prop, has been added, as due to the fact that upon opening a modal, `deploymentId` wasn't auto generated like on a different pages

The reason behind it is that on [DeploymentActionsModals.tsx](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/widgets/common/deployments/DeploymentActionsModals.tsx) we are mounting/unmounting modals, while the auto generating functionality is being triggered once the `open` prop value changes

As we are not displaying modals by manipulating `open` prop within [DeploymentActionsModals.tsx](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/widgets/common/deployments/DeploymentActionsModals.tsx) the issue was appearing

As modifying logic within [DeploymentActionsModals.tsx](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/widgets/common/deployments/DeploymentActionsModals.tsx) wouldn't be the easiest and quickest thing to do, I've decided to put this logic within the shared modal - it's not the ideal solution, but it seems like an acceptable one 💪

app/widgets/common/deployments/DeploymentActionsModals.tsx
```diff
@@ -27,6 +26,10 @@ const DeploymentActionsModals: FunctionComponent<DeploymentActionsModalsProps> =
     redirectToParentPageAfterDelete
 }) => {
     const commonProps = { deploymentId, deploymentName, open: true, onHide, toolbox };
+    const deploymentToDeployOn = {
```
Vorbert-Kruk, line 29: ### Note
I couldn't create this variable within the `switch case` block as eslint wasn't allowing me to do so

That's why I've declared it in the upper scope

qooban:
I'll review it tomorrow.

app/widgets/common/deployModal/GenericDeployModal.tsx
```diff
@@ -147,6 +147,19 @@ type GenericDeployModalProps = {
...
      */
     blueprintFilterRules?: FilterRule[];
+
+    /**
+     * Deployment on which submitted blueprint should be deployed on
+     */
+    deploymentToDeployOn?: {
+        id: string;
+        displayName: string;
+    };
```
qooban, line 157: Optional:

I'm wondering about the naming here. Maybe we should be more specific here and say `environmentToDeployOn`? 🤔

app/widgets/common/deployments/DeploymentActions.ts
```diff
@@ -161,9 +161,11 @@ export default class DeploymentActions {
         return this.manager.doGet('/labels/deployments?_reserved=true').then(({ items }) => items);
     }
 
-    // eslint-disable-next-line camelcase
-    doGetWorkflows(deploymentId: string): Promise<{ id: string; display_name: string; workflows: Workflow[] }> {
-        return this.manager.doGet(`/deployments/${deploymentId}?_include=id,display_name,workflows`);
+    doGetWorkflows(
```
qooban, line 164: Looks like it's no longer just `doGetWorkflows` as it also gets labels. Can we update the name?

app/widgets/common/deployments/DeploymentActionsMenu.consts.ts
```diff
@@ -0,0 +1,30 @@
...
+    install: 'install',
+    manageLabels: 'manageLabels',
+    setSite: 'setSite',
+    uninstall: 'uninstall',
+    update: 'update',
+    deployOn: 'deployOn'
+});
+
+export const permissions = {
+    executeWorkflowPermission: 'execution_start',
```
qooban, line 13: To keep consistency, probably better would be to have:
```suggestion
    executeWorkflow: 'execution_start',
```

app/widgets/common/deployments/DeploymentActionsMenu.utils.ts
```diff
@@ -0,0 +1,23 @@
+import type { Label } from '../labels/types';
+import type { Workflow } from '../executeWorkflow';
+import type { MenuItem } from './DeploymentActionsMenu.consts';
+import { permissions, actions } from './DeploymentActionsMenu.consts';
+
+function isDeployOnFunctionalityAvailable(deploymentLabels: Label[]) {
```
qooban, line 6: What do you think about shorter name `isEnvironment` or `hasEnvironmentLabel`?

app/widgets/common/deployments/DeploymentActionsMenu.utils.ts
```diff
@@ -0,0 +1,23 @@
+import type { Label } from '../labels/types';
```
qooban, line 1: General:

Are `DeploymentActionsMenu.utils.ts` and `DeploymentActionsMenu.consts.ts` used anywhere else than in `DeploymentActionsMenu` component? 🤔 I'm not saying it's bad to extract these kind of files, but seeing that merging it all together gives ~100 lines I'm not sure if it's not an overkill in this case.

app/widgets/common/deployModal/GenericDeployModal.tsx
qooban: Would it be possible to remove `generateDeploymentIdOnMount` prop if we generate ID always in both cases (mount and open)?

Vorbert-Kruk: I think so 🎉 
It would duplicate a functionality by a little, but not in the matter that should concern us - nice idea! 💪 

Unfortunately, there is one additional issue which may make this simplification a little bit harder
The situation is coming from the code presented below:

```typescript
    componentDidUpdate(prevProps: GenericDeployModalProps) {
        const { blueprintId, open } = this.props;

        if (!prevProps.open && open && typeof blueprintId === 'string') {
            // eslint-disable-next-line react/no-did-update-set-state
            this.setState({ ...GenericDeployModal.initialState, deploymentId: StageUtils.uuid() }, () =>
                this.selectBlueprint(blueprintId)
            );
        }
    }
```

As you can see, currently this functionality was working only for the cases when the `blueprintId` is provided via props
Implementing case is not making this requirement and because of that some additional indicator seems to be sadly required 😢 

But to be honest, I've added this functionality to mirror the situation in modal mention within RD-4899 - during testing I've seen that one additional step has to be done while using `Deploy On` functionality, and that typing `deploymentId` 😁 , that's why I've decided to share functionality of auto generating this id

Maybe it's unnecessary from the ticket perspective and could be removed, WDYT? 🤔

app/widgets/common/deployModal/GenericDeployModal.tsx
Vorbert-Kruk: Looking on the new conventions that are being introduced, it may be a nice idea 💪 
Implementing it right away 🚀

Vorbert-Kruk: Renamed it within 50f223aa55be9781b85d0a3f22435b25b829e8e4, but I haven't changed description of that prop to use new nomenclature, as I suspect that the `environment` may not be understandable in the same way by everyone 😄

app/widgets/common/deployments/DeploymentActions.ts
Vorbert-Kruk: Sure 💪

Vorbert-Kruk: Applied it as a part of c9d28eb36d9321268d663993f837e46ef5f3148e

app/widgets/common/deployments/DeploymentActionsMenu.consts.ts
Vorbert-Kruk: Oh yeah, you're right 😁 
As the adjustment requires also some changes across the file, I've implemented it as a part of 0f95837e69dba5552a737686bf59245235a69bf7, rather than suggestion button 🎉

app/widgets/common/deployments/DeploymentActionsMenu.utils.ts
Vorbert-Kruk: Yeah, `actions` from `DeploymentActionsMenu.consts.ts` are being used within `DeploymentActionModals`

To be honest, I've extracted them as it's more readable and convenient for me to work with the component that doesn't have bunch of code laying around on the "left and right" 😄 
So I'm looking on this situation more in terms of "I don't have to scroll 100 lines of code to see component logic which I have to modify" 
I know that it may be very personal thing (maybe it's a "zboczenie zawodowe", as we say in polish 😄), but as additional extraction is rarely a negative thing, I've decided to go into that direction

Let me know WDYT

app/widgets/common/deployModal/GenericDeployModal.tsx
qooban: I did quick check and it seems that `blueprintId` is always of type string, no matter if it is provided or not via props (as it's defaulted to `''` 🙂). So... I think we can remove it here and then simply add similar logic in `componentDidMount`. That'll make `generateDeploymentIdOnMount` unnecessary I believe. WDYT?

qooban (commented):
The only outstanding thing is https://github.com/cloudify-cosmo/cloudify-stage/pull/2358#discussion_r1066678858 - let's see how can we resolve it.

app/widgets/common/deployModal/GenericDeployModal.tsx
Vorbert-Kruk: Oh, you're right 😅 
In that case `generateDeploymentIdOnMount` indeed seems redundant - I'll remove it right away 💪

When it comes to removing this functionality from the  `componentDidUpdate`, I wouldn't recomment it
It's due to the fact that we could get into the situation when a user would submit modal and tried to do it again, he wouldn't have `deploymentId` regenerated and would get an error because of it - there are some places which are manipulating modal display via `open` prop, so upon opening it for the `n` time the logic written within `componentDidMount` wouldn't be applied

So for now I'll try to just remove `generateDeploymentIdOnMount` prop and make auto generation to be triggered via both `componendDidMount` and `componentDidUpdate`

If you'd like to go in a different direction, let me know 💪

qooban: I wasn't suggesting removal of the whole logic, I was just suggesting to remove ` && typeof blueprintId === 'string'` which is redundant in current implementation 🙂 Probably I did too many shortcuts in the description.

Anyways - remove `generateDeploymentIdOnMount` prop, remove ` && typeof blueprintId === 'string'` and initialize `deploymentId` with StageUtils.uuid() in `componentDidMount` - that's I think the best way to go.

app/widgets/common/deployModal/GenericDeployModal.tsx
```diff
@@ -273,17 +283,19 @@ class GenericDeployModal extends React.Component<GenericDeployModalProps, Generi
...
         const { blueprintId, open } = this.props;
-        if (!prevProps.open && open && typeof blueprintId === 'string') {
+
+        if (!prevProps.open && open) {
             // eslint-disable-next-line react/no-did-update-set-state
-            this.setState({ ...GenericDeployModal.initialState, deploymentId: StageUtils.uuid() }, () =>
-                this.selectBlueprint(blueprintId)
-            );
+            this.setState({ ...GenericDeployModal.initialState, deploymentId: StageUtils.uuid() }, () => {
+                this.selectBlueprint(blueprintId!);
```
Vorbert-Kruk, line 297: ### NOTE: 
Without `typeof blueprintId === 'string'` typescript is warning us, that the `blueprintId` may be an undefined value
It's not the case in this case, as the default value providing for `blueprintId` prop is `''`

We could resolve it without using `!` operator but it would require going through the file and creating a corresponding type for props and their default values
Currently I've decided to go with a `!` operator as the second option would require doing some additional refactors across the file

qooban (approved):
Looks good 👍🏻

PR #2372, opened 2023-01-11
RD-5770 Executions widget TS migration

kubama:
## Description
Includes some backend refactorings enforced by `backend.ts` migration.
The `backend.ts` had to be heavily refactored, but I can still see much more space for improvement there. `lodash` usage was stripped down to minimum due to lack of typing caused by using `require` instead of `import`.

## Screenshots / Videos
N/A

## Checklist

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/4301/

 

## Documentation
N/A


package.json
```diff
@@ -221,6 +223,7 @@
         "source-map-loader": "^1.1.3",
         "timekeeper": "^2.2.0",
         "ts-node": "^10.2.1",
+        "web-worker": "^1.2.0",
```
kubama, line 226: Required by `elkjs` but not installed automatically

PatrycjaKubiczek:
@kubama as I am a UI guardian and have a secret/secret provider ticket on my plate, could anyone else give you a code review this time? I don't want to slow down this PR, but I'm feeling a bit overwhelmed.

qooban:
I'll review it tomorrow.

qooban:
Change size seems big, so do you think it would be reasonable to add additional comments/description?

package.json
```diff
@@ -186,6 +187,7 @@
         "cypress-file-upload": "^5.0.8",
         "cypress-get-table": "^1.0.1",
         "cypress-localstorage-commands": "^2.2.1",
+        "elkjs": "^0.8.2",
```
kubama, line 190: Added to use its typings in the frontend

widgets/executions/src/backend.ts
```diff
@@ -64,6 +103,7 @@ export default r => {
          *
          * More information can be found here: https://github.com/OpenKieler/elkjs
          */
+        // eslint-disable-next-line import/no-extraneous-dependencies
```
kubama, line 106: Needed after adding `elkjs` to frontend optional dependencies

widgets/executions/src/backend.ts
```diff
@@ -434,12 +462,10 @@ export default r => {
...
-                    const operationsPromises = _.map(items, graph =>
-                        // eslint-disable-next-line camelcase
-                        helper.Manager.doGet(operationsFetchUrl, { params: { graph_id: graph.id }, headers })
-                    );
-
-                    const [{ items: operationsList }] = await Promise.all(operationsPromises);
+                    const { items: operationsList } = (await helper.Manager.doGet(operationsFetchUrl, {
+                        params: { graph_id: items[0].id },
+                        headers
+                    })) as unknown as PaginatedResponse<Operation>;
```
kubama, line 468: `Promise.all` was used but only the first resolved values was referenced - I optimized it

kubama:
> Change size seems big, so do you think it would be reasonable to add additional comments/description?

Added some more comments and extended the description.

widgets/executions/src/tasksGraph/GraphNodes.tsx
```diff
@@ -1,29 +1,24 @@
...
-
-GraphNodes.propTypes = {
-    graphNodes: PropTypes.arrayOf(GraphNodePropType).isRequired
-};
+const GraphNodes = ({ graphNodes, toolbox }: { graphNodes: ElkNode[]; toolbox: Toolbox }) => (
+    <>
+        {graphNodes.map(graphNode => (
+            <g key={graphNode.id} transform={`translate(${graphNode.x}, ${graphNode.y! + textVisualAdjustment})`}>
+                <GraphNode graphNode={graphNode} toolbox={toolbox} />
+                {!_.isEmpty(graphNode.children) && (
```
Vorbert-Kruk, line 13: WDYT about using explicit lodash import here, as stated in [Ongoing Refactors Guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Using-explicit-imports)? 🤔 

```suggestion
import type { ElkNode } from 'elkjs';
import type { Toolbox } from 'app/utils/StageAPI';
import { isEmpty } from 'lodash';
import GraphNode from './GraphNode';
import GraphEdges from './GraphEdges';

const textVisualAdjustment = 5;

const GraphNodes = ({ graphNodes, toolbox }: { graphNodes: ElkNode[]; toolbox: Toolbox }) => (
    <>
        {graphNodes.map(graphNode => (
            <g key={graphNode.id} transform={`translate(${graphNode.x}, ${graphNode.y! + textVisualAdjustment})`}>
                <GraphNode graphNode={graphNode} toolbox={toolbox} />
                {!isEmpty(graphNode.children) && (
```

widgets/executions/src/tasksGraph/GraphNode.tsx
```diff
@@ -75,11 +72,11 @@ const GraphNode = ({ graphNode, toolbox }) => {
                     </text>
                 ))}
             &gt;
-            {displayText && nodeInstanceId && operation && (state !== 'Pending' || labels.retry > 0) && (
+            {displayText && nodeInstanceId && operation && (state !== 'Pending' || (labels.retry && labels.retry > 0)) && (
```
Vorbert-Kruk, line 75: What do you think about extracting this condition as a separate variable?
It could help with documenting it's purpose and making it easier for us to read the `return` part of component 🎉

widgets/executions/src/tasksGraph/GraphNode.tsx
```diff
@@ -15,15 +12,15 @@ const colors = {
...
     const { Icon } = Stage.Basic;
-    const labels = graphNode.labels[0];
+    const labels = graphNode.labels![0];
 
     let currentTextPlacementY = 0;
 
     const title = labels.displayTitle || [labels.text];
     const { displayText, state } = labels;
-    const mappedState = _.findKey(states, stateArray => _.includes(stateArray, state));
+    const mappedState = _.findKey(states, stateArray => _.includes(stateArray, state)) as keyof typeof states;
```
Vorbert-Kruk, line 23: WDYT about using explicit lodash import across this file, as stated in [Ongoing Refactors Guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Using-explicit-imports)? 🤔 

It would require us to use the following import
```typescript
import { findKey, includes, size, isEmpty }  from 'lodash';
``` 

And removing `_.` strings from this file 🚀

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -128,11 +144,13 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
     }, [!!graphData]);
 
     function scrollToInProgress() {
-        const focusNode = _.find(graphData.children, containerNode =>
-            _.find(containerNode.children, subGraphNode => _.includes(states.inProgress, subGraphNode.labels[0].state))
+        const focusNode = graphData?.children?.find(containerNode =>
+            _.find(containerNode.children, subGraphNode =>
+                _.includes(states.inProgress, subGraphNode.labels?.[0].state)
```
Vorbert-Kruk, line 149: WDYT about using explicit lodash import across this file, as stated in [Ongoing Refactors Guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Using-explicit-imports)? 🤔 

It would require us to use the following import
```typescript
import { get, find, includes, noop }  from 'lodash';
``` 

And removing `_.` strings from this file 🚀

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -128,11 +144,13 @@ export default function ExecutionWorkflowGraph({ containerHeight, selectedExecut
     }, [!!graphData]);
 
     function scrollToInProgress() {
-        const focusNode = _.find(graphData.children, containerNode =>
-            _.find(containerNode.children, subGraphNode => _.includes(states.inProgress, subGraphNode.labels[0].state))
+        const focusNode = graphData?.children?.find(containerNode =>
```
Vorbert-Kruk, line 147: Nice to see the native `Array.prototype.find` function 🚀

widgets/executions/src/ExecutionsTable.tsx
```diff
@@ -1,12 +1,36 @@
...
+    execution?: Execution;
+    errorModalOpen: boolean;
+    executionParametersModalOpen: boolean;
+    deploymentUpdateModalOpen: boolean;
+    hoveredExecution?: string;
+    deploymentUpdateId: '';
+    error: null;
+}
+
+export default class ExecutionsTable extends React.Component<ExecutionsTableProps, ExecutionsTableState> {
```
Vorbert-Kruk, line 33: As stated in our [Ongoing Refactors Guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI) we should use explicit import for `React` package

To make it happen, we can modify this line in a following way:
```typescript
export default class ExecutionsTable extends Component<ExecutionsTableProps, ExecutionsTableState> {
```

Additionally, making the following import at the beginning of the file:
```typescript
import type { Component } from 'react'; 
```

app/widgets/common/executions/LatestExecutionStatusIcon.tsx
```diff
@@ -1,6 +1,6 @@
 // @ts-nocheck File not migrated fully to TS
```
Vorbert-Kruk, line 1: We should be able to remove this comment 😁

app/widgets/common/executions/LatestExecutionStatusIcon.tsx
```diff
@@ -12,16 +12,31 @@ import {
...
+
+export interface LatestExecutionStatusIconProps {
+    execution?: Execution;
+    onActOnExecution?: (execution: Execution, action: ExecutionAction, errorMessage: any) => void;
+    showLabel?: boolean;
+    labelAttached?: boolean;
+    toolbox: Toolbox;
+}
+
+export default class LatestExecutionStatusIcon extends React.Component<LatestExecutionStatusIconProps> {
```
Vorbert-Kruk, line 39: As stated in our [Ongoing Refactors Guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI) we should use explicit import for `React` package

To make it happen, we can modify this line in a following way:
```typescript
export default class LatestExecutionStatusIcon extends Component<LatestExecutionStatusIconProps> {
```

Additionally, making the following import at the beginning of the file:
```typescript
import type { Component } from 'react'; 
```

backend/handler/services/ManagerService.ts
```diff
@@ -5,69 +5,59 @@ import cookie from 'cookie';
 import { jsonRequest } from '../ManagerHandler';
 import { ALLOWED_METHODS_OBJECT, TOKEN_COOKIE_NAME } from '../../consts';
 import { getUrlWithQueryString } from '../../sharedUtils';
-import type { QueryStringParams } from '../../sharedUtils';
-import type { AllowedRequestMethod } from '../../types';
 import { getHeadersWithAuthenticationToken } from '../../utils';
+import type { ManagerResponse, ManagerService } from './ManagerService.types';
+
+const managerService: ManagerService = {
```
Vorbert-Kruk, line 11: I like this direction a lot 🎉

widgets/executions/src/backend.ts
```diff
@@ -1,5 +1,44 @@
...
+        containing_subgraph?: string;
+        current_retries: number;
+        retried_task?: unknown;
+        /* eslint-enable camelcase */
+    };
+    dependencies: unknown[];
+}
+
+export default (r: BackendServiceRegistrator) => {
     r.register('get_tasks_graph', 'GET', (req, res, next, helper) => {
```
Vorbert-Kruk, line 42: We could omit the usage of ominous letter in order of having meaningful variable names 💪 

```suggestion
export default (serviceRegistrator: BackendServiceRegistrator) => {
    serviceRegistrator.register('get_tasks_graph', 'GET', (req, res, next, helper) => {
```

I've double checked and it looks like that there are no more `r` occurences that would require adjustments from our side 🎉

Vorbert-Kruk (commented):
Nice piece of work! 
I've added some comments, mostly related to meeting [Ongoing Refactor](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI) criteria

Other than that I may have one comment related to components migration <in progress of writing - I've missclick a submit button 😅 >

app/widgets/common/executions/LatestExecutionStatusIcon.tsx
kubama: Unfortunately this file is still not migrated

app/widgets/common/executions/LatestExecutionStatusIcon.tsx
kubama: > As stated in our [Ongoing Refactors Guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI) we should use explicit import for React package

Actually this is already the case

widgets/executions/src/backend.ts
kubama: Done

widgets/executions/src/ExecutionsTable.tsx
kubama: Done

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: Done

widgets/executions/src/tasksGraph/GraphNode.tsx
kubama: Done

widgets/executions/src/tasksGraph/GraphNode.tsx
kubama: Done

widgets/executions/src/tasksGraph/GraphNodes.tsx
kubama: Done

kubama:
> As it seems to be a more common approach, I'd suggest going in a similar way as it would make our codebase more consistent

When props interface is not referenced anywhere I don't see any value in having it extracted as a named type. Actually I find it even easier to read when it's inlined, it's not necessary to navigate through the file to check given prop type.

Regarding consistency - depending on the context consistency can of course bring some values, but I don't think consistency by itself is always a value, at some point it can even become an unnecessary constraint that comes with an overhead.

widgets/executions/src/SystemWorkflowIcon.tsx
```diff
@@ -1,9 +1,9 @@
-// @ts-nocheck File not migrated fully to TS
+import type { Execution } from 'app/utils/shared/ExecutionUtils';
 
-export default function SystemWorkflowIcon({ execution }) {
+export default function SystemWorkflowIcon({ execution }: { execution?: Execution }) {
```
qooban, line 3:
```suggestion
export default function SystemWorkflowIcon({ execution }: { execution?: Pick<Execution, 'is_system_workflow'> }) {
```

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
```diff
@@ -24,22 +29,32 @@ const INITIAL_POSITION = {
     e: 0,
     f: GRAPH_MARGIN,
     version: 3
-};
-
-export default function ExecutionWorkflowGraph({ containerHeight, selectedExecution, showStatus, toolbox }) {
+} as unknown as Value;
```
qooban, line 32: Why we need `as unknown`? 🤔

widgets/executions/src/widget.tsx
```diff
@@ -13,8 +14,8 @@ export interface ExecutionsWidgetParams {
     /* eslint-enable camelcase */
 }
 
-export interface ExecutionsWidgetConfiguration {
-    fieldsToShow?: string;
+export interface ExecutionsWidgetConfiguration extends DataTableConfiguration {
+    fieldsToShow: string;
     showSystemExecutions?: boolean;
     singleExecutionView?: boolean;
 }
```
qooban, line 21: In L25 - can we adjust second parameter? (fetched data type)

widgets/executions/src/ExecutionsTable.tsx
```diff
@@ -289,22 +324,22 @@ export default class ExecutionsTable extends React.Component {
                                                     content="Show Execution Parameters"
```
qooban, line 324: Can we add translation support for this file?

Worth to check the other files from Executions widget.

widgets/executions/src/widget.tsx
qooban, line 21: I'm wondering would it be possible to improve `fieldsToShow` configuration definition so that we'll reuse values defined in translation file? (there's another comment about adding translation support to `ExecutionsTable.tsx`)

widgets/executions/src/backend.ts
qooban: Hmm... Aren't `widgets/*/src/backend.ts` files treated as backend files by TS configuration? 🤔 
They supposed to be treated that way by ESLint, but maybe we are missing it for TS.

widgets/executions/src/backend.ts
```diff
@@ -434,12 +462,10 @@ export default r => {
                         return;
                     }
 
-                    const operationsPromises = _.map(items, graph =>
-                        // eslint-disable-next-line camelcase
-                        helper.Manager.doGet(operationsFetchUrl, { params: { graph_id: graph.id }, headers })
-                    );
-
-                    const [{ items: operationsList }] = await Promise.all(operationsPromises);
+                    const { items: operationsList } = (await helper.Manager.doGet(operationsFetchUrl, {
```
qooban, line 465: I'm wondering can we improve `helper.Manager.doGet` so that it is defined as a generic and then by providing response type similarly to what we do in Toolbox related methods (`doGet<ResponseBody>`) we would be able to avoid casting to `unknown` in L568.

If it's too much work, let's leave it for the next stage, but then let's collect such improvements as another ticket.

app/widgets/common/executions/LatestExecutionStatusIcon.tsx
```diff
@@ -12,16 +12,31 @@ import {
...
 
 const t = StageUtils.getT('widgets.common.executions');
 
-export default class LatestExecutionStatusIcon extends React.Component {
+type ExecutionAction =
+    | typeof ExecutionUtils.FORCE_CANCEL_ACTION
+    | typeof ExecutionUtils.CANCEL_ACTION
+    | typeof ExecutionUtils.FORCE_RESUME_ACTION
+    | typeof ExecutionUtils.RESUME_ACTION
+    | typeof ExecutionUtils.KILL_CANCEL_ACTION;
```
qooban, line 29: Maybe more accurate would be to expose this type from `ExecutionUtils`? 🤔

qooban (commented):
Looks like it was tough topic and the result is a solid work 🙂 👍🏻

backend/handler/services/ManagerService.ts
```diff
@@ -5,69 +5,59 @@ import cookie from 'cookie';
...
-}
+    doPatch(url, requestOptions) {
+        return this.call(ALLOWED_METHODS_OBJECT.patch, url, requestOptions);
+    }
+};
 
-export function doPatch(url: string, requestOptions: RequestOptions) {
-    return call(ALLOWED_METHODS_OBJECT.patch, url, requestOptions);
-}
+export default managerService;
```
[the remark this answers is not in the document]
kubama: It's backward compatible, so no documentation update is necessary.

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
kubama: The typings for `react-svg-pan-zoom` seem to be not completely in line with the implementation. Value of `3` is not valid for `version`, however this is what we've been using since ever. I could experiment and check what would happen if I changed `version`, but I decided not to invest time into it and use `as unknown` to force casting. I can revisit it if you think it's worth it.

widgets/executions/src/backend.ts
kubama: Due to the physical placement of this file eslint inspects top-level `package.json` in its context, I don't think eslint or this particular rule provide configuration options to override this. It will be simplified if/when we merge this file into regular backend.

Vorbert-Kruk:
I'll address discussions tomorrow morning

widgets/executions/src/widget.tsx
kubama: Done

widgets/executions/src/ExecutionsTable.tsx
kubama: Done

widgets/executions/src/widget.tsx
kubama: Done

widgets/executions/src/backend.ts
kubama: Done

app/widgets/common/executions/LatestExecutionStatusIcon.tsx
kubama: Done

app/widgets/common/executions/LatestExecutionStatusIcon.tsx
Vorbert-Kruk: Ok, thanks for the info 👌

qooban: Be prepared @Vorbert-Kruk , it's part of your ticket about widgets common migration ;-)

app/widgets/common/executions/LatestExecutionStatusIcon.tsx
Vorbert-Kruk: Oh yeah, you're right 😅 
In the time of writing this comment, I've had in mind different context of explicit -> being explicit about the package parts that we are using
But as reading through the [Ongoing Refactors Guideliness](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Using-explicit-imports) just importing `React` package seems to be acceptable

As I'm looking at the subject right now, I could got confused with the refactors that we are doing with lodash (we're importing every used function explicitly, not by importing whole `lodash` package)
In that case maybe it would be worth to make [this rule](https://cloudifysource.atlassian.net/wiki/spaces/RD/pages/2148073473/Ongoing+Refactors+in+Cloudify+UI#Using-explicit-imports) more precise 🤔 
But that's a thought for another time 😄 

Sorry for this misunderstanding

backend/handler/services/ManagerService.types.ts
```diff
@@ -0,0 +1,27 @@
...
+    headers?: Record<string, string>;
+    params?: QueryStringParams;
+    timeout?: number;
+}
+
+export type ManagerResponse = { items: any[] };
+
+export interface ManagerService {
+    call(method: AllowedRequestMethod, url: string, requestOptions: RequestOptions): Promise<any>;
+    doGet<T>(url: string, requestOptions: RequestOptions): Promise<T>;
```
qooban, line 16: In the future we may also improve other methods - `doPost`,`doPut`, ...
For now we are good.

widgets/executions/src/DryRunIcon.tsx
```diff
@@ -1,22 +1,16 @@
-// @ts-nocheck File not migrated fully to TS
+import type { Execution } from 'app/utils/shared/ExecutionUtils';
 
-export default function DryRunIcon({ execution }) {
+const translate = Stage.Utils.getT('widgets.executions');
```
qooban, line 3: I see it's defined in many files, maybe extract `translateExecutionsWidget` function and export it from `widgets/executions/src/<something>`?

widgets/executions/src/ExecutionsTable.tsx
```diff
@@ -178,80 +245,43 @@ export default class ExecutionsTable extends React.Component {
...
                         return (
                             <DataTable.RowExpandable key={item.id} expanded={item.isSelected}>
                                 <DataTable.Row
                                     key={item.id}
                                     selected={item.isSelected}
                                     onClick={() => this.selectExecution(item)}
                                     onMouseOver={() => this.setHoveredExecution(item.id)}
-                                    onFocus={() => this.setHoveredExecution(item.id)}
                                     onMouseOut={() => this.unsetHoveredExecution(item.id)}
-                                    onBlur={() => this.unsetHoveredExecution(item.id)}
```
qooban, line 254: We don't need it? 🤔

widgets/executions/src/ExecutionsTable.tsx
```diff
@@ -165,7 +210,29 @@ export default class ExecutionsTable extends React.Component {
...
+                    name={defineName ? name : undefined}
+                    width={width}
+                    show={fieldsToShow.indexOf(label) >= 0 && showCondition}
+                />
+            );
+        }
+
+        function getLabelledNamedColumn(name: keyof Execution | 'attributes', width = '100%', showCondition = true) {
+            return getLabelledColumn(name, width, showCondition, true);
+        }
```
qooban, line 234: Nice 👍🏻

widgets/executions/src/tasksGraph/ExecutionWorkflowGraph.tsx
qooban: Understand. In that case let's leave it as is.

widgets/executions/src/backend.ts
qooban: > It will be simplified if/when we merge this file into regular backend.

Yep, having that in mind - let's keep it as is.

qooban (commented):
Added some more comments.

> When props interface is not referenced anywhere I don't see any value in having it extracted as a named type. Actually I find it even easier to read when it's inlined, it's not necessary to navigate through the file to check given prop type.

Regarding props definition - I've seen it for the first time recently in https://github.com/cloudify-cosmo/cloudify-blueprint-composer/pull/1163#discussion_r1060624205 and knowing that @Vorbert-Kruk is also pointing that out, I suggest again to discuss that during UI Weekly. Named type for props was unspoken pattern we followed so far, so let's discuss advantages/disadvantages of both approaches and see what would be our common conclusion. From technical perspective there might be no difference, but there's people aspect when working in the team and we have to take that into account. All in all - please add a point to weekly agenda. 

In specific I'm wondering e.g. if there is any difference in props handling by IDE.

backend/handler/services/ManagerService.ts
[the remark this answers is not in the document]
qooban: OK 👍🏻

Vorbert-Kruk:
> > As it seems to be a more common approach, I'd suggest going in a similar way as it would make our codebase more consistent
> 
> When props interface is not referenced anywhere I don't see any value in having it extracted as a named type. Actually I find it even easier to read when it's inlined, it's not necessary to navigate through the file to check given prop type.
> 
> Regarding consistency - depending on the context consistency can of course bring some values, but I don't think consistency by itself is always a value, at some point it can even become an unnecessary constraint that comes with an overhead.

In that case, we're having a 2 different opinions on that subject 😁 
For me having an extracted type is making the view less cluttered and more readable, but consistency (that comes with it) is making the shape of files more predictable and thus easier to read and to work with
Additional benefit to that, may be the newcomers that can quickly adapt to one type of pattern that is used across the codebase

But as it may the matter of a personal opinion, I won't make it anything close to a 'blocker' shape 😄 
(maybe right now I've been biased by repeatability of this pattern across our codebase and now I expect to find it in files 🤔 - I'll have to think about it 😁 )

Thanks for sharing your point of view, as I've got some points on my mind which I should think over and get under a consideration 💪 
In the future I'd be happy to have a discussion about this subject, as I can see that your position on this matter comes with a several benefits, but some of them are still a little bit foggy for me and I'd be glad to learn more about them 🎉 

But that's probably a subject for the future 😁 

Having that said, I'm approving this PR 🎉 
Thanks for addressing all comments and for sharing your POV on the mentioned subject 💪

Vorbert-Kruk (dismissed):

widgets/executions/src/ExecutionsTable.tsx
kubama: These props are not used by the component, so having them defined has no effect.

qooban: OK 👍🏻

widgets/executions/src/DryRunIcon.tsx
kubama: Done

kubama:
> Named type for props was unspoken pattern we followed so far

Actually, I don't think I was ever following such pattern :slightly_smiling_face: . I was just using named interface each time it was obviously beneficial to do so.

> please add a point to weekly agenda.

Done.

widgets/executions/src/widget.tsx
```diff
@@ -13,18 +16,19 @@ export interface ExecutionsWidgetParams {
...
 
-const t = Stage.Utils.getT('widgets.executions');
+export const translate = Stage.Utils.getT('widgets.executions');
```
qooban, line 25: I'm not fully sure we should use such generic names for exported functions that are doing something more specific.
That's why I suggested `translateExecutionsWidget`.

I know that:
1. it's longer, but isn't it something most of the team wanted to introduce (suggestion from Bartosz L.)? 🤔 
2. it's defined in specific widget and we are (at least currently) not importing anything exported by the widget from the outside of the widget

kubama: > it's longer, but isn't it something most of the team wanted to introduce?

As far as I understand the idea was to use the term 'translate` instead of 't' when naming such functions.

Anyway, the module from which an item is exported serves as its context. Thus it's common to assume that this context is known. For example, `react` module exports `Component` class, not `ReactComponent` to better emphasis what it actually is. Further, it has a method named `render`, not `renderReactComponent`. I can't think of any exception from such context-aware way of naming.
Of course when importing individual items this context is lost. If this leads to a confusion then `import ... as` is what should be used.
Back to our particular case - using 'translateExecutionsWidget' name would mean duplicating the information from the context, which is the name and placement of the containing file. If the name 'translate' is confusing in the context of the file where it's imported then `import ... as` should be used as opposed to different export name.

qooban: > As far as I understand the idea was to use the term 'translate` instead of 't' when naming such functions.

OK, in that case I think it's again worth discussing. Please add a point in UI weekly.
See what kind of functions were already introduced in `cloudify-stage` after we changed that rule. Notice that all exported translate functions are named to expose the context.

image

> Anyway, the module from which an item is exported serves as its context. Thus it's common to assume that this context is known. For example, react module exports Component class, not ReactComponent to better emphasis what it actually is. Further, it has a method named render, not renderReactComponent. I can't think of any exception from such context-aware way of naming.

I think you cannot compare `react` (very generic library) to Executions widget (very specific, and not generic widget).

> Back to our particular case - using 'translateExecutionsWidget' name would mean duplicating the information from the context, which is the name and placement of the containing file.

"name and placement of the containing file" speaks to me. It's valid argument, but I'm still not sure it's sufficient.

I'm approving the PR, but at the same time I'd like to think it over and discuss with the whole team during next UI weekly meeting.

kubama: > Please add a point in UI weekly.

Done

PR #2382, opened 2023-01-16
RD-5760 Part 2 - Restructure `app/components` directory

qooban:
## Description

Changed `app/components` directory structure to be easier to navigate and to understand relationships:

| Before | After |
|---|---|
| image | image |

Main changes:
* `basic/index.ts` file changed into `basic.ts` file (and `basic` folder removed as there were no other components there)
* `shared/index.ts` file changed into `shared.ts` file (and `shared` folder removed as all components were distributed)
* created/refactored feature folders - `maintenanceMode`, `license`, `templateManagement` 
* `common` folder contains miscellaneous components
* `overlays` - contact details, getting started modals
* `page` folder was created to collect all folders/files used on pages with widgets defined content + edit mode
* `routes` folder contains router related files + simple pages (we can consider moving simple pages into subfolder if needed)

Details in the PR. I encourage reviewers to checkout the branch and navigate to see the current result.
I'm open for further adjustments.

## Screenshots / Videos
N/A

## Checklist

## Tests
1. First version - [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4329/)
2. After code review adjustments - [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4335/) - failed on `getting_started_spec`, but seems unrelated (the same test passed in [Stage-UI-System-Test #4341](https://jenkins.cloudify.co/blue/organizations/jenkins/Stage-UI-System-Test/detail/Stage-UI-System-Test/4341))

## Documentation
Seems there's nothing to update.


app/components/AddButton.tsx
```diff
@@ -1,25 +0,0 @@
-// @ts-nocheck File not migrated fully to TS
```
qooban, line 1: This file was not used => removed.

app/components/shared.ts
```diff
@@ -0,0 +1,41 @@
+import ExecutionStatus from './misc/ExecutionStatus';
```
qooban, line 1: This is actually a file moved from `app/components/shared/index.ts`.

Just not detected properly by GitHub.

qooban:
@kubama This is a draft (as documentation and tests were not checked yet), but wanted to get your initial opinion on the changes.

kubama:
Some loose thoughts from me:
- I'm not sure about the 'widgetsPage' term which seems to be something new, not used anywhere else. Why not to use just the commonly used term 'page'? It seems to be referring to the same entity in the end.
- How about moving `PasswordModal.tsx` to `misc`?
- I'm not sure about `modals` being inside `widgetsPage`. These are some overlay components appearing on (first) login, not directly related to the 'page' entity. How about moving them one level up and naming them something like `initialOverlays` or maybe `firstLogin`?
- I'm also not sure about `editMode` and `sidebar` being under `widgetsPage`, as they seem to contain components of broader scope than just the 'page'

kubama:
> I can move them one level up, but regarding the name - probably `modals` or `overlays` would be more accurate, what do you think?

Maybe 'overlays' then? 'Modals' seem to be very generic.

> My thinking here was that currently they are only used on pages with widgets.

Didn't think of it this way, but you're right. Still a little bit counter-intuitive for me as logically a sidebar is something spanning multiple pages as if it was on a higher, not lower hierarchy level.

qooban:
@kubama 

> Maybe 'overlays' then? 'Modals' seem to be very generic.

OK, done.

> Didn't think of it this way, but you're right. Still a little bit counter-intuitive for me as logically a sidebar is something spanning multiple pages as if it was on a higher, not lower hierarchy level.

Right, makes sense. In that case I'm moving `sidebar` one level up.

Updated description and checked documentation (seems to me that there's nothing to update).

Vorbert-Kruk (commented):
From my side, I can also share some suggestions 😁 

### 1. `views` directory extraction
There are some directories which are extracted to indicate that the domain of those specific directories is a specific page in the application
To be specific, I'm referring to:
- `app/components/license`
- `app/components/maintenanceMode`
- `app/components/templateManagement`

Maybe we could extract a separate directory named e.g. `views` in which we could store those kind of pages? 🤔 

We can even go one level further and ask ourselves if we should even create a dedicated directory for those views directly in `app` directory (not in `app/components`) 🤔 

### 2. Modifying `misc` directory name
`misc` name seems to be very mysterious and I find it confusing as I don't have a glimpse of an idea what does it stand for 😅 
Maybe renaming it would be a good idea? 🤔 
If it's just containing shared components, we could name it `common` as we are doing in other parts of application, when something is shareable (e.g. `app/widgets/common`) 🚀

qooban:
@Vorbert-Kruk @kubama Thanks for initial comments, I updated description after changes. Let's do proper code review now. I'm open for further changes, but I'd like to limit it not to go outside `app/components` in this PR and RD-5760.

kubama (approved):

Vorbert-Kruk:
@qooban no problem 😉 

> page should also go there and this is quite big folder   

I was thinking about a different differentiation as `page` directory and components within are being used across not only `views` (I'll keep using that terminology from my previous comment 😁) but also across other components (such as e.g. [Home.tsx](https://github.com/cloudify-cosmo/cloudify-stage/tree/master/app/components/Home.tsx)) - so `page` directory seems to is matching `app/components` idea 🎉 

> license and maintenanceMode contain files used not only as License page and Maintenance Mode page (so they indeed are feature folders)

Good point, there may be some room for discussion on how strict we want to be with shareable components and their placement

> Maybe we could extract only *Page files to views folder, but probably then we should combine it with changes in routes folder (push it one level up). Regarding app directory structure refactoring - at least in this PR and probably at least for short-term I'd stay with app/components. I'm open to discuss this, but let's not consider it for this PR and for [RD-5760](https://cloudifysource.atlassian.net/browse/RD-5760). What do you think about creating a new ticket for the future?

I don't know about creating a dedicated ticket for that right now
As we are discussing some potential ideas, which would need some more conversation room (I can see that some ideas may be understood differently and the deeper synchronization would be beneficial), I don't think that creating ticket right now will gain as any value
Let's get back to the subject when similar discussions will appear in the future, okay? 😉 

> misc is a shortcut from miscellaneous, but I guess I can go with common here.   

Nice, thanks for both explanation (I'd never have never guessed that 😄) and for applying the suggestion 🎉

Vorbert-Kruk (approved):
LGTM 🎸 
Thanks for adding me to this PR and for addressing all comments 🎉

PR #2427, opened 2023-02-06
RD-6876 Update secret with secret provider in Secrets widget when provider connection is broken

PatrycjaKubiczek:
## Description

RD-6876

After the discussion with @qooban via Slack I decided to put this PR to review, however there is still one situation that the backend currently doesn't cover and it doesn't work: when you want to update the secret with the existing secret provider and you want to delete the secret provider and add only value. I think this requires work on the backend - but this PR will still improve the current status of Stage and Secrets Widget functionality. 

## Screenshots / Videos

n/a

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4477/) secret_spec ✔️ 
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4479/) secret_providers_spec ✔️ 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4489/) secret_spec ✔️ (09/02/23)

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4515/) secret_spec & secret_providers_spec  (15/02/23)
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4517/) all tests -(15/02/23)
[Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/4519/) rerun previous tests (16/02/23)

## Documentation

n/a


app/widgets/common/secrets/SecretActions.ts
```diff
@@ -30,6 +30,10 @@ export default class SecretActions {
         return this.manager.doGet(`/secrets/${key}`);
     }
 
+    doGetSkipValue(key: Secret['key']): Promise<Secret> {
+        return this.manager.doGet(`/secrets/${key}?_skip_value=true`);
+    }
+
```
qooban, line 36: 1. I think the type is not fully accurate (as you don't have `value` inside).
2. I suggest also to change function name

```suggestion
    doGetSecretWithoutValue(key: Secret['key']) {
        return this.manager.doGet<Omit<Secret, 'value'>>(`/secrets/${key}?_skip_value=true`);
    }

```

widgets/secrets/src/UpdateModal.tsx
```diff
@@ -44,18 +43,27 @@ export default function UpdateModal({ open, secret, toolbox, onHide }: UpdateMod
...
+                    });
             } else {
-                setUseSecretProvider(false);
+                unsetUseSecretProvider();
                 actions
                     .doGet(secret.key)
                     .then(({ value }) => {
                         setSecretValue(value);
+                        setSecretProvider('');
+                        setSecretProviderPath('');
```
qooban, line 66: Will
```suggestion
                        clearSecretProvider();
                        clearSecretProviderPath();
```
work here?

widgets/secrets/src/UpdateModal.tsx
```diff
@@ -44,18 +43,27 @@ export default function UpdateModal({ open, secret, toolbox, onHide }: UpdateMod
...
+                actions
+                    .doGetSkipValue(secret.key)
+                    .then(({ provider_name: providerName, provider_options: providerOptions }) => {
+                        setSecretProvider(providerName);
+
+                        try {
+                            const parsedSecretProvideOptions = JSON.parse(providerOptions as unknown as string);
+                            setSecretProviderPath(parsedSecretProvideOptions.path);
+                        } catch (e) {
+                            setSecretProviderPath('');
```
qooban, line 56: Will
```suggestion
                            clearSecretProviderPath();
```
work here?

widgets/secrets/src/UpdateModal.tsx
```diff
@@ -72,13 +80,22 @@ export default function UpdateModal({ open, secret, toolbox, onHide }: UpdateMod
 
     function updateSecret() {
         clearErrors();
-        if (isEmpty(secretValue) && !useSecretProvider) {
+        if (!useSecretProvider && isEmpty(secretValue)) {
```
qooban, line 83: Isn't `secretValue` of string type? 🤔 
If that's true then probably we could drop `isEmpty` here.

widgets/secrets/src/UpdateModal.tsx
```diff
@@ -72,13 +80,22 @@ export default function UpdateModal({ open, secret, toolbox, onHide }: UpdateMod
...
+        if (useSecretProvider) {
+            if (isEmpty(secretProvider) || isEmpty(secretProviderPath)) {
+                setErrors({
+                    secretProvider: isEmpty(secretProvider)
+                        ? translateForm('errors.validation.secretProvider')
+                        : undefined,
+                    secretProviderPath: isEmpty(secretProviderPath)
+                        ? translateForm('errors.validation.secretProviderPath')
+                        : undefined
+                });
```
qooban, line 96: `secretProvider` and `secretProviderPath` are strings, so we can probably drop `isEmpty`, right?

widgets/secrets/src/UpdateModal.tsx
```diff
@@ -110,7 +128,15 @@ export default function UpdateModal({ open, secret, toolbox, onHide }: UpdateMod
...
     function onSecretProviderChange() {
         clearErrors();
-        setUseSecretProvider(!useSecretProvider);
+
+        if (useSecretProvider) {
+            unsetUseSecretProvider();
+            clearSecretProvider();
+            clearSecretProviderPath();
+        }
+        if (!useSecretProvider) {
```
qooban, line 137:
```suggestion
        } else {
```

app/widgets/common/secrets/SecretActions.ts
Vorbert-Kruk: +1 on a different function name

`doGetSkipValue` seems related to its URL parameters but from another developer perspective, it may be confusing as use case of that parameter may not be known by the developer

Making the function/variable names related to their purpose, rather than their insides, seems to be a good strategy and a good rule of thumb 💪

widgets/secrets/src/UpdateModal.tsx
```diff
@@ -89,8 +106,9 @@ export default function UpdateModal({ open, secret, toolbox, onHide }: UpdateMod
         setLoading();
 
         const actions = new Stage.Common.Secrets.Actions(toolbox.getManager());
+        const formattedOptions = secretProviderPath ? { path: secretProviderPath } : undefined;
         actions
-            .doUpdate(secret.key, secretValue, secretProvider, secretProviderOptions)
+            .doUpdate(secret.key, secretValue, secretProvider, formattedOptions)
```
Vorbert-Kruk, line 111: ### Optional

In [CreateModal](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/widgets/secrets/src/CreateModal.tsx#L68) we have a similar line, but with a different name
As the name seems to be more accurate IMO, we could try to apply it also in the current example 😁 

```suggestion
        const secretProviderOptions = secretProviderPath ? { path: secretProviderPath } : undefined;
        actions
            .doUpdate(secret.key, secretValue, secretProvider, secretProviderOptions)
```

Vorbert-Kruk (commented):
Left one optional suggestion and some comments under @qooban discussions

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: 1. Changed name, you're absolutely right guys - that's way better than mine idea 😅 
2. I went for  `doGetWithoutValue(key: Secret['key']): Promise<Omit<Secret, 'value'>> {` because when I modified `return` - I got TS error

widgets/secrets/src/UpdateModal.tsx
PatrycjaKubiczek: yes, implemented https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/b28f9c3a558f11b5c80d9bbfc59a670c102503ba ✔️

widgets/secrets/src/UpdateModal.tsx
PatrycjaKubiczek: yes, implemented https://github.com/cloudify-cosmo/cloudify-stage/commit/b28f9c3a558f11b5c80d9bbfc59a670c102503ba ✔️

widgets/secrets/src/UpdateModal.tsx
PatrycjaKubiczek: implemented https://github.com/cloudify-cosmo/cloudify-stage/commit/b28f9c3a558f11b5c80d9bbfc59a670c102503ba ✔️

widgets/secrets/src/UpdateModal.tsx
PatrycjaKubiczek: implemented https://github.com/cloudify-cosmo/cloudify-stage/commit/b28f9c3a558f11b5c80d9bbfc59a670c102503ba ✔️

widgets/secrets/src/UpdateModal.tsx
PatrycjaKubiczek: implemented https://github.com/cloudify-cosmo/cloudify-stage/commit/b28f9c3a558f11b5c80d9bbfc59a670c102503ba ✔️

widgets/secrets/src/UpdateModal.tsx
PatrycjaKubiczek: thanks for noticing, implemented: https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/5a5235b789c1b0dec6319d39fb77591d35bd37e5 ✔️

Vorbert-Kruk (approved):
LGTM 💪

Vorbert-Kruk:
### Note
I can see that you've modified `formattedOptions` variable in `CreateModal` to appear the same as `secretProviderOptions` variable in `UpdateModal` (5a5235b789c1b0dec6319d39fb77591d35bd37e5)
But after that you've modified the shape of the `secretProviderOptions` in `UpdateModal` ([822296a](https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/822296a33f1e8ebd132855f0b3cfc46c3eafe6f0#diff-86dea181f436d9c1a482bb2cc809c0c0e4eb9d1f82c50e7d7e72dbbf19f8a20bL68))

I'm not certain why this change was made, but just wanted to point it out as I find it a little bit strange 😄 

Anyway, I don't think that this modification is a blocker for this PR, but just wanted to point it out 😁

qooban:
I'll review it tomorrow.

widgets/secrets/src/widget.utils.ts
```diff
@@ -1,2 +1,6 @@
 export const translateSecrets = Stage.Utils.getT('widgets.secrets');
 export const translateForm = Stage.Utils.getT('widgets.secrets.form');
+
+export const getSecretProviderOptions = (useSecretProvider: boolean, secretProviderPath: string) => {
+    return useSecretProvider ? { path: secretProviderPath } : undefined;
+};
```
Vorbert-Kruk, line 6: ## Optional

```suggestion
import type { ProviderOptions } from 'app/widgets/common/secrets/SecretActions';

export const getSecretProviderOptions = (useSecretProvider: boolean, secretProviderPath: string): ProviderOptions => {
    return useSecretProvider ? { path: secretProviderPath } : undefined;
};
```

widgets/secrets/src/UpdateModal.tsx
```diff
@@ -44,18 +43,25 @@ export default function UpdateModal({ open, secret, toolbox, onHide }: UpdateMod
...
+                setUseSecretProvider();
+                actions
+                    .doGetWithoutValue(secret.key)
+                    .then(({ provider_name: providerName, provider_options: providerOptions }) => {
+                        setSecretProvider(providerName);
+
+                        if (providerOptions) {
+                            const parsedSecretProvideOptions = JSON.parse(providerOptions);
+                            setSecretProviderPath(parsedSecretProvideOptions.path);
+                        }
```
qooban, line 55: If `providerOptions` is a string which we have to parse, then probably safer would be to wrap `JSON.parse` with `try-catch` block.

```suggestion
                        if (providerOptions) {
                            try {
                                const parsedSecretProvideOptions = JSON.parse(providerOptions);
                                setSecretProviderPath(parsedSecretProvideOptions.path);
                            } catch (error) {
                                setErrors( . . . );
                            }
                        }
```

Ref.: https://stackoverflow.com/questions/4467044/proper-way-to-catch-exception-from-json-parse

widgets/secrets/src/widget.utils.ts
qooban: I would even go further and use named parameters here to avoid using boolean argument.

```suggestion

export const getSecretProviderOptions = ({useSecretProvider: boolean, secretProviderPath: string}): ProviderOptions => {
    return useSecretProvider ? { path: secretProviderPath } : undefined;
};
```

Ref.: https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/

app/widgets/common/secrets/SecretActions.ts
qooban: Ad 2. 
```
    doGetWithoutValue(key: Secret['key']) {
        return this.manager.doGet<Omit<Secret, 'value'>>(`/secrets/${key}?_skip_value=true`);
    }
```
^ it doesn't work? 🤔 I just checked locally and have no issues. Am I missing something?

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -15,7 +15,7 @@ export type Secret = {
     value: string;
     schema: string;
     provider_name?: string;
-    provider_options?: ProviderOptions;
+    provider_options?: string;
```
qooban, line 18: Question: So in `GET /secrets` we will receive `string`, but in `PUT /secrets` or `PATCH /secrets` it is expected to be provided as `dict` right?

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -15,7 +15,7 @@ export type Secret = {
     value: string;
```
qooban, line 15: Optional:

I'm wondering about `?` signs in this type. If it's used with `GET /secrets` without `_include` then we are sure that this value is present or am I missing something?

test/cypress/integration/widgets/secrets_spec.ts
```diff
@@ -74,9 +75,19 @@ describe('Secret store management widget', () => {
         });
 
         cy.getWidget('secrets').within(() => {
-            cy.getSearchInput().type(secretName);
+            cy.createSecretProvider({ name: newSecretProviderName, type: 'vault', visibility: 'global' });
+            cy.get('.rowActions').find('i[title="Update secret"]').click();
+        });
+
+        cy.get('.modal').within(() => {
+            cy.setSingleDropdownValue(secretProviderName, newSecretProviderName);
```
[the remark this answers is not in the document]
PatrycjaKubiczek: Cypress says it works :) 
so:
```
setSingleDropdownValue: (fieldName: string, value: string) =>
        cy
            .getField(fieldName)
            .click()
            .within(() => cy.contains('div[role=option]', value).click()),
```
`setSingleDropdownValue` has `getField` which is basically `cy.contains('.field', fieldName)` so that grabs the whole field which contains Secret_Provider_1, and then within that field we're looking for a specific value (Secret_Provider_1), and clicking on it. Should I change it? I think it can work both ways

test/cypress/integration/widgets/secrets_spec.ts
```diff
@@ -74,9 +75,19 @@ describe('Secret store management widget', () => {
         });
 
         cy.getWidget('secrets').within(() => {
-            cy.getSearchInput().type(secretName);
+            cy.createSecretProvider({ name: newSecretProviderName, type: 'vault', visibility: 'global' });
```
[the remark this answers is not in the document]
PatrycjaKubiczek: nope, modified it - https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/a2c62f0f777ac057f9ca4e873dd4480d36872b53

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: Zrzut ekranu 2023-02-14 125911
I think that's is correct

widgets/secrets/src/UpdateModal.tsx
PatrycjaKubiczek: I even added try catch block earlier (https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/5163db5a5532e80dea7bbd55bb87da62aa37e813) :) but then modify it too many times so I lost it somewhere between commits - add it again: https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/e360c5395f2b5c874f7eec8c863a4d556c8cbbbc

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: okay, I'm pretty sure that when I was checking earlier it was throwing a TS error, however it's working now 🤔 I might have missed some type earlier, but since it's working right now I updated it accordingly: https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/10585cbba8c7cb6f4871c7caf838a6f657d3dde2 ✔️

widgets/secrets/src/widget.utils.ts
PatrycjaKubiczek: hmm I need to change the code a bit so the second option would be an empty `{}` not undefined ( https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/10585cbba8c7cb6f4871c7caf838a6f657d3dde2 ✔️ ) - however, I'm not sure how should I implement @qooban's suggestion in that scenario 🤔

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: hmm you're correct, however I think in order to make it work I need to add some example data for Secret in SecretsTable state https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/10585cbba8c7cb6f4871c7caf838a6f657d3dde2 I'm not sure if that's right way? 🤔

qooban: Probably we could play with `secret` state variable in `SecretsTable` and allow having it `undefined` (`secret?: Secret;` in `SecretsTableState`) => You won't need to create empty Secret object then. 

You would just need to update `deleteSecret`. Looking at the code it seems safe to have `const secretKey = secret!.key;` in L150.

test/cypress/integration/widgets/secrets_spec.ts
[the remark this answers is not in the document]
qooban: Oh, this way... Hm... I'm afraid that might be not clear for others at first glance, so I would change it to:

```suggestion
            cy.setSingleDropdownValue("Secret provider", newSecretProviderName);
```

widgets/secrets/src/widget.utils.ts
qooban: Hmm... That's interesting 🙂 

Actually looking at that function and it's uses I'm wondering shouldn't we simplify it (then probably we won't have here any issue boolean arguments). 

Right now `doCreate` and `doUpdate` have two optional arguments. 
(BTW Side note: `doCreate` has `providerOptions?: ProviderOptions | null` while `doUpdate` has `providerOptions?: ProviderOptions`. I guess we can drop `null` in the first type.)

What if `providerName` is empty string or undefined - does `providerOptions` matter at all? I guess it shouldn't.
How about doing all that provider options transformations (`string` -> `{ path: string }`) in `SecretActions` and only if provider name is specified - that should allow you to avoid unnecessary conditions and always generate provider options in the same shape:

```
export const getSecretProviderOptions = (path: string): ProviderOptions => ({ path });
```

qooban (commented):
Replied in open discussions.

test/cypress/integration/widgets/secrets_spec.ts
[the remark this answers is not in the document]
PatrycjaKubiczek: sure thing, https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/e4bd416cedb6a9ab166ce7cfac99ee76583017cb updated it 💪

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: that's a good idea, thanks! :) implemented it: https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/6adb0d38273375f793a58c6d41d83523db401a16 ✔️

widgets/secrets/src/widget.utils.ts
PatrycjaKubiczek: implemented these changes: https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/0d2da85c3c3478c850586ef0f2227aa5a556696f is this what you had in mind @qooban ? :)

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -48,23 +59,34 @@ export default class SecretActions {
...
+            body = {
+                ...body,
                 provider_name: providerName,
-                provider_options: providerOptions
-            }
-        });
+                provider_options: getSecretProviderOptions(providerPath)
+            };
+        }
+        return this.manager.doPut(`/secrets/${key}`, { body });
```
qooban, line 77: Optional:

I guess you will be able to get better type security with:
```suggestion
        return this.manager.doPut<Secret, UpdateSecretRequestBody>(`/secrets/${key}`, { body });
```

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -48,23 +59,34 @@ export default class SecretActions {
...
+    doUpdate(key: Secret['key'], value?: string, providerName?: string, providerPath?: string) {
+        let body: any = { value };
+        if (providerName && providerPath) {
+            body = {
+                ...body,
+                provider_name: providerName,
+                provider_options: getSecretProviderOptions(providerPath)
+            };
+        }
+        return this.manager.doPatch(`/secrets/${key}`, { body });
```
qooban, line 89: Optional:

I guess you will be able to get better type security with:
```suggestion
        return this.manager.doPatch<Secret, UpdateSecretRequestBody>(`/secrets/${key}`, { body });
```

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -48,23 +59,34 @@ export default class SecretActions {
...
-        });
+    doUpdate(key: Secret['key'], value?: string, providerName?: string, providerPath?: string) {
+        let body: any = { value };
+        if (providerName && providerPath) {
+            body = {
+                ...body,
+                provider_name: providerName,
+                provider_options: getSecretProviderOptions(providerPath)
+            };
+        }
```
qooban, line 88: Optional:

Or simply:
```suggestion
        if (providerName && providerPath) {
            body.provider_name = providerName;
            body.provider_options = getSecretProviderOptions(providerPath);
        }
```

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -48,23 +59,34 @@ export default class SecretActions {
...
-        providerOptions?: ProviderOptions | null
+        providerPath?: string
     ) {
-        return this.manager.doPut(`/secrets/${key}`, {
-            body: {
-                value,
-                visibility,
-                is_hidden_value: hidden,
+        let body: SecretRequestBody = {
+            key,
```
qooban, line 65: I think we don't need `key` in the body, it's passed in URL.

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -17,6 +19,15 @@ export type Secret = {
...
 };
+
+interface SecretRequestBody {
+    key: string;
+    value: string;
+    visibility: Visibility;
+    is_hidden_value: boolean;
+    provider_name?: string;
+    provider_options?: ProviderOptions;
+}
```
qooban, line 30: Optional:

If you introduced it here, then I think we should be more precise. I went through https://docs.cloudify.co/api/staging/dev/#create-secret and https://docs.cloudify.co/api/staging/dev/#update-secret, so WDYT about making it more accurate (in terms of fields presence and required/optional indicator):
```suggestion
type CreateSecretRequestBody = Pick<Secret, 'value'> & Optional<Pick<Secret, 'visibility' | 'is_hidden_value' | 'schema' | 'provider_name' | 'provider_options'>>
```
There's also `update_if_exists`, but I think we don't have to include it for now.

If we are introducing it for create, then probably we could also do it for update:
```
type UpdateSecretRequestBody = Omit<CreateSecretRequestBody, 'schema'>;
```
There's `creator` also, but similarly to `update_if_exists` I don't think we need it now.

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -17,6 +19,15 @@ export type Secret = {
     provider_name: string;
```
qooban, line 19: Optional:

L18 - according to https://docs.cloudify.co/api/staging/dev/#the-secret-resource `schema` seems to be a dict, so I would go with `schema: Record<string, any>`

app/widgets/common/secrets/SecretActions.ts
```diff
@@ -48,23 +59,34 @@ export default class SecretActions {
...
+        }
+        return this.manager.doPut(`/secrets/${key}`, { body });
     }
 
-    doUpdate(key: Secret['key'], value?: string, providerName?: string, providerOptions?: ProviderOptions) {
-        return this.manager.doPatch(`/secrets/${key}`, {
-            body: { value, provider_name: providerName, provider_options: providerOptions }
-        });
+    doUpdate(key: Secret['key'], value?: string, providerName?: string, providerPath?: string) {
+        let body: any = { value };
```
qooban, line 81: Optional:

```suggestion
        let body: UpdateSecretRequestBody = { value };
```

qooban (approved):

widgets/secrets/src/widget.utils.ts
qooban: Yep 👍🏻 Nice, thx.

app/widgets/common/secrets/SecretActions.ts
qooban: Updated my previous comment. Not `Optional`, but `Partial` 🙂 Sorry.

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: removed it https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/671183b9f06cb0bdf4e05515089934bb4e895100 ✔️

qooban (approved):
I'm approving in advance as I need to go and I'm not sure when I'll be back.
Thanks for patience 🙂 

As alwyas - please verify with system tests.

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: implemented: https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/0fa559952b71187d19369d99957c8610211a7d73 ✔️

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/0fa559952b71187d19369d99957c8610211a7d73 ✔️

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: it should CreateSecretRequestBody not UpdateSecretRequestBody in that case :) https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/fed30641b28375a29e8da15227b22f287c0c8e0a - done ✔️

app/widgets/common/secrets/SecretActions.ts
PatrycjaKubiczek: done: https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/0fa559952b71187d19369d99957c8610211a7d73 ✔️

app/widgets/common/inputs/fields/ValueListInputField.tsx
```diff
@@ -23,7 +23,7 @@ const parseOptionValue = (value: any) => {
     return isStringifiedNumber(value) ? `"${value}"` : value;
 };
 
-function getOptionsFromSecret(secretValue: string, schema: string) {
+function getOptionsFromSecret(secretValue: string, schema: Record<string, any>) {
     if (schema) {
```
Vorbert-Kruk, line 27: If `schema` is now an object, statement `if (schema)` will always return true 😄 
So maybe some adjustments may be needed here 🤔

test/cypress/integration/widgets/secrets_spec.ts
```diff
@@ -68,15 +69,27 @@ describe('Secret store management widget', () => {
...
+        cy.createSecretProvider({ name: newSecretProviderName, type: 'vault', visibility: 'global' });
+
+        cy.getWidget('secrets').within(() => {
+            cy.getSearchInput().clear().type(secretName);
+            cy.get('.rowActions').find('i[title="Update secret"]').click();
+        });
+
+        cy.get('.modal').within(() => {
+            cy.setSingleDropdownValue('Secret provider', newSecretProviderName);
+            cy.getField('Path at the provider').find('input').type('new/path');
```
Vorbert-Kruk, line 86: ### Optional
There is `cy.typeToFieldInput` command, which does the same job as `cy.getField(...).find('input').type(...)` 🎉 
We could try to reuse it here, but it's not mandatory 😉  - just wanted to spread knowledge about this command 💪

widgets/secrets/src/UpdateModal.tsx
```diff
@@ -72,25 +80,29 @@ export default function UpdateModal({ open, secret, toolbox, onHide }: UpdateMod
...
         }
-        if (isEmpty(secretProvider) && useSecretProvider) {
-            setErrors({ secretProvider: translateForm('errors.validation.secretProvider') });
-            return;
-        }
-
-        if (!isEmpty(errors)) {
-            return;
+        if (useSecretProvider) {
+            if (!secretProvider || !secretProviderPath) {
```
Vorbert-Kruk, line 88: Nested `if` statements have been introduced, but the root one (`if (useSecretProvider)`) is not having any else statements or operations other than those executed within `if (!secretProvider || !secretProviderPath)`

Because of that, I think that we should be able to merge those ifs 💪 

While doing so, we may try to extract it as a variable, as reading 3 conditions inside the render function may be less confusing than reading a descriptive variable name 🚀

PatrycjaKubiczek: added addional variables - I hope it's much cleaner now 💪 https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/72a6c9aac1924ef984c4e95b4914ffc307cf078e

test/cypress/integration/widgets/secrets_spec.ts
PatrycjaKubiczek: change it to `typeToFieldInput` https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/72a6c9aac1924ef984c4e95b4914ffc307cf078e ✔️ always forget about this command :)

app/widgets/common/inputs/fields/ValueListInputField.tsx
PatrycjaKubiczek: added `!isEmpty` https://github.com/cloudify-cosmo/cloudify-stage/pull/2427/commits/72a6c9aac1924ef984c4e95b4914ffc307cf078e to check object ✔️

qooban (approved):

test/cypress/integration/widgets/secrets_spec.ts
Vorbert-Kruk: No worries, I have the same issue with similar commands 😄

PR #2436, opened 2023-02-15
RD-6319 implemented special validation on csys-location-lat and long

Blewin:
## Description

special validation for labels `csys-location-lat` and `csys-location-long`

Part of the story is to add validation for those labels:

The value must be a number (decimal)

Latitude must be a number between -90 and 90, if the value is out of range an error message The latitude must be a number between -90 and 90 should be presented 

longitude must be a number between -180 and 180, if the value is out of range an error message longitude between -180 and 180 should be presented

No multiple values (list) are allowed, a user tries to add multiple values the error message longitude can't have multiple values or latitude can't have multiple values

## Screenshots / Videos

image

## Checklist

## Tests

 

[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4548/)


Vorbert-Kruk:
Will do the review tomorrow

app/translations/en.json
```diff
@@ -728,6 +728,8 @@
                 },
                 "fetchingLabelsError": "Cannot fetch labels for deployment '{{deploymentId}}'",
                 "labelDuplicationError": "Cannot add the same label twice",
+                "validationNumber": "This label should be a number between {{ from }} and {{ to }}",
```
kubama, line 731:
```suggestion
                "labelNumberValidationError": "This label should be a number between {{ from }} and {{ to }}",
```

app/widgets/common/labels/LabelsInput.tsx
```diff
@@ -41,6 +43,34 @@ function useReservedKeys(toolbox: Stage.Types.Toolbox) {
     return { reservedKeys, fetchingReservedKeys };
 }
 
+const getLabelNumberValidation =
+    (minmax: number) =>
```
kubama, line 47: So it's min or max? :slightly_smiling_face:

app/widgets/common/labels/LabelsInput.tsx
```diff
@@ -41,6 +43,34 @@ function useReservedKeys(toolbox: Stage.Types.Toolbox) {
...
+        }
+
+        return undefined;
+    };
+
+/**
+ * keys of this record will be validated,
+ * additionally, there will be a check if they are added more than once
+ */
+const specialValidation: Record<string, (v: string) => (string | undefined) | undefined> = {
```
kubama, line 62: Very mysterious name for a variable :slightly_smiling_face:. Can we be more specific so it's obvious what's special about it?

kubama (commented):
Would be good to see a test case covering the added logic.

app/widgets/common/labels/LabelsInput.tsx
Vorbert-Kruk, line 62: Besides [kubama comment](https://github.com/cloudify-cosmo/cloudify-stage/pull/2436/files#r1108110066) I may also point out that we are using an acronym as a parameter name 

```typescript
(v: string)
```

We should be able to make this name more meaningful 💪

app/widgets/common/labels/LabelsInput.tsx
```diff
@@ -77,9 +107,20 @@ const LabelsInput: FunctionComponent<LabelsInputProps> = ({
         const allLabels = [...labels, ...(hideInitialLabels ? initialLabels : [])];
         return !!_.find(allLabels, newLabel);
     })();
+    const specialValidationError =
+        specialValidation[newLabelKey] &&
+        (specialValidation[newLabelKey](newLabelValue) ||
+            checkUniqueness(
+                newLabelKey,
+                labels.map(({ key }) => key)
+            ));
```
Vorbert-Kruk, line 116: Maybe it's due to a naming but I find this check to be confusing

- `specialValidation[newLabelKey]` we are checking if label should be checked for a numeric constraints
- `specialValidation[newLabelKey](newLabelValue)` is checking if `newLabelValue` is a value which meets the numeric contraints - if not, `checkUniqueness` function is being executed to check if `newLabelKey` is unique

For me it took a while to get all code pieces together, to understand what's going on - because of that I may have some suggestions

1. We are already checking if the label was duplicated by via `newLabelIsAlreadyPresent` variable, which is used to display `DuplicationErrorPopup`    
  Maybe we can then omit having additional `checkUniqueness` check (within `specialValidationError`) as every variable is being checked for that via `newLabelIsAlreadyPresent` variable? 

2. Maybe we could omit one of the condition checks
  Correct me if I'm wrong, but we could write
    ```typescript
        specialValidation[newLabelKey]?.(newLabelValue)
    ```
    instead of
    ```typescript
    specialValidation[newLabelKey] &&
          (specialValidation[newLabelKey](newLabelValue)
    ```
    which could make this code piece more readable 💪 


2. Maybe we could create a function for checking `specialValidation` for a certain label?
  Argumentation is that objects returning functions are not a common approach within this repository, thus it may be confusing for a developer
  To have the same result, we could just create a function that is taking 2 arguments: key and value of the new label 🎉

app/widgets/common/labels/LabelsInput.tsx
Blewin: both

app/widgets/common/labels/LabelsInput.tsx
Blewin: ad 1st: the newLabelIsAlreadyPresent only toggles when both key and value are the same. i think adding ifs to that procedure will be less readable than making a separate check

app/widgets/common/labels/LabelsInput.tsx
Blewin: changed to boundary

app/widgets/common/labels/LabelsInput.tsx
Blewin: very good suggestions,implemented, i overengineered :D

Vorbert-Kruk:
I'll do the review tomorrow

Vorbert-Kruk (commented):
Left some comments

But overall @Blewin, I may have one request
I can see that not all review comments have been addressed
After the investigation I can see that some of them were no longer valid after your changes
But IMHO it should be somehow indicated, by a simple comment or by resolving no longer valid discussion
From my POV, I have to invest my time into validating "why this discussion was unattended", which could be easily omited
In addition I assume, that points from previous review have been addressed in one way or another - without it, I could just conclude review with a "Please address open discussions" comment, which is rather not desired 😄 
So the request from my side would be to address all non-optional discussions before asking for re-review
It will safe some time for changes initiator and a reviewer
In addition, it won't make an impression on a reviewer that his work was ignored, which is also nice

app/widgets/common/labels/LabelsInput.tsx
Vorbert-Kruk: Happy to hear that the suggestions were beneficial 🎉

test/cypress/support/labels.ts
```diff
@@ -12,12 +12,15 @@ declare global {
 const commands = {
     typeLabelKey: (key: string) => cy.get('div[name=labelKey] > input').clear().type(key),
     typeLabelValue: (value: string) => cy.get('div[name=labelValue] > input').clear().type(value),
-    addLabel: (key: string, value: string) => {
+    prepareAddingLabels: (key: string, value: string) => {
```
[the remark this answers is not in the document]
Blewin: haha, i was struggling so much with this name, THANK YOU

app/translations/en.json
Blewin: done

app/widgets/common/labels/LabelsInput.tsx
```diff
@@ -41,6 +43,27 @@ function useReservedKeys(toolbox: Stage.Types.Toolbox) {
     return { reservedKeys, fetchingReservedKeys };
 }
 
+const latLongKeys = ['csys-location-lat', 'csys-location-long'];
+
+function validateLatLong(newValue: string, newLabelKey: string, existingLabelKeys: string[]) {
```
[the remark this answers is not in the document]
Blewin: very good idea, implemented

Blewin:
Noted, I will respond to every discussion, (it will help me not miss anything too :D)

test/cypress/integration/widgets/labels_spec.ts
```diff
@@ -59,4 +59,28 @@ describe('Labels widget', () => {
         });
         cy.contains('There are no Labels defined');
     });
+
+    it('should properly validate lat-long labels', () => {
```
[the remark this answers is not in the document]
Blewin: did it, but it required some ugly code, please validate (maybe https://stackoverflow.com/questions/65928822/cypress-how-to-temporarily-escape-from-a-cy-within ) would be better, but im not sure

test/cypress/integration/widgets/labels_spec.ts
```diff
@@ -59,4 +59,28 @@ describe('Labels widget', () => {
...
+            cy.typeLabelValue('101');
+            cy.get('.add').parent().should('have.attr', 'disabled');
+
+            cy.typeLabelValue('0');
+            cy.get('.add').parent().should('not.have.attr', 'disabled');
+
+            cy.typeLabelValue('sample_valeu');
+            cy.get('.add').parent().should('have.attr', 'disabled');
+
+            cy.contains('button', 'Cancel').click();
```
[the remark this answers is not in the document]
Blewin: implemented

test/cypress/integration/widgets/labels_spec.ts
```diff
@@ -67,20 +67,33 @@ describe('Labels widget', () => {
         cy.get('.modal').within(() => {
             getCreatedLabel().should('not.exist');
             cy.contains('button', 'Add').should('have.attr', 'disabled');
+        });
 
-            cy.prepareAddingLabels(CHECKED_KEY, '90');
+        cy.get('.modal').within(() => {
+            cy.fillLabelInputs(CHECKED_KEY, '90');
             cy.get('.add').parent().should('not.have.attr', 'disabled');
```
kubama, line 74: How about extracting it to a function with `boolean` argument?

Vorbert-Kruk (approved):
Looks great 💪  
Thanks for addressing all discussions and for having a little brainstorming on implementation ideas 🎉

test/cypress/integration/widgets/labels_spec.ts
Blewin: done

test/cypress/integration/widgets/labels_spec.ts
```diff
@@ -62,6 +61,10 @@ describe('Labels widget', () => {
 
     it('should properly validate lat-long labels', () => {
         const CHECKED_KEY = 'csys-location-lat';
+        const checkValidationErrorVisbility = (shouldBeVisible: boolean) =>
+            shouldBeVisible
+                ? cy.contains('This label should be a number').should('be.visible')
+                : cy.contains('This label should be a number').should('not.exist');
```
kubama, line 67:
```suggestion
            cy.contains('This label should be a number').should(shouldBeVisible ? 'be.visible' : 'not.exist');
```

kubama (approved):

PR #2437, opened 2023-02-15
RD-6900 Create basic compatibility layer for custom user configuration

qooban:
## Description

The following changes were implemented:

- conversion of `saml` section into `auth` section
- lodash import refactor according to Ongoing Refactors

## Screenshots / Videos
N/A

## Checklist

## Tests

Added backend test.

## Documentation

https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2517


backend/config.ts
```diff
@@ -10,25 +11,60 @@ import logging from '../conf/logging.json';
...
+function initUserConfig() {
+    try {
+        const userDataConfigPath = getResourcePath('userConfig.json', true);
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        let userDataConfig: UserDataConfig = require(userDataConfigPath);
+        let convertedSamlSection = false;
+
+        // Backward-compatibility layer - convert `saml` section to `auth` section
+        if (userDataConfig.saml?.enabled) {
+            console.warn(
```
qooban, line 30: Used `console` as it's currently not possible to use `logger` in this file (circular deps).

backend/config.ts
```diff
@@ -10,25 +11,60 @@ import logging from '../conf/logging.json';
...
+                certPath,
+                loginPageUrl: ssoUrl,
+                afterLogoutUrl: portalUrl
+            };
+
+            convertedSamlSection = true;
+        }
+
+        // Security reason - get only allowed parameters
+        userDataConfig = pick(userDataConfig, keys(flatten(userConfig, { safe: true })));
```
kubama, line 48: Not sure how it affects security... what bad would happen if non-supported parameter was provided and the line above was missing?

backend/config.ts
```diff
@@ -10,25 +11,60 @@ import logging from '../conf/logging.json';
...
+            console.warn(
+                `WARNING: User configuration was migrated automatically and it is ` +
+                    `available at ${newUserDataConfigPath}. Please verify it and use it to replace ${userDataConfigPath}.`
+            );
+        }
+
+        // Create full user configuration
+        userConfig = defaultsDeep(userDataConfig, userConfig);
+    } catch (err: any) {
+        if (err.code !== 'MODULE_NOT_FOUND') {
```
kubama, line 62: Isn't it just for `require` call? If so then how about placing try-catch just around it? I think the `if` could be then even dropped.

backend/config.ts
qooban: In L84 we are merging all configurations and that line secures accessing configuration fields that are not exposed through `userConfig.json`. IMHO we should not allow users to modify internal configuration, so that's why this kind of sanitization.

backend/config.ts
qooban: Seems like it's only for `require` call. Updated.

backend/config.ts
kubama: I don't  see any security value here. System administrator can edit other configuration files as well (even if it's not something officially supported).

kubama (approved):

PR #2447, opened 2023-02-22
RD-6890 Extended login error messages mapping mechanism 

Vorbert-Kruk:
## Checklist

### DIFF
--- a/widgets/blueprints/README.md
+++ b/widgets/blueprints/README.md
@@ -95,11 +95,12 @@ The default visibility is "Tenant", and according to the logged-in user's permis
 
 1. Click the deploy icon ![Deploy icon]( /images/ui/icons/deploy-icon.png ).   
 2. In the Deploy Blueprint dialog, specify a name for your deployment.
-3. Specify the required deployment inputs.   
+3. Specify environment on which the blueprint should be deployed on - this option is presented only when the blueprint is using [get_environment_capability](/developer/blueprints/spec-intrinsic-functions#get_environment_capability) intrinsic function.
+4. Specify the required deployment inputs.   
    The names of the default input values appear in the inputs fields. You can leave these defaults or override them with new values.
    Input's description (on hovering help icon ![Help icon]( /images/ui/icons/help-icon.png )) might help you understand how to fill-in the proper value.
    An alternative for providing the inputs is to specify a .yaml file containing the relevant values.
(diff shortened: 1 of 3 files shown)


qooban (dismissed):
Hmm... I'm approving not to block it, but I suspect it's a bit overcomplicated. Probably we could avoid validation state variables and should stop using login error from Redux store in `LoginPage` component.

I did a quick try to simplify it - not fully tested, just to share general idea - https://github.com/cloudify-cosmo/cloudify-stage/pull/2451
Could you let me know what do you think? 

Alternatively you can use derived state from props (and pass `loginError` to some kind of state variable only when `loginError` changes), but AFAIR that's not recommended by React - https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops.

Vorbert-Kruk:
> Hmm... I'm approving not to block it, but I suspect it's a bit overcomplicated. Probably we could avoid validation state variables and should stop using login error from Redux store in `LoginPage` component.
> 
> I did a quick try to simplify it - not fully tested, just to share general idea - #2451 Could you let me know what do you think?
> 
> Alternatively you can use derived state from props (and pass `loginError` to some kind of state variable only when `loginError` changes), but AFAIR that's not recommended by React - https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops.

@qooban I think that the code share by you have indeed a useful simplification, which we could use 🎉 
To be honest, I was a little bit shortsighted on the solution, because of the previous way of implementation 😄 
But we can totally go with your idea and stop using data from store in order of using the local state 💪 
Thanks a lot for spotting that direction and for providing initial implementation of it 💪 
I'll adjust the code in a similar way 🎉

Vorbert-Kruk:
@qooban I've applied mentioned simplification 💪 
Also I've adjusted and refactored it a little bit, to make the solution fully working 🎉 
Additionally, new round of system tests have been executed 🚀 

Besides that, I can see that `auth.error` redux field is not being used anywhere except in [ManagerStatePersister file](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/utils/ManagerStatePersister.ts#L23)
Because of that, I'll probably create a ticket for removing this property, as it doesn't seem to bring any value 💪

Vorbert-Kruk:
Unit test were failing after adding `throw error` in `login` action
I've adjusted them within b03475621bb6fd6e0be47b034ac21fc884498e6f 🎉

PR #2489, opened 2023-03-20
RD-6965 hidden open in composer buttons when user doesnt have bluepri…

Blewin:
…nt_upload permission

## Checklist


app/utils/stageUtils.ts
```diff
@@ -199,6 +200,10 @@ export default class StageUtils {
         return _.intersection(authorizedRoles, userRoles).length > 0;
     }
 
+    static shouldHideComposerLink(managerData: ManagerData) {
+        return !StageUtils.isUserAuthorized(Consts.permissions.COMPOSER_LINK, managerData);
+    }
+
```
kubama, line 206: Looking at other functions in this class, which are pretty generic 'utils', this one is in comparison very specific. Thus, I'm not sure if it's the best place to put it in.
I can also see it's only referenced by blueprints widget code base, so I'd say it should be placed there.

app/utils/stageUtils.ts
```diff
@@ -199,6 +200,10 @@ export default class StageUtils {
         return _.intersection(authorizedRoles, userRoles).length > 0;
     }
 
+    static shouldHideComposerLink(managerData: ManagerData) {
```
kubama, line 203: Each invocation of this function is preceded with `!` (negation). It would be more reasonable to have `shouldShowComposerLink` then.

kubama (commented):
1. Shouldn't blueprint action buttons widget be updated as well?
2. Can we have system test scenario to cover the implemented change?
3. The blueprints widget has configuration option to show/hide composer link - shouldn't this option also be altogether hidden if user has no composer access?

kubama:
I also think it would be good to update relevant widgets documentation - just a minor update providing information, that certain options are available depending on assigned user roles.

app/utils/stageUtils.ts
Blewin: I just introduced it, and thought that burying such buisness knowledge in widget is not a good practice,

app/utils/stageUtils.ts
Blewin: haha, good idea

Blewin:
> The blueprints widget has configuration option to show/hide composer link - shouldn't this option also be altogether hidden if user has no composer access?

- i think that this would be too confusing to the user, having different options for no apparent reason, i propose adding explanation text

Blewin:
> Shouldn't blueprint action buttons widget be updated as well?

Done

Blewin:
Docs update:
https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2537

app/utils/stageUtils.ts
Blewin: done

app/utils/stageUtils.ts
qooban: I see it's now referenced in 2 widgets, which supports the idea of sharing it, but what Kuba M. mentioned (there are pretty generic utility functions in `app/utils/stageUtils.ts` and `shouldShowComposerLink` is quite specific) is also still valid. Maybe widgets common would be a better place? 🤔

Blewin: moved it into /widget/common/actions, feels good, wdyt kuba?

app/widgets/common/blueprints/BlueprintActions.ts
```diff
@@ -191,6 +193,10 @@ export default class BlueprintActions {
         return Object.values(CompletedBlueprintStates).includes(blueprint.state);
     }
 
+    static shouldShowComposerLink(managerData: ManagerData) {
```
kubama, line 196: I think this function can be potentially used in broader context than just deciding on composer link visibility - can we then name it in a more generic way? Like `hasComposerPermission` or something similar?

app/utils/stageUtils.ts
kubama: I think it's not exactly blueprint related in the way other functions in this file are. But TBH I don't have a better proposal - `roles/utils.ts` could be used, but it's also not a perfect fit.
Alternatively, if we renamed this function as I suggested in another comment I think it would suit better in `StageUtils` and could then stay there.

app/widgets/common/blueprints/BlueprintActions.ts
Blewin: This name was done on purpose, since i wanted to keep business info "what permissions are needed to show composer link" in one place. naming it hasComposerPermission would dilute it

kubama: I think such "business info" should be kept in the related component, as it's strictly related to the view. Seeing such named function in the context of blueprint model related logic seems messy to me.

Blewin: The problem is that it is in few places, not only one component.

app/utils/stageUtils.ts
Blewin: youp, did that!

kubama (dismissed):

Blewin:
stage: [Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/5010/)

kubama (approved):

PR #2509, opened 2023-03-29
RND-57: Disabled buttons when there are no blueprints in the system

Vorbert-Kruk:
## Checklist


widgets/deploymentButton/src/widget.tsx
```diff
@@ -33,8 +29,18 @@ Stage.defineWidget({
...
+        const searchActions = new SearchActions(toolbox);
+        return searchActions.doListBlueprints(widget.configuration.blueprintFilterRules, {
+            _include: 'id',
+            _size: 1
+        });
+    },
+
+    render(widget, data, _error, toolbox) {
         const { basic, color, icon, label, blueprintFilterRules } = widget.configuration;
+        const disableDeploymentButton = isEmpty(data) || data!.items.length === 0;
```
Vorbert-Kruk, line 42: ### Note
I know that this line could be shorter:
```typescript
const disableDeploymentButton = !data?.items?.length;
```

But I decided to go with the longer version as it may indicate cases which are being validated
1. When `data` is not existing - it appears before `data` is being fetch and while `data` is being fetched
2. When there are no blueprints in the system

Same logic has been applied to the `widgets/environmentButton/src/widget.tsx`

app/widgets/common/filters/FilterActions.ts
```diff
@@ -1,10 +1,15 @@
...
+import type { QueryStringParams } from 'backend/sharedUtils';
 import type { Filter, FilterRule } from './types';
 
 export default class FilterActions {
     constructor(private toolbox: Stage.Types.Toolbox) {}
 
-    doList(params: Record<string, any>): Promise<Stage.Types.PaginatedResponse<Filter>> {
+    doGet(filterId: string, params?: QueryStringParams) {
+        return this.toolbox.getManager().doGet<Filter>(`/filters/deployments/${filterId}`, params);
+    }
```
Vorbert-Kruk, line 10: ### Note
It's possible to optimize amount of getting data here as we could implement similar solution as in [DeploymentActions](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/app/widgets/common/deployments/DeploymentActions.ts#L63-L65)

But for the sake of simplicity and as we are not sending many of those requests (currently once per page load) I thought that it could be omited for now

But for sure it's something that we could implement in the future 💪

Vorbert-Kruk: Btw, maybe it could be a valuable refactor to have in mind, while going through the repo files

From one point of view it increases type safety in our repo and from the other point of view it could help with optimizing requests (as we could detect places which are trying to fetch not used data)

Just a thought 😁

test/cypress/integration/widgets/deployment_button_spec.ts
```diff
@@ -126,6 +126,30 @@ describe('Create Deployment Button widget', () => {
         );
     };
 
+    it('disables clicking deployment button when there are no blueprints available', () => {
```
qooban, line 129: Maybe
```suggestion
    it('is disabled when there are no blueprints available', () => {
```
?

test/cypress/integration/widgets/environment_button_spec.ts
```diff
@@ -18,6 +34,28 @@ describe('Environment button widget', () => {
         cy.clickButton('Cancel');
     });
 
+    it('disables clicking environment button when there are no blueprints available', () => {
```
qooban, line 37: Maybe
```suggestion
    it('is disabled when there are no blueprints available', () => {
```
?

widgets/deploymentButton/src/widget.tsx
qooban: Maybe
```suggestion
        const disableDeploymentButton = Stage.Utils.isEmptyWidgetData(data) || data!.items.length === 0;
```
?

widgets/environmentButton/src/EnvironmentButton.tsx
```diff
@@ -22,6 +23,7 @@ const EnvironmentButton: FunctionComponent<EnvironmentButtonProps> = ({ configur
             <Dropdown
                 fluid
                 icon={null}
+                disabled={disabled}
```
qooban, line 26: Hmm... Shouldn't we disable just one option in dropdown (L39) and not the whole button? 🤔 

As AFAIR even if we don't have blueprints we still can create environment with recently added environment wrapper.

widgets/deploymentButton/src/widget.types.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export declare namespace DeploymentButtonWidget {
+    type DataItem = Pick<FullBlueprintData, 'id'>;
+
+    export interface Configuration extends ButtonConfiguration {
+        toolbox: Stage.Types.Toolbox;
+        blueprintFilterRules: FilterRule[];
+    }
+
+    export type Data = Stage.Types.WidgetData<Stage.Types.PaginatedResponse<DataItem>>;
```
qooban, line 13: Hmm... I didn't know about `Stage.Types.WidgetData` generic type. Are we using it anywhere else? 🤔 Do we actually need it?

widgets/environmentButton/src/widget.types.ts
```diff
@@ -1,8 +1,13 @@
...
 import type { PostEnvironmentBlueprintRequestBody } from 'backend/routes/blueprints/Environment.types';
+import type { FullBlueprintData } from 'app/widgets/common/blueprints/BlueprintActions';
 
 export interface DefaultableLabel extends Label {
     blueprintDefault: boolean;
 }
 
 export type Capability = PostEnvironmentBlueprintRequestBody['capabilities'][number];
+
+export type DataItem = Pick<FullBlueprintData, 'id'>;
```
qooban, line 11: Do we need `DataItem` to be exported? 🤔 
Especially that the name is quite generic I would limit the scope if possible.

widgets/environmentButton/src/widget.types.ts
```diff
@@ -1,8 +1,13 @@
...
+
+export type WidgetData = Stage.Types.WidgetData<Stage.Types.PaginatedResponse<DataItem>>;
```
qooban, line 13: Maybe it would be good to encapsulate the type inside namespace as you did in Deployment Button widget? 🤔 

(so that it's less generic)

widgets/environmentButton/src/widget.tsx
```diff
@@ -19,7 +24,31 @@ Stage.defineWidget<never, never, ButtonConfiguration>({
...
 
-    render(widget, _data, _error, toolbox) {
-        return <EnvironmentButton toolbox={toolbox} configuration={widget.configuration} />;
+    async fetchData(_widget, toolbox) {
+        const searchActions = new SearchActions(toolbox);
+        const filterActions = new FilterActions(toolbox);
+
+        try {
+            const filterRules = await filterActions.doGet(environmentFilterId);
+            return searchActions.doListBlueprints(filterRules.value || [], {
```
qooban, line 33: We can simplify it. It's possible to pass `filter_id` to `GET /searches/*` endpoint, so what do you think about extending SearchActions to not only support `filter_rules`, but also `filter_id`? 

That way you'll get here very similar code to the one in `widgets/deploymentButton/src/widget.tsx` and maybe that would be a trigger to extract it and use common function in both? 🤔

widgets/environmentButton/src/widget.tsx
```diff
@@ -19,7 +24,31 @@ Stage.defineWidget<never, never, ButtonConfiguration>({
...
+                _include: 'id',
+                _size: 1
+            });
+        } catch (err) {
+            log.error(err);
+            return {};
+        }
```
qooban, line 40: I see you don't catch errors in `widgets/deploymentButton/src/widget.tsx`, only here. Is it on purpose? 🤔

widgets/environmentButton/src/widget.tsx
```diff
@@ -19,7 +24,31 @@ Stage.defineWidget<never, never, ButtonConfiguration>({
...
+    },
+
+    render(widget, data, _error, toolbox) {
+        const disableEnvironmentButton = isEmpty(data) || data!.items.length === 0;
```
qooban, line 44: Maybe:
```suggestion
        const disableEnvironmentButton = Stage.Utils.isEmptyWidgetData(data) || data!.items.length === 0;
```
?

widgets/environmentButton/src/widget.tsx
```diff
@@ -19,7 +24,31 @@ Stage.defineWidget<never, never, ButtonConfiguration>({
     permission: Stage.GenericConfig.WIDGET_PERMISSION('environmentButton'),
     categories: [Stage.GenericConfig.CATEGORY.DEPLOYMENTS, Stage.GenericConfig.CATEGORY.BUTTONS_AND_FILTERS],
 
-    render(widget, _data, _error, toolbox) {
-        return <EnvironmentButton toolbox={toolbox} configuration={widget.configuration} />;
+    async fetchData(_widget, toolbox) {
+        const searchActions = new SearchActions(toolbox);
```
qooban, line 28: Do we need `async` here? I don't see it in `widgets/deploymentButton/src/widget.tsx`.

widgets/deploymentButton/src/widget.tsx
```diff
@@ -33,8 +29,18 @@ Stage.defineWidget({
     permission: Stage.GenericConfig.WIDGET_PERMISSION(widgetId),
     categories: [Stage.GenericConfig.CATEGORY.DEPLOYMENTS, Stage.GenericConfig.CATEGORY.BUTTONS_AND_FILTERS],
 
-    render(widget: Stage.Types.Widget<DeploymentButtonConfiguration>, _data, _error, toolbox) {
+    fetchData(widget, toolbox) {
+        const searchActions = new SearchActions(toolbox);
+        return searchActions.doListBlueprints(widget.configuration.blueprintFilterRules, {
+            _include: 'id',
+            _size: 1
+        });
```
qooban, line 37: Optional:

Maybe we should return number of blueprints to the widget from here instead of returning only a response and do calculations later (in `render` method)? 🤔 

Same about the other widget.

app/widgets/common/filters/FilterActions.ts
qooban: Definitely makes sense to me as we discussed yesterday.

Regarding that specific function (`FilterActions.doGet`) - if we improve `SearchActions` (by adding support for `filter_id` I suggested in another comment) we probably won't need this.

qooban (commented):
Added some comments, mostly minor code-style improvements.

From the UI/UX perspective the most important seems: Are we sure we should disable the whole dropdown in Environment Button widget when there are no blueprints?

widgets/environmentButton/src/widget.tsx
Vorbert-Kruk: Sure, seems like a good idea :)

widgets/environmentButton/src/widget.types.ts
Vorbert-Kruk: No there is no need of doing that - I'm removing `export` annotation right away 💪

widgets/environmentButton/src/widget.types.ts
Vorbert-Kruk: Sure

I was wondering about it, but as there wasn't much types related to widget, I've decided to not go with that strategy
But as you think that it would be a good approach to take, I'll implement it 💪

widgets/deploymentButton/src/widget.types.ts
Vorbert-Kruk: Currently we are using it in 3 places:
- widgets/deploymentButton/src/widget.types.ts
- widgets/environmentButton/src/widget.types.ts
- widgets/tokens/src/widget.types.ts

While working on this ticket I was looking for a common patterns in [Tokens widget](https://github.com/cloudify-cosmo/cloudify-stage/tree/master/widgets/tokens/src/widget.types.ts)
That's why `Stage.Types.WidgetData` has appeared in both `deploymentsButton` and `environmentButton`

Do you think that it can be left like that or do we need an adjustment?
If we'd need an adjustment, what solution would you propose?

test/cypress/integration/widgets/deployment_button_spec.ts
Vorbert-Kruk: Oh, it definitely seems to be simpler 😄 
Let's go in that direction 💪

widgets/environmentButton/src/EnvironmentButton.tsx
Vorbert-Kruk: I've been talking about it with Illan, and he said that it should be the expected results

Argumentation was that during environment creation we want user to select blueprint (via manually providing blueprint name), because of that, we don't to have this option enabled when there are no blueprints available.

I didn't know if it's possible to create environment without providing blueprint, but still the product was asked about this specific case and the decision was made 😄

widgets/environmentButton/src/widget.tsx
Vorbert-Kruk: Yes
The reason is that in `environmentButton` 2 subsequent requests are being sent:
- `filterActions.doGet`
- `searchActions.doListBlueprints`

As we are waiting for the first response with `await`, we can get into the situation when we'd have unhandled error which would lead to render error within the widget:
image

As `deploymentButton` widget is sending request and returning promise from `fetchData` function, its errors are being handled by the inner widget mechanisms and returned within `errors` parameter on `render` function 💪 

That's why the implementation differs on both places

widgets/environmentButton/src/widget.tsx
Vorbert-Kruk: Yes, as in [L31](https://github.com/cloudify-cosmo/cloudify-stage/pull/2509/files#diff-042f309e1fbf8d88d99741a5e731b69592c3bcc7a39c2cb8783cbe77c90e87a9R32) we are awaiting for `filterRules` results

widgets/environmentButton/src/widget.types.ts
Vorbert-Kruk: Addressed in bee5366bc75924f6ac9312718e4c7a5062c922de

widgets/deploymentButton/src/widget.tsx
Vorbert-Kruk: I would prefer to stay with the current approach
The reasoning is that currently it's explicit why the buttons are being disabled (`Stage.Utils.isEmptyWidgetData(data) || data!.items.length === 0;`) 

By trying to migrate calculations to the `fetchData` function, we would need to complicate code a little bit, as data would consists of just number - which is not a common pattern in our repository 😁 
In addition we could try to handle success/error fetching state, which would also increase the complicity of the code

Because of that, I'd stay with the current solution
Would that be okay with you?

widgets/environmentButton/src/widget.tsx
Vorbert-Kruk: After discussing the suggestion on slack, I've applied the adjustment within f5348fb827e3da5b82aa4c84ad333359b466cfee

widgets/environmentButton/src/widget.tsx
Vorbert-Kruk: After applying [one of your suggestions](https://github.com/cloudify-cosmo/cloudify-stage/pull/2509#discussion_r1154320691) `async` were no longer needed, so I've removed it 💪

widgets/environmentButton/src/widget.tsx
Vorbert-Kruk: After applying [one of your suggestions](https://github.com/cloudify-cosmo/cloudify-stage/pull/2509#discussion_r1154320691), fetching data from multiple endpoints is no longer required
Because of that `try catch` block seems to be obsolete and was removed 💪

app/widgets/common/filters/FilterActions.ts
Vorbert-Kruk: Agreed 💪 
After adjusting `SearchActions`, I've removed this piece of code within 23d86a6dbce04ff485dac238487e7c21b92f6e31 🎉

widgets/deploymentButton/src/widget.types.ts
qooban: Looking at [app/utils/widgetDataFetcher.ts](https://github.com/cloudify-cosmo/cloudify-stage/blob/f4cf0566f6c2bbeb2d34d2f1e8fd46b4e41e8664/app/utils/widgetDataFetcher.ts) the return type of `fetchByFunc` function is `Promise<any>`, so I'm not sure `Stage.Types.WidgetData` is acurrate for `data` in widget's `render` method.

Anyways, for the sake of simplicity in this PR, I'm OK to keep your solution.

widgets/environmentButton/src/widget.tsx
qooban: OK 👍🏻

widgets/deploymentButton/src/widget.tsx
qooban: OK.

qooban (approved):
Thanks for addressing comments.

Adding `filter_id` support looks useful IMHO and we may use it soon. If system tests after that are passing, I'm OK to merge.

widgets/deploymentButton/src/widget.types.ts
Vorbert-Kruk: Kk, in that case I'll omit it for the current PR
But in the convenient time I'll try to do an investigation, to check what would be the optimal solution 💪

PR #2521, opened 2023-04-06
RND-146: Given ability to modify workflow parameters while running workflow

Vorbert-Kruk:
## Description

Changes consists of:
- Updating modal header
- Introducing form, which enables user to override workflow parameters values
- Introducing new test scenario
- Adjusting existing translations
- Splitting and extracting parts of `RunWorkflowModal` to dedicated files

### Before
image

### After
image

## Checklist

## Tests

 

1. Tests execution after introducing new features
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4948/)

2. Tests execution after introducing new test scenario
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4953/)

3. Tests execution after addressing CR comments
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4977/)

## Documentation

https://github.com/cloudify-cosmo/docs.getcloudify.org/pull/2541


app/widgets/common/deploymentsView/header/RunWorkflowModal.utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    );
+
+    return filteredWorkflowParameters;
+};
+
+export const initializeWorkflowParameters = (
+    workflowParameters: FetchedWorkflow['parameters']
+): SimplifiedWorkflowParameter[] => {
+    const mappedParameters = mapFetchedWorkflowParameters(workflowParameters);
+    return filterSupportedWorkflowParameters(mappedParameters);
```
Vorbert-Kruk, line 61: ### Note
Doing mapping first and then filtering was intentional

Reason behind it is to reduce amount of `Array.prototype.reduce` usages, so that the code could be more readable and easier to understand

It's not the most optimal solution (as the rule of thumb is to filter out unnecessary items first, before doing any sort of manipulations) but I thought that in this case performance difference would be unnoticeable (in contrast to code readability 😁 )

app/widgets/common/deploymentsView/header/RunWorkflowModal.utils.ts
```diff
@@ -0,0 +1,57 @@
...
+    type OptionsGroupName = 'enabledOptions' | 'disabledOptions';
+    type GroupedOptions = Partial<Record<OptionsGroupName, DropdownItemProps[]>>;
+
+    const { enabledOptions = [], disabledOptions = [] } = chain(workflows)
+        .sortBy('name')
+        .map(workflow => ({
+            text: getWorkflowOptionText(workflow),
+            value: workflow.name,
+            disabled: workflow.disabled
+        }))
```
Vorbert-Kruk, line 19: ### Note

Previously there was filter, which were disabling selection of workflows which had parameters requiring an input from user (see bbeae1067a227081ae79e50012748a59ed058ed1)
Removed line was presented like:
```typescript
.filter(workflow => !find(workflow.parameters, parameter => parameter.default === undefined))
```

As previously we've no way of requiring user to provide parameters data, I thought that this mechanism was implemented in order of not executing workflows which would be failing (as required parameter values wouldn't be provided)
That's why I decided to remove it, as currently we've got a strategy for dealing with such cases

While keeping this line, whole mechanism for validating parameter inputs would be redundant and could be removed - I've noticed that, just after implementing form validation

I wanted to emphasize this change, as it may be left unnoticed due to it's extraction to a separate file

If my understanding was incorrect and the decision wasn't the right one, please let me know!

qooban:
I'll review it tomorrow.

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -141,17 +168,42 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
+                                >
+                                    <InputField
+                                        onChange={setParametersInputs}
+                                        toolbox={toolbox}
+                                        error={!!errors[parameters.name]}
+                                        value={parametersInputs[parameters.name]}
+                                        input={{
+                                            type: parameters.type as Input['type'],
+                                            name: parameters.name,
+                                            display: {},
```
qooban, line 192: This property is only valid for `textarea` type, but on https://docs.cloudify.co/staging/dev/developer/blueprints/spec-workflows/ I don't see it in parameters schema while it's present on https://docs.cloudify.co/staging/dev/developer/blueprints/spec-inputs/.

Can you maybe check with someone from the backend - do we support `display` for workflow parameters? 🤔 
If we support it, we should probably pass it through here as well.

app/widgets/common/deploymentsView/header/RunWorkflowModal.utils.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+};
+
+const filterSupportedWorkflowParameters = (
+    workflowParameters: EnhancedWorkflow['parameters']
+): EnhancedWorkflow['parameters'] => {
+    const supportedParameterTypes = ['string', 'integer', 'float', 'boolean', 'list', 'textarea'];
+
+    const filteredWorkflowParameters = workflowParameters.filter(
+        parameter => parameter.type === undefined || supportedParameterTypes.includes(parameter.type as string)
```
qooban, line 46: 1. Casting here should not be necessary if we fix `Workflow.parameters` type definition. Mentioned that in one of the comments.
2. How about adding `undefined` to `supportedParameterTypes`? 🤔

app/widgets/common/deploymentsView/header/useParametersInputs.ts
```diff
@@ -0,0 +1,21 @@
+import { useState } from 'react';
+import { Form } from 'cloudify-ui-components';
```
qooban, line 2: I think we should import `Form` from `basic` instead, right?

app/widgets/common/deploymentsView/header/useParametersInputs.ts
```diff
@@ -0,0 +1,21 @@
...
+    };
+
+    return [
+        inputs,
+        (values: any, field?: any) => setInputs({ ...inputs, ...(field ? Form.fieldNameValue(field) : values) }),
+        resetInputs
+    ] as const;
+}
+
+export default useParametersInputs;
```
qooban, line 21: Hmm... Can't we use `useInputs` hook here? Looks very similar.

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -1325,6 +1325,77 @@ describe('Deployments View widget', () => {
...
         });
 
+        it('should allow to manipulate workflow parameters', () => {
+            const parameterNames = {
+                notVisible: 'not_visible',
+                visibleAndRequired: 'visible_and_required',
+                visibleAndNotRequired: 'visible_and_not_required'
+            } as const;
+
+            const mockedParameters = {
```
qooban, line 1335: I would ad here parameter without `type`, so the case currently broken (see my other comment) will be covered 🙂

test/cypress/integration/widgets/deployments_view_spec.ts
```diff
@@ -1325,6 +1325,77 @@ describe('Deployments View widget', () => {
...
+
+                cy.getField(parameterNames.visibleAndRequired).should('have.class', 'error');
+                cy.typeToFieldInput(parameterNames.visibleAndRequired, 'test');
+
+                cy.clickButton('Run');
+
+                cy.interceptSp('POST', '/execution-groups', request => {
+                    const { default_parameters: defaultParameters } = request.body;
+
+                    expect(defaultParameters[parameterNames.notVisible]).to.be.an('undefined');
```
qooban, line 1390: Optional:

Very minor. 

I think
```suggestion
                    expect(defaultParameters[parameterNames.notVisible]).to.be('undefined');
```
is more correct from spoken language perspective.

If we had "undefined something", then "an" preposition would make sense here.

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -125,12 +130,34 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
         turnOffLoading();
     }
 
+    const initializeParametersInputs = () => {
+        const defaultParametersData = selectedWorkflow?.parameters
+            ? selectedWorkflow.parameters.reduce((parameters, parameter) => {
+                  parameters[parameter.name] = parameter.default;
+                  return parameters;
+              }, {} as Record<string, unknown>)
```
qooban, line 138: Optional:

Depending on how you address comments in `useParametersInputs` maybe worth to reuse type from there:
```suggestion
              }, {} as ParameterInputs)
```
? 🤔

qooban (commented):
Looks quite good. I did some basic UI tests and added comments, hope they'll help to make it even better.

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -141,17 +168,42 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
+                                    required={parameters.required}
+                                    error={errors[parameters.name]}
+                                >
+                                    <InputField
+                                        onChange={setParametersInputs}
+                                        toolbox={toolbox}
+                                        error={!!errors[parameters.name]}
+                                        value={parametersInputs[parameters.name]}
+                                        input={{
+                                            type: parameters.type as Input['type'],
```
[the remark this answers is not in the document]
Vorbert-Kruk: Just checked and it's not required - probably it was required during one of the previous phases of working on this PR
In that case I'll remove it, good catch! 💪

app/widgets/common/deploymentsView/header/RunWorkflowModal.utils.ts
```diff
@@ -0,0 +1,57 @@
...
+    workflowParameters: FetchedWorkflow['parameters']
+): SimplifiedWorkflowParameter[] => {
+    return map(workflowParameters, (parameterFields, parameterName) => {
+        return {
+            ...parameterFields,
+            name: parameterName,
+            required: parameterFields.default === undefined
+        };
+    });
+};
```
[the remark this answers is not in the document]
Vorbert-Kruk: If that wouldn't be an issue, I'd leave it as it is
I know that `return` is not required in multiple places, but I'm leaving it intentionally, as it's making the code more readable for me 
One additional benefit for me is that during development phase it's much easier to adjust/enhance pieces of code that are having `return` statements

I've seen multiple comments similar to this one, that's why I wanted to share my perspective on this topic, so that the thought behind those decisions could be understood

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -141,17 +168,42 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
                             options={workflowsOptions}
-                            onChange={(_event, { value }) => setWorkflowId(value as string)}
-                            value={workflowId}
+                            onChange={handleSelectWorkflow}
+                            value={selectedWorkflow?.name}
                         />
                     </Form.Field>
+                    {selectedWorkflow &&
+                        selectedWorkflow.parameters.map(parameters => {
+                            return (
```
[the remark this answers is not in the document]
Vorbert-Kruk: Sure 👌

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -101,20 +86,40 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
+        const validationErrors: Record<string, string> = {};
+
+        if (!selectedWorkflow) {
+            validationErrors.error = tModal('errors.noWorkflowError');
+        }
+
+        if (selectedWorkflow) {
+            const requiredParameters = selectedWorkflow.parameters.filter(parameter => parameter.required);
+
+            requiredParameters.forEach(parameter => {
```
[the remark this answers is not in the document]
Vorbert-Kruk: Sure, seems like a nice suggestion 💪 
I'll just add alias for the `name` to be `parameterName` if that'd be okay with you
Reason being is that I find it helpful from readability perspective - while reading a code we don't need to search a reference between "name" and any other resource, which makes specific lines self explanatory 🎉 

I know that I may be overthinking this stuff 😄 
But if that approach doesn't bring up any other issue, then why not go with this direction 😁

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -101,20 +86,40 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
     async function runWorkflow() {
-        if (!workflowId) {
-            setErrors({ error: tModal('errors.noWorkflowError') });
+        const validationErrors: Record<string, string> = {};
+
+        if (!selectedWorkflow) {
+            validationErrors.error = tModal('errors.noWorkflowError');
+        }
+
+        if (selectedWorkflow) {
```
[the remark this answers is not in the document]
Vorbert-Kruk: Good idea 💪 
I'll implement it right away

app/widgets/common/deploymentsView/header/RunWorkflowModal.utils.ts
Vorbert-Kruk: First suggestion I'll implement right away 💪 

When it comes to the second, I'd prefer to leave it as it is
Reason being is that `undefined` doesn't seem to me like a valid type of the parameter - rather than that, it's occurrence seems to be sending a "missing type" message to a developer 😁 
That's why I'd prefer to leave it as it is, to have a differentiation between having a specific parameter type and not having it at all 

Let me know if that'd be okay with you

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
Vorbert-Kruk: Implemented within b6ee5a7027288bc21736c646b37efb7cbf426023

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
Vorbert-Kruk: Implemented within b6ee5a7027288bc21736c646b37efb7cbf426023

app/widgets/common/deploymentsView/header/RunWorkflowModal.utils.ts
Vorbert-Kruk: First suggestion was implemented within b6ee5a7027288bc21736c646b37efb7cbf426023

app/widgets/common/deploymentsView/header/useParametersInputs.ts
Vorbert-Kruk: Oh yeah, you're right
Adjusting it right away

Vorbert-Kruk: Adjusted within 5250da6f4420c43aede931ad70cd3a607f982280

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
Vorbert-Kruk: Addressed within rameters.type as

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -141,17 +168,42 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
+                    {selectedWorkflow &&
+                        selectedWorkflow.parameters.map(parameters => {
+                            return (
+                                <Form.Field
+                                    label={parameters.name}
+                                    key={parameters.name}
+                                    required={parameters.required}
+                                    error={errors[parameters.name]}
+                                >
+                                    <InputField
```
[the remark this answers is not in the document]
Vorbert-Kruk: On my environment input for `operation` is being presented in the following way:

image

Could you verify it again? Maybe you had some old code version on your machine and now the issue doesn't appear 🤔

test/cypress/integration/widgets/deployments_view_spec.ts
Vorbert-Kruk: It was adjusted within dee6c5b44896a0ec3b640708c001d54d6f475943

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
Vorbert-Kruk: Addressed within b6ee5a7027288bc21736c646b37efb7cbf426023

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
Vorbert-Kruk: Sure 💪 , extracted the type and reused it within b36a865bf3ecbe7dcca77b7ff5579e3fdb5121a4

app/widgets/common/deploymentsView/header/useParametersInputs.ts
Vorbert-Kruk: I'll address this comment on a slack

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -141,17 +168,42 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
                             selection
                             options={workflowsOptions}
-                            onChange={(_event, { value }) => setWorkflowId(value as string)}
-                            value={workflowId}
+                            onChange={handleSelectWorkflow}
+                            value={selectedWorkflow?.name}
                         />
                     </Form.Field>
+                    {selectedWorkflow &&
+                        selectedWorkflow.parameters.map(parameters => {
```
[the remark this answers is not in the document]
Vorbert-Kruk: I'll address this comment on a slack

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -141,17 +168,42 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
+                                    <InputField
+                                        onChange={setParametersInputs}
+                                        toolbox={toolbox}
+                                        error={!!errors[parameters.name]}
+                                        value={parametersInputs[parameters.name]}
+                                        input={{
+                                            type: parameters.type as Input['type'],
+                                            name: parameters.name,
+                                            display: {},
+                                            constraints: [],
```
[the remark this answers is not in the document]
Vorbert-Kruk: `display` property was adjusted while addressing other comment
But as `constraints` property wasn't discussed on a Friday's daily (with Mateus)z and would need a further investigation, I'll leave it as it is, so that we would be able to proceed with this PR

test/cypress/integration/widgets/deployments_view_spec.ts
Vorbert-Kruk: Unfortunately this small change won't be easy to implement

At first I've also tried to go without `an`, but then I've started getting the following error:
```typescript
TypeError: expect(...).to.be is not a function
```

So when it comes to language perspective, I think that `Chai` has it on perspective on that 😄

Vorbert-Kruk:
After addressing CR comments, I've executed system tests once more
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/4977/)

(I've also added it to the PR description)

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
qooban: Yes, it's the same. You see - there's `{}` initial value set on the field in the form, but `operation` parameter does not have `default` value set 🙂

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
qooban: I think the scope where it's used explains it, but OK, let it be `parameterName` 🙂

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
qooban: `constraints` property is related to complex types IIRC (not types which we are going to support for bulk execution), so we should be good leaving it.

test/cypress/integration/widgets/deployments_view_spec.ts
qooban: I suspect it's still valid. See my comment - https://github.com/cloudify-cosmo/cloudify-stage/pull/2521#discussion_r1168550003.

test/cypress/integration/widgets/deployments_view_spec.ts
qooban: OK, let's leave it then.

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
```diff
@@ -101,20 +86,40 @@ const RunWorkflowModal: FunctionComponent<RunWorkflowModalProps> = ({ filterRule
...
+            clearErrors();
             setLoadingMessage(tModal('messages.creatingDeploymentGroup'));
             const groupId = getGroupIdForBatchAction();
             const deploymentGroupsActions = new DeploymentGroupsActions(toolbox);
             await deploymentGroupsActions.doCreate(groupId, { filter_rules: filterRules });
 
             setLoadingMessage(tModal('messages.startingExecutionGroup'));
             const executionGroupsActions = new ExecutionGroupsActions(toolbox);
-            await executionGroupsActions.doStart(groupId, workflowId);
+            await executionGroupsActions.doStart(groupId, selectedWorkflow!.name, parametersInputs);
```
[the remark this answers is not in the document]
qooban: You don't have to create new `if (selectedWorkflow)` block, there's one available earlier in the code AFAIR.

Anyways, it's minor issue, so I won't insist.

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
qooban: Oh, but wait... It's different from what you shared 🙂 I just double checked and it looks better after your recent changes:

image

so `operation` property has no initial value set. All good then!

test/cypress/integration/widgets/deployments_view_spec.ts
qooban: It seems that after your recent changes (probably when you started using `InputFields`) we don't have any issues with `type === undefined`, so OK, let's leave it like that.

qooban:
> Please just take a look at `operation` parameter issue.

Tested UI with the latest changes and it seems like there's no more issue with parameter without `type` and without `default` values.

app/widgets/common/deploymentsView/header/RunWorkflowModal.tsx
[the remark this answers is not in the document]
Vorbert-Kruk: Oh, sorry in that case, I've interpreted the suggestion differently 😅 

Yeah, there is `if` statement that we can use, but in between both of those code pieces we are having additional condition which is returning from function if any error has been set

Implementing suggestion and rearranging code pieces would add some complexity (and pyramid like code structure), which may be unnecessary

So if you don't insist, I'd leave it as it is

But nonetheless, nice suggestion 💪 
For sure I'll have in mind similar cases in the future, just so they could be implemented if possible 🎉

qooban (approved):

PR #2531, opened 2023-04-12
RD-5771 Migrate Resource Filter widget to TypeScript

qooban:
## Description
Full TS migration of Resource Filter widget with translation support and using explicit lodash imports.

## Screenshots / Videos
N/A

## Checklist

## Tests
1. [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/4938/) - failed on known issues only (RD-4589 and RD-6901), so I'm considering it as PASSED
2. [Build Status](https://jenkins.cloudify.co/job/Stage-UI-System-Test/4949/) - full rerun after code review adjustments, failed on known issues only (RD-6631), so I'm considering it as PASSED

## Documentation
N/A

### DIFF
--- a/app/utils/StageAPI.ts
+++ b/app/utils/StageAPI.ts
@@ -250,6 +250,7 @@ declare global {
             nodeInstanceId: any;
             executionStatus: any;
             siteName: any;
+            depNodeId: string | null;
         }
 
         const i18n: typeof import('i18next').default;
--- a/widgets/filter/src/Filter.tsx
+++ b/widgets/filter/src/Filter.tsx
@@ -1,21 +1,37 @@
-// @ts-nocheck File not migrated fully to TS
+import type { FilterConfiguration } from 'widgets/filter/src/types';
+import type { DynamicDropdownProps } from 'app/widgets/common/components/DynamicDropdown';
+
 const deploymentFilter = { deployment_id: 'deploymentId' };
 const blueprintFilter = { blueprint_id: 'blueprintId' };
 const blueprintDeploymentFilter = { ...deploymentFilter, ...blueprintFilter };
 
-const filterFields = [
+type FilterField = keyof Stage.ContextEntries;
+const filterFields: FilterField[] = [
     'blueprintId',
     'deploymentId',
     'nodeId',
     'nodeInstanceId',
     'executionId',
     'executionStatus',
-    'siteName'
+    'siteName',
+    'depNodeId'
 ];
 
-export default class Filter extends React.Component {
-    constructor(props, context) {
-        super(props, context);
+interface FilterProps {
+    configuration: FilterConfiguration;
+    toolbox: Stage.Types.Toolbox;
+}
+type FilterState = Stage.ContextEntries;
+
+export default class Filter extends React.Component<FilterProps, FilterState> {
+    private eventHandlers: {
+        'deployments:refresh': (deploymentIds?: any) => void;
+        'filter:refresh': () => void;
+        'blueprints:refresh': (blueprintIds?: any) => void;
+    };
+
+    constructor(props: FilterProps) {
+        super(props);
 
         this.state = {
             ...this.getStateFromContext()
@@ -33,12 +49,12 @@ export default class Filter extends React.Component {
         _.each(this.eventHandlers, (handler, eventName) => toolbox.getEventBus().on(eventName, handler, this));
     }
 
-    shouldComponentUpdate(nextProps, nextState) {
+    shouldComponentUpdate(nextProps: FilterProps, nextState: FilterState) {
         const { configuration } = this.props;
         return !_.isEqual(configuration, nextProps.configuration) || !_.isEqual(this.state, nextState);
     }
 
-    componentDidUpdate(prevProps) {
+    componentDidUpdate(prevProps: FilterProps) {
         const { configuration } = this.props;
         const oldAllowMultipleSelection = prevProps.configuration.allowMultipleSelection;
         const newAllowMultipleSelection = configuration.allowMultipleSelection;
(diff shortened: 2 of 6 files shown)


widgets/filter/src/Filter.tsx
```diff
@@ -51,70 +68,81 @@ export default class Filter extends React.Component {
...
+            };
+        }, {} as FilterState);
     }
 
-    setValue(name, value) {
+    setValue(name: FilterField, value: Stage.ContextEntries[FilterField]) {
         const { toolbox } = this.props;
         toolbox.getContext().setValue(name, value);
-        this.setState({ [name]: value });
+        this.setState((prevState: FilterState) => ({ ...prevState, [name]: value }));
```
qooban, line 88: I changed this use of it in order to avoid `@ts-ignore` as typing of React's `setState` is not perfect, see: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18365#issuecomment-351884578 for details.

app/utils/StageAPI.ts
```diff
@@ -250,6 +250,7 @@ declare global {
             nodeInstanceId: any;
             executionStatus: any;
             siteName: any;
```
qooban, line 252: I tried to change `any`s here to proper types at once, but it turned out to be much more work, so I'm leaving it for RD-6559.

widgets/filter/src/Filter.tsx
```diff
@@ -138,29 +166,53 @@ export default class Filter extends React.Component {
...
-            filter,
-            flushOnRefreshEvent
-        }) => {
+            valueProp
+        }: FilterDropdownProps) => {
             const { DynamicDropdown } = Stage.Common.Components;
             const { appendQueryParam } = Stage.Utils.Url;
             const { configuration, toolbox } = this.props;
 
-            const joinedEntityName = entityName.replace(' ', '');
```
qooban, line 159: Adding type-safety with all that names constructed using string manipulation would make the code quite complex IMHO, so I ended up with making `stateProp`, `enabledConfigurationKey` to be required and added `onChange` prop.

widgets/filter/src/Filter.tsx
```diff
@@ -138,29 +166,53 @@ export default class Filter extends React.Component {
...
-            if (configuration[enabledConfigurationKey || `filterBy${joinedEntityName}s`]) {
-                const camelCaseEntityName = _.lowerFirst(joinedEntityName);
-                const { [stateProp || `${camelCaseEntityName}Id`]: value } = this.state;
+            if (configuration[enabledConfigurationKey]) {
+                const camelCaseEntityName = lowerFirst(entityName.replace(' ', ''));
+                const { [stateProp]: value } = this.state;
```
Vorbert-Kruk, line 213: `|| ${camelCaseEntityName}Id` was removed on purpose?

widgets/filter/src/Filter.tsx
```diff
@@ -51,70 +68,81 @@ export default class Filter extends React.Component {
...
         const { allowMultipleSelection } = configuration;
         const context = toolbox.getContext();
 
         if (!allowMultipleSelection) {
-            if (!_.isEmpty(selectedDeploymentId) && !_.isEmpty(selectedNodeId)) {
+            if (
+                !Array.isArray(selectedNodeId) &&
+                !Array.isArray(selectedDeploymentId) &&
+                !!selectedDeploymentId &&
+                !!selectedNodeId
```
kubama, line 144: How about creating `isSingleValue` helper function to avoid repetition?

widgets/filter/src/Filter.tsx
```diff
@@ -138,29 +166,53 @@ export default class Filter extends React.Component {
...
+            fetchManagerEndpoint?: string;
+            filter?: DynamicDropdownProps['filter'];
+            flushOnRefreshEvent?: boolean;
+            pageSize?: DynamicDropdownProps['pageSize'];
+            searchParams?: DynamicDropdownProps['searchParams'];
+            textFormatter?: DynamicDropdownProps['textFormatter'];
+            valueProp?: DynamicDropdownProps['valueProp'];
+        }
+
         const createDropdown = ({
```
Vorbert-Kruk, line 192: ### Optional 

WDYT about extracting `createDropdown` function (along with `FilterDropdownProps`) outside of `Filter` component body?
As I've checked, it can be moved to the top of the file without any errors appearing 🎉 

As it looks like a standard component function (even the name of it's typing is indicating that 😁) I think that we could declare it outside of the `Filter` component 

Additional thought may be that instead of using it like that
```ts
                        createDropdown({
                            entityName: 'Blueprint',
                            fetchManagerEndpoint: 'blueprints?state=uploaded',
                            flushOnRefreshEvent: true
                        }),
```

we could probably use it in the following way

```ts
                        <createDropdown
                            entityName="Blueprint"
                            fetchManagerEndpoint="blueprints?state=uploaded"
                            flushOnRefreshEvent
                        >
```

But then probably `createDropdown` name should be adjusted

I know that those kind of changes are not related to the migration on which you are working, that's why I've marked this comment as an optional one

WDYT? 🤔

widgets/filter/src/Filter.tsx
```diff
@@ -138,29 +166,53 @@ export default class Filter extends React.Component {
...
+            enabledConfigurationKey: keyof FilterConfiguration;
+            fetchAll?: DynamicDropdownProps['fetchAll'];
+            fetchIncludeExtra?: string;
+            fetchManagerEndpoint?: string;
+            filter?: DynamicDropdownProps['filter'];
+            flushOnRefreshEvent?: boolean;
+            pageSize?: DynamicDropdownProps['pageSize'];
+            searchParams?: DynamicDropdownProps['searchParams'];
+            textFormatter?: DynamicDropdownProps['textFormatter'];
+            valueProp?: DynamicDropdownProps['valueProp'];
```
kubama, line 189: How about extending from `DynamicDropdownProps` with `Pick`?

widgets/filter/src/Filter.tsx
qooban: Yes. Now `stateProp` is mandatory, so we don't need to have alternative here.

widgets/filter/src/Filter.tsx
qooban: Makes sense. Will do it.

widgets/filter/src/Filter.tsx
qooban: Good idea 👍🏻 Actually, I was thinking about extracting it at the beginning, but dropped it to focus on the main change (adding TS support) 🙂 

Extracted!

widgets/filter/src/Filter.tsx
qooban: Good idea. Done.

Vorbert-Kruk (approved):
LGTM 💪 
Thanks for addressing all review comments 🎉

kubama (approved):

PR #2398, opened 2023-01-19
RD-6631 System test fix?

kubama:
## Description
The test execution hangs on `cy.visit`, even though test recording reveals that the page loads.
Having no better idea, I tried to make the `before` hook be as similar as possible to what's on L59 (which does not have the same issue). Hopefully it will fix the issue :crossed_fingers:.

## Screenshots / Videos
N/A

## Checklist

- [X] My code follows the [UI Code Style](https://cloudifysource.atlassian.net/l/c/x8fq902p).
- [x] I followed the [UI Testing Guidelines](https://cloudifysource.atlassian.net/l/cp/udCVfvrx) and verified that all tests/checks have passed.
- [X] I verified if that change requires any update in the [documentation](https://cloudifysource.atlassian.net/l/c/4XKtPocy).
- [X] I added proper labels to this PR.

## Tests
https://jenkins.cloudify.co/view/UI/job/Stage-UI-System-Test/4357/

 

## Documentation
N/A

PR #2550, opened 2023-04-25
RD-6631: Updated initially visited url

Vorbert-Kruk:
## Description

Changes consists of modifying initial url which is being used to open the login page

Looking at the video attached to the [RD-6631](https://cloudifysource.atlassian.net/browse/RD-6631) I suspect that cypress can sometimes have difficulties when it comes to the redirection on the same site, that he is already
As when test is failing, we are trying to be redirected to `/console`, when the application is rendered under `/console/login` URL, I thought that there may be some issue with using a path name that is partially available in the already presented URL
Because of that and the fact that we are using in other places `/console/login` path name, I thought that it may make a nice try 💪 

If this solution won't work, we could tempt ourselves with using similar strategy as in [commands.ts](https://github.com/cloudify-cosmo/cloudify-stage/blob/master/test/cypress/support/commands.ts#L230-L234):
```typescript
        cy.location('pathname').then(pathname => {
            if (forceVisitLoginPage || pathname !== '/console/login') {
                cy.visit('/console/login');
            }
        });
```

Which would make sure that we are not trying to do a redirection when we are being presented with the desired page

## Screenshots / Videos

N/A

## Checklist

## Tests

 

1. After implementing initial suggestion
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/5025/)

2. After adjusting suggestion
[Build Status](https://jenkins.cloudify.co/view/UI%20Health%20View/job/Stage-UI-System-Test/5042/)

## Documentation

N/A

[RD-6631]: https://cloudifysource.atlassian.net/browse/RD-6631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/cypress/integration/user_configuration_spec.ts
```diff
@@ -30,7 +30,7 @@ describe('User configuration', () => {
     describe('allows to customize Login page', () => {
         before(() => {
             mockConfigResponse();
-            cy.visit('/console');
+            cy.visit('/console/login');
```
qooban, line 33: Hmm... Didn't we have `/console/login` here before?

See https://github.com/cloudify-cosmo/cloudify-stage/pull/2398.

Vorbert-Kruk: Oh, you are right!
In that case I'll implement suggestion mentioned in the PR description 💪

Vorbert-Kruk: Implementation has been updated as a part of 5a4d6b83aed8526c2ac3d4ca951a9bdde92acc07

test/cypress/integration/user_configuration_spec.ts
```diff
@@ -30,7 +30,7 @@ describe('User configuration', () => {
     describe('allows to customize Login page', () => {
         before(() => {
             mockConfigResponse();
-            cy.visit('/console');
+            cy.visitUrlIfRedirectionIsRequired('/console/login');
```
qooban, line 33: I'm not sure it can help as here we don't have redirection, it's actually first time visit, so IMHO there's another problem. 

According to the error message it seems like `load` event was not triggered for some reason or Cypress missed it somehow 🤔 This is hard to detect.

Workaround which comes to my mind is to convert `before` statement into `beforeEach` and make it possible to add test retries mechanism for `allows to customize Login page`. Actually all 3 tests inside that `describe` block can be combined into one for time optimization.

BTW Did we experience RD-6631 recently? (according to tracking page it was observed at the beginning of month) Maybe it's something we should not be worried about too much? 🤔

Vorbert-Kruk: When it comes to redirection I also was surprised about that idea, but as I was looking on the video attached to the [RD-6631](https://cloudifysource.atlassian.net/browse/RD-6631) I've seen that we are trying to redirect to the page that is already loaded

That's why I've thought about implementation which ensures that we are on a proper page

I suspect that cypress is watching for `onload` event, while executing `cy.visit` functionality
In our case there are 2 issues with that:
- as we were presented with a `login` page before we've been trying to visit it, `onload` event was triggered before we were checking it
- as it seems that we are unable to revisit site with the provided URL, site is not being refreshed and therefore `onload` event is not being triggered

So I think that the solution still may be worth a try! 😁 
But I agree that the root cause of the issue seems to be strange indeed

 
> BTW Did we experience [RD-6631](https://cloudifysource.atlassian.net/browse/RD-6631) recently? (according to tracking page it was observed at the beginning of month) Maybe it's something we should not be worried about too much? 🤔

Oh you're right 😅 
There were 2 issues related to authentication - one was more frequent that the other
And it seems that I've started working on the not most recent one, probably by mistake 😅

Let me know WDYT about the situation
We can for sure leave the code unchanged or try to implement suggested solution, just for sake of trying to increase stability of the test

qooban: > I was looking on the video attached to the [RD-6631](https://cloudifysource.atlassian.net/browse/RD-6631) I've seen that we are trying to redirect to the page that is already loaded

> as we were presented with a login page before we've been trying to visit it, onload event was triggered before we were checking it

I think you are making wrong assumptions here. There's no other code in https://github.com/cloudify-cosmo/cloudify-stage/blob/161a7c33512d86830ea4185bff4e06162c450847/test/cypress/integration/user_configuration_spec.ts responsible for opening login page than `cy.visit` in `before` hook. There's no other magic there, login page was opened by `cy.visit` call.

What happened IMHO is that `cy.visit` called sth to change location in the browser (browser reacted and opened login page) and then started waiting for `onload` event and never received it.

`cy.visit` seems quite complex - https://github.com/cypress-io/cypress/blob/master/packages/driver/src/cy/commands/navigation.ts - so I'm not planning to dig into details ;-)

qooban (commented):
I think it's not worth to merge this PR. Reasoning in open discussion. WDYT?

Vorbert-Kruk:
Closed due to reasons mentioned in discussions

Vorbert-Kruk:
Closed due to reasons mentioned in discussions

PR #2647, opened 2024-08-16
Remove default host from server.listen to allow by default to listen on ipv6 and ipv4

fstaniulamartus:
## Description

From node documentation:

> If host is omitted, the server will accept connections on the [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) when IPv6 is available, or the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0) otherwise.
>
> In most operating systems, listening to the [unspecified IPv6 address](https://en.wikipedia.org/wiki/IPv6_address#Unspecified_address) (::) may cause the net.Server to also listen on the [unspecified IPv4 address](https://en.wikipedia.org/wiki/0.0.0.0) (0.0.0.0).

We were setting `localhost` by default in the configuration, which on some systems where ipv6 was available, it would listen only on ipv6 interface. Omitting the default host, would in turn allow to listen by default on both interfaces if not specified in the configuration by user.

This cherry-picks changes from #2646 to master branch.

## Screenshots / Videos
N/A

## Checklist

## Tests
N/A

## Documentation
N/A

### DIFF
--- a/backend/config.ts
+++ b/backend/config.ts
@@ -117,5 +117,5 @@ export function getClientConfig(mode: Mode): ClientConfig {
 
 export function getBackendConfig() {
     const { backend } = getConfig().app;
-    return { host: backend?.host ?? 'localhost', port: backend?.port ?? 8088 };
+    return { host: backend?.host, port: backend?.port ?? 8088 };
 }
--- a/backend/routes/Config.types.ts
+++ b/backend/routes/Config.types.ts
@@ -11,7 +11,12 @@ export type UserConfig = typeof userConfig;
 
 export type AppConfig = typeof app &
     typeof root &
-    typeof logging & { db: { options: typeof dbOptions } } & typeof userConfig;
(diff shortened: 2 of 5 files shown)


Vorbert-Kruk (dismissed):
Do you think that this change would be worth to reflect on `fusion`?

fstaniulamartus:
> Do you think that this change would be worth to reflect on `fusion`?

I think it's a good change, in case fusion would be started on machines that support both interfaces ipv4 and ipv6. I will move this change to fusion.

This change does not have negative effects on consumers of fusion, since you can configure `app.backend.host` in the `conf/me.json`, so I'd treat it as enhancement.

ykamartus (dismissed):
test approval

Vorbert-Kruk (dismissed):

ykamartus (approved):

PR #2662, opened 2024-09-16
Getting started system test fix

ykamartus:
## Description

Added:
 - loading all yaml files instead of one
 - increasing plugin upload service timeout as we started loading more items to the manager
 - providing tests improvements after changes
 - replacing AWS with GCP in blueprint catalog spec because one of the tests is not passing due to types issues in blueprint yaml file, GCP works correctly 

## Screenshots / Videos

N/A

## Checklist

## Tests

 

[all_tests](https://jenkins.cloudify.co/job/Stage-UI-System-Test/5839/)

### DIFF
--- a/app/components/overlays/gettingStarted/installation/process.ts
+++ b/app/components/overlays/gettingStarted/installation/process.ts
@@ -1,5 +1,5 @@
 import log from 'loglevel';
-
+import isEmpty from 'lodash/isEmpty';
 import type { PostPluginsUploadQueryParams } from 'backend/routes/Plugins.types';
 import StageUtils from '../../../../utils/stageUtils';
 import type Internal from '../../../../utils/Internal';
@@ -29,7 +29,7 @@ const sleep = async (milliseconds: number) => new Promise(resolve => setTimeout(
 
 // TODO(RD-1874): use common api for backend requests
 export const installPlugin = async (internal: Internal, plugin: PluginInstallationTask) => {
-    if (_.isEmpty(plugin.yamlUrls) || !plugin.wagonUrl) {
+    if (isEmpty(plugin.yamlUrls) || !plugin.wagonUrl) {
         return false;
     }
     const params = {
@@ -167,7 +167,7 @@ export const createResourcesInstaller = (
 
         const runInstallPluginStep = async (scheduledPlugin: PluginInstallationTask) => {
             let result = false;
-            if (!_.isEmpty(scheduledPlugin.yamlUrls) && scheduledPlugin.wagonUrl) {
+            if (!isEmpty(scheduledPlugin.yamlUrls) && scheduledPlugin.wagonUrl) {
                 triggerProgressEvent(TaskType.Plugin, scheduledPlugin.name, TaskStatus.InstallationProgress);
                 result = await installPlugin(internal, scheduledPlugin);
                 if (destroyed) return;
--- a/app/components/overlays/gettingStarted/installation/process.ts
+++ b/app/components/overlays/gettingStarted/installation/process.ts
@@ -29,14 +29,14 @@ const sleep = async (milliseconds: number) => new Promise(resolve => setTimeout(
 
 // TODO(RD-1874): use common api for backend requests
 export const installPlugin = async (internal: Internal, plugin: PluginInstallationTask) => {
-    if (!plugin.yamlUrl || !plugin.wagonUrl) {
+    if (_.isEmpty(plugin.yamlUrls) || !plugin.wagonUrl) {
         return false;
     }
     const params = {
         visibility: 'tenant',
         title: plugin.title,
         iconUrl: plugin.icon,
-        yamlUrl: plugin.yamlUrl,
+        yamlUrl: plugin.yamlUrls,
         wagonUrl: plugin.wagonUrl
     };
     try {
@@ -167,7 +167,7 @@ export const createResourcesInstaller = (
 
         const runInstallPluginStep = async (scheduledPlugin: PluginInstallationTask) => {
             let result = false;
-            if (scheduledPlugin.yamlUrl && scheduledPlugin.wagonUrl) {
+            if (!_.isEmpty(scheduledPlugin.yamlUrls) && scheduledPlugin.wagonUrl) {
                 triggerProgressEvent(TaskType.Plugin, scheduledPlugin.name, TaskStatus.InstallationProgress);
                 result = await installPlugin(internal, scheduledPlugin);
                 if (destroyed) return;
--- a/app/utils/shared/PluginUtils.ts
+++ b/app/utils/shared/PluginUtils.ts
@@ -1,12 +1,8 @@
-import { find, last } from 'lodash';
+import { find } from 'lodash';
 import type { PluginCatalogEntry } from 'backend/routes/Plugins.types';
 
-const latestDslVersion = 'cloudify_dsl_1_5';
-export function getYamlUrl(plugin: Pick<PluginCatalogEntry, 'yaml_urls'>) {
-    const latestDslYamlUrl = find(plugin.yaml_urls, { dsl_version: latestDslVersion })?.url;
-    const fallbackYamlUrl = last(plugin.yaml_urls)!.url;
-
-    return latestDslYamlUrl || fallbackYamlUrl;
+export function getYamlUrls(plugin: Pick<PluginCatalogEntry, 'yaml_urls'>) {
+    return plugin.yaml_urls.map(item => item.url);
 }
 
 const fallbackDistributionPrefix = 'manylinux';
@@ -19,5 +15,5 @@ export function getWagon(plugin: Pick<PluginCatalogEntry, 'wagon_urls'>, current
 
 export default {
     getWagon,
-    getYamlUrl
+    getYamlUrls
 };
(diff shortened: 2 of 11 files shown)


app/components/overlays/gettingStarted/installation/process.ts
```diff
@@ -29,14 +29,14 @@ const sleep = async (milliseconds: number) => new Promise(resolve => setTimeout(
 
 // TODO(RD-1874): use common api for backend requests
 export const installPlugin = async (internal: Internal, plugin: PluginInstallationTask) => {
-    if (!plugin.yamlUrl || !plugin.wagonUrl) {
+    if (_.isEmpty(plugin.yamlUrls) || !plugin.wagonUrl) {
```
Vorbert-Kruk, line 32: We can utilize explicit import of `isEmpty` lodash function, instead of utilizing a global `_` for it.

app/components/overlays/gettingStarted/installation/process.ts
```diff
@@ -167,7 +167,7 @@ export const createResourcesInstaller = (
 
         const runInstallPluginStep = async (scheduledPlugin: PluginInstallationTask) => {
             let result = false;
-            if (scheduledPlugin.yamlUrl && scheduledPlugin.wagonUrl) {
+            if (!_.isEmpty(scheduledPlugin.yamlUrls) && scheduledPlugin.wagonUrl) {
```
Vorbert-Kruk, line 170: We can utilize explicit import of `isEmpty` lodash function, instead of utilizing a global `_` for it.

app/utils/shared/PluginUtils.ts
```diff
@@ -1,12 +1,8 @@
...
 import type { PluginCatalogEntry } from 'backend/routes/Plugins.types';
 
-const latestDslVersion = 'cloudify_dsl_1_5';
-export function getYamlUrl(plugin: Pick<PluginCatalogEntry, 'yaml_urls'>) {
-    const latestDslYamlUrl = find(plugin.yaml_urls, { dsl_version: latestDslVersion })?.url;
-    const fallbackYamlUrl = last(plugin.yaml_urls)!.url;
-
-    return latestDslYamlUrl || fallbackYamlUrl;
+export function getYamlUrls(plugin: Pick<PluginCatalogEntry, 'yaml_urls'>) {
+    return plugin.yaml_urls.map(item => item.url);
```
Vorbert-Kruk, line 5: Good to see that we're simplifying stuff on our end 👍 
We may say that is a benefit of spotting this defect 😄

backend/routes/Plugins.types.ts
```diff
@@ -42,5 +42,12 @@ export interface PostPluginsUploadQueryParams {
     title: string;
     visibility: string;
     wagonUrl?: string;
-    yamlUrl?: string;
+    yamlUrl?: string | string[];
 }
+
+export type FileDetails =
```
Vorbert-Kruk, line 48: As this type is exported, maybe we can try to give it more contextual name OR to migrate it to `backend/routes/Plugins.ts` file (without `export` keyword) as it's being utilized only there.
Currently it may be misleading when trying to utilize it to other purposes.

WDYT? Does it make sense or in this example it wouldn't matter as much?

backend/routes/Plugins.ts
```diff
@@ -23,17 +25,43 @@ const upload = multer();
 const logger = getLogger('Plugins');
 const getFiles = (req: Request<any, any, any, any, any>) => req.files as { [fieldname: string]: Express.Multer.File[] };
 
+const getYamlUrls = (yamlUrl?: string | string[]) => {
```
Vorbert-Kruk, line 28: Just to increase type safety, we can add a return type
```suggestion
const getYamlUrls = (yamlUrl?: string | string[]): string[] => {
```

backend/routes/Plugins.ts
```diff
@@ -23,17 +25,43 @@ const upload = multer();
...
+
+const getFileDetails = async (fileSource: Buffer | string, filename = '') => {
+    if (fileSource instanceof Buffer) {
+        return {
+            file: fileSource,
+            name: filename
+        };
+    }
+
+    const fileBuffer: Buffer = await downloadFile(fileSource);
```
Vorbert-Kruk, line 46: ### Optional
We could set a return type of `downloadFile` to be `Buffer`.
It would ensure that we don't need to guess it and assign it on every usage 👍

backend/routes/Plugins.ts
```diff
@@ -54,17 +82,23 @@ function downloadFile(url: string) {
...
 ) {
     const archive = archiver('zip');
-    archive.append(wagonFile, { name: wagonFilename });
-    archive.append(yamlFile, { name: 'plugin.yaml' });
+
+    if (wagonFile) {
+        archive.append(wagonFile.file, { name: wagonFile.name });
+    }
+
+    if (!_.isEmpty(_.compact(yamlFiles))) {
```
Vorbert-Kruk, line 96: We can safely migrate away from utilizing `_` global lodash object, to utilize explicit exports.

backend/routes/Plugins.ts
```diff
@@ -130,45 +164,48 @@ router.post<never, any | GenericErrorResponse, any, PostPluginsUploadQueryParams
...
-            promises.push(Promise.resolve(files.yaml_file[0].buffer));
+            promises.push(undefined);
         }
 
-        if (req.query.iconUrl) {
-            promises.push(downloadFile(req.query.iconUrl));
-        } else if (_.get(req.files, 'icon_file')) {
-            promises.push(Promise.resolve(files.icon_file[0].buffer));
+        const yamlUrls = getYamlUrls(req.query.yamlUrl);
+        if (!_.isEmpty(yamlUrls)) {
```
Vorbert-Kruk, line 183: We can use explicit import for `isEmpty` lodash function.

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -24,12 +19,16 @@ export interface PluginsCatalogItem extends Omit<PluginDescription, 'wagon_urls'
 
 const translate = Stage.Utils.getT('widgets.pluginsCatalog');
 
-function toUploadData(item: PluginsCatalogItem) {
+const getPluginKey = (plugin: PluginsCatalogItem) => {
+    return `${plugin.display_name.replaceAll(' ', '')}_${plugin.uploadedVersion}`.toLowerCase();
```
Vorbert-Kruk, line 23: We can make this concatenation more readable

```suggestion
    const pluginName = plugin.display_name.replaceAll(' ', '').toLowerCase();
    return `${pluginName}_${plugin.uploadedVersion}`;
```

Not sure if `toLowerCase()` needs to be casted also on `plugin.uploadedVersion`, so I've executed it only for the plugin name.

Let me know WDYT 👍

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
```diff
@@ -70,63 +68,66 @@ const PluginsCatalogList: FunctionComponent<PluginsCatalogListProps> = ({ toolbo
...
+            if (pluginUploading) {
+                return translate('uploadButton.uploading');
+            }
+
+            if (recentVersionUploaded) {
+                return translate('uploadButton.disabled');
+            }
+
+            return translate('uploadButton.enabled');
+        };
```
Vorbert-Kruk, line 119: Could we assign it to a variable?
As this code piece is just for rendering text, probably reexecuting this function on every render is not necessary.

We could do something like
```typescript
        const uploadPluginButtonTitle = useMemo(() => {
            if (pluginUploading) {
                return translate('uploadButton.uploading');
            }

            if (recentVersionUploaded) {
                return translate('uploadButton.disabled');
            }

            return translate('uploadButton.enabled');
        }, [pluginUploading, recentVersionUploaded]);
```

Let me know WDYT 👍

Vorbert-Kruk (commented):
Looks good
Left some small comments

As the target of the fix is `7.0.5-build` it's worth to remember that it would be good to `cherry-pick` it to `master`

backend/routes/Plugins.ts
ykamartus: @Vorbert-Kruk it's already inferred to string [] type so there is no need for explicit type assignment

backend/routes/Plugins.types.ts
ykamartus: @Vorbert-Kruk Good catch, renamed it as PluginFileDetails as all of the files are plugins related

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
ykamartus: @Vorbert-Kruk it does not provide any calculations or expensive operations so we will get less readable code and almost no performance improvement that's why it's not wrapped with useMemo here

I wish we already use React 19 as it will handle such performance issues under the hood on compiler level and useMemo/useCallback will be deprecated features

backend/routes/Plugins.ts
Vorbert-Kruk: Yeah, it is but as right now there are multiple points of return, was wondering if making the return type explicit would be a good idea

But as you've mentioned it's not something required, so I'm leaving the decision up to you 👍

widgets/pluginsCatalog/src/PluginsCatalogList.tsx
Vorbert-Kruk: Performance wise you're right
Probably the view of it just was strange for me, as usually we're just passing variables to components.
But certainly it shouldn't be a blocker to leave it as it is right now 👍

Vorbert-Kruk (approved):

